Issue #4, opened 2020-07-31
How should we deal with math options/rendering/html?


rowanc1:
I think there are a couple things to think about:

* Using `markdown-it-texmath` has a soft dependency on `katex` which I am pretty sure we do not want to bundle
  * Is the output the raw latex? Or should we have an option to actually put out the katex output (only requiring css)
  * My instinct at this stage is to leave it as latex (not rendered), similar to jupyter book at the moment
      * Note that this is different than the code, which is pre-parsed.
* We will have to have some defaults on what the math in markdown is configured as (`$`, `\(` `\[`, etc.)
  * I think the default is `$..$`, but overrides possible in options
* What is the html wrapper for math?

## HTML Wrappers:

I think in JupyterBook/sphinx currently this is done:

Without a label:
```html
<div class="math notranslate nohighlight">\[
\int_0^\infty \frac{x^3}{e^x-1},dx = \frac{\pi^4}{15}
\]</div>
```

With labels:
```html
<div class="math notranslate nohighlight" id="equation-eqn-best">
<span class="eqno">(2)<a class="headerlink" href="#equation-eqn-best" title="Permalink to this equation">¶</a></span>\[
e = mc^2
\]</div>
```

And inline:
```html
<span class="math notranslate nohighlight">\(x_{hey}=it+is^{math}\)</span>
```

choldgraf:
the guiding principle that I _think_ we've taken so far is to be as close to the Jupyter Notebook experience as possible. That way users aren't surprised when they do stuff w/ jupyter book vs. jupyter notebook. Does that help guide the decision-making here?

rowanc1:
Thanks @choldgraf - I think that makes sense.

I have gone with default `$`s at this stage - similar to jupyter, looks like Jupyter also picks up things with `\begin{equation}`. And we also want to pick up ` {math}`` ` as a role and the directives as well.

I will get the basic functionality in through #5, and then we can always come back and fix/add things.

I am looking through `markdown-it-texmath` and it assumes that we render with katex and we can't template, without a full overwrite of the rules in `markdown-it`, so that might be something to contribute back to or spin off into a small sub-package. Not sure on your thoughts on that.

choldgraf:
I don't have strong opinions on katex, not super familiar with it. I'm also really curious to see if MathJax 3 significantly improves on some of the downsides of MathJax (though it'll be a little while before we switch I assume, though I'll note that `master` on Sphinx now supports MathJax 3)

[...]

PR #5, opened 2020-07-31
✨  NEW: Markdown math

rowanc1:
- [x] Add equation numbers
- [x] Test targets of equation numbers (likely adding types)

See #4


rowanc1:
Basic equation numbering.

Note that the example doesn't yet render math.

image

choldgraf:
This sounds cool! Is there some way that I can test this out? Maybe we can set up a little sphinx build that loads the JS/HTML libraries for this?

rowanc1:
Put this together:

https://markdown-it-myst.glitch.me/

Won't have the math stuff in it, but the latest deployed version!

choldgraf:
awesome! (though in this case I specifically wanted to try out the new behavior of this PR myself.

for those of us that don't speak JS very well, do you think you could add a simple set of instructions for how to compile this library locally, and a little lightweight HTML file that lets us demo locally?

though I don't think this kinda stuff should block merging since this codebase is so young, so I'm happy if you wanna 🚢 it

rowanc1:
Sounds good. I will detail that in another PR!

Ran out of time tonight .. will finish Sunday!

rowanc1:
I will add a bit more to the readme around starting up the dev environment, it should be straight forward after you get [yarn installed](https://classic.yarnpkg.com/en/docs/install), on a mac - something like this:

```
brew install yarn
cd markdown-it-myst
yarn install
yarn start
```

Install the packages and start the server, there is also a `yarn test` in there that goes through all of the fixtures at the moment.

PR #6, opened 2020-08-02
📖  DOCS: Add developer install instructions

rowanc1:
Add install instructions, please modify or ask questions to get up and running and we can clarify the readme!


choldgraf:
woo - these worked for me!

choldgraf:
One thought - we could separate the three `dev` commands into three sections, and add more explanatory text about what to expect (rather than using in-line comments for them). E.g., I think that running `yarn start` also runs `yarn build` first? Or maybe not? Either way, just giving people a bit more idea of what to expect could help for a non-seasoned JS-er

rowanc1:
Cool, added a bit more explanation for each!

The `start` does an in-memory build, and isn't quite the same as the build command, I added this note to the docs.

choldgraf:
LGTM - thanks! 🎉

choldgraf:
oops - forgot to squash-merge...oh well. I just activate it as the default for our PRs

Issue #8, opened 2020-08-07
Parsing directive options


rowanc1:
Two supported syntaxes:

````
```{directive_name} arguments
---
option1: name
option2: |
    Longer text block
---
content...
```
````

or

````
```{directive_name} arguments
:option1: name
:option2: other
content...
```
````

Currently the library is not parsing this, leaving myself some pointers to the python code here:
https://github.com/executablebooks/MyST-Parser/blob/master/myst_parser/parse_directives.py#L100

rowanc1:
I have implemented the quick `:options:` not the yaml yet.

Still need to have some validation/type checking on them!

PR #10, opened 2020-11-20
✨ NEW: General improvements

rowanc1:
There are a number of general improvements and reorganizing that are in this PR as I get back into development. The PR as it stands touches most things as the library is still quite new.

I think it is worth pushing ahead on this and as we move closer to something that covers the basics of myst parsing/setup etc. we can start to move to more targeted commits. Perhaps we should move to a dev branch model? As this isn't really ready for a version bump either?

image

## New Features

- block breaks +metadata +++ {meta:data}
- inline comments % comment
- admonitions
- references to headers ()=
- figures, num ref and default reference names
- math and numbered math blocks
- equation numbers

## Still lacking

* The extension of new directives especially is pretty confusing, and it would be better to have more separation in the file structures as knowing which file to jump into is hard at the moment.
* Am not parsing yaml #8 although I did get a start at the other options. These are not typed or checked at all against directives, which they should be.
* There are a few too many passes on numbering, I think that needs to be simplified and centralized. I have started a pattern of putting the target on the `token.meta.target` but haven't pushed it all the way through yet.


src/roles.ts
```diff
@@ -74,10 +74,31 @@ const knownRoles: Record<string, RoleConstructor> = {
...
+      if (match == null) return { attrs: { ref: content }, content: '' };
+      const [, modified, ref] = match;
+      return { attrs: { ref: ref.trim() }, content: modified.trim() };
+    },
+    renderer: (tokens, idx, opts, env) => {
+      const token = tokens[idx];
+      const ref = token.attrGet('ref') ?? '';
+      if (!env.targets[ref]) {
+        return `<span style="background-color:red;" title="No reference '${escapeHtml(ref)}' found.">${escapeHtml(token.content || `"${escapeHtml(ref)}" reference not found!`)}</span>`;
+      }
```
rowanc1, line 96: Testing

choldgraf:
I love it - the sample page looks much nicer and the styling looks really good 👍 

Couple quick thoughts:

1. Personally I'd just merge this into master and keep iterating that way. We designate this repo as "pre-alpha" which IMO means that users can't expect *any* kind of continuity of functionality or API.
2. I like the red background for refs that don't resolve, though I worry that this may confuse some folks who are referencing a different page. Not saying it's a dealbreaker for the red highlighting, but wanted to note it.
3. I wonder if we could build some simple Sphinx docs for this package? The `index.html` could be hosted along with the Sphinx package if we wish, no? I guess we're gonna run into the same issues as https://github.com/executablebooks/thebe/pull/298 though :-/

rowanc1:
That all sounds good. In the middle of a bit of refactoring at the moment. Will push through on that and then merge it in.

I have changed it to an `error` class just now - so can play with the styling. Right now the references aren't well exposed, so difficult to create the external/cross-page refs. I will be working on that too. 👍 

Had the same thoughts for the sphinx docs, can look into some alternative ways of building the docs. Maybe take the build onto gh-actions and host on gh-pages? Not sure.

choldgraf:
🚀

PR #15, opened 2020-11-22
✨ NEW: Fence directives

rowanc1:
Now supports *not* parsing the internals of a directive, which is helpful for things like `math` directives.
Supported by the `skipParsing` option on the directive.

### DIFF
--- a/fixtures/directives.math.md
+++ b/fixtures/directives.math.md
@@ -1,0 +1,22 @@
+Math directive:
+.
+```{math}
+:label: my_label
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+```
+.
+<div class="math numbered" id="eq-my_label" number="1">
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+</div>
+.
+
+Math directive:
+.
+```{math}
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+```
+.
+<div class="math">
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+</div>
+.
--- a/src/directives/index.ts
+++ b/src/directives/index.ts
@@ -1,11 +1,12 @@
 /* eslint-disable no-param-reassign */
 import MarkdownIt from 'markdown-it';
+import Renderer from 'markdown-it/lib/renderer';
 import container from 'markdown-it-container';
 import Token from 'markdown-it/lib/token';
 import { RuleCore } from 'markdown-it/lib/parser_core';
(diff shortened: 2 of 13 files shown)


fixtures/directives.math.md
```diff
@@ -0,0 +1,22 @@
+Math directive:
+.
+```{math}
+:label: my_label
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+```
+.
+<div class="math numbered" id="eq-my_label" number="1">
+w_{t+1} = (1 + r_{t+1}) s(w_t) + y_{t+1}
+</div>
```
rowanc1, line 10: Right now the math doesn't add `\[` around it. I am not sure we actually want this? The alternative is to use something else to notify mathjax.

src/directives/index.ts
```diff
@@ -96,11 +142,20 @@ const numbering = (directives: Directives): RuleCore => (state) => {
...
+        const target = newTarget(state, name || label, directive.numbered);
+        token.meta.target = target;
+      }
+    }
+    if (token.type === DirectiveTokens.fence) {
+      const directive = getDirective(directives, token.attrGet('kind'));
+      const { name, label } = token.meta?.opts;
+      // Only number things if the directive supports numbering AND a name or label is provided
+      if (directive?.numbered && (name || label)) {
+        const target = newTarget(state, name || label, directive.numbered);
```
rowanc1, line 158: @choldgraf I am using `name` or `label` for things like equations a bit interchangably. The RST docs seem to say `name` is common, but EB suggests `label` for things like math? Let me know what the convention is!! This seems to work for now.

choldgraf: Hmmm - my personal understanding is that label is the "thing you reference" elsewhere (e.g. `myeq1`). E.g., the [rst quickstart](https://docutils.sourceforge.io/docs/user/rst/quickref.html) uses `label` quite a lot, very little "name". Where do you see `name` recommended in rst?

rowanc1: This is where I was going to:
https://docutils.sourceforge.io/docs/ref/rst/directives.html#math

rowanc1: The other place I have been looking is here:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#math

Which does say `label` -- missed that before.

rowanc1: image

rowanc1: Upon further conversation:

* Use name. Label is directive specific (e.g. equation).

rowanc1: See also https://github.com/sphinx-doc/sphinx/issues/8476

PR #33, opened 2021-07-06
Test out and move to new packages

rowanc1:
## Running notes (mostly looking at other packages):
* Have overridden default renderers at a higher level to keep the labelling/cross-references working.
* Ideally the defaults from this package can be removed and updated to sensible defaults below: https://github.com/executablebooks/markdown-it-amsmath/issues/1
* ~We are still escaping html, this should be passed in escaped: https://github.com/executablebooks/markdown-it-amsmath/issues/2~
* math_block_eqno --> math_block_label, however ideally they are just the same with meta information on the tags: https://github.com/executablebooks/markdown-it-amsmath/issues/3
* No longer need the "_end" tags which were created before
   * Is there any situation where these would be necessary --- if we parse anything inside of the math.
   * For example: https://curvenote.dev/article/equation (that might be a custom extension though)
* There is inconsistency between the AMS math package and the dollars -- ideally since we now control both, these can work the same.
* Ideally there are named metadata on the tokens that we can use (numbered, labelled) and this is documented.
* Ideally there is consistency with the math directive (TBD, but currently it is the same token)
* Update the css classes to deal with `math amsmath` etc

## Backwards incompatible:
- [ ] `math_block_eqno` --> `math_block_label` (ideally we just move straight to `math_block` instead)


rowanc1:
This is superseded by #35.

Issue #34, opened 2022-02-03
Repository Evolution


rowanc1:
## Background

Over the past few months we have slowly started pulling repositories out of this repository into adjacent repos, this repository originally had the goal of being what has now become `markdown-it-docutils` and the majority of the code has now been duplicated (with the exception of state management) and improved (e.g. recursive directive parsing). That repository now has the core `directives` and `roles` that support the myst spec, and there have been other repositories that support some aspects of math parsing, and other requirements to support myst (e.g. block breaks, comments, etc.).

## Repository Responsibilities

In discussion with @chrisjsewell and @fwkoch today, we are planning to change this repo into a container repository that has the following responsibilities:

* parse MyST into a AST. To start with this will be markdown-it token streams, but will be more formalized into mdast over time.
* serialize AST into a default HTML tokens
* expose an opinionated set of markdown-it plugins (e.g. that can be used in `jupyterlab-myst`, `myst-vs-code`, etc.)
* directly expose functionality to add new roles/directives that will be passed into docutils, as well as any other markdown-it plugins to add
* Handle cross-reference management through a plugin system. This will expose the ability to do cross-document reference resolution, but that is not a direct responsibility of this library.

In parallel to this @chrisjsewell and @fwkoch are looking into standardizing testing and serialization into a mdast spec, which will likely form the basis of many unit tests for "myst as a spec" that can be implemented in many different libraries/languages. Initial experimentation on this is happening in https://github.com/chrisjsewell/myst-spec and https://github.com/curvenote/schema/pull/44.

## Naming repositories
As a part of this, we discussed the fact that neither `markdown-it-docutils` nor `markdown-it-myst` were good names:
1. `markdown-it-docutils` is not actually doing the responsibility of docutils in python, which has a much wider scope. The scope of that repository is strictly parsing (using markdown-it) roles and directives. The base directives are those used in docutils/sphinx and likely common extensions to the jupyterbook ecosystem.
2. `markdown-it-myst` should be able to switch out the dependence on markdown-it and have no changes to the public interface (e.g. moving to https://unifiedjs.com/, which has more uptake in the MDX community etc.).

We discussed in our meeting today that we could live with (1) for now (another option thought of later is below); and (2) we should change the repository name to `mystjs`. `myst` is taken, and this nicely does not confuse it with the python implementation.

## Repository Dependencies:
* [markdown-it-docutils](https://github.com/executablebooks/markdown-it-docutils) - directives and roles
[...]

PR #35, opened 2022-02-08
Mystjs

rowanc1:
This change reflects #34 which changes this repository over to mystjs and will be a bit of a test branch while we get things over to the new package, and make sure all of the various packages work together nicely.

- As part of this, I have also moved to npm, to better reflect the other packages in use in this org
- Package has been renamed already
- markdown-it-myst has been deprecated on npm
- Merge in state management PR https://github.com/executablebooks/markdown-it-docutils/pull/14

@fwkoch I have taken a few of your changes into this branch. Feel free to push as you are adding and moving test around in the various sub repos.

- [x] Delete all roles
- [x] Delete all directives
- [ ] Move to myst-extra
- [ ] Update tests and any new HTML structures


tests/myst/commonmark.headings.yml
```diff
@@ -12,7 +12,7 @@ cases:
       ### Heading!
     html: |-
       <h3>Heading!</h3>
-  - title: CommonMark setext headings
```
fwkoch: This is not a typo: https://spec.commonmark.org/0.30/#setext-headings

rowanc1 (commented):
Saw one of your recent commits, I think spaces after figure numbering is a styling concern, and shouldn't be mdast.

fwkoch:
I think you are talking about here in the hast: https://github.com/executablebooks/mystjs/pull/35/files#diff-5f90869d685f1edebcfbaf78e1d1875b5c24e7e3eb13ac7d863d5d0fa04ba8d1R53 ? Makes sense to take it out if styling will deal with it.

src/mdast/tokensToMyst.ts
```diff
@@ -0,0 +1,506 @@
...
+    type: 'inlineCode',
+    noCloseToken: true,
+    isText: true,
+  },
+  hardbreak: {
+    type: 'break',
+    noCloseToken: true,
+    isLeaf: true,
+  },
+  link: {
```
fwkoch, line 123: We need to make sure any links like `[](my-ref) are converted to `contentReferences` if `my-ref` is an existing identifier.

I think we can just do this in state resolution: visit all `links`, if `url in state.targets`, change the node to `contentReference`.

tests/myst/roles.references.equations.yml
```diff
@@ -0,0 +1,101 @@
...
+      children:
+        - type: paragraph
+          children:
+            - type: text
+              value: 'see '
+            - type: contentReference
+              kind: eq
+              identifier: matrix
+              label: '  matrix  '
+              resolved: false
```
rowanc1, line 14: @fwkoch I don't think `resolved` is part of the AST? But are part of the transform section?


We can have multiple asts that are parsed, and then brought together.

fwkoch: Yeah, I agree with this - will fix it up.

PR #40, opened 2022-03-31
Use attrGet for node attributes

fwkoch:
Previously we pulled these from token.meta, but meta is no longer being blindly copied from the directive in `markdown-it-docutils`


fwkoch:
Closing as this is very stale.

PR #42, opened 2022-04-08
Add header numbering and save numbers to nodes during transform

fwkoch:
- `addNumbersToNodes` transform determines numbers for containers, math, and headers and adds them to the respective nodes
- header numbering follows `1`, `1.1`, `1.2`, `2` etc pattern based on header depth
- `numrefs` and `eq` refs are only resolved if node is numbered
- tests!

This is dependent on updates to `myst-spec` here:
* https://github.com/executablebooks/myst-spec/pull/36


src/mdast/tokensToMyst.ts
```diff
@@ -248,10 +248,17 @@ const defaultMdast: Record<string, Spec> = {
     type: 'container',
     getAttrs(token): Container {
       const name = token.meta?.name || undefined;
+      // TODO: Use unnumbered on container and eliminate this logic
+      let num;
+      if (token.meta?.numbered === false) {
```
rowanc1, line 253: Should we have `|| token.unnumbered` in the if statement here?

src/mdast/tokensToMyst.ts
```diff
@@ -248,10 +248,17 @@ const defaultMdast: Record<string, Spec> = {
...
+        num = false;
+      } else {
+        num = name ? true : undefined;
+      }
       return {
         kind: 'figure',
         ...normalizeLabel(name),
-        numbered: name ? true : undefined,
+        numbered: num,
```
rowanc1, line 261: unnumbered?

src/mdast/tokensToMyst.ts
```diff
@@ -266,10 +273,17 @@ const defaultMdast: Record<string, Spec> = {
...
+      if (token.meta?.numbered === false) {
+        num = false;
+      } else {
+        num = name ? true : undefined;
+      }
       return {
         kind: undefined,
         ...normalizeLabel(name),
-        numbered: name ? true : undefined,
+        numbered: num,
```
rowanc1, line 286: Flip logic here as well.

rowanc1 (commented):
Looks great. I think we can introduce warnings about missing heading numbers later, can we address a few of the TODOs now though. Otherwise looks solid!

rowanc1:
Nice, new changes look good. Will wait til monday morning for https://github.com/executablebooks/myst-spec/pull/36 to merge!

package-lock.json
```diff
@@ -76,6 +76,11 @@
         "typescript": "^4.6.3"
       }
     },
+    ".yalc/myst-spec": {
+      "version": "0.0.3",
+      "extraneous": true,
+      "license": "MIT"
+    },
```
rowanc1, line 83: Can you remove this!

rowanc1 (commented):
Looks good. Just cleaning up the package lock!

PR #43, opened 2022-04-08
Transform for converting html nodes to mdast

fwkoch:
This transform turns all HTML nodes in a tree to mdast, e.g.

```yaml
before:
  type: root
  children:
    - type: html
      value: <img src="example.jpg" title="example">
after:
  type: root
  children:
    - type: paragraph
      children:
        - type: image
          url: example.jpg
          title: example
          alt: ''
```


rowanc1:
Can you throw a table example in?!

PR #44, opened 2022-04-12
Add option to not update enumerator values when counting targets

fwkoch:
~Enumerator value is still saved on the state for reference resolution etc. but now it may be left off the target node.~

3 options to disable heading, figure, and equation enumeration. If enumeration is disabled, numbered references will not be resolved.

### DIFF
--- a/src/mdast/transforms.ts
+++ b/src/mdast/transforms.ts
@@ -7,18 +7,19 @@ import { remove } from 'unist-util-remove';
 import { map } from 'unist-util-map';
 import { Admonition, AdmonitionKind, GenericNode } from './types';
 import { admonitionKindToTitle, normalizeLabel } from './utils';
-import { State, enumerateTargets, resolveReferences } from './state';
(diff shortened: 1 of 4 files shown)


src/mdast/transforms.ts
```diff
@@ -7,16 +7,18 @@ import { remove } from 'unist-util-remove';
 import { map } from 'unist-util-map';
 import { Admonition, AdmonitionKind, GenericNode } from './types';
 import { admonitionKindToTitle, normalizeLabel } from './utils';
-import { State, addEnumeratorsToNodes, resolveReferences } from './state';
+import { State, enumerateTargets, resolveReferences } from './state';
 
 export type Options = {
   addAdmonitionHeaders?: boolean;
   addContainerCaptionNumbers?: boolean;
+  updateEnumeratorValues?: boolean;
```
rowanc1, line 15: enumerateFigures is often true, while headings can be false!

rowanc1 (approved):
Looks good from the read of the code!

PR #49, opened 2022-04-29
🧱 Handle html breaks within markdown text

fwkoch:
This pr only affects the `convertHtmlToMdast` transform. Markdown documents may have small html snippets inside them, e.g.:

```
# Example

Here is a break<br />in the middle of a paragraph.
```

When we convert these snippets to mdast, however, they are treated as full html documents, in which `html-util-to-mdast` strips leading/trailing `<br />` breaks. https://github.com/syntax-tree/hast-util-to-mdast/blob/50eda792602f44a6576226a4c14143c5fafb0c63/lib/all.js#L26-L44

This PR adds an option to `keepBreaks` which bypasses the underlying break behavior and directly translates `<br />` to a mdast `break` node. By default, `keepBreaks` is true.


rowanc1 (approved):
🚀

PR #52, opened 2022-06-20
Move from node to index lookup

rowanc1:
I was running into all sorts of bugs when I was running this on the jupyter-book documentation, this is one of them. I couldn't trace it down, but this seems to be equivalent.

image


fwkoch (approved):
Hmm, yeah, this seems fine. `findAfter` just looks up index from node, so handing it the index is more direct.

PR #53, opened 2022-06-21
🔗 Enable linkify in markdownit config

fwkoch:
Previously setting `linkify: true` in `MarkdownIt` options had no effect since it was disabled at the config level. Now, it's enabled at the config level (though still must be set to 'true' in options).

`config.ts` is copied from here: https://github.com/markdown-it/markdown-it/blob/12.3.2/lib/presets/commonmark.js and only `'linkify'` is added to `core.rules`

### DIFF
--- a/src/myst.ts
+++ b/src/myst.ts
@@ -1,4 +1,11 @@
 import MarkdownIt from 'markdown-it';
+import { Root } from 'mdast';
+import type { Plugin } from 'unified';
+import { unified } from 'unified';
+import rehypeStringify from 'rehype-stringify';
+import { directivesDefault, rolesDefault } from 'markdown-it-docutils';
+import { MARKDOWN_IT_CONFIG } from './config';
+import { formatHtml, mystToHast, tokensToMyst, transform, State } from './mdast';
 import {
   mathPlugin,
   convertFrontMatter,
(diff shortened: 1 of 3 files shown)


src/myst.ts
```diff
@@ -1,4 +1,11 @@
 import MarkdownIt from 'markdown-it';
+import { Root } from 'mdast';
+import type { Plugin } from 'unified';
+import { unified } from 'unified';
+import rehypeStringify from 'rehype-stringify';
+import { directivesDefault, rolesDefault } from 'markdown-it-docutils';
+import { MARKDOWN_IT_CONFIG } from './config';
```
fwkoch, line 7: This is the only change, the others are just re-order...

src/myst.ts
```diff
@@ -104,7 +105,7 @@ export class MyST {
 
   _createTokenizer() {
     const exts = this.opts.extensions;
-    const tokenizer = MarkdownIt('commonmark', this.opts.markdownit);
+    const tokenizer = MarkdownIt(MARKDOWN_IT_CONFIG as any, this.opts.markdownit);
```
fwkoch, line 108: The `MarkdownIt` type requires a `presetName` string, but passing a config JSON works. This is only slightly risky when we bump `markdown-it` version.

rowanc1 (approved):
Nice, looks good!

PR #62, opened 2022-10-11
Changes needed to support importing as an ESM module

wlach:
* Change the type of the package
* Link to compiled distribution in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "module": "dist/esm/index.js",
   "unpkg": "dist/index.umd.min.js",
   "types": "dist/types/index.d.ts",
+  "type": "module",
   "files": [
     "src",
     "dist"
@@ -16,7 +17,7 @@
   "exports": {
     ".": {
       "require": "./dist/index.umd.js",
-      "import": "./dist/esm/index.js"
+      "import": "./dist/index.esm.js"
     }
   },
   "keywords": [


package.json
```diff
@@ -9,14 +9,15 @@
   "module": "dist/esm/index.js",
   "unpkg": "dist/index.umd.min.js",
   "types": "dist/types/index.d.ts",
+  "type": "module",
```
wlach, line 12: This may have unwanted downstream impacts

wlach:
Some small things I needed to do to make https://github.com/wlach/mystjs-svelte work. I'm still far from being a node expert, so I'm not 100% sure if the `"type": "module",` part is free of unwanted downstream consequences. Probably could use testing.

rowanc1:
So cool. It may take a bit to test this out in all of the various places, are you able to move forward in your work without this being merged released in the next two weeks?

wlach:
> So cool. It may take a bit to test this out in all of the various places, are you able to move forward in your work without this being merged released in the next two weeks?

Sorry for late reply. Yup, this is extreme prototype code so am not in any rush.

rowanc1:
Hi @wlach, we are finally getting to some of these things. As part of the updates recently we removed the `exports` dict, which was causing some problems with the module loading. We also took some pieces of the module pointers which no longer point to the bundle.

I have not been able to get the `"type": "module"` working in downstream packages with out a ton of other changes.

I *think* that the current changes should be enough to improve your downstream experience (esm can consume cjs, but not the other way around) and I think that previously our ESM module was just wrong.

We are planning a new release in the coming days, and I would be curious if it fixes your use case! I will ping here when we have released.

rowanc1:
Hi @wlach -- I am going to close this, as it is now quite out of date!

Happy to fix up install issues if you find any in the future. We have gotten better at our base scripts, so hopefully less problems next time. :)

PR #67, opened 2022-11-01
Use jtex to validate docx templates

fwkoch:
See template here: https://github.com/myst-templates/docx_default


rowanc1:
Awesome, tested locally and the word templates work as expected. The default now has the myst logo!

PR #70, opened 2022-11-10
🐛 Remove dependence on @curvenote/schema

rowanc1:
This was a hangover from using prosemirror-docx as the starting place for myst-to-docx.


fwkoch (approved):
Nice! 🧹

PR #71, opened 2022-11-10
Include a bundle for mystjs that works with jest

rowanc1:
This fixes a downstream packaging issue with mystjs to allow it to work in an easier way with jest.

### DIFF
--- a/packages/citation-js-utils/package.json
+++ b/packages/citation-js-utils/package.json
@@ -4,7 +4,6 @@
   "description": "Utilities for citation-js.",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
-  "unpkg": "dist/index.umd.min.js",
   "types": "dist/types/index.d.ts",
   "files": [
     "src",
--- a/packages/mystjs/docs/Makefile
+++ b/packages/mystjs/docs/Makefile
@@ -7,13 +7,13 @@ fix-default-js:
 
 build:
 	cd ..; npm run build
-	cp ../dist/index.umd.min.js _static/myst.min.js
+	cp ../dist/index.cjs.js _static/myst.min.js
 	jupyter-book build .
 	make fix-default-js
 
 build-watch:
 	jupyter-book build .
-	cp ../dist/index.umd.min.js _build/html/_static/myst.min.js
+	cp ../dist/index.cjs.js _build/html/_static/myst.min.js
 	make fix-default-js
 
 watch: clean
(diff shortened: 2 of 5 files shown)


packages/citation-js-utils/package.json
```diff
@@ -4,7 +4,6 @@
   "description": "Utilities for citation-js.",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
-  "unpkg": "dist/index.umd.min.js",
```
rowanc1, line 7: These were copy-paste hangovers and were never built (for `citation-js-utils` and `jtex`)

packages/mystjs/package.json
```diff
@@ -33,11 +33,12 @@
     "test:watch": "jest --watchAll",
     "lint": "eslint \"src/**/*.ts\" -c .eslintrc.js",
     "lint:format": "prettier --check src/*.ts src/**/*.ts",
-    "build:bundle": "esbuild src/index.ts --bundle --outfile=dist/index.umd.min.js --platform=browser",
+    "build:bundle:cjs": "esbuild src/index.ts --bundle --outfile=dist/index.cjs.js --platform=node",
+    "build:bundle:browser": "esbuild src/index.ts --bundle --outfile=dist/index.umd.min.js --platform=browser --minify",
```
rowanc1, line 37: This is the main change, I have added a CJS dependency, as well as actually minified the umd build. There likely needs to be more testing on the mystjs for the browser as I am not convinced that it will work, but this can go in now as it is an improvement.

rowanc1 (commented):
This is an addition of the cjs build that works with Jest, I think there is some more testing to ensure the browser bundle works in a follow up PR.

packages/mystjs/docs/Makefile
```diff
@@ -7,13 +7,13 @@ fix-default-js:
 
 build:
 	cd ..; npm run build
-	cp ../dist/index.umd.min.js _static/myst.min.js
+	cp ../dist/index.cjs.js _static/myst.min.js
```
fwkoch, line 10: Why is the `cjs` build used here, instead of the browser build?

fwkoch (approved):
All the builds look good - just one question regarding the makefile

packages/mystjs/docs/Makefile
rowanc1: Ah, yeah, I flipped that back and forth and didn't check that. Will fix now.

PR #72, opened 2022-11-11
💇‍♀️ Modify extractPart: support multiple parts, no longer support tags

fwkoch:
`extractParts` is used when building exports for tex templates; if a template requires a specific `part` (e.g. "abstract"), this function finds blocks corresponding to that part and pulls it out of the mdast tree.

Previously, this function looked at metadata `tags` and `part`. If either of those included the desired part, the block was removed.

Now, this function only looks at metadata `parts`. Also, it no longer removes the block from the source tree if there are multiple `parts` (this means a single block could potentially populate multiple parts of a tex template - certainly an edge case, but better than deleting a block when additional `parts` metadata is still present and unused).

### DIFF
--- a/packages/myst-common/src/extractParts.ts
+++ b/packages/myst-common/src/extractParts.ts
@@ -9,10 +9,9 @@ import { copyNode } from './utils';
  */
 export function selectBlockParts(tree: Root, part: string) {
   const blockParts = selectAll('block', tree).filter((block) => {
-    if (!block.data?.tags && !block.data?.part) return false;
-    if (block.data?.part === part) return true;
+    const parts = (block.data?.parts as string[]) ?? [];
     try {
-      return (block.data.tags as any).includes(part);
+      return parts.includes(part);
     } catch {
       return false;
     }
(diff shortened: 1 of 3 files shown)


rowanc1:
In conversation with @fwkoch, I think that we should stick to a single `part` (i.e. `abstract`) and not allow a block to have multiple `parts`. I can't think of a use case where a single block is repeated in different places (in abstract and acknowledgements, for example). We can always walk that choice back in the future, but I think staying simpler for now is better.

I will clean up the PR and get it in!

packages/myst-common/src/extractParts.ts
```diff
@@ -9,10 +9,9 @@ import { copyNode } from './utils';
  */
 export function selectBlockParts(tree: Root, part: string) {
   const blockParts = selectAll('block', tree).filter((block) => {
-    if (!block.data?.tags && !block.data?.part) return false;
```
rowanc1, line 12: I think that this is actually the only line of this PR that we want to retain, deprecating being able to control the part of the document using a tag. I think it easier to create a new PR and close this one @fwkoch.

PR #76, opened 2022-11-11
💇‍♀️ Modify extractPart: no longer support tags

rowanc1:
Replaces #72


fwkoch:
Thanks, this looks great. I love `part` instead of `parts`.

PR #85, opened 2022-11-22
🌠 Add transform to CLI to convert svg/gif images to png on export

fwkoch:
SVGs and animated GIFs are not supported in PDF/Word exports. This adds a transform to convert these image types to PNG on export.

SVG transformation requires [imagemagick](https://imagemagick.org/index.php) or [inkscape](https://inkscape.org/) CLI; GIF transformation requires imagemagick. If these libraries are not available, source images will be unmodified.

### DIFF
--- a/packages/myst-cli/src/build/docx/single.ts
+++ b/packages/myst-cli/src/build/docx/single.ts
@@ -130,6 +130,8 @@ function defaultWordRenderer(
     vfile,
     {
       getImageBuffer(image: string) {
+        // This extra read somehow prevents an error when buffer-image-size tries to get image dimensions...
+        fs.readFileSync(image);
         return Buffer.from(fs.readFileSync(image).buffer);
       },
       useFieldsForCrossReferences: false,
--- a/packages/myst-cli/src/transforms/images.ts
+++ b/packages/myst-cli/src/transforms/images.ts
@@ -11,6 +11,8 @@ import {
   addWarningForFile,
   computeHash,
   hashAndCopyStaticFile,
+  imagemagick,
+  inkscape,
   isUrl,
 } from '../utils';
 import type { ISession } from '../session/types';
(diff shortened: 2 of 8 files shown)


packages/myst-cli/src/build/docx/single.ts
```diff
@@ -130,6 +130,8 @@ function defaultWordRenderer(
     vfile,
     {
       getImageBuffer(image: string) {
+        // This extra read somehow prevents an error when buffer-image-size tries to get image dimensions...
+        fs.readFileSync(image);
```
rowanc1, line 134: How frustrating! Super super weird as well.

rowanc1 (commented):
Looks great, love the cleanup you did to merge the various image formats to a single walk of the tree.

Does this do the logic for `svg --> pdf` for tex outputs and `svg --> png` for word yet?

I couldn't find that logic?

packages/myst-cli/src/transforms/images.ts
```diff
@@ -147,6 +151,143 @@ export async function transformImages(
...
+) {
+  const images = selectAll('image', mdast) as GenericNode[];
+  // Ignore all images with supported extension types
+  const validExts = opts?.imageExtensions ?? DEFAULT_IMAGE_EXTENSIONS;
+  const unsupportedImages = images.filter((image) => {
+    return !validExts.includes(path.extname(image.url));
+  });
+  if (unsupportedImages.length === 0) return;
+
+  // Gather unsupported SVG & GIF images that may be converted to supported types
```
fwkoch, line 171: All this logic of deciding which unsupported image types may be converted to a different supported type (and whether to use imagemagick or inkscape) is already quite complex and bloated. Likely any additional format handling will require us to refactor this...

fwkoch (commented):
Alright - I added `imageExtensions` options to `transformMdast`. Nos we can warn on unknown/unsupported image extensions as well as make better decisions about what formats to transform into (e.g. svg to pdf vs svg to png).

PR #86, opened 2022-11-22
👩‍🔬 Initial JATS implementation for body content

rowanc1:
This implements JATS for most body content, and tests. This is purely additional content, so safe to add.

In action, this allows real-time editing of JATS.
image

The _major_ parts of JATS will be the `front` and `back` sections, which will be dealt with in a different PR. Because of this we are not doing terms, citations, or footnotes in this pass. Otherwise this is complete and tested.


packages/myst-to-jats/tests/basic.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+describe('Basic JATS', () => {
+  test.each(cases.map((c): [string, TestCase] => [c.title, c]))('%s', (_, { tree, jats }) => {
+    const pipe = unified().use(mystToJats);
+    pipe.runSync(tree as any);
+    const vfile = pipe.stringify(tree as any);
+    expect((vfile.result as JatsResult).value).toEqual(jats);
+  });
+});
+
+// describe('myst-to-jats', () => {
```
rowanc1, line 34: Should remove this.

packages/myst-to-jats/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  text?: string;
+  cdata?: string;
+  attributes?: Attributes;
+  elements?: Element[];
+};
+
+/**
+ * https://jats.nlm.nih.gov/archiving/tag-library/1.3/attribute/ref-type.html
+ */
+export enum RefType {
```
rowanc1, line 18: I think these will likely move to a jats utility library in the future.

packages/myst-to-jats/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  'table' = 'table',
+  /** Table footnote */
+  'table-fn' = 'table-fn',
+}
+
+export type GenericNode = {
+  type: string;
+  value?: string;
+  children?: GenericNode[];
+} & Record<string, any>;
```
rowanc1, line 73: It would be nice to have these in a better place than mystjs -- allowing overriding types that are forgiving. @fwkoch not sure where you think this would be best to be put?

packages/myst-to-jats/src/transforms/definitions.ts
```diff
@@ -0,0 +1,35 @@
+import type { Plugin } from 'unified';
+import type { Root } from 'mdast';
+import type { Parent } from 'myst-spec';
+import { selectAll } from 'unist-util-select';
+
+export type DefinitionList = Parent & { type: 'definitionList' };
+export type DefinitionItem = Parent & { type: 'definitionItem' };
+export type DefinitionTerm = Parent & { type: 'definitionTerm' };
+export type DefinitionDescription = Parent & { type: 'definitionDescription' };
```
rowanc1, line 9: I think we need a staging ground for mystjs-used types that are not in spec yet. These are copied from a few other places, and it would be nice to be consistent. @fwkoch thoughts on a library for that?

rowanc1 (commented):
Initial implementation with a few notes.

packages/myst-to-jats/src/transforms/definitions.ts
fwkoch: Yeah, that would be good.

My first thought is to just chuck it into a folder in `myst-common`... but having a separate package is probably nicer so we can reference it independently (i.e. in `myst-spec` readme, MEPs, etc). `myst-spec-incubator` `myst-spec-proposals` `myst-spec-alpha`...

packages/myst-to-jats/src/types.ts
fwkoch: _This_ feels more like a `myst-common` thing...

fwkoch (approved):
This is really nice. Great to have something working, independent of more nuanced discussion/changes around `front` and `back`.

PR #88, opened 2022-11-23
🎇 Improve fallback when inkscape is not available

rowanc1:
This is to fix an error when `inkscape` is not available, but `imagemagik` is. We now fallback from PDF --> PNG, rather than error.

### DIFF
--- a/packages/myst-cli/src/transforms/images.ts
+++ b/packages/myst-cli/src/transforms/images.ts
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import type { Root } from 'mdast';
 import mime from 'mime-types';
-import type { GenericNode } from 'mystjs';
+import type { GenericNode } from 'myst-common';
 import { selectAll } from 'unist-util-select';
 import fetch from 'node-fetch';
 import path from 'path';
@@ -16,6 +16,7 @@ import {
   isUrl,
 } from '../utils';
 import type { ISession } from '../session/types';
+import chalk from 'chalk';
 
 const DEFAULT_IMAGE_EXTENSIONS = ['.png', '.jpg', '.jpeg'];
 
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/transforms/images.ts
```diff
@@ -220,13 +221,24 @@ export async function transformImageFormats(
       session.log.info(`${logPrefix} PNG using inkscape`);
       svgConversionFn = inkscape.convertSvgToPng;
     } else if (svgToPng && imagemagickAvailable) {
+      if (svgToPdf) {
+        addWarningForFile(
```
fwkoch, line 225: Nice, this is good. I think there's an edge case where only `pdf` is in the allowed image formats and not `svg` or `png` - so in that specific case we are converting from one unsupported format to another (when we might just prefer to leave the original unsupported `svg`).

packages/myst-cli/src/utils/inkscape.ts
```diff
@@ -1,9 +1,30 @@
 import fs from 'fs';
 import path from 'path';
 import { sync as which } from 'which';
+import type { LoggerDE } from 'myst-cli-utils';
 import { makeExecutable } from 'myst-cli-utils';
 import type { ISession } from '../session/types';
 
+function createInkscpapeLogger(session: ISession): LoggerDE {
```
fwkoch, line 8: 👍

fwkoch (approved):

packages/myst-cli/src/transforms/images.ts
rowanc1: Hm yeah, good point. I am good to kick that down the road, as it is a pretty unlikely scenario.

Issue #84, opened 2022-11-22
jtex: Allow for nested files structure


kai-tub:
### Context

Hey, I would like to extract the individual issues discussed in https://github.com/kai-tub/latex-beamer-pure-minimalistic/discussions/75 for clarity and easier communication.

As discussed in the thread, I am more than happy to try to give my template first-class support for mystjs/jtex.
As part of the first trial, I would strongly prefer not to be forced to have a flat file hierarchy when creating the template.
My workflow would be as follows:
- Create the `template.tex` and `template.yml` files at the root of my _template_ repository
- Create a git submodule of my actual beamter repository, as this guarantees to have the source _fixed_ and keep the myst-interface separate from the beamer template itself

My reasoning is as follows:
- My CI/CD pipeline for the beamer presentation would have to be updated to _exclude_ changes to the `template.tex` to trigger updates/releases/tests and I don't think that this is helpful because the repository contains the _theme_ and I think a template for something other users aren't interested in using, just adds noise and additional complexity that isn't necessary
- I am the maintainer of the beamer presentation theme and would also be the maintainer of the template, but I don't think that it is reasonable to assume that this will be the case for all other nice-looking beamer themes.
  - Of course, one can always _fork_ the original theme and inject the template code and maintain that separately but it would duplicate code that could easily become out-of-sync and produce confusing results, as using the theme "directly" might produce different outputs than the template itself if the sources are different 
  - This would also happen with my suggestion, but then it would be clear that the template _just_ needs to point to the newest commit of the theme
- It makes it easier to keep separate changelogs/releases for both components
  - If I update the theme interface to support more options, I would need an additional way of communicating these changes _only_ to the template users 
 - Last practical reason: In my theme, I 'utilize' folders to keep logos in a separate directory. So I would already _need_ support for folders to correctly render my template. This would also be a change I wouldn't be happy to do, because it adds noise to the folder structure and might be confusing for future template creators

### Proposal

Allow the `template.tex` and `template.yml` files to be in a separate directory from the other theme components.
Allow files to be in separate directories to not restrict the 'expected' structure of themes.

Please let me know if I should further clarify my points :)

fwkoch:
Hey @kai-tub - thanks for breaking this out into a bite-sized issue!

[...]

PR #89, opened 2022-11-24
🗃 Allow template.yml#files to be nested

fwkoch:
Addresses #84 

Files defined in the `template.yml` may now be nested in arbitrary subfolders.

- `template.tex` must still be at the top level of the template directory
- Any other files may be defined with relative unix-style paths
- Folders and wildcard patters are still not allowed in `files`

Example:
```
jtex: v1
...
files:
  - template.tex
  - logos/logo.png
  - styles/myst.sty
```

### DIFF
--- a/.changeset/old-pants-nail.md
+++ b/.changeset/old-pants-nail.md
@@ -1,0 +1,5 @@
+---
+'jtex': patch
+---
+
+Allow nested files in template.yml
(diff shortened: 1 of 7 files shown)


rowanc1 (approved):
Nice, this is great -- enforces us to put things in the files list or they aren't going to be copied!

PR #93, opened 2022-11-25
🧙‍♂️ Improve source-maps for development

rowanc1:
`option-click` now brings you to the source location rather than the type


fwkoch (approved):
Don't fully get `declarationMap` vs `sourceMap, but if it improves linking in dev, that's great. PR looks good.

rowanc1:
The `.d.ts.map` is moved to the types folder, which is what is picked up in vscode!

image

PR #95, opened 2022-11-25
Library for reading `JATS`

rowanc1:
Types and utilities for working with JATS XML documents in Node and Typescript.

Read and write JATS XML from node or see summries from the command line.

To use from the command line, use the `-g` to create a global install, which will provide a `jats` CLI:

```
npm install -g jats-xml
jats -v
```

## What is JATS?

JATS is a NISO standard for Journal Article Tags Schema, which is a way to define the XML structure of a scientific article semantically. This includes the `front`-matter (authors, funding, title, abstract, etc.), the `body` of the article (sections, figures, equations, tables, etc.), and `back`-matter (references, footnotes, etc.). The JATS can also contain `sub-articles`.

The standard documents are hosted by the NIH <https://jats.nlm.nih.gov/>. There are three flavours, this library currently uses in most cases the most precriptive tag set (for article authoring). Another helpful resource is <https://jats4r.org/>, which provides other examples and recomendations for JATS.

Note that most publishers do **not** provide the XML as a first class output - they should, it is an important part of open-science to have the content programatically accessible and interoperable. It is only [FAIR](https://www.go-fair.org/fair-principles/) 😉.

## From the command line

Commands available:

`download`: attempt to find the JATS file and download it locally.

```bash
jats download https://docs.python.org/3.7 article.jats
```

Note, currently this just downloads the XML, **not** the associated files.

`sumamry`: summarize the contents of the JATS, given a URL, DOI, or local file

```bash
jats summary https://elifesciences.org/articles/81952
jats summary 10.1371/journal.pclm.0000068
jats summary /local/article.jats
```

This will provide a summary, including a list of what the JATS file contains.

## Working in Typescript

All tags are accessible as types/enums. There is also documentation from each node-type

```typescript
import { Tags } from 'jats-xml';
Tags.journalId;
```

## Reading JATS in Node

```typescript
import 'fs' from 'fs';
import { Inventory, toDate } from 'jats-xml';
import { toText } from 'myst-common';
import { select, selectAll } from 'unist-util-select';
const data = fs.readFileSync('article.jats').toString();
const jats = new JATS(data);
// Easy access to properties
jats.doi
jats.body // A tree of the body (or front/back)
toDate(jats.publicationDate) // as a Javascript Date object
select('[id=fig1]', jats.body) // select a figure by an ID
selectAll('fig', jats.body) // Or selectAll figures
```


packages/jats-xml/src/cli/parse.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+  if (checkdoi.validate(file) || isUrl(file)) {
+    const data = await downloadJatsFromUrl(session, file, DEFAULT_RESOLVERS);
+    session.log.debug(toc(`Downloaded and parsed JATS file in %s`));
+    return new Jats(data);
+  }
+  throw new Error(`Could not find ${file} locally, and it doesn't look like a URL or DOI`);
+}
+
+function formatLongString(data: string, offset = 0, length = 88 - offset): string {
```
rowanc1, line 56: These utilities are maybe throw away or something we can bring to `myst-cli-utils`. Might be worth looking into some slightly better CLI tools for formatting.

packages/jats-xml/src/download.ts
```diff
@@ -0,0 +1,37 @@
+import checkdoi from 'doi-utils';
+import fetch from 'node-fetch';
```
rowanc1, line 2: There might be a bit extra to do here to override fetch to make this work client side.

packages/jats-xml/src/resolvers.ts
```diff
@@ -0,0 +1,58 @@
+import checkdoi from 'doi-utils';
+import fetch from 'node-fetch';
+import type { ISession } from './types';
+
+export interface Resolver {
+  test: (url: string) => boolean;
+  jatsUrl: (url: string) => string;
+}
```
rowanc1, line 8: Right now only have three resolvers, I think we should follow the advice of Alf and look to cross ref and see if it is there.

https://twitter.com/invisiblecomma/status/1596095274828824576

> If the publisher has deposited a URL for the article XML with Crossref, yes:
> https://www.biglist.com/lists/lists.mulberrytech.com/jats-list/archives/202204/msg00009.html
> https://www.biglist.com/lists/lists.mulberrytech.com/jats-list/archives/202204/msg00010.html

That can probably happen in a next step.

packages/jats-xml/src/session.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { Logger } from 'myst-cli-utils';
+import type { ISession } from './types';
+
+export class Session implements ISession {
+  API_URL = 'https://api.myst.tools';
+  log: Logger;
+  constructor(opts?: { logger?: Logger }) {
+    this.log = opts?.logger ?? chalkLogger(LogLevel.debug);
+  }
+}
```
rowanc1, line 11: This should be cleaned up, don't need that URL in the Session.

packages/jats-xml/src/types/refType.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * https://jats.nlm.nih.gov/archiving/tag-library/1.3/attribute/ref-type.html
+ */
+export enum RefType {
```
rowanc1, line 4: Need to do a better job at auto-creating these from the schema. That is worth a bit of time I think, but out of scope for this PR!

rowanc1 (commented):
Some comments to clean up!

packages/jats-xml/src/cli/parse.ts
```diff
@@ -0,0 +1,201 @@
...
+import { select, selectAll } from 'unist-util-select';
+import { downloadJatsFromUrl } from '../download';
+import { DEFAULT_RESOLVERS } from '../resolvers';
+import { findDoi, formatDate, toDate } from '../utils';
+
+function hasValidExtension(output: string) {
+  return ['.xml', '.jats'].includes(extname(output).toLowerCase());
+}
+
+async function downloadJats(session: ISession, urlOrDoi: string, output: string): Promise<string> {
```
fwkoch, line 21:
```suggestion
async function downloadAndSaveJats(session: ISession, urlOrDoi: string, output: string): Promise<string> {
```

Maybe? I just got a little lost interpreting the scope of the several different `download` functions

packages/jats-xml/src/cli/parse.ts
```diff
@@ -0,0 +1,201 @@
...
+  if (fs.existsSync(urlOrDoi)) {
+    throw new Error(`File "${urlOrDoi}" is local and cannot be downloaded!`);
+  }
+  if (!(checkdoi.validate(urlOrDoi) || isUrl(urlOrDoi))) {
```
fwkoch, line 25: This check also happens in `downloadJatsFromUrl`, can we delete it here?

fwkoch (approved):
Very cool. Looks great and easy to understand.

packages/jats-xml/src/download.ts
fwkoch: What do you mean here - What wouldn't work?

rowanc1: I am not sure if node-fetch is designed for the browser. Will check and see if it fails.

packages/jats-xml/src/cli/parse.ts
rowanc1: Nice.

packages/jats-xml/src/cli/parse.ts
rowanc1: Yep, done.

packages/jats-xml/src/session.ts
rowanc1: Done.

Issue #90, opened 2022-11-24
Allow multiple templates for MyST websites


rowanc1:
## Problem

Currently there is only a single site template for mystjs, which is currently the curvenote [book-theme](https://github.com/curvenote/book-theme). We want to generalize and upstream that template (or most parts of it at least), and allow for multiple different templates to exist.

This work should ideally support the "template-er" to be written in any language (e.g. python or javascript), which works on a structured data directory with all assets needed. The tempalte-er can then take that structured data and create a website (e.g. static html, or a more complex application like the current themes that use remix to serve/transform the data on the fly).

## Goals

- Get the base remix theme over to executablebooks, these will be opinionated and can be used in react components for other remix themes (maybe generalizing parts to support any react themes, e.g. nextjs)
- Have an initial theme in EBP that follows the JupyterBook theme
- Use and extend the `template.yml` syntax developed for static exports to create structured data
- Allow the exports to be picked up in the static site folder, so we can download PDFs etc from the site
- Consider and study the current Sphinx templates to see if we can use that template ecosystem easily (or design it in a way that we can pull on it in the future, including all sorts of design extensions).

## Considerations

The current `SiteConfig` has all sorts of design pieces in it at the moment that are defined centrally (currently in @curvenote/site-common), these should instead be defined by the `template.yml` (e.g. `hide_header_block` or `analytics`). I think we can suggest some base things or naming conventions, and/or provide ways in the future to inherit these (mixins), but all of the template data/customizations should be defined in the template. However, because these are structured, they can still be *validated* by myst, which means nice error messages, etc. when using the CLI.

## Creating a template server

The template should define how to start the service on the command line (e.g. `npm run start`) which should provide a watch task. MyST will providing the content server (serves structured JSON) and a socket on that server that pushes information about when to reload/rebuild the site. Other than those two constraints, I don't think there is too much else we are putting on the template server/process, which should leave things wide open for many different templates.

cc-ing @echarles @choldgraf who were interested as @fwkoch and I start to move forward on this in the next few weeks. Would love any questions/thoughts to prompt more of our thinking into the open. 🚀

echarles:
Thank you for starting the discussion. To change the theme, I have been hacking that line https://github.com/curvenote/curvenote/blob/6827be474a1760017280c2d10dab9f989a356e6a/packages/curvenote/src/web/index.ts to clone another repository and it worked OK.

[...]

PR #99, opened 2022-11-29
🌎 Prepare for site templating

fwkoch:
- Jtex may now clone templates (in addition to download/unzip)
- Site frontmatter / config modified:
  - Removed: analytics, design, twitter, logo, logo_text
  - Added: authors, github
- Site config may now specify arbitrary template options
- Projects / pages no longer inherit site frontmatter

Context:
- #90


packages/jtex/src/download.ts
```diff
@@ -7,18 +7,21 @@ import unzipper from 'unzipper';
...
 export enum TemplateKinds {
   tex = 'tex',
   docx = 'docx',
+  site = 'site',
 }
 
 const DEFAULT_TEMPLATES = {
   tex: 'tex/myst/curvenote',
   docx: 'docx/myst/default',
+  site: 'site/myst/default',
```
fwkoch, line 24: This does not exist yet.

packages/jtex/src/download.ts
```diff
@@ -47,13 +50,21 @@ function defaultUrl(session: ISession, template: string) {
...
   const subdirs: string[] = [];
   if (buildDir) subdirs.push(buildDir);
-  subdirs.push(
-    'templates',
-    hash ? createHash('sha256').update(template).digest('hex') : join(...template.split('/')),
-  );
+  subdirs.push('templates');
+  if (hash) {
+    if (kind) subdirs.push(kind);
+    subdirs.push(createHash('sha256').update(template).digest('hex'));
```
rowanc1, line 64: What is this hash for? Is this where the template is downloaded to?

packages/jtex/src/download.ts
```diff
@@ -122,12 +133,28 @@ function unnestTemplate(path: string) {
...
+    downloadUrl = templateUrl;
+  } else {
+    downloadUrl = await fetchTemplateDownloadLink(session, opts);
+  }
+  if (downloadUrl.endsWith('.git')) {
+    await cloneTemplate(session, downloadUrl, opts);
+  } else {
+    await downloadAndUnzipTemplate(session, downloadUrl, opts);
+  }
+  session.log.info(`💾 Saved template to path ${templatePath}`);
```
rowanc1, line 152: 👍

packages/myst-config/src/site/types.ts
```diff
@@ -23,33 +23,19 @@ export interface SiteAction extends SiteNavPage {
...
-    'projects',
-    'nav',
-    'actions',
-    'domains',
-    'twitter',
-    'logo',
-    'logo_text',
-    'favicon',
-    'analytics',
-  ].concat(SITE_FRONTMATTER_KEYS),
```
rowanc1, line 42: Do these getting removed effect the downstream stuff right away?

packages/myst-config/src/site/validators.ts
```diff
@@ -153,35 +141,48 @@ export function validateSiteConfigKeys(
...
 
+/**
+ * Validate and return common, non-template attributes of SiteConfig
+ */
 export function validateSiteConfig(input: any, opts: ValidationOptions) {
   const value = validateObjectKeys(input, SITE_CONFIG_KEYS, {
     ...opts,
     returnInvalidPartial: true,
+    // Template options will appear as extra keys here; no need for warnings.
+    suppressWarnings: true,
```
rowanc1, line 161: Do the key/values also get dropped?

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -32,19 +32,23 @@ import type {
...
 } from './types';
 
-export const SITE_FRONTMATTER_KEYS = ['title', 'description', 'venue', 'keywords', 'design'];
-export const PROJECT_FRONTMATTER_KEYS = [
+export const SITE_FRONTMATTER_KEYS = [
+  'title',
+  'description',
   'authors',
+  'venue',
+  'github',
```
rowanc1, line 42: Are we removing these from the site? Or maybe I am not sure about what `SITE_FRONTMATTER_KEYS` is?

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -153,24 +149,6 @@ export function validateVenue(input: any, opts: ValidationOptions) {
   return output;
 }
 
-export function validateSiteDesign(input: any, opts: ValidationOptions): SiteDesign | undefined {
```
rowanc1, line 156: Nice, these are all template options now!

rowanc1 (commented):
This looks great - a few questions!

packages/myst-config/src/site/validators.ts
fwkoch: Yes, that is why we need the other `getSiteTemplateOptions` function, to get all the non-explicit keys.

packages/jtex/src/download.ts
fwkoch: Yes, if the `template` specified is a URL, it is downloaded to a URL hash folder (rather than trying to extract a nice-looking folder path from an arbitrary URL).

We could have like a special case for github urls or something, if that gets matched, extract `org/repo`? But there needs to be some fallback that works for all URLs.

packages/myst-config/src/site/types.ts
fwkoch: The effects of removing these could be:
- Validation errors - I think these will not happen because site config now allows any arbitrary key/value template options
- Fields getting ignored in site build - Need to be a little careful here. Logo, logo_text, twitter can all be added as template options, so those should continue to work as before once everything is said and done. Analytics is a structured field so we may need some backwards compatibility massaging to pull that into template options... (`analytics.google` -> `analytics_google`) or whatever.

packages/myst-frontmatter/src/frontmatter/validators.ts
fwkoch: `SITE_FRONTMATTER_KEYS` is just a list of the keys defined on the `SiteFrontmatter` type. The only change here is _adding_ `authors` and `github` to site frontmatter.

fwkoch: Corresponding type change here: https://github.com/executablebooks/mystjs/pull/99/files#diff-0efda934c1ea585a450b89b6d34781b6e78c12ada734af7d546f40970eb1730eR96-R102

packages/jtex/src/download.ts
rowanc1: Ah ok, clever, yeah that is totally fine. People won't be looking in that folder I don't think, so this should be fine!

packages/myst-config/src/site/types.ts
rowanc1: Nice. Ok, thanks for the explanation.

packages/myst-config/src/site/validators.ts
rowanc1: Sounds good. Didn't pick that up the first time.

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -93,31 +93,22 @@ export type Export = {
...
   title?: string;
   description?: string;
+  authors?: Author[];
   venue?: Venue;
+  github?: string;
   keywords?: string[];
-  design?: SiteDesign;
 };
 
 export type ProjectFrontmatter = SiteFrontmatter & {
```
rowanc1, line 105: It might be clearer at some point to break the type inheritance here?

packages/myst-frontmatter/src/frontmatter/validators.ts
rowanc1: Ok, that is much clearer. I left a note about maybe changing the type structure (not sure if the validators would also be clearer if they were a bit more repeated rather than inherited now that site --> project isn't a thing?)

rowanc1:
Ok this is looking good, a few optional things, but overall looks great to merge. As discuss offline, we want to merge the jats work first and do a release to fix a `myst-transforms` issue (due to `doi-utils`).

rowanc1 (commented):
One thing I am noticing now is that the build site requires the template (to know if things are files etc to copy.).

I wonder if there is a lite way to do this so as to keep CI tasks fast. (just need the template.yml, not the whole template).

PR #103, opened 2022-11-30
🚀 jats-to-myst

rowanc1:


packages/jats-to-myst/src/transforms/typography.ts
```diff
@@ -0,0 +1,65 @@
...
+ * ```
+ * [ ..., {paragraph: [child1, math1, child2, child3, math2]}, ... ]
+ * [ ..., {paragraph: [child1]}, math1, {paragraph: [child2, child3]}, math2, ... ]
+ * ```
+ *
+ * @param tree
+ * @param parentNodeType The parent (e.g. a paragraph)
+ * @param childNodeType The child (e.g. a list or math)
+ */
+function unnestTransform(tree: GenericParent, parentNodeType: string, childNodeType: string) {
```
rowanc1, line 21: This is something that should eventually be reused in other places, came from myst-transforms.

packages/jats-to-myst/src/index.ts
```diff
@@ -0,0 +1,536 @@
...
+    case RefType.table:
+      return 'table';
+    case RefType.custom:
+      return undefined;
+  }
+}
+
+type Attributes = Record<string, any>;
+
+const handlers: Record<string, Handler> = {
```
rowanc1, line 32: Lots of things still to turn on here, but the following is going quite well on much of the JATS I have found.

rowanc1 (commented):
Some notes.

fwkoch (approved):
🚀 Looks great!

packages/jats-to-myst/src/transforms/typography.ts
rowanc1: I think we should move this over to myst-transforms and deduplicate, then it should be good to go!

PR #110, opened 2022-12-19
Introduce a latex parser/converter

rowanc1:
This allows for parsing of latex (using `unified-latex`) and then conversion to mdast. There is a fair bit of interpretation of the latex token stream --> sensible choices for the AST. Want to get the frontmatter a bit better and tables, then should be good for a v1!


fwkoch (approved):
I think this looks great! Cool how tiny the commit is to actually add tex support to the build.

packages/myst-cli/src/build/init.ts
```diff
@@ -8,24 +8,24 @@ import type { ISession } from '../session';
...
-  # license: 
+  # license:
   # doi:
   # date:
   # index:
   # subject:
   keywords: []
   authors: []
-  bibliography: []
+  # bibliography: []
```
rowanc1, line 24: Commenting this out means that we actually look for them by "default".

packages/myst-cli/src/transforms/images.ts
```diff
@@ -280,17 +292,16 @@ export async function transformImageFormats(
...
-    ];
-    addWarningForFile(
-      session,
-      file,
-      `Unsupported image extension${
-        badExts.length > 1 ? 's' : ''
-      } may not correctly render: ${badExts.join(', ')}`,
-      'error',
-    );
+    unconvertableImages.forEach((image) => {
```
rowanc1, line 294: This emits more warnings, but now with explicit node positions.

rowanc1 (commented):
Ok, I think this is good to go!

Still a few latex environments to work out, but good enough for a 0.0.1 for *sure*.

Issue #115, opened 2022-12-31
Improve compatibility with windows (e.g. can't init)


riziles:
### Describe the bug

**context**
When I run `myst init`, I get an error message:

```console
$ myst init
SyntaxError: Invalid regular expression: /c:\LocalTemp\Repos\Local\MystTest\/: \ at end of pattern
```

**problem**
I realize this project is still in beta, but it looks awesome. I'm a big JupyterBook user and love this idea. Maybe my bug is specific to Windows?

### Reproduce the bug

run `myst init` from Windows command line. Same error in Powershell and Git Bash as well. Also tried manually typing in the path.

### List your environment

node 18.12.1
Windows 10

rowanc1:
So glad you are jumping in. :)

Can you try running with `myst init -d` for debug mode?

Most of our testing is on mac and Linux so would love your help getting up and started with some more windows usage!

riziles:
Here is the output of `myst init -d`:

```
SyntaxError: Invalid regular expression: /C:\LocalTemp\Repos\Local\MystTest\/: \ at end of pattern
    at new RegExp (<anonymous>)
    at C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:216265:25
    at Array.map (<anonymous>)
    at replaceCwd (C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:216263:15)
    at Object.info (C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:216280:50)
    at C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:224651:19
    at Generator.next (<anonymous>)
    at C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:78:61
    at new Promise (<anonymous>)
    at __async (C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:62:10)
    at init (C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:224608:10)
    at C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:225052:13
    at Generator.next (<anonymous>)
    at fulfilled (C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:65:24)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)


C:\LocalTemp\Miniconda\node_modules\myst-cli\dist\myst.js:216265
[...]

PR #116, opened 2022-12-31
🐛 Path replacement on windows for logging

rowanc1:
See #115


packages/myst-cli-utils/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     // module is overridden from the build:esm/build:cjs scripts
     "module": "es2015",
     "jsx": "react-jsx",
-    "lib": ["es2020"],
+    "lib": ["es2021"],
```
rowanc1, line 7: This is necessary to use the `replaceAll` string method (introduced in node 15).

Issue #118, opened 2023-01-01
SPDX licenses in myst-frontmatter


rowanc1:
https://www.npmjs.com/package/spdx-correct

This is better than our current correct algorithm!

I think we still need the titles and other metadata, but this could be moved over to get the IDs from this library.

rowanc1:
This is now merged!

PR #119, opened 2023-01-02
Improve myst-frontmatter

rowanc1:
This has a number of improvements for working with frontmatter. The biggest changes are to outsource to the `orcid` and `credit-roles` libraries that are now on npm.

**Planned for this PR:**
- [x] #118


packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
```diff
@@ -27,7 +27,7 @@ import {
 const TEST_AUTHOR: Author = {
   userId: '',
   name: 'test user',
-  orcid: 'https://orcid.org/0000-0000-0000-0000',
+  orcid: '0000-0000-0000-0000',
```
fwkoch, line 30: Looking at the `orcid` library, looks like the original URL input would still be accepted, and just coerced to the id itself. Just changed the test since it compares before and after to the same object, so coercion breaks the tests.

fwkoch (approved):
Looks good, more smaller libraries is nice for reuse of credit-roles / orcid validation. Alias stuff in the validators also seems useful.

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
rowanc1: Yep, I am starting to build out more of the tests in YAML, which are a bit easier to understand and capture more of the usecases I think. (e.g. author/authors).

PR #132, opened 2023-01-16
🗂 Document tab behaviour

rowanc1:
This is tracking some improvements made to tabs and adds the `selected` property to allow you to open a tag to a certain area. The UI is shown here:
https://curvenote.github.io/theme-base/?path=/story/components-tabs--semi-synced-tabs

I have also moved the `spec-ext` type definition to the appropriate place for now.

### DIFF
--- a/packages/myst-cli/src/process/myst/directives.ts
+++ b/packages/myst-cli/src/process/myst/directives.ts
@@ -378,6 +378,47 @@ const TabSet: IDirective = {
   hast: (h, node) => h(node, 'div', { class: 'margin' }),
 };
 
+const TabItem: IDirective = {
+  myst: class TabItem extends Directive {
+    public required_arguments = 1;
+
+    public optional_arguments = 0;
+
+    public final_argument_whitespace = true;
+
+    public has_content = true;
+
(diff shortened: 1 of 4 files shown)


packages/myst-cli/src/process/myst/directives.ts
```diff
@@ -378,6 +378,47 @@ const TabSet: IDirective = {
...
+
+    public optional_arguments = 0;
+
+    public final_argument_whitespace = true;
+
+    public has_content = true;
+
+    public option_spec = {
+      sync: directiveOptions.unchanged,
+      selected: directiveOptions.flag,
```
rowanc1, line 393: This line is the only change in this file. Moved it up closer to the `tab-set` definition.

rowanc1 (commented):
A few comments for the review.

packages/myst-cli/src/process/myst/directives.ts
fwkoch: 👍 Makes sense aligning with sphinx

fwkoch (approved):

Issue #107, opened 2022-12-12
🧹 Prevent new folder creation on `myst clean`


fwkoch:
The `webObjectFileFactory` aggressively copies notebook outputs to files very early in the processing stage (i.e. when going from notebook file to pre-processed mdast). This is opposed to markdown processing where everything is kept in memory at that stage.

This only becomes apparent when running `myst clean`: markdown and notebook files are processed to read frontmatter and find all the exports which need cleaning. Just in doing that the `_build/site/public/_static` folder is created and some notebook files are copied over...

Likely the solution to this is modifying the `webObjectFileFactory`, maybe aligning more with how markdown files are processed? or maybe just making the file copying happen a bit later?

rowanc1:
This is related to #104, which some in progress work is starting to remove dependence on blocks, and have a smaller surface area for dealing with notebooks.

Issue #104, opened 2022-12-07
Remove dependence on `@curvenote/blocks`


rowanc1:
There is a lingering dependency on `@curvenote/blocks` for type definitions around notebooks.

These should be moved out of myst and into the base dependencies.

This currently causes problems with myst-frontmatter, which is always a step behind:

![image](https://user-images.githubusercontent.com/913249/206294604-5618d56c-8bbb-43f9-8a67-f04e838b1062.png)

PR #133, opened 2023-01-16
Update notebook handling to use new nbtx

fwkoch:
See nbtx updates here: https://github.com/curvenote/nbtx/pull/15

Behaviour is mostly unchanged, file writing just happens more explicitly during mdast transform, rather than implicitly via nbtx and file factory.

See also #104 #107


packages/myst-cli/src/process/notebook.ts
```diff
@@ -25,43 +24,36 @@ export async function processNotebook(
...
 
-  const items = await children?.slice(0, end).reduce(async (P, item: TranslatedBlockPair) => {
+  const items = await cells?.slice(0, end).reduce(async (P, cell: ICell) => {
     const acc = await P;
-    if (item.content.kind === KINDS.Content) {
-      if (item.content.format === ContentFormatTypes.md)
-        return acc.concat(asString(item.content.content));
-      if (item.content.format === ContentFormatTypes.txt)
-        return acc.concat(`\`\`\`\n${asString(item.content.content)}\n\`\`\``);
+    if (cell.cell_type === CELL_TYPES.markdown) {
```
rowanc1, line 44: Are these cell type enums defined in nbformat?

packages/myst-cli/src/process/notebook.ts
```diff
@@ -25,43 +24,36 @@ export async function processNotebook(
...
-      if (item.output && item.output.original) {
+    if (cell.cell_type === CELL_TYPES.code) {
+      const code = `\`\`\`${language}\n${asString(cell.source)}\n\`\`\``;
+      if (cell.outputs && (cell.outputs as IOutput[]).length > 0) {
         const minified: MinifiedOutput[] = await minifyCellOutput(
-          fileFactory,
-          item.output.original as CellOutput[],
-          { basepath: '' }, // fileFactory takes care of this
+          cell.outputs as IOutput[],
+          cache.$outputs,
```
rowanc1, line 55: Nice, this will make all sorts of things so much easier now that this is centralized!

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -1,34 +1,46 @@
...
-      useHash: true,
-    });
     await Promise.all(
       outputs.map(async (output) => {
-        output.data = await minifyCellOutput(
-          fileFactory,
-          output.data as CellOutput[],
-          { basepath: '' }, // fileFactory takes care of this
-        );
+        output.data = await minifyCellOutput(output.data as IOutput[], cache.$outputs);
```
rowanc1, line 18: No more writing here which should clean up #107 right?

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -1,34 +1,46 @@
...
 
   outputs.forEach((node) => {
-    walkPaths(node.data, (p: string, obj: any) => {
-      obj.path = `/${p}`;
-      obj.content = `/${obj.content}`;
+    walkOutputs(node.data, (hash: string, obj: any) => {
+      if (!cache.$outputs[hash]) return undefined;
+      const [content, { contentType, encoding }] = cache.$outputs[hash];
+      const folder = session.staticPath();
+      const filename = `${hash}${extFromMimeType(contentType)}`;
```
rowanc1, line 28: do we want a prefix on here? `output-`, thoughts?

packages/myst-cli/src/utils/webFileObject.ts
```diff
@@ -23,7 +22,7 @@ const FileExtensionMap: Record<string, FileExtension> = {
   [KnownCellOutputMimeTypes.AppJson]: FileExtension.json,
 };
 
-export class WebFileObject implements IFileObject {
+export class WebFileObject {
```
rowanc1, line 25: Is this used elsewhere? Looks like maybe in writing base64 images? I wonder if that could be turned into a single function?

packages/myst-cli/package.json
```diff
@@ -53,6 +53,7 @@
   },
   "dependencies": {
     "@curvenote/blocks": "^1.5.17",
```
rowanc1, line 55: Can we remove this now?

rowanc1 (commented):
A few questions -- looking awesome. Can we remove blocks completely now (#104)?

I also wonder if we can remove the web-file object/factory and just rely on a function to write? Might clean a few things up, but not critical!

packages/myst-cli/src/process/notebook.ts
fwkoch: Not super nicely, just as a union of strings type: https://github.com/jupyterlab/jupyterlab/blob/master/packages/nbformat/src/index.ts#L144

packages/myst-cli/package.json
rowanc1: 💥 
image

packages/myst-cli/src/process/notebook.ts
rowanc1: Sounds good!

packages/myst-cli/src/utils/webFileObject.ts
rowanc1: I have removed this!

rowanc1:
Ok, I think this is good to go. A few more things to test before we go in, but should be good!

rowanc1 (approved):
Gone through and tested this in as many of the downstream places as possible! All looks good.

Issue #135, opened 2023-01-16
Target propagration doesn't work for directives


rowanc1:
For example, labelling a `{tab-set}` the label gets lost as the `mystDirective` wrapping node (which holds the label) is removed during processing.

PR #136, opened 2023-01-16
🐛 Target propagations after lifting `mystDirective`s

rowanc1:
See #135


packages/myst-transforms/src/targets.ts
```diff
@@ -23,8 +23,11 @@ import { normalizeLabel, toText } from 'myst-common';
 export function mystTargetsTransform(tree: Root) {
   visit(tree, 'mystTarget', (node: Target, index: number, parent: Parent) => {
     // TODO: have multiple targets and collect the labels
-    const nextNode = findAfter(parent, index) as any;
+    const child = findAfter(parent, index) as any;
     const normalized = normalizeLabel(node.label);
+    // A mystDirective will be removed in a later step
+    // Put the target on the first child if it exists
+    const nextNode = child?.type === 'mystDirective' ? child.children?.[0] : child;
```
fwkoch, line 30: This makes sense - calls slightly into question the order these transforms take place, i.e. why don't we remove `mystDirectives` before propagating targets?

But this makes sense to get things working now.

fwkoch (approved):

packages/myst-transforms/src/basic.ts
```diff
@@ -15,13 +15,17 @@ import { blockquoteTransform } from './blockquote';
...
+  // Some specifics about the ordering are noted below
   captionParagraphTransform(tree);
   mathNestingTransform(tree, file);
+  // Math labelling should happen before the target-transformation
   mathLabelTransform(tree, file);
+  // Target transformation must happen after lifting the directives, and before the heading labels
   mystTargetsTransform(tree);
+  // Label headings after the targets-transform
   headingLabelTransform(tree);
-  mystCleanupTransform(tree);
```
rowanc1, line 24: Moving this up to the top fixes the issue, the rest is docs and types!

packages/myst-transforms/src/targets.ts
rowanc1: Fixed up taking this suggestion.

fwkoch (approved):
Yeh, this makes more sense. Nice renaming / documentation too.

packages/myst-transforms/src/targets.ts
```diff
@@ -3,29 +3,34 @@ import { findAfter } from 'unist-util-find-after';
...
+ * Note, this should happen after `mystDirective`s have been lifted,
+ * and other structural changes to the tree that don't preserve labels.
  */
 export function mystTargetsTransform(tree: Root) {
   visit(tree, 'mystTarget', (node: Target, index: number, parent: Parent) => {
     // TODO: have multiple targets and collect the labels
     const nextNode = findAfter(parent, index) as any;
     const normalized = normalizeLabel(node.label);
     if (nextNode && normalized) {
+      // TODO: raise error if the node is already labelled
```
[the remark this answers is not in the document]
rowanc1: I think we would over-write it if it is implicit? Not quite sure about how that shakes out at this stage! Kicking it down the road a bit more .. ! 👢 💨 ... ⚽️ .......

PR #143, opened 2023-01-20
🛏 Embed notebook cells in a page

fwkoch:
image

- Pull `metadata` from notebook cells onto blocks
- Pull metadata defined in `code` nodes onto blocks
- Use labels in that metadata to add block-level targets
- Add `embed` directive which pulls labeled nodes directly into the page

### DIFF
--- a/.changeset/forty-pillows-retire.md
+++ b/.changeset/forty-pillows-retire.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Update figure/image nodes to embed images if given a target instead of a file path / url
--- a/.changeset/ninety-ducks-reflect.md
+++ b/.changeset/ninety-ducks-reflect.md
@@ -1,0 +1,6 @@
+---
+'myst-cli': patch
+'myst-transforms': patch
+---
+
+Embed block content (including notebook cells) based on block or code cell label in metadata
--- a/.changeset/young-spies-hunt.md
+++ b/.changeset/young-spies-hunt.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Fix inline DOIs to update references in tex/pdf export
--- a/package-lock.json
+++ b/package-lock.json
@@ -11740,6 +11740,16 @@
         "url": "https://opencollective.com/unified"
       }
     },
+    "node_modules/unist-util-filter": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/unist-util-filter/-/unist-util-filter-4.0.0.tgz",
+      "integrity": "sha512-H4iTOv2p+n83xjhx7eGFA3zSx7Xcv3Iv9lNQRpXiR8dmm9LtslhyjVlQrZLbkk4jwUrJgc8PPGkOOrfhb76s4Q==",
+      "dependencies": {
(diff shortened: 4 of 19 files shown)


rowanc1:
@choldgraf @mmcky I would love a targeted review on this. We are aiming to do something similar to the `glue` functionality in JupyterBook, but _without_ access to the python kernel (in this case), something that is ideally python agnostic, and also is workable in VSCode.

What @fwkoch has pulled together in this PR is the ability to add a directive (`embed`) and is currently bootstrapping off the syntax developed to solve the above constraints in Quarto (using comments at the top of a Jupyter cell (`#| label: my-cell`)). It would be pretty trivial to pull from a `cell_id`, or some other specific metadata in the notebook cell.

I think for the context of _this_ review, I would love some perspective on how you see folks using embed functionality in JupyterBook in the wild.

Perhaps the easiest way to bring this in soon would be to adopt the `{glue}` terminology for the directive and role now? And then we experiment a bit more and bring this up in a MEP -- I would also like to think about this in the context of jupyterlab-myst as well. 🤔

Looking forward to your thoughts!

choldgraf:
a few quick thoughts:

glue actually originally comes from [scrapbook](https://github.com/nteract/scrapbook). We were trying to follow the same pattern they used. However, their implementation only dealt with embedding content in the notebook, not pasting it. So we reused the glue verb for the "pasting" action to avoid checking terminology too much, but IMO it is more understandable to use a word like "embed" or "paste" instead. So +1 to that name.[^1]

[^1]: As an aside, i wonder if this could be the target for a Myst syntax one day...i note that obsidian has a dedicated embed syntax by putting a ! In front of the reference to a block (eg `[[mylabel]]` creates a link to the label, and `![[mylabel]]` instead embeds it)

For this PR, i like the idea of starting with cell content variables to unblock us and experiment. Better to follow Quarto rather than invent a new in-cell-content syntax for now as well. Perhaps in the future this can still be supported, but if cell level metadata also exists then it takes precedence.

I bet that we could do a search for "glue" in the issues in JB and the forum, and get a good idea of some ways people are using it. I think statistics and figures are the two big ones. I also think people would like to use it to reuse content (eg write it one place and then just insert it many places...similar to how the substitution syntax in the MyST parser works (with `{{}}`)

rowanc1:
Ah, knowing about scrapbook is super helpful! It looks like that saves to the JSON output, which is nice from being able to access these without rerunning the notebook (e.g. different if saving in markdown). I think that MyST should be able to do a good-enough job at rendering Quarto and ideally the opposite as well -- compatibility in the ecosystem in syntax is good for everyone.

I think we will plan to move ahead with the embed syntax, so as not to overload "glue". Will get to a place of experimenting, and then aim to put a MEP together!!

Thanks @choldgraf for your insights. 🚀

rowanc1:
Small bug found on frontmatter parsing! Related to the block structure changes in the mdast:

image

rowanc1:
Nice tested locally!!

rowanc1:
This is fantastic, the addition to the figure syntax is really nice @fwkoch -- i.e. the source of the figure "image" is just the output link. I think that idea is really really powerful and can lead to all sorts of integrations down the line.

Issue #144, opened 2023-01-21
myst-transforms, myst-to-tex, myst-to-jats was not found in your node_modules


choldgraf:
I just tried to build a local jupyter book site with `myst-cli`, and got these errors:

```
The path "myst-transforms" is imported in ../theme-base/packages/myst-to-react/src/myst.tsx but "myst-transforms" was not found in your node_modules. Did you forget to install it?
The path "myst-to-tex" is imported in ../theme-base/packages/myst-to-react/src/myst.tsx but "myst-to-tex" was not found in your node_modules. Did you forget to install it?
The path "myst-to-jats" is imported in ../theme-base/packages/myst-to-react/src/myst.tsx but "myst-to-jats" was not found in your node_modules. Did you forget to install it?
```

I'm running myst-cli v0.1.5 (what was installed after running `npm update -g myst-cli`.

rowanc1:
This is a spurious warning that has leaked through due to the underlying package changing their message slightly!

rowanc1:
Aiming to do another release on Monday or Tuesday that should fix these up, thanks @choldgraf for the report. Looking forward to more. :)

Issue #141, opened 2023-01-19
Lots of issues trying LaTeX


drscotthawley:
### Describe the bug

(**EDIT:** edited to show standard/reproducible JAES template, rather than my specific document.)

**context**
Hi guys. Saw [Steve's post about "Writing a Scientific Paper in MyST Markdown"](https://twitter.com/stevejpurves/status/1616018826910437377?s=20&t=amcH0PLDTnVKkz2MVZVXJw) and Rowan's seminar next week. Since I'm starting two new scientific papers, I decided to try MyST.  

Followed the instructions https://curvenote.com/blog/how-to-use-latex-with-myst-markdown, ...downloaded a ZIP of my stuff from Overleaf, using the template files from the Journal of the Audio Engineering society. 

Installed myst via npm, ran it in the directory where my .tex file is and....

**expectation**
I expected to get a wonderful rendering of my scientific document. 

**bug**
But instead lots of errors occur .

```console
/tmp/aes-template$ cp ~/Downloads/JAES_Latex_template.zip .
/tmp/aes-template$ unzip JAES_Latex_template.zip 
Archive:  JAES_Latex_template.zip
  inflating: jaes-sample1-bibtex.tex  
  inflating: jaes.bib                
  inflating: jaes.bst                
  inflating: jaes.cls                
  inflating: cuted.sty               
  inflating: readme.txt              
  inflating: jaes-sample2-bibitem.tex  
  inflating: jaes-guide.pdf          
  inflating: headshot2.jpg           
  inflating: headshot1.jpg           
  inflating: fig1_example.eps        
  inflating: fig2_example.eps        
  inflating: jaes-sample1-bibtex.pdf  
  inflating: jaes-sample2-bibitem.pdf  
/tmp/aes-template$ myst init
💾 Writing new project and site config file: myst.yml
/tmp/aes-template$ myst start
🐕 Fetching template metadata from https://api.myst.tools/templates/site/myst/book-theme
🌎 Cloning template from https://github.com/curvenote/book-theme.git
💾 Saved template to path _build/templates/site/myst/book-theme
⤵️  Installing web libraries (can take up to 60 s)
📦 Installed web libraries in 13 s
⛔️ jaes-sample1-bibtex.tex:21:1 Unhandled TEX conversion for node of "macro_markboth"
⚠️  jaes-sample1-bibtex.tex:21:15 Unknown character 
⚠️  jaes-sample1-bibtex.tex:21:19 Unknown character 
[...]

PR #145, opened 2023-01-21
🐛 Catch and mute spurious warnings

rowanc1:
See #144, #141


packages/myst-cli/src/build/site/logger.ts
```diff
@@ -29,6 +29,7 @@ export function createServerLogger(session: ISession): LoggerDE {
       if (!line) return;
       // This is a spurious Remix warning https://github.com/remix-run/remix/issues/2677
       if (line.includes('is not listed in your package.json dependencies')) return;
+      if (line.includes('was not found in your node_modules')) return;
```
rowanc1, line 32: The phrasing of this spurious warning changed in a recent release. This is still a problem when working in npm workspaces, but isn't actually real.

PR #146, opened 2023-01-22
Overhaul Docs and many details

rowanc1:
I have gone through the onboarding process and improved a number of commands:

- myst templates (removed `jtex list`)
- myst build
- myst (aliwas for simplified `myst init`)
- improved frontmatter validation

I have also added three quickstart guides! ⚡️⚡️⚡️


packages/myst-cli/src/frontmatter.ts
```diff
@@ -67,6 +70,34 @@ export async function getRawFrontmatterFromFile(session: ISession, file: string)
   await loadFile(session, file);
   const result = cache.$mdast[file];
   if (!result || !result.pre) return undefined;
-  const frontmatter = getFrontmatter(result.pre.mdast);
+  // Copy the mdast, this is not a processing step!
+  const frontmatter = getFrontmatter(copyNode(result.pre.mdast));
```
rowanc1, line 74: @fwkoch this fixes that random bug I was seeing, where it was changing the page structure!

packages/myst-cli/src/session/types.ts
```diff
@@ -13,8 +13,8 @@ export type ISession = {
   configFiles: string[];
   store: Store<RootState>;
   log: Logger;
+  reloadConfigs(): void;
```
rowanc1, line 16: There is likely a better way to do this, the configs are still cached on TOC changes, which is pretty annoying. :)

rowanc1 (commented):
I think that each commit is likely better to go through, I have made them all atomic.

I am going to plan to bring this in and update the documentation such that @choldgraf can take a look at the tutorials/guides live and provide a bit of feedback on an update CLI version.

 @fwkoch if you have thoughts on anything here, we can address in on monday or Tuesday, I would most especially like you review on the build options, which I changed -- and the way to reload the config files in a session.

choldgraf:
Quick q: can we set up a netlify or a GH actions preview of PRs? This would make it easier to quickly see what the changes in docs look like

PR #149, opened 2023-01-24
✍️ Write bibtex from citation renderers in tex export

fwkoch:
image

This fixes up inline DOI citations for tex/pdf exports. It is using the `cache.$citationRenderers` to write the bibtex, instead of just copying existing bibtex files (inline DOIs did not have existing bibtex files).

It's still a little klugey: we are using `renderer.cite._graph.find((item) => type === '@biblatex/text').data` (i.e. the raw text input data), rather than rebuilding the bibtex item based on structured citation data found directly on `renderer.cite`.


fwkoch:
This addresses https://github.com/executablebooks/mystjs/issues/140

rowanc1:
Oh, didn't realize the target was the other branch not main?

Should we revert?

PR #150, opened 2023-01-24
👓 suggested changes for guides review

stevejpurves:
some changes to the guides after initial walkthrough and review. Some typos corrected but also some additional links and content. We use tutorial and guide interchangeably which is a little confusing a first -- that motivated some of the earlier changes.


rowanc1:
Thanks @stevejpurves, I have taken the commit from here and moved it over to a branch: #152

I will take these and make some improvements directly rather than through a fork (I think you have edit rights on this repo?!). There are a couple other inconsistencies that I noticed as well!

PR #152, opened 2023-01-24
📖 Improve the quickstart tutorials

rowanc1:
- [x] use "tutorial" (i.e. quickstart tutorial), not guide (in line with https://diataxis.fr/)
- [x] Do the website quickstart first (this is more in line with getting an early preview that works out of the box)
- [x] Encourage the user to start the webserver in the document tutorial, but focus on the live-preview rather than the customization of the website.

### DIFF
--- a/.changeset/fair-ravens-grab.md
+++ b/.changeset/fair-ravens-grab.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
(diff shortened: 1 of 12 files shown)


rowanc1:
Still many things to do to address the issues that @stevejpurves raised in #153, but this has iterative improvements towards those goals!!

PR #154, opened 2023-01-25
📤 Support embeded cross references and output images/text in file exports

fwkoch:
#143 added the ability to embed output images and other blocks from one page into another with `embed`, `image`, or `figure` directives. However, for static PDF/docx exports, we were only processing a single page, so references to content on other pages did not resolve.

Now, we process all files in a project even for single page exports. While `crossReference` links still don't work (since the target page doesn't exist), `embed` content now works (since it is pulled from the target page onto the exported page).

Furthermore, notebook `outputs`, which previously only worked in a web context, are now converted to static `images` or `code` for rendering in a static document. Rich HTML outputs still are not supported, but simple image/text outputs are.

image

### DIFF
--- a/.changeset/few-beers-shop.md
+++ b/.changeset/few-beers-shop.md
@@ -1,0 +1,7 @@
+---
+'myst-cli': patch
+'myst-to-docx': patch
+'myst-to-tex': patch
+---
+
+Embeded content from other pages resolves on single page PDF/docx export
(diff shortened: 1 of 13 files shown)


rowanc1 (approved):
This looks awesome. Works really well on CLI!

rowanc1:
Really having fun with this feature. :)

image

PR #159, opened 2023-01-26
Fix  readme example, mystjs undefined use MyST directly.

Carreau:


rowanc1:
Thank you! This section and workflow of using parser components directly needs some love. Are you planning to use this directly somewhere? If so I could spend a bit of time improving the experience!!

Carreau:
Mostly exploring, I'll likely use something akin to 

```
const html = myst.renderMdast({type: "text", value:"Hello"})
```

Directly, but I'm not a JS person, so I struggle a bit. I finaly have someone helping to convert my papyri AST to MyST Ast,and then I can use myst js to render,

rowanc1:
That is so awesome -- I would love to catch up again soon.

Currently the full render process is a bit more than just what is in `mystjs`, which we really need to fix. A lot of the transformations are in `myst-transforms`, and it is a bit piecemeal to bring them all together. This is the best demo of how to use it right now:
https://github.com/curvenote/theme-base/blob/main/packages/myst-demo/src/index.tsx#L44

We want to make that a *lot* better, and actually have what you wrote do the thing you expect it to do. That is going to take a bit of work to make the packages be a bit more sensible -- I will open an issue this afternoon and tag you in it!

Issue #161, opened 2023-01-26
Include files from subfolders won't render on PDF


eurunuela:
### Describe the bug

**context**
I separate the sections of my manuscript into different md files inside of the `sections` subfolder to make it easier to read the main file.

**expectation**
I expected the contents of the files to appear on the PDF when using the `myst build main.md --pdf` command, just as they appear on the web.

**bug**
But instead the contents are not shown on the PDF.

**problem**
This is a problem for people separating the sections of their document into files inside a subfolder.

### Reproduce the bug

1. Create a subfolder called `sections`
2. Create a md file inside of the `sections` folder.
3. Include the md file from the `sections` folder into the main md file.
4. Run the `myst start` command to see the contents appear on the web version of the document.
5. Run the `myst build --pdf` command to see the contents do NOT appear on the PDF.

### List your environment

```
❯ jupyter-book --version
Jupyter Book      : 0.13.1
External ToC      : 0.2.4
MyST-Parser       : 0.15.2
MyST-NB           : 0.13.2
Sphinx Book Theme : 0.3.3
Jupyter-Cache     : 0.4.3
NbClient          : 0.5.13
```

rowanc1:
Thanks so much for the bug report @eurunuela! Glad myst is going well so far. :)

Looks like `include` isn't supported in docx/tex, but for some reason the error isn't being reported on the tex side.

![image](https://user-images.githubusercontent.com/913249/214964847-dfb658c6-a521-499d-946c-4ecbbbab0f7d.png)

@fwkoch is already looking at a fix! :)

fwkoch:
@eurunuela - thanks for the issue! This PR should be all we need to get the `include` directive working for PDF exports: https://github.com/executablebooks/mystjs/pull/163

We will get this released soon, so you can try it out!

---

I think this issue hints at a bigger feature at the back of our minds around multi-page PDF exports. Rather than forcing everything together using `include` directives, it would be great to define the different sections and pull them together nicely with a table of contents and some sort of multi-page latex template, maybe?

rowanc1:
PR looks great -- @fwkoch do you have any idea why the warnings aren't showing up for the latex export side?

fwkoch:
Yes! We were recording the errors but not actually logging them - added that fix to the PR: https://github.com/executablebooks/mystjs/pull/163/files#diff-809f642b8b08c50fa59434f4dd5e68c8063c8be90e8022b375c13b1edb262a6bR48

[...]

PR #163, opened 2023-01-26
📄 Render include node children during tex/docx export

fwkoch:
This addresses #161 - When using the `include` directive, a new node of `type: include` is added to the document AST. However, this node type was simply ignored by `myst-to-tex` and `myst-to-docx`, so it's children disappeared 😱

This PR simply tells `myst-to-tex` and `myst-to-docx` to render the children of `include` nodes.

### DIFF
--- a/.changeset/friendly-dodos-think.md
+++ b/.changeset/friendly-dodos-think.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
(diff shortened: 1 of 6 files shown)


rowanc1:
Nice, can you add a changelog please! `npm run changeset`

rowanc1:
🚀

PR #173, opened 2023-01-30
🎇 Support for PDF images

rowanc1:
This converts PDFs to PNGs in the build. They now work for Word and web export.


fwkoch (approved):
Cool, yeah, this is nice. Pretty boilerplate compared to the other image conversions.

rowanc1:
Nice -- I think this is good enough for now, next time we support an image type, we may want to have a better way to introduce image transforms.

PR #174, opened 2023-01-30
🐛 Validate subtitle & short_title on project/site

rowanc1:
Support for validation to match the new types. subtitle and short_title are now available for the project and site.


fwkoch (approved):
Looks good - pretty straightforward. Doesn't change any of the fallback behaviour for these properties, only adds them to site / project.

PR #179, opened 2023-02-03
📦 Pull tab/grid/card directives into external packages

fwkoch:
This starts pulling custom directives out of `myst-cli` into smaller, reusable subpackages. Next we will need to improve (simplify!) how these external directive plugins are defined.


rowanc1:
This is looking great! Excited to be able to get some more attention at this level, and have reuse in jupyterlab as well. 🚀

Issue #181, opened 2023-02-03
Refactor directive and role parsing


fwkoch:
**Context:**
Over the past year we have been coming up against some of the limitations of early design around the markdown-it tokenizer. There were many things that are in that library that are now either duplicated or deprecated (e.g. references and state management), there are also challenges in having errors being reported to the CLI, jupyter or web context in a consistent way. Some errors are not possible to report until much later (i.e. after transformations), and some errors need to be much more lenient (e.g. parser errors). We also need to start thinking about adding myst language extensions (e.g. tabs, diagrams) in a way that work across all serializers and are agnostic to the tokenizer being used (e.g. unified or markdown-it).

**Current state:**

- `markdown-it-docutils` - markdown-it plugin to handle tokenizing roles and directives. In addition to generic roles/directives, it introduces special token stream behaviour for admonitions, code, math, etc etc.
- `markdown-it-myst-extras` - markdown-it parsing support for additional markdown syntax features including colon fence, block breaks, targets, comments
- `mystjs` - creates a parser with markdown-it-docutils plugin and extensibility to add more _markdown-it tokenizers_ for new directives/roles, defines tokens-to-myst, including the special roles/directives from markdown-it-docutils, defines basic transforms, includes myst-to-hast
- `myst-cli` - consumes and extends mystjs, makes it usable from CLI, defines a bunch of new directives/roles (in a markdown-it, token-y way, i.e. has to define parsing to token stream then transforming to mdast)
- `myst-transforms` - a bunch of additional mdast transforms consumed by myst-cli
- `myst-ext-card/grid/tabs` - directives previously defined in myst-cli, pulled into separate subpackages

**Desired state:**

- `markdown-it-docutils` - continues to exist as-is to support existing vscode integration, unused in mystjs
- `markdown-it-myst-extras` - unchanged, continues to be used by mystjs as-is
- `markdown-it-myst` - pulls out basic role/directive tokenizing from markdown-it-docutils. No custom tokens for any specific roles/directives, instead, they are all just `mystRole/Directive` with `args`, `options`, `value` and `parsed_args/options/value`
- `mystjs` - no longer exists in it's previous capacity. Lets us potentially rename myst-cli to mystjs? 
[...]

PR #184, opened 2023-02-06
💥 Role / Directive Refactor

fwkoch:
`markdown-it-myst` is a new package to generically parse roles/directives into tokens. It does a nested parse to tokenize args/options/content as well, passing along both the raw role/directive content and the tokenized content.

This package does not define custom role/directive types; those are defined downstream as AST transforms on content made available here.

This package is the first step in addressing #181

---

# Update

This PR has consolidated the work on `markdown-it-myst` with `myst-parser` #193 and `myst-roles/directives` #206 - it addresses all the refactoring described in #181


.changeset/brown-onions-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Pull tab/grid/card directives into external packages
```
rowanc1, line 5: Should this be in here? Did this get picked up somehow accidentally?

fwkoch: Hmm, yeah not sure how it got picked up here. This changeset was already resolved on main from this commit: https://github.com/executablebooks/mystjs/commit/037c8d5808077a1f34c4e37b8c4fecdf6ba59031

I'll delete the file from here...

rowanc1:
I think that the only bugs that are left are around the card headers.

image

It would be great to have tests on the remaining directives, but also want to get this in and out the door! I think there is another pass on validation we need to do, but that can come in next, I don't think we actually loose that much?

rowanc1:
Some known things:
- Alignment of images is a regression.
- And better validation of width/height.
- figure classes did a bit more on pulling off 'center' and making that the align key.

Anything else?

I think those can easily come in a validation pass.

Issue #183, opened 2023-02-06
Website can't make JIT changes if markdown files are inside of a folder


eurunuela:
### What version of `mystjs` are you using?

v0.1.9

### How did you install myst?

npm

### What operating system are you using?

Mac

### Which area is this feature request for?

Website User Interface

### Describe the Bug

The sections in my manuscript are written in different markdown files inside of the `sections/` folder.

When I use `myst start`, the website renders just fine. However, whenever I edit my section files, the website does not update with the changes and I see the following warning message in the console:

`File is not in project: sections/methods.md`

### Expected Behavior

The website should update to show the changes.

### To Reproduce

1. Create a folder called `sections/`
2. Create markdown file inside the folder, e.g., `sections/methods.md`, and write something in it
3. Run `myst start`
4. Edit the `sections/methods.md` file
5. The website will not show the changes and the warning message should appear in the terminal

fwkoch:
@eurunuela - I could not recreate this bug exactly as you describe; editing the existing `sections/methods.md` while `myst start` was running worked as expected (site update, no warning). However, creating a _new_ markdown file in the `sections/` folder did match your description - warning and no update - until I cancelled and re-ran `myst start`.

I'll fix up that bug and see if that resolves things for you.

eurunuela:
Hey @fwkoch! I'm still seeing this bug in v0.1.17...

Here's a screen recording: https://www.loom.com/share/5c38a68c9d7b475bad2066c62534f30a

eurunuela:
Turns out I'm having this issue even when the different files are in the main folder as opposed to being in a `sections` folder...

eurunuela:
Any updates on this @rowanc1 @fwkoch ?

PR #186, opened 2023-02-07
🔄 Reload project on all created/deleted files during watch

fwkoch:
This addresses #183 - creating or deleting markdown files now reloads the entire project, adding/removing pages for these files. Previously this full reload only happened when the `_toc.yml` file was updated or a new `bib` file was added.

This likely introduces some inefficiencies - we are reloading everything when we really only need to reload the project structure... but I suspect in practice, things are fast enough to not worry - this still happens infrequently compared to page edits. Something to keep an eye on, though.

One other small change is renaming the `triggerReload` argument to `serverReload`, hopefully eliminate some confusion with page vs. project vs. server reload.

### DIFF
--- a/packages/myst-cli/src/build/site/watch.ts
+++ b/packages/myst-cli/src/build/site/watch.ts
@@ -16,11 +16,7 @@ type TransformOptions = {
   defaultTemplate?: string;
 };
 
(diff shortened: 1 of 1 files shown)


rowanc1:
Looks great!

Issue #194, opened 2023-02-08
Numbering field in frontmatter has no effect on headings


stevejpurves:
### What version of `mystjs` are you using?

0.1.11

### How did you install myst?

npm

### What operating system are you using?

Mac

### Which area is this feature request for?

Export: LaTeX or PDF

### Describe the Bug

the following are both valid frontmatter

```
---
numbering: true
---
```

```
---
numbering:
   heading_1: true
  heading_2: true
---
```

but tex is always produced without numbering `\section*`, etc...

### Expected Behavior

non starred `*` headings should be generated when numbering is applied.

### To Reproduce

Export any myst file with numbering set to true in the frontmatter

alcrene:
It seems to me this issue is resolved ? I used run into it, and since I updated `mystmd` I no longer do so.

PR #195, opened 2023-02-08
#️⃣ unstar sections when numbering=true

stevejpurves:
addresses issue #194

### DIFF
--- a/.changeset/twenty-rabbits-repeat.md
+++ b/.changeset/twenty-rabbits-repeat.md
@@ -1,0 +1,5 @@
+---
+'myst-to-tex': patch
+---
+
+Assume enumerated is default true for "star" check in myst-to-tex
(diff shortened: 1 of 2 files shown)


stevejpurves:
I didn't see the original `enumerated` field on the node at all, but noted that `enumerator` is added when numbering is enabled -- so using a truth value on that field to remove the stars from sections.

fwkoch:
@stevejpurves - I took a look at this again - It didn't feel quite right to use `enumerator` instead of `enumerated` since the former field is computed during processing (sounds like you were also hesitant about that). I looked around a bit and found that in other transforms we treat `enumerated` as `true` unless it is explicitly set to `false`, leading to checks like `node.enumerated !== false`:
- https://github.com/executablebooks/mystjs/blob/main/packages/myst-transforms/src/enumerate.ts#L255
- https://github.com/executablebooks/mystjs/blob/main/packages/myst-to-html/src/state.ts#L103

I've updated this PR to have the same check as those (rather than relying on `enumerator`).

This implicit truthiness is a bit odd feeling, potentially we want to be more careful about explicitly setting `enumerated = true`? But for now, this makes `myst-to-tex` consistent with the rest of the codebase.

Issue #196, opened 2023-02-08
Add option to *not* build citations from links to dois


stevejpurves:
### Which area is this feature request for?

LaTeX Export

### Describe the feature you'd like to request

The current DOI scanning & lookup feature is a great way to build references from inline DOI links 👏 but if an author has done something different from their referencing, it is not always desirable.

Currently, there is no way to disable this.

### Describe the solution you'd like

A setting (in `frontmatter`) that allows DOI lookup for citations to be disabled.

PR #197, opened 2023-02-08
🔨 forcibly removing DOI citation lookup & insertion

stevejpurves:
Trying to address issue #196 here.

Currently, this is intended for discussion on how to best provide an option in frontmatter.

### DIFF
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -136,13 +136,13 @@ export async function transformMdast(
     new WikiTransformer(),
     new GithubTransformer(),
     new RRIDTransformer(),
-    new DOITransformer(), // This also is picked up in the next transform
+    // new DOITransformer(), // This also is picked up in the next transform
     new MystTransformer(intersphinx),
   ];
   linksTransform(mdast, vfile, { transformers, selector: LINKS_SELECTOR });
 
   // Initialize citation renderers for this (non-bib) file
-  cache.$citationRenderers[file] = await transformLinkedDOIs(log, mdast, cache.$doiRenderers, file);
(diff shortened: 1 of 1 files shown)


packages/myst-cli/src/process/mdast.ts
```diff
@@ -136,13 +136,13 @@ export async function transformMdast(
     new WikiTransformer(),
     new GithubTransformer(),
     new RRIDTransformer(),
-    new DOITransformer(), // This also is picked up in the next transform
+    // new DOITransformer(), // This also is picked up in the next transform
```
stevejpurves, line 139: Removing this here actually seems ot have no effect. What is the role of this transformer?

packages/myst-cli/src/process/mdast.ts
```diff
@@ -136,13 +136,13 @@ export async function transformMdast(
...
     new MystTransformer(intersphinx),
   ];
   linksTransform(mdast, vfile, { transformers, selector: LINKS_SELECTOR });
 
   // Initialize citation renderers for this (non-bib) file
-  cache.$citationRenderers[file] = await transformLinkedDOIs(log, mdast, cache.$doiRenderers, file);
+  // cache.$citationRenderers[file] = await transformLinkedDOIs(log, mdast, cache.$doiRenderers, file);
```
stevejpurves, line 145: removing the renderer here stopped lookup and rendering of `\cite{}` commands.

stevejpurves:
@fwkoch so would adding a frontmatter field, and checking that in these two locations, adding the transformers only when `true` be the best way forward?

```
convert_doi_links: true|false
```
?

rowanc1:
We can certainly come back to this! My main hesitation at this stage is we don't have any centralized options yet for controlling parse behaviour. I think this is an obvious thing we need to add, but it will take some work. :)

PR #229, opened 2023-02-17
Improve admonitions to support GitHub and Quarto

rowanc1:
This improves the documentation around the admonitions and adds explicit support for github and quarto-style admonitions.

Slightly nicer explainer I think:
image

Both of these work, but not yet in the demo (needs a release!)

image

### DIFF
--- a/packages/markdown-it-myst/src/directives.ts
+++ b/packages/markdown-it-myst/src/directives.ts
@@ -12,10 +12,10 @@ const COLON_OPTION_REGEX = /^:(?<option>[^:\s]+?):(\s*(?<value>.*)){0,1}\s*$/;
 function replaceFences(state: StateCore): boolean {
   for (const token of state.tokens) {
     if (token.type === 'fence' || token.type === 'colon_fence') {
-      const match = token.info.match(/^\{([^\s}]+)\}\s*(.*)$/);
+      const match = token.info.match(/^\s*\{([^}]+)\}\s*(.*)$/);
       if (match) {
         token.type = 'directive';
-        token.info = match[1];
+        token.info = match[1].trim();
         token.meta = { arg: match[2] };
       }
     }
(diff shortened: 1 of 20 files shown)


packages/markdown-it-myst/src/directives.ts
```diff
@@ -12,10 +12,10 @@ const COLON_OPTION_REGEX = /^:(?<option>[^:\s]+?):(\s*(?<value>.*)){0,1}\s*$/;
 function replaceFences(state: StateCore): boolean {
   for (const token of state.tokens) {
     if (token.type === 'fence' || token.type === 'colon_fence') {
-      const match = token.info.match(/^\{([^\s}]+)\}\s*(.*)$/);
+      const match = token.info.match(/^\s*\{([^}]+)\}\s*(.*)$/);
```
fwkoch, line 15: This allows spaces _inside_ the directive name too. Probably don't want that?

Maybe `^\s*\{\s*([^}\s]+\s*)\}\s*(.*)$` instead...?

packages/myst-transforms/src/admonitions.ts
```diff
@@ -67,6 +70,41 @@ export function admonitionHeadersTransform(tree: Root, opts?: Options) {
   });
 }
 
+/**
+ * Visit all blockquote notes and add headers if necessary, support GitHub style admonitions
+ */
+export function admonitionBlockquoteTransform(tree: Root) {
```
fwkoch, line 76: Nice 👍

fwkoch (commented):
This looks good - only comment is, you guessed it, about regex.

packages/markdown-it-myst/src/directives.ts
rowanc1: Nice. Yeah, good point. I will update -- If we go with something more like pandoc in the future, this whole area is going to need an overhaul. But that is for later!

rowanc1: Done!

PR #233, opened 2023-02-17
◻️ Allow whitespace around role name inside brackets

fwkoch:
Should be a small change, allowing whitespace around role names: `{ my_role }` - but it does change regex, which always feels a little risky.


rowanc1 (approved):
Changes look good!

PR #234, opened 2023-02-17
🆕 Add mystToHtml as a simple wrapper function

rowanc1:


packages/myst-to-html/src/renderMdast.ts
```diff
@@ -0,0 +1,32 @@
...
+import rehypeStringify from 'rehype-stringify';
+import { unified } from 'unified';
+import { formatHtml } from './format';
+import { mystToHast } from './schema';
+import { State } from './state';
+import { transform } from './transforms';
+
+export function mystToHtml(
+  tree: Root,
+  opts?: {
```
fwkoch, line 11: These options are pretty limited - only cover what we needed in the `myst-parser` tests. In the future we could expand them to match all options for each pipeline... Also could provide some defaults...?

But I think for now this is fine.

fwkoch (approved):

packages/myst-to-html/src/renderMdast.ts
rowanc1: Yeah, I didn't put any thought into this!! We will probably get back to this in the future, and can put a bit more thought in then. :)

PR #236, opened 2023-02-17
#️⃣ tex-to-myst: Only insert figure number for latex \ref

fwkoch:
Parsing latex `Figure \ref{fig:...}` resulted in `Figure _Figure 1_`. Now it correctly inserts `Figure _1_`.

To insert `_Figure 1_` you can still use `\autoref` in the latex.


rowanc1:
Nice! Thank you!

PR #242, opened 2023-02-18
🐛 Update package dependencies for standalone use

rowanc1:


package-lock.json
```diff
@@ -11781,8 +11782,9 @@
       }
     },
     "node_modules/unist-util-find-after": {
-      "version": "4.0.0",
-      "license": "MIT",
+      "version": "4.0.1",
```
rowanc1, line 11785: This is the only version bump. Should be fine!

PR #247, opened 2023-02-19
Use reproducable image in docs

kolibril13:
Currently, the docs are using random images by https://source.unsplash.com/random/500x150?sunset

This is a personal preference, but I'd prefer reproducible images in the docs, as they have a higher recognition factor.

I've just replaced the links with the first link that was produced by my browser when I replaced https://source.unsplash.com/random/500x150?sunset by it.

As the new link is a bit long, it might be worth looking intro another option to get these images. 
Is there something like
https://source.unsplash.com/001/500x150?sunset ?

### DIFF
--- a/docs/figures.md
+++ b/docs/figures.md
@@ -29,7 +29,7 @@ There are two directives that can be used to add additional information about th
 ## Image directive
 
 ````{myst}
-```{image} https://source.unsplash.com/random/500x150?sunset
+```{image} https://images.unsplash.com/photo-1564731119111-2debdd39229c?crop=entropy&cs=tinysrgb&fit=crop&fm=jpg&h=150&ixid=MnwxfDB8MXxyYW5kb218MHx8c3Vuc2V0fHx8fHx8MTY3NjgzNDAzNg&ixlib=rb-4.0.3&q=80&utm_campaign=api-credit&utm_medium=referral&utm_source=unsplash_source&w=500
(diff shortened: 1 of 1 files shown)


rowanc1:
I like the idea of this. I wonder if we host a few `CC-0` images on our server (not in git) to allow for use in these demos.

I need to move https://github.com/curvenote/myst-tools-theme into the `myst.tools` repository in EBP, but I wonder if that is a better place to host these images - which keeps the git repo size small for mystjs.

Then we could have `https://myst.tools.org/i/mountain-1` as the URL or something clean.

kolibril13:
I think it would be nice if the url has also an ending that indicates that it's an image. 
E.g.
https://myst.tools.org/i/mountain-1
can instead be
https://myst.tools.org/i/mountain-1.png

GitHub images could be accessed via an url like this
https://raw.githubusercontent.com/kolibril13/mobject-gallery/main/preview_image.png
Or a slightly shorter URL on GitHub pages like this:
https://kolibril13.github.io/mobject-gallery/preview_image.png

To keep the repo size small, the images could be hosted in an extra repo, like mystjs-image-assets.
But hosting them on the curvenote server sounds also like a good idea!

rowanc1:
Yeah, all this will be over in EBP shortly, so no curvenote servers involved, just redirects of URLs to be myst-tools branded.

Good call on including the extension!

PR #251, opened 2023-02-21
Small improvements to latex support

fwkoch:
- basic implementation of `renewcommand` (same as `newcommand` except no warning on override)
- multiple `\title` values warn
- `adjustbox`, `makecell`, `textwidth`, `onecolum` all pass without warnings

### DIFF
--- a/packages/tex-to-myst/src/tex.ts
+++ b/packages/tex-to-myst/src/tex.ts
@@ -18,9 +18,10 @@ function parseArgument(node: GenericNode, next: GenericNode): boolean {
     }
     const lastArg = node.args[node.args.length - 1];
     if (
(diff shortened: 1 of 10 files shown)


packages/tex-to-myst/src/tex.ts
```diff
@@ -94,6 +97,10 @@ const macros: Record<string, number> = {
   v: 1,
 };
 
+// These macros allow a mix of {} and [] arguments, rather than only allowing
+// a single leading [] argument followed by {} arguments.
+const mixed_arg_macros = ['multirow'];
```
rowanc1, line 102: Nice, this looks great.

PR #256, opened 2023-02-23
🦶🎵 Implement footnotes in myst-to-tex

fwkoch:


packages/myst-to-jats/src/index.ts
```diff
@@ -393,6 +408,10 @@ const plugin: Plugin<[Options?], Root, VFile> = function (opts) {
     const tree = copyNode(node) as any;
     basicTransformations(tree, file);
     const state = new JatsSerializer(file, opts ?? { handlers });
+    const footnotes = selectAll('footnoteDefinition', tree) as FootnoteDefinition[];
```
fwkoch, line 411: We handle `footnoteDefinition` nodes left in the tree here. This is nice and what we should move to everywhere...

packages/myst-to-tex/src/index.ts
```diff
@@ -247,20 +248,30 @@ const handlers: Record<string, Handler> = {
   include(node, state) {
     state.renderChildren(node, true);
   },
+  footnoteReference(node, state) {
+    if (!node.identifier) return;
+    const footnote = state.references.footnotes?.[node.identifier];
```
fwkoch, line 253: Here, we assume `footnoteDefinition` nodes have been pulled off the tree and are passed in separately. This is how `myst-cli` currently operates but is not super nice. We should move towards just leaving the footnote nodes in the tree rather than passing around a separate `references` object.

packages/myst-to-jats/src/index.ts
rowanc1: Nice, looking forward to that.

rowanc1:
Looks good! Looking forward to implementing this properly in JATS, but will take another pass on the XML side. 🔜

PR #259, opened 2023-02-23
📖 Initial jupyterlab-myst docs

rowanc1:
This is a very rough first attempt at docs that can change rapidly over the coming days. I want to get some visibility on the excellent work being done by our team over in https://github.com/executablebooks/jupyterlab-myst before the jupyter format meeting next week.

cc @agoose77


rowanc1:
Change & help wanted on this page and in the linked repo:
- https://github.com/executablebooks/jupyterlab-myst-quickstart

Going to push ahead for now.

agoose77:
This is looking fantastic. We can also get a jupyterlite distribution working if you think it would work well in this format?

rowanc1:
So into that, I am not quite sure where to start there, so very very open to pointers!!

Another thing that @stevejpurves is working on is the thebe integrations, which will bring this work back into the myst websites and JupyterBook. That should be using JupyterLite, but a headless version.

agoose77:
Right! The idea behind using jupyterlite is that we can embed a JupyterLite distribution with jupyterlab-myst installed, so that people can try it out without launching Binder!

PR #264, opened 2023-02-24
🐛 Resolution & rendering of citations

fwkoch:


packages/tex-to-myst/src/citations.ts
```diff
@@ -13,7 +13,7 @@ function createCitation(
   state.openParagraph();
   const value = texToText(getArguments(node, 'group'));
   const citations = value.split(',').map((l) => l.trim());
-  const isGroup = citations.length > 1;
+  const isGroup = true; // citations.length > 1;
```
fwkoch, line 16: This just makes all citations `citeGroup` nodes; myst theme components were not correctly rendering individual parenthetical `cite` nodes. Probably this should be fixed here eventually: https://github.com/executablebooks/myst-theme/blob/main/packages/myst-to-react/src/cite.tsx#L27 but for now, using cite groups is fine.

Issue #283, opened 2023-03-01
Support for TIFF images in conversion


rowanc1:
TIFF images don't work in many web browsers and need to be converted to some other format, e.g. PNG.

This is in the same code area as #282.

PR #285, opened 2023-03-01
🌠 Support TIFF images

rowanc1:
See #283.


fwkoch (approved):
I gave this a spin and it seems to work fine. Looking at this PR and the entire transform itself, this code is in desperate need of a refactor, but that can be tackled separately. Issue for that is here: https://github.com/executablebooks/mystjs/issues/287

PR #286, opened 2023-03-01
Prevent rebuild loop

fwkoch:
This attempts to address https://github.com/executablebooks/mystjs/issues/266 (though the repeated site build is difficult to reliably reproduce, so we will see if this helps over time).

This PR starts by adding a simple check: `If we are currently reprocessing and a new reprocess is requested, do not reprocess again.`

However, so intentional changes are not lost, we need one more step: `If we are currently reprocessing and a new reprocess is requested, do not reprocess again right away, but *do* reprocess one more time when we are done.`

As we start considering more and more simultaneous reprocess requests, it gets harder to keep track of what exactly needs to be processed. For example, while reprocessing `file1`, if requests to reprocess `file2` and `file3` do we keep track of all of them...? To avoid this complexity, if a reprocess is requested, we simply reprocess _everything_ once the initial reprocess is complete.

We will see if this helps or just leads to more problems. As far as I can tell, there is no regression in the "happy-path" reprocessing behaviour. 😅

### DIFF
--- a/packages/tex-to-myst/src/tables.ts
+++ b/packages/tex-to-myst/src/tables.ts
@@ -127,7 +127,7 @@ export const TABLE_HANDLERS: Record<string, Handler> = {
       currentNode.rowspan = rowspan;
     }
     state.renderChildren(node.args[node.args.length - 1]);
-    state.closeNode();
+    state.closeParagraph();
   },
   macro_multicolumn(node, state) {
     // This macro is defined as:
--- a/.changeset/early-carpets-sleep.md
+++ b/.changeset/early-carpets-sleep.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Prevent unwanted repeated site reload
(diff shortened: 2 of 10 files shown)


packages/tex-to-myst/src/tables.ts
```diff
@@ -127,7 +127,7 @@ export const TABLE_HANDLERS: Record<string, Handler> = {
       currentNode.rowspan = rowspan;
     }
     state.renderChildren(node.args[node.args.length - 1]);
-    state.closeNode();
+    state.closeParagraph();
```
fwkoch, line 130: This PR also sneaks in a small bug fix with correctly closing multirow macro. It is closed differently than multicolumn macro below.

rowanc1:
I tried hard to break this. From the code it is certainly an improvement, and I _think_ that it should address the problem, I couldn't get it recreated in trying in multiple projects that I have seen this before.

Issue #284, opened 2023-03-01
Myst parser does not handle the `code-cell` directive


stevejpurves:
A `code-cell` directive is used in MyST notebook. 

This is defined [here in the jupyterbook docs](https://jupyterbook.org/en/stable/file-types/myst-notebooks.html#create-a-myst-notebook-with-jupytext) in terms of converting from `.ipynb`  to MyST `.md` using `juyptertext` that appears to be where code-cell is defined?

currently mysjs does not parse this directive, byt myst-nb does. Being able to parse and support existing MyST notebooks would be very cool.

stevejpurves:
This probably also means extending the docs in terms of adding a subsection on the directive, and a section on working with MyST notebooks. In fact, if this is a migration pathway for people onto the js cli, then perhaps there should be a guide?

rowanc1:
This was released in 0.1.18! 🚀

PR #289, opened 2023-03-03
⌨️ Add code-cell directive

rowanc1:
See #284

Follows the implementation here: https://jupyterbook.org/en/stable/file-types/myst-notebooks.html


packages/myst-directives/src/code.ts
```diff
@@ -137,3 +139,53 @@ export const codeBlockDirective: DirectiveSpec = {
...
+    const code: Code = {
+      type: 'code',
+      lang: data.arg as string,
+      executable: true,
+      value: data.body as string,
+    };
+    let tags: string[] | undefined;
+    // TODO: this validation should be done in a different place
+    // For example, specifying that the attribute is YAML,
+    // and providing a custom validation on the option.
```
rowanc1, line 167: @fwkoch I think picking this up in a validation pass on the directives would be very nice. :)

PR #291, opened 2023-03-08
docs: add warning about spec

agoose77:
This PR starts to add a few admonitions that indicate where syntax is not in the MyST spec. I think the idea here would be that over time we make MEPs to introduce this syntax, and we gradually remove the warnings :)

Thoughts?

### DIFF
--- a/docs/cross-references.md
+++ b/docs/cross-references.md
@@ -171,6 +171,12 @@ See [](#my-math-label) for an equation!
 
 `````{tip}
 # Labelling dollar math
+
+
+````{warning}
+The following syntax for labelling dollar-math equations is in `beta`, as it is not yet described by the [MyST Specification](https://myst-spec.readthedocs.io/en/latest/index.html). As such, it may change in the future.
+````
+
 You can also use the dollar-math label, which is included in after an equation with `$$ (label)`.\
 This can even all be on a single line!
 ````{myst}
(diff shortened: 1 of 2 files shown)


docs/cross-references.md
```diff
@@ -171,6 +171,12 @@ See [](#my-math-label) for an equation!
 
 `````{tip}
 # Labelling dollar math
+
+
+````{warning}
+The following syntax for labelling dollar-math equations is in `beta`, as it is not yet described by the [MyST Specification](https://myst-spec.readthedocs.io/en/latest/index.html). As such, it may change in the future.
+````
```
rowanc1, line 178: I think that we might actually just want remove dollar math labels from the documentation, there are other ways to label that are more consistent (i.e. a directive or a target assignment `()=`)

agoose77:
@rowanc1 I've updated the spec link to the internal myst-tools, and removed the mention of dollar-math labels entirely. Is that better?

BTW, feel free to edit as you see fit. I'll pop back to this in the near future.

rowanc1:
Nice -- I think this is honestly pretty close.

There need to be a few other things that we should clean up in just removing the `(label)` from the docs. For example:

image

image

I think updating the third example to use latex math labels still showcases the points well:

```
$$
\label{maxwell}
\begin{aligned}
\nabla \times \vec{e}+\frac{\partial \vec{b}}{\partial t}&=0 \\
\nabla \times \vec{h}-\vec{j}&=\vec{s}\_{e}
\end{aligned}
$$

$$ Ax=b \label{one-liner} $$

See [](#maxwell) for enlightenment and [](#one-liner) to do things on one line!
```

agoose77:
image

I'm going to log off and become a hermit crab.

rowanc1:
I had to stare at that for a while to convince myself!!!

Thank you for the doc fixes, appreciate the help, looking forward to more in the future! 🚀

rowanc1:
Live:
https://myst-tools.org/docs/mystjs/math#dollar-math-equations

Thank you!

PR #303, opened 2023-03-14
✨ New package: myst-to-md

fwkoch:
`myst-to-md` is a new plugin to convert a MyST mdast tree to MyST markdown. It extends [mdast-util-to-markdown](https://github.com/syntax-tree/mdast-util-to-markdown) to include MyST features (roles, directives, blocks, crossReferences, frontmatter, etc) and gentler error logging (rather than total failure on invalid node and other validation errors).


packages/myst-to-md/src/misc.ts
```diff
@@ -0,0 +1,56 @@
+import type { Handle, Info, State } from 'mdast-util-to-markdown';
+import { fileWarn } from 'myst-common';
+import type { VFile } from 'vfile';
+import type { Parent, Validator } from './types';
+
+function comment(node: any): string {
+  return `% ${node.value}`;
```
rowanc1, line 7: Do we do any new line handling here?

rowanc1 (commented):
I mostly focused on the tests in the review, this is good to merge as is as there are not dependencies yet used, and we can start to pick this up as we integrate it in and use it more!

Nice work @fwkoch -- huge amount of work here. 🚀

packages/myst-to-md/src/misc.ts
rowanc1: I think I was maybe referring to `node.value = 'something with \n newelines'`, I don't think this can happen easily, so probably not important at the moment!

PR #307, opened 2023-03-15
Add option to copy template files on export and write relative paths

### LINKED ISSUE #301: absolute paths to files elsewhere on users system found in latex `_build`
### What version of `mystjs` are you using?

0.1.17 + dev

### How did you install myst?

npm

### What operating system are you using?

Mac

### Which area is this feature request for?

Export: LaTeX or PDF

### Describe the Bug

If I have myst latex template that has `frontmatter` options that are expected to be file paths, these will be resolved as absolute paths on the user's system when a build is made.

i.e. an excerpt from the `template.yml`

```
options:
  - type: file
    id: coverpage_image
    description: >
      Path to an image that is used on the report's cover page. This will should
      be a high quality image as it will occupy the upper two thirds of the
      first page.
    required: false
  - type: file
    id: partners_image
    description: |
      Path to an image containing partner organisation logos and relted text.
    required: true
  - type: file
    id: support_image
    description: >
      Path to an image containing logos and test for supporting and financing
      organisatons.
    required: true
```

usage of the template options in `template.tex`

```
...
\partners{[-options.partners_image-]}
\support{[-options.support_image-]}
...
```

part of the frontmatter section from the `main.md` file.

```
  - format: tex
    template: ../templates/report
    partners_image: partners.png
    support_image: logo.png
```

produces the following lines in the exported `main.tex` file.

```
...
\partners{/Users/username/dev/org/report/partners.png}
\support{/Users/username/dev/org/report//logo.png}
...
```

### Expected Behavior

I would expect the paths to remain relative and the files are copied into place, such that the contents of `_build/export/main` folder remains portable and can be zipped up and distributed.

### To Reproduce

_No response_

fwkoch:
This should resolve #301 

When exporting tex, all files specified as options in `template.yml` are copied to a folder adjacent to the content (where other static images, etc are saved for the export). The paths written to the output file are now relative as well so the tex content is portable.


rowanc1 (commented):
Nice, lots of drilling session to the right function!

fwkoch:
Yeah - I guess we probably want to move this to passing around the `vfile` and logging to that, rather than the session.

But for now, validation-specific warnings/errors are not yet tied to vfile, so this is consistent...

stevejpurves:
cool - just tested this on the case that threw up the original issue --- works nicely!

PR #315, opened 2023-03-17
🏗 Stop hoisting deep headings into frontmatter title

### LINKED ISSUE #281: Frontmatter should be more careful about hoisting header
In jupyterlab-myst we got a bug report from @tavin on how we are pulling frontmatter together, currently if it doesn't find a YAML block, it will try to find a header from the content, this leads to unexpected results.

We should instead either make that much picker (first node, h1 only) or turn it off in most cases. See:

https://github.com/executablebooks/jupyterlab-myst/issues/61

Thanks @tavin for the report!

tavin:
Fixes #281 maybe?

@rowanc1 I have no idea if someone needs this behavior. Obviously it could be controlled by a switch. For me it seems wrong to dig beyond the first H1 though.

### DIFF
--- a/packages/myst-transforms/src/frontmatter.spec.ts
+++ b/packages/myst-transforms/src/frontmatter.spec.ts
@@ -283,43 +283,6 @@ describe('getFrontmatter', () => {
     });
     expect(frontmatter).toEqual({ title: 'My Title' });
   });
-  it('title is extracted from any heading if not in frontmatter', () => {
-    const input = {
-      type: 'root',
-      children: [
-        {
-          type: 'code',
-          lang: 'yaml',
-          value: 'subtitle: My Subtitle',
-        },
-        {
-          type: 'text',
-          value: 'hello',
-        },
-        {
-          type: 'text',
-          value: 'hello',
-        },
-        {
-          type: 'heading',
-          depth: 5,
-          children: [
-            {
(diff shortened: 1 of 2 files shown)


rowanc1:
Thanks for the PR @tavin, sorry it took me a few days to get to this!

I think that you are right, we shouldn't be looking through the first title/node in the document to create the frontmatter. The thinking here is that it was better to have some title than nothing (e.g. for the sidebar or the tab-title). But I think based on the behaviour in jupyterlab-myst especially, that doesn't really make sense. If users want a title -- put it in the yaml block or as the first node!

fwkoch:
I agree that this makes sense. Just to clarify my impression:
- Explicit title in the frontmatter is the preferred source of truth.
- If there is no title in the frontmatter, only a header (H1-only, even?) that falls before any paragraph or other flow content may be implicitly interpreted as title. (This more nuanced implicit title is outside the scope of this PR.)
- This implicit title will never be explicitly removed from the document and added to frontmatter. (This is the scope of this PR.)

tavin:
@rowanc1 I removed the test case for this behavior and I think you're clear to rerun .

rowanc1:
Thanks @tavin for looking into this and the other PRs, appreciate the help. :)

🚀

PR #317, opened 2023-03-17
🎆 Refactor image conversion transform for extensibility and code deduplication

fwkoch:
Previously the `transformImageFormats` function was a big pile of if statements for each extension type, repeated code, and opaque logic for prioritizing conversion preference.

Now, the conversion functions and preferred conversion orders are provided declaratively with centralized logic.

Addresses https://github.com/executablebooks/mystjs/issues/287


packages/myst-cli/src/transforms/images.ts
```diff
@@ -196,257 +195,254 @@ export async function transformImages(
...
+  return null;
+}
+
+/**
+ * These are all the available image conversion functions
+ *
+ * Get the function to convert from one extension to another with
+ * conversionFnLookup[from][to]
+ */
+const conversionFnLookup: Record<string, Record<string, ConversionFn>> = {
```
fwkoch, line 293: When implementing a new conversion, you just need to implement the conversion function itself then add it here.

rowanc1 (approved):
Tested locally and added a bunch of docs!

fwkoch:
Thanks, docs look good!

PR #322, opened 2023-03-21
✅ Parse tasklists to an AST

rowanc1:
This parses tasklists into an AST, previously they were going in as HTML. This adds a `checked` property to the AST.

image

### DIFF
--- a/.changeset/happy-hornets-impress.md
+++ b/.changeset/happy-hornets-impress.md
@@ -1,0 +1,5 @@
+---
+'myst-parser': patch
+---
+
+Update readme example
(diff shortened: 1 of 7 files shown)


fwkoch (approved):
Nice, definitely better than leaving the HTML. It's a pretty gross path to get here... markdown -> tokens with "task list" classes -> child html for the checkboxes -> regex on that html to determine attributes to lift up to the list item -> delete html and fix up list item text. This is a case where working at the AST level is more complex than if we just modified the parsing logic itself - but that logic isn't very accessible.

Anyway, this works great, and the test cases are there!

rowanc1:
Yeah, I looked at unified here and .. it is just super easy. :)

I think that the test cases are the most important here for catching things if we change in the future!

PR #328, opened 2023-03-27
👩‍🔬 Inline citation improvements

rowanc1:
- Allow macros in tex parsing
- Prefix and suffix for inline citations

cite-prefix-suffix

### DIFF
--- a/packages/tex-to-myst/src/citations.ts
+++ b/packages/tex-to-myst/src/citations.ts
@@ -17,15 +17,6 @@ function createCitation(
   const type: string = isGroup ? 'citeGroup' : 'cite';
   const cite = u(type) as GenericNode;
   if (kind) cite.kind = kind;
-  const args = getArguments(node, 'argument');
-  if (args.length > 0) {
-    // If there is only one argument, it is the suffix
-    const [prefix, suffix] = args.length === 1 ? [undefined, args[0]] : args;
-    const prefixText = texToText(prefix).trim();
-    const suffixText = texToText(suffix).trim();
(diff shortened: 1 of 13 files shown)


packages/tex-to-myst/src/citations.ts
```diff
@@ -34,6 +25,16 @@ function createCitation(
   } else {
     cite.label = citations[0];
   }
+  const args = getArguments(node, 'argument');
+  if (args.length > 0) {
+    // If there is only one argument, it is the suffix
+    const [prefix, suffix] = args.length === 1 ? [undefined, args[0]] : args;
+    const prefixText = texToText(prefix).trim();
+    const suffixText = texToText(suffix).trim();
+    const citation = isGroup ? cite.children?.[0] : cite;
```
fwkoch, line 34: This feels a little funny if there are multiple citations and prefix/suffix, but maybe this is just the expected behaviour and not a problem.

fwkoch (approved):
This looks nice! Clean and pretty straightforward, prefix/suffix on `Cite` makes sense to me.

packages/tex-to-myst/src/citations.ts
rowanc1: Yep, you are correct. This should be on the first and last child of the group:

image

image

Will fix that up.

rowanc1: Ok, pushed. I think there are actually quite a few intricacies when you get into textual/narrative citations. Where the prefix is repeated, but not the suffix. I am going to leave as is for now, and fix when we get complaints. Lots of weird work-arounds on the internet for this type of stuff.

image

image

PR #331, opened 2023-03-28
🏷 Add title/subtitle to JATS

rowanc1:
For example:
```xml
<title-group>
  <article-title>
    Systematic review of day hospital care for elderly people
  </article-title>
</title-group>
```

See https://jats.nlm.nih.gov/archiving/tag-library/1.3/element/article-title.html

### DIFF
--- a/.changeset/calm-llamas-fly.md
+++ b/.changeset/calm-llamas-fly.md
@@ -1,0 +1,5 @@
+---
+'myst-to-jats': patch
+---
+
+Add subtitle and title to JATS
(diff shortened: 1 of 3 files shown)


fwkoch:
This is another point where getting the title from the content could come up if it's not defined in frontmatter (i.e. first H1)... but I think in this case where a title isn't mandatory, we just need to look in frontmatter. Keep structured metadata all the way through.

rowanc1:
Sounds good! Also hopefully pulling that title out could happen in a transform before this export!

PR #332, opened 2023-03-28
💻 Extending project frontmatter to accept `thebe` options

stevejpurves:
The `thebe`, `thebe-core` & `thebe-react` libraries accepts a number of options allowing connections to compute resources and rendering options to be configured. 

These options are currently documented [here](https://thebe-core.curve.space/docs-core/a-configuration) and [here](https://thebe-core.curve.space/docs-core/coreoptions).

This PR adds support for including these options in project frontmatter with the associated validation.

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/types.ts
+++ b/packages/myst-frontmatter/src/frontmatter/types.ts
@@ -26,41 +26,41 @@ export type Biblio = {
 };
 
 export type Thebe = {
-  useBinder: boolean;
-  useJupyterLite: boolean;
-  requestkernel: boolean;
-  binderOptions: ThebeBinderOptions;
-  serverSettings: ThebeServerSettings;
-  kernelOptions: ThebeKernelOptions;
-  savedSessionOptions: ThebeSavedSessionOptions;
-  mathjaxConfig: string;
-  methjaxUrl: string;
+  useBinder?: boolean;
+  useJupyterLite?: boolean;
+  requestKernel?: boolean;
+  binderOptions?: ThebeBinderOptions;
+  serverSettings?: ThebeServerSettings;
+  kernelOptions?: ThebeKernelOptions;
+  savedSessionOptions?: ThebeSavedSessionOptions;
+  mathjaxConfig?: string;
+  methjaxUrl?: string;
 };
 
 export type ThebeBinderOptions = {
-  binderUrl: string;
-  ref: string;
-  repo: string;
-  repoProvider: string;
+  binderUrl?: string;
+  ref?: string;
+  repo?: string;
+  repoProvider?: string;
 };
 
 export type ThebeServerSettings = {
-  baseUrl: string;
-  token: string;
-  wsUrl: string;
-  appendToken: boolean;
+  baseUrl?: string;
+  token?: string;
+  wsUrl?: string;
+  appendToken?: boolean;
 };
 
 export type ThebeKernelOptions = {
-  kernelName: string;
-  name: string;
-  path: string;
+  kernelName?: string;
+  name?: string;
+  path?: string;
 };
 
 export type ThebeSavedSessionOptions = {
-  enabled: boolean;
-  maxAge: number;
-  storagePrefix: string;
+  enabled?: boolean;
+  maxAge?: string | number;
+  storagePrefix?: string;
 };
 
 export type Numbering = {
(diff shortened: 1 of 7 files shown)


packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -1,4 +1,5 @@
 import type { CreditRole } from 'credit-roles';
+import { SupportOptionRange } from 'prettier';
```
fwkoch, line 2: Should this be here?

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -24,6 +25,44 @@ export type Biblio = {
   last_page?: string | number;
 };
 
+export type Thebe = {
+  useBinder?: boolean;
```
fwkoch, line 29: Can you use both `binder` and `jupyter-lite`, i.e. can both of these be `true`?

fwkoch: Ah, I see, all of these come directly from the `thebe` options, so they probably should continue to match that.

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -24,6 +25,44 @@ export type Biblio = {
...
+  serverSettings?: ThebeServerSettings;
+  kernelOptions?: ThebeKernelOptions;
+  savedSessionOptions?: ThebeSavedSessionOptions;
+  mathjaxConfig?: string;
+  mathjaxUrl?: string;
+};
+
+export type ThebeBinderOptions = {
+  binderUrl?: string;
+  ref?: string;
```
fwkoch, line 42: This is the `branch`? Probably leave it matching the thebe options, but it's not clear what this is without looking at the docs.

fwkoch (approved):
This looks good - lots of boilerplate to do the validation, but that's just following the pattern. All the options follow directly from `thebe` options, which is nice.

packages/myst-frontmatter/src/frontmatter/types.ts
stevejpurves: can be any git `ref` including `HEAD`, which also includes tags & commits?
in terms of naming, it pulls directly from the language binder uses so you need to be familar with that, but I think that's ok
image

packages/myst-frontmatter/src/frontmatter/types.ts
stevejpurves: Yes both can be true, and from `thebe-react` binder would win. probably something that could be made better but the `useBinder` is a legacy thebe option.

stevejpurves: @fwkoch redefining the types here is right? i mean rather than adding `thebe-core` as a type only dependency?

packages/myst-frontmatter/src/frontmatter/types.ts
stevejpurves: looks like a VSCode overzealous-auto-add! removing.

packages/myst-frontmatter/src/frontmatter/types.ts
fwkoch: `myst-frontmatter` is quite lightweight and imported in many places, so we do not want to import `thebe-core` here.

You could go the other way - make this the source of truth for the types and add `myst-frontmatter` as a dependency to `thebe-core`? Or just redefine them - doing that gives some flexibility to be different (maybe useful if, say, the theme is setting a thebe option that we do not want frontmatter to control, or something...).

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -87,6 +125,7 @@ export type SiteFrontmatter = {
   keywords?: string[];
 };
 
+// TODO extend
```
rowanc1, line 128:
```suggestion
```

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -458,6 +609,7 @@ export function validateGithubUrl(value: any, opts: ValidationOptions) {
   });
 }
 
+// TODO extend
```
rowanc1, line 612:
```suggestion
```

rowanc1 (commented):
Nice! Just cleaning up a few comments.

PR #334, opened 2023-03-29
📒 Track notebook source for embedded content

fwkoch:
`figure`/`embed` directives may embed code/output cells from other notebooks. To recompute these, we need to know where they come from.

This PR adds url `source` and `sourceKey` to all embed nodes to locate the source notebook cell. This also adds `dependencies` at the top level of the page config with all url sources required to compute embedded content in the page.

### DIFF
--- a/packages/myst-cli/src/process/site.ts
+++ b/packages/myst-cli/src/process/site.ts
@@ -179,7 +179,7 @@ export async function writeFile(
   const toc = tic();
   const selectedFile = selectFile(session, file);
   if (!selectedFile) return;
-  const { frontmatter, ...mdastPost } = selectedFile;
+  const { frontmatter, mdast, kind, sha256, slug, references, dependencies } = selectedFile;
   const exports = await Promise.all([
     resolvePageSource(session, file),
     ...(await resolvePageExports(session, file, projectPath)),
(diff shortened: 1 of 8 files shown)


packages/myst-cli/src/process/site.ts
```diff
@@ -179,7 +179,7 @@ export async function writeFile(
   const toc = tic();
   const selectedFile = selectFile(session, file);
   if (!selectedFile) return;
-  const { frontmatter, ...mdastPost } = selectedFile;
+  const { frontmatter, mdast, kind, sha256, slug, references, dependencies } = selectedFile;
```
fwkoch, line 182: This is just removing `file` from the written page config. I think this is ok? Previously this was writing local file paths... not useful at all for the site. (The original file is still available as a download in the export lists.)

rowanc1: My thought on `file` is that it will be helpful when we sync up with an 'Edit on GitHub' which will require a local file path.

rowanc1:
Code changes look good. I think I would prefer to keep `file` in the exported config, but we can also put this when we need it if you feel strongly.

Another note:
We also need to traverse the URLs and remap them in the theme CDN code. I think that code might actually be better in common.

packages/myst-cli/src/process/site.ts
stevejpurves: ok so sounds like this is one mini change to revert!

rowanc1: In talking through with @fwkoch I think we should actually remove `file`, it is non-relative at the moment. We can bring this back soon with a feature of `edit on github`!

PR #335, opened 2023-03-29
🌀 Change to `SourceFileKind` (from KINDS)

rowanc1:
Move enum to myst-common

### DIFF
--- a/packages/myst-common/src/index.ts
+++ b/packages/myst-common/src/index.ts
@@ -13,14 +13,14 @@ export {
 } from './utils';
 export { selectBlockParts, extractPart } from './extractParts';
 export { TemplateKind, TemplateOptionType } from './templates';
-export { ParseTypesEnum } from './types';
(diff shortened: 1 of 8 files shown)


rowanc1:
@stevejpurves this should make the kind available to thebe (or that part of the theme), without importing from the CLI.

fwkoch (approved):
Looks good - just moving around and renaming.

packages/myst-common/src/index.ts
```diff
@@ -13,14 +13,14 @@ export {
...
-export { ParseTypesEnum } from './types';
+export { ParseTypesEnum, SourceFileKind } from './types';
 
 export type { MessageInfo } from './utils';
 export type {
+  Dependency,
   GenericNode,
   GenericParent,
   Citations,
-  Footnotes,
```
rowanc1, line 23: This shouldn't need to be used anymore!

PR #338, opened 2023-03-30
⛹️ always include outputs with code cells

stevejpurves:
This always adds an output array to blocks that represent `code-cell`s in the ast, even when the outputs array is empty. This is consistent with `ipynb` format, where the `outputs` array is always present.


rowanc1:
Nice, looks good.

PR #339, opened 2023-03-30
🐛 Transform Wikipedia links properly

MridulS:
This fixes https://github.com/executablebooks/mystjs/issues/224 (maybe in a bit hacky way).

The wiki link string was being manipulated by adding zero width space by https://github.com/executablebooks/mystjs/blob/8615f221874abb8542176ceeb0e8f8e2da42777e/packages/myst-transforms/src/links/plugin.ts#L20 
which in turn messed up updates to the text.

Also fixed a regex, so `The_big_bang_theory`, correctly transforms to `The big bang theory` and not `The big_bang_theory`.

### DIFF
--- a/packages/myst-transforms/src/links/wiki.ts
+++ b/packages/myst-transforms/src/links/wiki.ts
@@ -88,7 +88,7 @@ export class WikiTransformer implements LinkTransformer {
       lang: result.lang,
     };
     link.internal = false;
-    const title = page.replace(/_/, ' ');
+    const title = page.replace(/_/g, ' ');
     updateLinkTextIfEmpty(link, title);
     return true;
   }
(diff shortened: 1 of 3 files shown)


packages/myst-transforms/src/links/wiki.ts
```diff
@@ -88,7 +88,7 @@ export class WikiTransformer implements LinkTransformer {
       lang: result.lang,
     };
     link.internal = false;
-    const title = page.replace(/_/, ' ');
+    const title = page.replace(/_/g, ' ');
```
rowanc1, line 91: Wonderful. Good catch. :)

Thank you for updating the tests!

rowanc1 (approved):
This looks good!

rowanc1:
Will merge in when the tests pass! Excited to have this fixed up, thank you so much for your help @MridulS!!

PR #345, opened 2023-04-06
🗂 Begin to support multipage export

fwkoch:
- JATS now allows multi article export with `article` and `sub-article`
- `myst-cli` allow specifying `article` and `sub_articles` in `myst-frontmatter` export options - sub-articles only are used for JATS export
- File exports may now be defined in the project config in `myst.yml`
- This PR also deduplicates some of the export option resolution code. It's still a bit messy, but at least not messy and duplicated...
- It also add an end-to-end cli test! We need more of these!


rowanc1:
I have gone through and tested this, I think it is a massive step forward on the exports front, and I haven't found any regressions in my exploratory testing!

I will compile a different issue of some of the other JATS-related things that we need to clean up over the coming months. This PR is good to go in though!

packages/myst-cli/src/build/utils/getFileContent.ts
```diff
@@ -0,0 +1,87 @@
...
+  // Consolidate all citations onto single project citation renderer
+  combineProjectCitationRenderers(session, projectPath);
+
+  const extraTransforms: TransformFn[] = [];
+  if (imageExtensions.includes(ImageExtensions.webp)) {
+    extraTransforms.push(transformWebp);
+  }
+  // if (opts?.extraTransforms) {
+  //   extraTransforms.push(...opts.extraTransforms);
+  // }
```
rowanc1, line 56: Should this explicitly be commented out?

packages/myst-to-jats/tests/basic.spec.ts
```diff
@@ -29,19 +34,22 @@ describe('Basic JATS body', () => {
     const pipe = unified().use(mystToJats);
     pipe.runSync(tree as any);
     const vfile = pipe.stringify(tree as any);
-    expect((vfile.result as JatsResult).value).toEqual(jats);
+    expect(vfile.result).toEqual(jats);
```
rowanc1, line 37: Note to self: we will have to update the JATS demo in the website/theme.

rowanc1 (commented):
Taking a final look through. :)

PR #354, opened 2023-04-17
🎉 Add `siunitx` package to latex, JATS and MyST

rowanc1:
This implements the majority of the `siunitx` package in latex:

https://texdoc.org/serve/siunitx/0

This also creates alt-text and keeps other information about numbers and units in the mdast node.


agoose77:
I think you're right that the angle-bracket syntax is the most MyST-like. In my view, sometimes the problem is just verbose, rather than the markup. I think the recent MEP for links is a useful one; links are already implemented in Markdown, and a notion of a scheme applies well to our use case. However, in this case, we don't have any prior art to improve upon, so going with something ReST-like is probably the right call?

With an si-unit parser, the verbose `` {si}`5 <\kilo\gram\metre\per\square\second>` `` becomes `` {si}`5 <kg.m/s^2>` ``, which is very readable.

rowanc1:
Love it. Ok, I will go on this path and let you know where I end up!! (Will also open an issue to track this!)

PR #356, opened 2023-04-18
🧐 Add JATS DTD validation to myst-to-jats tests

fwkoch:
This also fixes up a few instances of previously-invalid JATS XML including:
- `contrib-group` now wraps contribs
- `front` with `article-meta` is always present
- `element-citations` are never empty

Also this PR adds XML declaration and doctype headers.


rowanc1:
This looks good -- lots of little improvements to the JATS export, and now actually testing against the DTD!

PR #361, opened 2023-04-21
📒 Handle notebook cell ids and outputs in jats

fwkoch:
- Give notebook cell `blocks` an id and type (code or content)
- Give notebook cell code/output ids
- Make code cells `sec` elements in JATS
- Add cell outputs to JATS


rowanc1:
Tested out locally, this is a massive improvement!

PR #364, opened 2023-04-24
🤖 Specify output from the command line

rowanc1:
This allows us to directly specify the output from the CLI, it only allows you to have a single file if you are specifying the output. It also infers the export type and turns on `--force`.

```
myst build example.ipynb -o example.xml
```

### DIFF
--- a/packages/myst-cli/src/build/build.ts
+++ b/packages/myst-cli/src/build/build.ts
@@ -17,18 +17,19 @@ export type BuildOpts = {
   xml?: boolean;
   all?: boolean;
   force?: boolean;
+  output?: string;
   checkLinks?: boolean;
 };
 
-export function getExportFormats(opts: BuildOpts & { explicit?: boolean }) {
-  const { docx, pdf, tex, xml, all, explicit } = opts;
+export function getExportFormats(opts: BuildOpts & { explicit?: boolean; extension?: string }) {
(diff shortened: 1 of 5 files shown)


packages/myst-cli/src/build/build.ts
```diff
@@ -54,14 +55,33 @@ export async function collectAllBuildExportOptions(
...
+  if (output && formats.length !== 1) {
+    throw new Error(`Unrecognized file extension for output: ${path.extname(output)}`);
+  }
   session.log.debug(`Exporting formats: "${formats.join('", "')}"`);
   let exportOptionsList: ExportWithInputOutput[];
   if (files.length) {
     exportOptionsList = await collectExportOptions(session, files, formats, {
-      force,
+      // If there is an output and file specified, force is implied
+      force: force || !!output,
```
fwkoch, line 75: Do we also want implied `force` if, like `(files.length === 1 && (opts.docx || opts.pdf || opts.tex || opts.xml))`? I think that would be the case for `myst build index.md --jats`

fwkoch (approved):
One more case where we can add implied `force` I think, but other than that, looks good.

packages/myst-cli/src/build/build.ts
rowanc1: Updated!

PR #365, opened 2023-04-24
🧪 JATS Validation

rowanc1:
This improves some minor things for jats validation (nesting in paragraphs mostly!!).


packages/myst-spec-ext/src/types.ts
```diff
@@ -32,11 +32,13 @@ export type CaptionNumber = Parent & {
 export type FootnoteDefinition = FND & {
   /** @deprecated this should be enumerator */
```
fwkoch, line 33: 🙌 Nice to act on this.

fwkoch (approved):

PR #367, opened 2023-04-26
📂 Update thebe frontmatter options

fwkoch:
This updates thebe options in frontmatter to (1) remove unnecessary advanced options, (2) add local config options, and (3) allow some option objects to also be overridden with true/false.

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
+++ b/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
@@ -55,7 +55,7 @@ const TEST_THEBE: Thebe = {
     provider: 'github' as any,
   },
   server: {
-    baseUrl: 'https://my.server.org',
+    url: 'https://my.server.org',
     token: 'legit-secret',
   },
   kernelName: 'python3',
@@ -64,7 +64,7 @@ const TEST_THEBE: Thebe = {
   mathjaxConfig: 'TeX-AMS_CHTML-full,Safe',
   mathjaxUrl: 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js',
   local: {
-    baseUrl: 'http://localhost:8888',
+    url: 'http://localhost:8888',
     token: 'test-secret',
     kernelName: 'python27',
     sessionName: 'another-path',
--- a/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
+++ b/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
@@ -47,25 +47,28 @@ const TEST_BIBLIO: Biblio = {
 };
 
 const TEST_THEBE: Thebe = {
-  useBinder: false,
-  useJupyterLite: false,
-  requestKernel: true,
-  binderOptions: {
-    binderUrl: 'https://my.binder.org/blah',
+  lite: false,
+  binder: {
+    url: 'https://my.binder.org/blah',
     ref: 'HEAD',
     repo: 'my-repo',
-    repoProvider: 'github',
+    provider: 'github',
   },
-  serverSettings: {
-    baseUrl: 'http://localhost:8888',
-    token: 'test-secret',
-    wsUrl: 'ws://localhost:8888',
-    appendToken: true,
+  server: {
+    baseUrl: 'https://my.server.org',
+    token: 'legit-secret',
   },
-  kernelOptions: { kernelName: 'python 3', name: 'python3', path: 'some-path' },
-  savedSessionOptions: { enabled: true, maxAge: 383000, storagePrefix: 'thebe' },
+  kernelName: 'python3',
+  sessionName: 'some-path',
+  disableSessionSaving: true,
   mathjaxConfig: 'TeX-AMS_CHTML-full,Safe',
   mathjaxUrl: 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js',
+  local: {
+    baseUrl: 'http://localhost:8888',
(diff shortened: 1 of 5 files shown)


packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
```diff
@@ -47,25 +47,28 @@ const TEST_BIBLIO: Biblio = {
...
-    appendToken: true,
+  server: {
+    baseUrl: 'https://my.server.org',
+    token: 'legit-secret',
   },
-  kernelOptions: { kernelName: 'python 3', name: 'python3', path: 'some-path' },
-  savedSessionOptions: { enabled: true, maxAge: 383000, storagePrefix: 'thebe' },
+  kernelName: 'python3',
+  sessionName: 'some-path',
+  disableSessionSaving: true,
```
fwkoch, line 63: I modified this from `sessionSaving` so `undefined` provides an appropriate default (the docs called for default `sessionSaving = true`, so `disableSessionSaving = undefined` is equivalent).

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
```diff
@@ -47,25 +47,28 @@ const TEST_BIBLIO: Biblio = {
...
     repo: 'my-repo',
-    repoProvider: 'github',
+    provider: 'github',
   },
-  serverSettings: {
-    baseUrl: 'http://localhost:8888',
-    token: 'test-secret',
-    wsUrl: 'ws://localhost:8888',
-    appendToken: true,
+  server: {
```
fwkoch, line 57: I modified this from `serverSettings` to just `server`. This is equivalent to `binder` and `local` as opposed to `binderSettings` and `localSettings`.

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
```diff
@@ -47,25 +47,28 @@ const TEST_BIBLIO: Biblio = {
...
+    baseUrl: 'https://my.server.org',
+    token: 'legit-secret',
   },
-  kernelOptions: { kernelName: 'python 3', name: 'python3', path: 'some-path' },
-  savedSessionOptions: { enabled: true, maxAge: 383000, storagePrefix: 'thebe' },
+  kernelName: 'python3',
```
fwkoch, line 61: Should `kernelName` and `sessionName` be defined at the top level here or should the be nested under `server`?

Alternatively, do we even need the `server` key or should we lift `baseUrl` and `token` up a level?

I'm totally fine with this as-is, just asking questions...

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -25,41 +25,32 @@ export type Biblio = {
...
-  useJupyterLite?: boolean;
-  requestKernel?: boolean;
-  binderOptions?: ThebeBinderOptions;
-  serverSettings?: ThebeServerSettings;
-  kernelOptions?: ThebeKernelOptions;
-  savedSessionOptions?: ThebeSavedSessionOptions;
-  mathjaxConfig?: string;
+  lite?: boolean;
+  binder?: boolean | ThebeBinderOptions;
+  server?: boolean | ThebeServerOptions;
```
fwkoch, line 30: Should `server` be allowed to be a boolean? It can't really be `true`; we have to know the URL. And if it's `false`, just leave it undefined.

fwkoch (commented):
@stevejpurves - Here are a few more questions that came up during implementation.

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
stevejpurves: yeah this is good!

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
stevejpurves: I was toying with that one when writing this but yeah there is no good reason to keep the `serverSettings` here - it's a jupyterlab naming convention for that data fragment but no need to keep it here

stevejpurves: so maybe we can go with `url` for `baseUrl` after all too

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
stevejpurves: I defined them at top level as they can equally apply to the `binder` configuration, as well as the `server` configuration.
If `binder` is specified that will win over anything supplied in `server` (actually in `thebe` connecting to binder causes any `serverSettings` provided to be overridden`

packages/myst-frontmatter/src/frontmatter/types.ts
stevejpurves: hmm.... the only `server: true` thing that makes sense if it also defaults to the same configuration as `local`...
not sure we should do that though?

fwkoch: I see, makes enough sense to leave it!

packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
fwkoch: Ok, I'll update to these simpler names.

If we also want to cater to folks intimately familiar with jupyterlab, we could leave these other names as aliases as well 🤷‍♀️ but I guess there's also the `path` vs `sessionName` stuff which may not just be an alias but actually have a slightly different meaning.

PR #368, opened 2023-04-26
🧮 Inline user_expression support

rowanc1:
This supports the eval role in jupyterlab_myst in the CLI and exporting out to formats like JATS, docx, tex, and word for a subset of mimetypes. Lots of work still to expand support for other mimetypes, but a good start!

image


rowanc1:
This is a good start, certainly needs more, but going to merge so we can move the theme forward and get thebe introduced.

PR #377, opened 2023-05-11
🤖 Build static html export

rowanc1:
This creates a new command `myst build --html` which exports the site as a static HTML page that can be hosted on GitHub pages or similar.

**Requires:**
- https://github.com/executablebooks/myst-theme/pull/68

There is a bit more work that needs to be done to actually set up for using this for real, but getting close!


corentinmorice1:
Thanks a lot, this is exactly what we were looking for after hearing the Myst talk at JupyterCon! I just tried this on Gitlab, by moving _build/html/ to public/. No error is raised, but I get a 404 error. Does _build/html/ contain the whole website?

rowanc1:
It is likely that the project is in a nested folder and there is no index page, you can confirm by seeing what is inside of `_build/html` (likely a folder, that then has an `index.html`?!). To fix that, you can remove the `site.projects` list entirely in your `myst.yml`.

```diff
site:
  title: 'Pixels and Their Neighbours: Finite Volume'
-  projects:
-    - slug: pixels-and-their-neighbours
-      path: .
  actions:
    - title: Download PDF
      url: ./cockett-heagy-oldenburg-2016.pdf
      static: true
  logo_text: Pixels and their neighbours
```

This will need the most up to date site templates (as of yesterday), which can be pulled/updated with a `myst clean --templates` and then `myst start`.

I think that up next we should change the default to being `1 site == 1 project`, where as currently the default is that `1 site >= 1 project`, which requires some sort of folder nesting and redirects, which are not preserved in the static build.

Let me know if that works?!

jeremyneveu:
Thanks for the help. So with @corentinmorice1 we created a repo https://gitlab.dsi.universite-paris-saclay.fr/corentin.morice/myst-test from the myst website tutorial, added a index.md file and removed the site.projects section from the `myst.yml` file. Now we are able to get a static website from gitlab pages, but the website is badly shaped

https://corentin.morice.gitlab.dsi.universite-paris-saclay.fr/myst-test/

I have the same building the html on my laptop, so it does not come from gitlab pages process. Something seems missing but what ? Any idea ?

rowanc1:
I think that you need to set the `BASE_URL` environment variable to `/myst-test`.

`export BASE_URL="/myst-test"`

That will need to be updated for any repository, Github actions adding this works:

```
env:
  BASE_URL: /${{ github.event.repository.name }}
```

If you know something similar in gitlab, let me know and I will add to the docs I am writing now!

jeremyneveu:
Thank you for the new doc about deployment. I tried the `npx serve _build/html` command and the web site was nice on my laptop. But I still fail to have a static website through giltab pages, despite your advice. I was able to set `BASE_URL` but it still fails:
```$ mv _build/html/ public/
$ echo $CI_PAGES_URL
https://corentin.morice.gitlab.dsi.universite-paris-saclay.fr/myst-test
$ echo $BASE_URL
/my-test
```
I haven't found an equivalent to the variable you are mentionning in github actions. Also, I don't really understand this sentence in the docs: "The build and site assets are in the /build folder, which would point outside of the current repository to a repository called 'build', which probably doesn't exist!" Maybe if you rephrase it to me I can guess how to fix gitlab pages.

PR #389, opened 2023-05-15
🔗 Add support for github issue previews

rowanc1:
github-previews


rowanc1:
I think that there are many things to improve in the theme side - but I think this is good enough on the parsing (which is all mystjs is doing) to get out the door!

The myst-theme can be done incrementally!

PR #393, opened 2023-05-20
⌨️ Add code block environment in myst-to-tex

finsberg:
In this PR I would like to address https://github.com/executablebooks/mystjs/issues/369

My first thought here is that `verbatim` perhaps should be the default code environment since this might be what is preferred by most journals. Any thoughts?

Also, to begin with I wanted to use the existing options from myst to implement a different code environment. I went for the `class` option so that with in order to specify that you want to use `minted` you should attach the `:class: minted` to the code block, e.g

````
```{code-cell}  python
1 + 2
```
````
will be rendered as
```
\begin{verbatim}
1+2
\end{verbatim}
```
while

````
```{code-cell}  python
:class: minted
1 + 2
```
````
will be rendered as
```
\begin{minted}{python}
1+2
\end{minted}
```
I don't really like to add these "magic" class names, but the alternative I guess is to update the spec for `Code`:
https://github.com/executablebooks/myst-spec/blob/ef03d930d7f7d260f1cd88c3629de35601256dbb/schema/commonmark.schema.json#L186 
Unless you have another idea?

We could also use `identifier` in order to be able to refer to the code block, i.e by adding a label to the code block. It would also be good to add a caption to the code block. I am not sure the best way to do this with the current spec. Any thoughts on this?

### CODE REVIEW

File: packages/myst-to-tex/src/index.ts [hunk 1]
```diff
@@ -80,10 +80,18 @@ const handlers: Record<string, Handler> = {
   definitionDescription(node, state) {
     state.renderChildren(node, true);
   },
-  code(node, state) {
-    state.write('\\begin{verbatim}\n');
+  code(node: Code, state) {
+    let start = '\\begin{verbatim}\n';
+    let end = '\n\\end{verbatim}'
+
+    if (node.class === 'minted' && node.lang !== undefined) {
```
rowanc1, line 87: I think the only improvements I would suggest to this is that the class might be a space-separated (i.e. css) string.

Something similar to:

```typescript
function getClasses(className?: string): string[] {
  const classes =
    className
      ?.split(' ')
      .map((s) => s.trim().toLowerCase())
      .filter((s) => !!s) ?? [];
  return [...new Set(classes)];
}
```

and then a `getClasses(node.class).includes('minted')` should work well for this use case I think.

packages/myst-to-tex/tests/basic.yml
```diff
@@ -265,3 +265,50 @@ cases:
...
+      1+2
+      \end{verbatim}
+  - title: code-block-minted
+    mdast:
+      type: root
+      children:
+        - type: mystDirective
+          name: code-block
+          args: python
+          class: minted
```
rowanc1, line 298: We could update the test a tiny bit here just to add a new class name:

```suggestion
          class: some-class minted
```

rowanc1 (changes requested):
@finsberg this looks great, and is a good way to unblock your use case!

I think that we should still likely consider this or `listings` to be the default code environment in a future PR, as well as having this configuration from the `myst.yml` or frontmatter as well. Both of those changes are complimentary to this though, and we can adapt to any changes relatively easily.

The only change I would request is allowing the class attribute to be a CSS-style string with a list of class names. I added a bit of code from the `myst-theme` that could be used here.

Let me know if you are up for making that improvement, or I can jump in and help.

Thanks for the PR! 🚀

finsberg:
@rowanc1 Thanks for your input. I think this is a good idea. I have changed the PR according to your suggestions.

Let me know if it looks good or if you want additional changes. I also added an option to use `listings` instead of `minted` using en `if-else if` clause. 

I can also give it a try to create a new PR for adding the configuration to the `myst.yml` or frontmatter if this is of interest?

rowanc1:
Thanks @finsberg for the PR -- happy to have another one targeting the `myst.yml`.

We don't currently have anything like that (config options that make it to the writers), so there might be some poking around/discussion to be had (I know that @fwkoch has thoughts there!).

Let me know if you are up for it and we can put together a bit of a roadmap for the feature and try to think through where some of the hard parts are!

I will plan to release this class-based highlighting in the next few days.

finsberg:
> Thanks @finsberg for the PR -- happy to have another one targeting the `myst.yml`.
> 
> We don't currently have anything like that (config options that make it to the writers), so there might be some poking around/discussion to be had (I know that @fwkoch has thoughts there!).
> 
> Let me know if you are up for it and we can put together a bit of a roadmap for the feature and try to think through where some of the hard parts are!
> 
> I will plan to release this class-based highlighting in the next few days.

@rowanc1  Sounds great! I would definitely like to contribute to a feature like that. I really like this project and I think it has a great potential to change the way we write academic papers. I will also start adopting mystjs for my future academic papers and will try to contribute the best I can in order to push this project forward. Thanks again for a great project.

rowanc1:
This is released in v`0.1.31`! Coming to a live demo near you soon 🚀 

code-latex-class

Thanks for the kind words, really excited about MyST, I think it can change a lot of things in science communication!

PR #395, opened 2023-05-26
📦 Add missing package for `myst-transforms`

GYHHAHA:


rowanc1:
Thank you, just changed it to `^3.1.0` to match `myst-parser` and updated the package-lock. Will get it in when the tests pass!

Thanks again for your help!

PR #396, opened 2023-05-26
Proposal: markdown-it-htmyst

tavin:
Hi @rowanc1 as a followup of sorts to https://github.com/executablebooks/markdown-it-docutils/issues/48 and https://github.com/executablebooks/mystjs/issues/181, I'd like to propose adding something like this to mystjs.

It's a first cut at what you said is the "best path".

The code I have so far supports admonitions and prf directives in a rudimentary way. Of course you also need a bit of CSS to see that.

Could we maybe discuss this in a bit more depth? I could find a way to demonstrate it. Right now I can personally confirm it solves the problem of rendering myst directives in e.g. wiki software that currently supports markdown-it... @sneakers-the-rat.

### DIFF
--- a/packages/markdown-it-htmyst/package.json
+++ b/packages/markdown-it-htmyst/package.json
@@ -1,11 +1,11 @@
 {
-  "name": "markdown-it-myst",
+  "name": "markdown-it-htmyst",
   "version": "0.1.3",
   "sideEffects": false,
   "license": "MIT",
-  "description": "markdown-it tokenizer for MyST roles and directives",
-  "author": "Franklin Koch <franklin@curvenote.com>",
-  "homepage": "https://github.com/executablebooks/mystjs/tree/main/packages/markdown-it-myst",
+  "description": "markdown-it HTML hinter for MyST roles and directives",
+  "author": "Tavin Cole <7685034+tavin@users.noreply.github.com>",
+  "homepage": "https://github.com/executablebooks/mystjs/tree/main/packages/markdown-it-htmyst",
   "main": "./dist/cjs/index.js",
   "module": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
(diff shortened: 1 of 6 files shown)


rowanc1:
Nice @tavin -- I think I have a sense of where you are going from looking through the code, and happy to include it. It would be helpful to see a small test or readme example to really see where you are going! It is super helpful to have the `markdown-it --> html` path revived and supported, thanks for your help with that!

@fwkoch @tavin I would be curious on your read if this should be a standalone package or if it belongs with any of the other packages that are already existing?

I suppose we could have an example that revives some minimal css as well, that exists somewhere in the git history.

tavin:
> Nice @tavin -- I think I have a sense of where you are going from looking through the code, and happy to include it. It would be helpful to see a small test or readme example to really see where you are going! It is super helpful to have the `markdown-it --> html` path revived and supported, thanks for your help with that!

Sure I can add something like that. I have been traveling lately but I will have some time over the summer.

> @fwkoch @tavin I would be curious on your read if this should be a standalone package or if it belongs with any of the other packages that are already existing?

I think I can guess why HTML presentation has been separated from markdown-it-myst but nevertheless I also think at least a usable reference implementation should be part of mystjs next to markdown-it-myst. If it's too hard for people to use... they won't. I was initially excited about using myst markdown for some personal projects but have run into a lot of practical obstacles. This would remove one of them.

packages/markdown-it-htmyst/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "markdown-it-htmyst",
+  "version": "0.1.3",
+  "sideEffects": false,
+  "license": "MIT",
+  "description": "markdown-it HTML hinter for MyST roles and directives",
+  "author": "Tavin Cole <7685034+tavin@users.noreply.github.com>",
+  "homepage": "https://github.com/executablebooks/mystjs/tree/main/packages/markdown-it-htmyst",
+  "main": "./dist/cjs/index.js",
```
rowanc1, line 9: In #417 we made quite a few changes for working as ESM packages only (and switching to vitest). I am happy to help out with upgrading here as well when you need it.

rowanc1:
Thanks @tavin -- agree with your sentiment, there needs to be a good default html export. We have focused on React heavily at the moment as well as usage through the CLI, so many of these extensibility and package-reuse parts are certainly underbaked. Looking forward to your help/feedback to make it better! Happy travelling - let us know when you want to pick this up again.

Issue #392, opened 2023-05-19
Extend math macros to enable nested macros


stevejpurves:
### Which area is this feature request for?

LaTeX Export

### Describe the feature you'd like to request

Given the math macros:

```
'\aNrm': "a"
'\aMat': '[\mathrm{\aNrm}]'
```

When only `\aMat` is used in the document body, it seems that only the macro defining `'\aMat': '[\mathrm{\aNrm}]'` is added to the tex preamble, this means that the symbol `\aNrm` is still undefined and causes and error when `\aMat` is used.

### Describe the solution you'd like

Check for nesting of macros and include all dependencies.

### Describe alternatives you've considered

Or we could just add all the defined math macros to the document, even if they are not used.

[originally reported by @alanlujan91]

cc @rowanc1

rowanc1:
Should be released later this week!! Thanks @alanlujan91 for the bug report.

PR #397, opened 2023-05-27
📡 Recursive math commands for LaTeX

rowanc1:
See #392


rowanc1:
@alanlujan91 @stevejpurves this should fix up the math recursion issues you were running into in #392! We will release it sometime this week!

Issue #189, opened 2023-02-07
Audit of Features missing in JupyterBook


rowanc1:
Although mystjs and JupyterBook/sphinx are targeted at slightly different use-cases, there is overlap in creating websites (see [background](https://myst-tools.org/docs/mystjs/background#how-do-jupyter-book-and-mystjs-relate)). Below is a list of missing features/directives/roles that are not yet implemented in `mystjs`, we hope to have parity at the rendering level first, then move on to extensibility (see #181 as a major step).

We will try to keep this updated as we audit more jupyterbooks.

**Directives**

- [ ] example
- [ ] toggle (show option)
- [ ] literalinclude
- [ ] prf
- [ ] only
- [ ] panels
- [ ] epigraph (replicated mostly with quote improvements)
- [ ] glossary / terms
- [ ] contents
- [ ] doc
- [ ] field lists
- [ ] tasks #187 
- [x] girds (grid-item-card)
- [x] cards (header/footer)
- [x] dropdown
- [ ] sphinx proof

**Extensibility**

See #181 as a roadmap.

- [ ] transforms turned on (e.g. dollarmath, etc.)
- [ ] directives / roles

**Execution**

- [ ] Linking artifacts from a notebook (glue) - we are supporting this using ids/imports ([see docs](https://myst-tools.org/docs/mystjs/cross-references#targeting-cells))
- [ ] running code execution for a notebook

**Improvements**

- [ ] code-cell: tags (output_scroll, hide)
- [x] admonition: classes
- [x] admonition: dropdown class
- [x] intersphinx

**UI**
- [ ] TOC - collapse
- [ ] logo more like JB
- [ ] toc footer
- [ ] page footer
- [ ] banner
- [ ] search #100 
- [ ] internationalization #166 

URL:
* Collapsing things down to a `project/slug` isn't how JB works, but probably ok

Export
- [ ] static html #188

Jan-David-Black:
@rowanc1 To implement a feature on this list we need to simultaneously consider LaTeX and the Web output, correct? How could we go about sidebar/margin content? Using \minipage or similar does not really float my boat. At the moment I would be in favor of adding the sidebar content to the appendix, or to add it below the current paragraph as a content box.
What do you think?

rowanc1:
[...]

PR #407, opened 2023-06-02
Propagate block metatags to code node and output node

GYHHAHA:
This is the first part of implementation for jupyter meta tags support, as discussed in [here](https://github.com/executablebooks/myst-theme/issues/50).

After merging this, there are at least 4 things on the list:
- [x] release new myst-cli and myst-spec-ext version
- [ ] change according components in myst-theme
- [ ] create new block component to control the visibility of the whole cell
- [ ] update doc on the executable content section

For this pr, there are still 2 questions, which are mentioned in the comments:
- [x] can we ensure that there will only have single code node and output node in a block node?
- [x] will we raise when `hide-xx` and `remove-xx` appear at the same time?


packages/myst-cli/src/transforms/code.spec.ts
```diff
@@ -114,3 +114,65 @@ describe('liftCodeMetadataToBlock', () => {
...
+
+function build_mdast_by_tags(tags: string[]) {
+  const mdast: any = {
+    type: 'root',
+    children: [
+      {
+        type: 'block',
+        children: [
+          {
+            type: 'code',
```
rowanc1, line 126: I think adding an `executable: true` here is reasonable!

perhaps extending this test to `not` have an output as well, which I think would break things at the moment!

packages/myst-cli/src/transforms/code.ts
```diff
@@ -83,3 +83,48 @@ export function liftCodeMetadataToBlock(session: ISession, filename: string, mda
     }
   });
 }
+
+/**
+ * Traverse mdast, propagate block tags to code and output
+ */
+export function propagateBlockDataToCode(session: ISession, filename: string, mdast: Root) {
+  const blocks = selectAll('block', mdast) as GenericNode[];
```
rowanc1, line 91: I think it is possible that this might grab too many blocks (e.g. the markdown ones, which aren't going to have the code/output). I think we put a kind or something on the block, not sure where that is at the moment!

rowanc1 (commented):
Thanks for the PR, this is looking great! I added a few minor comments about searching through the MDAST and handling a few other cases of the markdown cells, blocks without output (or perhaps code, but that would be weird).

I also think we want to leave tags on the block that we haven't processed here.

rowanc1:
For `will we raise when hide-xx and remove-xx appear at the same time?` I think that we could raise a warning, and then go with the most strict -- i.e. `remove > hide > show`.

packages/myst-cli/src/transforms/code.ts
GYHHAHA: If this is changed, it's reasonable to change the `liftCodeMetadataToBlock` with the same way. This worth a separate pr to do since two tests will be added and their functionality do not have direct link to the meta tag implementation.

packages/myst-cli/src/transforms/code.spec.ts
```diff
@@ -114,3 +114,93 @@ describe('liftCodeMetadataToBlock', () => {
...
+
+// describe('checkMetaTags', () => {
+//   it('duplicate tag warn', async () => {
+//     const consoleSpy = jest.spyOn(console, 'warn');
+//     for (const action of ['hide', 'remove']) {
+//       for (const target of ['cell', 'input', 'output']) {
+//         const tag = `${action}-${target}`;
+//         const mdast = build_mdast([tag, tag], true);
+//         propagateBlockDataToCode(new Session(), '', mdast);
+//         expect(consoleSpy).toHaveBeenCalledWith(`tag '${tag}' is duplicated`);
```
GYHHAHA, line 150: I'm not quite sure how mystjs do warning assertion test. The standard jest spy strategy seems not to work with the session warning. The error report is weird also, the Expected and Received actually have been the same.
```
 FAIL  src/transforms/code.spec.ts (6.465 s)
  propagateBlockDataToCode
    ✕ duplicate tag warn (32 ms)

  ● propagateBlockDataToCode › duplicate tag warn

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: "tag 'hide-cell' is duplicated"
    Received: "tag 'hide-cell' is duplicated"

    Number of calls: 1
```

packages/myst-cli/src/transforms/code.ts
rowanc1: That sounds good -- let me know when you are a bit further on the feature and I can go in and test against a few other projects!

.changeset/nervous-plants-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+'myst-cli': minor
+'myst-spec-ext': minor
```
rowanc1, line 3: Thank you! I think that we can have a `patch` change for these as they are backwards compatible additions?

```suggestion
'myst-cli': patch
'myst-spec-ext': patch
```

GYHHAHA: That's true. I should get myself more familiar with changeset tools. :eyes:

rowanc1: All good! It is super helpful in these monorepos, it builds the changelog as well as bumps all of the dependencies between packages. That and turbo repo are pretty big gamechangers for mono-repo development. :)

rowanc1:
From a functionality point of view do you think the feature is now complete? Core-review wise it looks solid! 
Let me know and I can take it for a test drive. :)

GYHHAHA:
For `mystjs` side, it is complete. But as you see from my task list, changes should also be made in `myst-to-react` which now I'm going to do after I'm able to use this new myst-cli version.

rowanc1:
Perfect -- I will try to get to this today!

GYHHAHA:
Hi @rowanc1, if you are planning to merge this in and do a patch release today, I will have some spare time to finish the rest of tasks in myst-theme at this weekend. Not rushing, just want to know is there any blocker. Thanks :blush:

rowanc1:
Hi @GYHHAHA I was stuck doing some major code renovations over in #417. I think that will help a lot of things, including the dev experience and downstream packaging. But it will probably mean a few issues with the merge, I will try to deal with them and then get on to testing this shortly!!

GYHHAHA:
Sure! I think this will be great. Take your time!

rowanc1:
Didn't quite get to actually testing, I hope I can get to it this eve -- tests are now passing again though! :)

packages/myst-cli/src/transforms/code.ts
```diff
@@ -83,3 +83,91 @@ export function liftCodeMetadataToBlock(session: ISession, filename: string, mda
...
+  blocks.forEach((block) => {
+    if (!block.data || !block.data.tags) return;
+    if (!Array.isArray(block.data.tags)) {
+      session.log.error(`tags in code-cell directive must be a list of strings`);
```
rowanc1, line 139: I am tweaking this a bit to pass in the node and file information for better console tracking:

image

rowanc1:
This looks great! I am going to merge and release once the tests go green. We should add some docs after we get the frontend working!

Thanks again for your help on this feature, another check on #189. 🚀

rowanc1:
This is released as `v1.0.2`. 🚀 Let me know how the theme work goes or if you need any help!

PR #413, opened 2023-06-08
💬 Remove `mystComment` in favour of `comment`

rowanc1:
In a future version of the spec, hopefully this is removed. There isn't anything myst-like about the comments.


rowanc1:
Tracking spec change here:
https://github.com/executablebooks/myst-spec/issues/62

PR #419, opened 2023-06-12
🔧 Fixes for JATS validation

fwkoch:
This PR makes a bunch of not-super-nice changes to identifiers before and during JATS export so there are no duplicates in the JATS xml. It also fixes a few picky structural aspects of how the xml was constructed.

### DIFF
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -21,7 +21,8 @@ function createOutputDirective(): { myst: string; id: string } {
 }
 
 function blockDivider(cell: ICell, index: number) {
-  const id = cell.metadata.id ?? `nb-cell-${index}`;
+  const id =
+    cell.metadata.id ?? `nb-cell-${index}-${computeHash(JSON.stringify(cell)).substring(0, 10)}`;
   const type = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
   return `+++ ${JSON.stringify({ id, type, ...cell.metadata })}\n\n`;
 }
--- a/packages/myst-to-jats/src/backmatter.ts
+++ b/packages/myst-to-jats/src/backmatter.ts
@@ -1,7 +1,8 @@
 import type { CitationRenderer, CitationJson } from 'citation-js-utils';
-import type { Element } from './types.js';
+import type { Element, IJatsSerializer } from './types.js';
+import { notebookArticleSuffix } from './utils.js';
 
-export function citeToJatsRef(key: string, data: CitationJson): Element {
+export function citeToJatsRef(state: IJatsSerializer, key: string, data: CitationJson): Element {
   const publicationType = !data.type || data.type === 'article-journal' ? 'journal' : data.type;
   const elements: Element[] = [];
   const authors: Element[] | undefined = data.author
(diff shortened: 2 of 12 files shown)


packages/myst-cli/src/process/notebook.ts
```diff
@@ -21,7 +21,8 @@ function createOutputDirective(): { myst: string; id: string } {
 }
 
 function blockDivider(cell: ICell, index: number) {
-  const id = cell.metadata.id ?? `nb-cell-${index}`;
+  const id =
+    cell.metadata.id ?? `nb-cell-${index}-${computeHash(JSON.stringify(cell)).substring(0, 10)}`;
```
fwkoch, line 25: This hash is pretty gross - at this point in the transform we do not have `slugs` so this is required to make the `nb-cell-0` labels unique across multiple notebooks. There could still be a collision if the cells are identical... Maybe it would be better to update these to use slugs during post-process transformation.

packages/myst-to-jats/src/backmatter.ts
```diff
@@ -112,7 +113,7 @@ export function citeToJatsRef(key: string, data: CitationJson): Element {
   return {
     type: 'element',
     name: 'ref',
-    attributes: { id: key },
+    attributes: { id: `${key}${notebookArticleSuffix(state)}${slugSuffix(state)}` },
```
fwkoch, line 116: The article suffix (`-article`) comes in on the article representation of the notebook; slug suffix is only used on citations so there is not collision on the reference ids across different backmatters.

packages/myst-to-jats/src/index.ts
```diff
@@ -27,9 +27,14 @@ import type { SupplementaryMaterial } from './transforms/containers.js';
 import type { Section } from './transforms/sections.js';
 import { sectionAttrsFromBlock } from './transforms/sections.js';
 import { inlineExpression } from './inlineExpression.js';
+import { notebookArticleSuffix, slugSuffix } from './utils.js';
 
 type TableCell = SpecTableCell & { colspan?: number; rowspan?: number; width?: number };
 
+function escapeForXML(text: string) {
```
fwkoch, line 34: Pretty simple function, but & and < are the only two characters that must be escaped in text.

packages/myst-cli/src/process/notebook.ts
rowanc1: I think that is probably the right way to go, is to do a full reset of all of the IDs across JATS.

I am running into problems that the sections are named the same, which breaks things.

packages/myst-to-jats/src/index.ts
rowanc1: Nice -- that is awesome.

packages/myst-to-jats/src/index.ts
```diff
@@ -127,7 +136,9 @@ const handlers: Record<string, Handler> = {
     state.renderChildren(node);
   },
   blockquote(node, state) {
+    state.openNode('p');
     state.renderInline(node, 'disp-quote');
+    state.closeNode();
```
rowanc1, line 141: Is this necessary?

image

fwkoch: I added this when testing a quote inside a figure caption, which isn't allowed. That's too much of an edge case though, so I reverted this change.

PR #423, opened 2023-06-15
🤖 Meca frontmatter

fwkoch:
This is the frontmatter part of the MECA export. It adds a new export format `meca` as well as `requirements` and `resources` to project frontmatter to list corresponding files to each of those categories.

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/validators.ts
+++ b/packages/myst-frontmatter/src/frontmatter/validators.ts
@@ -78,7 +78,7 @@ export const PAGE_FRONTMATTER_KEYS = [
 ].concat(PROJECT_FRONTMATTER_KEYS);
 
 // These keys only exist on the project.
-PROJECT_FRONTMATTER_KEYS.push('references');
+PROJECT_FRONTMATTER_KEYS.push('references', 'requirements', 'resources');
 
 export const USE_PROJECT_FALLBACK = [
   'authors',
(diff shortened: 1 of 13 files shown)


stevejpurves:
@fwkoch how does the `meca: export` relate to the project-wide export of a single meca bundle? 

i.e. if i want to build and deploy a site, and want a meca bundle to be built for it (i.e. the root article in the `_tcc.yml`) would be the entrypoint, where would I specify that I want the meca built?

is this a case of adding `exports.format: meca` to the root article's fromtmatter? and if so will everything in the toc be included?

stevejpurves:
also `validateProjectFrontmatter` (https://github.com/executablebooks/mystjs/blob/92fd39dfee982edbeb51008cbbe37cda41cf09ab/packages/myst-frontmatter/src/frontmatter/validators.ts#L853) doesn't seem to be ever called in myst-cli, the only usage of the validation is when checking page frontmatter and via the `validateProjectFrontmatterKeys` functon. 

So it seems like there is no point currently where project frontmatter from `myst.yml` is validated?

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -816,6 +818,8 @@ export function validateProjectFrontmatterKeys(
       validateThebe,
     );
   }
+
+  console.log('REQUIREMENTS', value);
```
stevejpurves, line 822: to remove ☝🏼 :(

rowanc1:
@fwkoch there are some new changes in `main` that are likely needed for tests to pass -- SPDX released a new license list that we test against, and the jtex template download has been given a larger timeout.

fwkoch:
Superseded by #431

PR #429, opened 2023-06-16
🕵🏽‍♀️ LaTeX improvements for `\ce` & `listings`

rowanc1:
See also: https://github.com/executablebooks/myst-theme/pull/128

### DIFF
--- a/packages/tex-to-myst/src/chem.ts
+++ b/packages/tex-to-myst/src/chem.ts
@@ -1,0 +1,15 @@
+import type { Handler } from './types.js';
+import { getArguments, texToText } from './utils.js';
+
+const CHEM_HANDLERS: Record<string, Handler> = {
+  macro_ce(node, state) {
+    state.openParagraph();
+    const value = texToText(getArguments(node, 'group')[0]);
+    state.addLeaf('chemicalFormula', {
+      value,
+    });
+    return;
+  },
+};
+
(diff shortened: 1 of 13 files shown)


packages/tsconfig/base.json
```diff
@@ -4,7 +4,7 @@
   "compilerOptions": {
     "composite": false,
     "declaration": false,
-    "declarationMap": false,
+    "declarationMap": true,
```
rowanc1, line 7: Improvements for local development!

packages/tex-to-myst/src/chem.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getArguments, texToText } from './utils.js';
+
+const CHEM_HANDLERS: Record<string, Handler> = {
+  macro_ce(node, state) {
+    state.openParagraph();
+    const value = texToText(getArguments(node, 'group')[0]);
+    state.addLeaf('chemicalFormula', {
+      value,
+    });
+    return;
```
rowanc1, line 11: This is really basic, but works!

rowanc1 (commented):
A few comments. Would be nice to do a better pass on improving chemical formulas!

rowanc1:
I think we need to increase the test timeout for `vitest`, right now it is only 5s and that is a bit short to download the tex templates.

PR #442, opened 2023-06-27
🪵 Return temp folders / log files from export functions

fwkoch:
This weaves in some return values to various export functions. These report temp folders and log files that got created during export. This can subsequently be cleaned up, moved around, etc. (None of this yet happens in `myst-cli` - though it would probably make sense to add a `--clean` option to directly clean as we go, rather than needing to invoke `myst clean`).

This also includes a small fix to imagemagick cli args.


rowanc1 (approved):
Looks good!

Issue #139, opened 2023-01-19
Install MyST CLI using tools Pythonistas are familiar with (e.g. conda)


rowanc1:
It is possible to install node dependencies through Conda, including activating the correct version of nodejs. @stevejpurves and I tried this a few months back and couldn't figure it out.

If anyone from the community has experience with this, we would absolutely love the help!! I believe that having this in a more familiar package manager would certainly help with onboarding Python users.

@choldgraf maybe you have some pointers on who can do this sort of stuff in their sleep?!

choldgraf:
In my experience - there are many in the conda-forge community who are willing to help maintain packages and who know much more about it than I do :-)   lots of folks help maintain other parts of the myst python ecosystem as well. 

Maybe @ocefpaf has thoughts on anybody that knows how to manage javascripty things in the conda-forge world?

Another option is that I think we *could* actually make mystjs installable with python as well, I know that the sphinx-theme-builder uses nodeenv to bundle its own version of node, and that might be an option ([ref `pyproject.toml`](https://github.com/pradyunsg/sphinx-theme-builder/blob/main/pyproject.toml))...cc @pradyunsg if he has any thoughts there

ocefpaf:
> Maybe @ocefpaf has thoughts on anybody that knows how to manage javascripty things in the conda-forge world?

I don't know the first thing about JS but we do package node on conda-forge and folks use it to later install JS stuff. Maybe @bollwyvl did some recipes with that? Sorry bollwyvl  if I'm misremembering.

pradyunsg:
Is there a reason the CLI was built on the JS toolchain rather than the Python one? But more broadly, we could use nodeenv under the hood of a PyPI package if we really wanted to do that; but it might actually be cleaner to provide a `myst-cli` on PyPI that uses `myst-parser` rather than trying to adapt a npm package to a PyPI package (it works for pyright though, so it's not strictly out of the picture).

bollwyvl:
Some thoughts on packaging and distributing `npm` confections as useful CLI tools:

- use [`ncc`](https://www.npmjs.com/package/@vercel/ncc) single-file distributions
  - nobody wants the 1000s of `node_modules` files crapped all over their env, much less their path length, space, and case (in)sensitivity on windows
- try to get stuff to install with `--ignore-optionals`, `--ignore-scripts`, and anything else that reduces the surface area of drive-by npm modulejacking, etc.
[...]

PR #447, opened 2023-06-30
📖 Update to `mystmd` branding

rowanc1:
This is a follow on task from executablebooks/team-compass#21.

This will eventually allow a consistent install experience across pip and npm.

See #139


rowanc1:
@choldgraf tagging you in here based on our meeting from last week on this!

The change here is that the download would change from `npm install -g myst-cli` to `npm install -g mystmd`. As we talked about last week this allows us to align with eventually installing from conda and PyPI with the same name in the future (#139).

The install process I have also sped up about 17x. 🚀

As a part of this change, this would also rename the git repo to `mystmd` and then we have alignment across most of our various pages and doesn't turn off users who see the `js` part and think it isn't for them! This will also allow us to bring back some of the js-specific docs with less confusion.

Let me know if you have any thoughts on this, I would like to get this out before the webinar on Wednesday as we already looking to have quite a few folks signed up there!

choldgraf:
Go for it. I think that we should double check our explanation in the future to make sure it matches the mental model people bring (eg MyST is a flavor of markdown but it is also a documentation engine?!) bit I think that will be best done after running these ideas by people, getting feedback, and making adjustments.

rowanc1:
Sounds good - will get this out today!

PR #448, opened 2023-07-03
🏛 Improve MECA export structure and contents

fwkoch:
MECA exports now maintains source directory hierarchy, and includes all source files, exports, and dependencies.


rowanc1:
Tested locally and been through a few demos with @fwkoch, this is looking great!

PR #463, opened 2023-07-06
💡proposed `thebe` frontmatter changes

stevejpurves:
These changes remove the `local` options for now and change the `server` and default `thebe: true` behaviour to remove the possibility of inadvertently connecting to a local server. 

We can discuss the frontmatter layout here, and once we are happy with that we can implement changes in the `myst-theme`'s and upstream in `thebe` to support it.

## summary of changes in this PR

* `local` section was removed
* `jupyter.server` can no longer be a boolean, it must be an object and must contain a `url` and `token` -- purpose here is to make connecting to a `localhost` server a more deliberate action
* doc changes
  * tried to make the docs simpler (based on feedback)
  * docs now promote use of the `jupyter` property name instead of `thebe` - `jupyter` has been added as an alias

### TODO

* [x] issue #557

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/validators.ts
+++ b/packages/myst-frontmatter/src/frontmatter/validators.ts
@@ -728,7 +728,11 @@ export function validateThebe(input: any, opts: ValidationOptions): Thebe | unde
   if (value.binder) {
     output.binder = validateBinderHubOptions(
       value.binder === true ? {} : (value.binder as BinderHubOptions),
-      incrementOptions('binder', opts),
+      {
+        ...incrementOptions('binder', opts),
+        errorLogFn: (msg) =>
+          validationError(msg.split('object').join('an object or boolean'), opts),
+      },
     );
   }
   if (defined(value.server)) {
--- a/packages/myst-frontmatter/src/frontmatter/validators.ts
+++ b/packages/myst-frontmatter/src/frontmatter/validators.ts
@@ -742,7 +742,7 @@ export function validateThebe(input: any, opts: ValidationOptions): Thebe | unde
     if (server) {
       output.server = validateJupyterServerOptions(server, serverOpts);
     } else {
-      output.server = server;
+      return undefined;
     }
   }
   if (defined(value.kernelName)) {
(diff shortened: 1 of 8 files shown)


stevejpurves:
we should take care of this https://github.com/executablebooks/mystmd/issues/473 here

stevejpurves:
still needs a manual testing pass

stevejpurves:
ok - after manual testing I'm happy with this except for a double error message when the `myst.yml` specifies `jupyter.server: true`. I'm not sure why it's being logged twice, maybe that's a common gotcha @fwkoch ?

image

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -705,29 +703,46 @@ export function validateBiblio(input: any, opts: ValidationOptions) {
...
-  const value: Thebe | undefined = validateObjectKeys(input, { optional: THEBE_KEYS }, opts);
   if (value === undefined) return undefined;
   const output: Thebe = {};
   if (defined(value.lite)) {
     output.lite = validateBoolean(value.lite, incrementOptions('lite', opts));
   }
   if (defined(value.binder)) {
-    output.binder = validateBooleanOrObject(
-      value.binder,
+    const isBoolean = validateBoolean(value.binder, {
```
fwkoch, line 728:
```suggestion
    const asBoolean = validateBoolean(value.binder, {
```

Maybe? Since this resolves to the boolean value, if it's `false`, then `isBoolean === false` which feels wrong?

fwkoch (approved):
Looks good - the double logging of errors is a myst bug, made an issue here: https://github.com/executablebooks/mystmd/issues/644

fwkoch:
@stevejpurves - I went to get the tests working on this and realized the logic wasn't quite right. We were not gracefully handling the case of `value.binder = false`. I updated the validation there a bit so:
1. any falsy value is ignored
2. `true` is explicitly converted to `{}`
3. everything else is treated as an object

I think this makes sense? It's all isolated in that last commit so you can take a look.

stevejpurves:
@fwkoch ok that makes sense. The additional validation function was in place there to try and get a better error message to the console - otherwise on error the cli reports that binder should be an object, and doesn't mention object or boolean. I think I have addressed that in the last commit.

PR #468, opened 2023-07-08
🔀 Consolidate id deduplication/simplification for JATS

fwkoch:
JATS requires no duplicate ids across all identifiers (sections, citations, footnotes, etc etc). Further, most examples of JATS files have simple identifiers like `ref-1`, `sec-5` etc.

To solve the first problem, I previously scattered suffixes, hashes, etc throughout the `myst-cli` and `myst-to-jats` codebase... it was messy code, made messy ids, and didn't solve all the problems (duplicate ids still popped up in realistic examples).

This PR eliminates the mess described in the previous paragraph and brings all deduplication/simplification to a couple transforms isolated to `myst-to-jats`: (1) first, a transform that replaces all the identifiers on the targets and creates a lookup of the replacements, then (2) a transform that updates all the references based on the lookup.


rowanc1:
Yay!

image

PR #474, opened 2023-07-11
🏁 Add project and page banners

rowanc1:
This adds support for tracking banners in the same way that we do thumbnails. For example:

image


packages/myst-cli/src/transforms/images.ts
```diff
@@ -500,6 +500,40 @@ export async function transformThumbnail(
   }
 }
 
+export async function transformBanner(
```
fwkoch, line 503: Maybe some deduplication we can do between this and thumbnails, but for now, looks great!

fwkoch (approved):

packages/myst-cli/src/transforms/images.ts
rowanc1: Yeah, agree. :)

PR #475, opened 2023-07-11
📤 Put url/filename on project exports in site config

fwkoch:
Previously exports in the `config.json` site file were just copied straight from the original frontmatter. This pr updates the exports to include filename/url, like we do for page-level exports:

image

### DIFF
--- a/packages/myst-cli/src/process/site.ts
+++ b/packages/myst-cli/src/process/site.ts
@@ -145,7 +145,7 @@ async function resolvePageSource(session: ISession, file: string) {
   return { format: extname(file).substring(1), filename: basename(file), url: `/${fileHash}` };
 }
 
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/process/site.ts
```diff
@@ -145,7 +145,7 @@ async function resolvePageSource(session: ISession, file: string) {
   return { format: extname(file).substring(1), filename: basename(file), url: `/${fileHash}` };
 }
 
-async function resolvePageExports(session: ISession, file: string) {
+export async function resolvePageExports(session: ISession, file: string) {
```
rowanc1, line 148: Just needs meca in the list!

rowanc1 (approved):

PR #477, opened 2023-07-12
🛌 Improve directive/mdast for figures with embedded content

fwkoch:
- `figure` directive now takes `remove-input`/`remove-output`, same as `embed` directive
- `figure` directive now takes `placeholder` image used in cases where main content (output, unsupported image) cannot be used
- `embed` directive takes label reference as arg, same as how `figure` takes it (no longer `:label:` option - this is nicely deprecated)
- mdast structure simplified:
  - `embed > block > children` is reduced to `embed > children`
  - `figure > embed > children` is reduced to `figure > children`
  - ids/labels are removed from `embed` children

todo - respect placeholder images in
- [x] jats - make placeholder an alternative
- [x] pdf/tex/docx - use placeholder if it's needed, ignore if not
- [x] md - reverse placeholder image back to `figure` directive option

See: https://github.com/executablebooks/mystmd/issues/359


packages/myst-cli/src/process/mdast.ts
```diff
@@ -273,14 +286,17 @@ export async function postProcessMdast(
...
+  linksTransform(mdast, state.file as VFile, {
     transformers,
     selector: LINKS_SELECTOR,
   });
-  resolveReferencesTransform(mdastPost.mdast, state.file as VFile, { state });
-  embedDirective(session, mdastPost.mdast, mdastPost.dependencies, state);
+  resolveReferencesTransform(mdast, state.file as VFile, { state });
+  embedDirective(session, mdast, dependencies, state);
+  if (simplifyFigures) {
+    transformPlaceholderImages(mdast, { imageExtensions });
```
fwkoch, line 296: This does need to happen in post-process because figure embed content need to be resolved alongside placeholder images.

packages/myst-cli/src/transforms/embed.ts
```diff
@@ -16,20 +17,31 @@ export function embedDirective(
   dependencies: Dependency[],
```
fwkoch, line 17: Rename this to a transform not a directive `transformEmbedNodes`

packages/myst-cli/src/transforms/embed.ts
```diff
@@ -16,20 +17,31 @@ export function embedDirective(
...
   state: IReferenceState,
 ) {
-  const embedNodes = selectAll('embed', mdast) as GenericNode[];
+  const embedNodes = selectAll('embed', mdast) as Embed[];
   embedNodes.forEach((node) => {
     const normalized = normalizeLabel(node.label);
```
fwkoch, line 22: Put label in embed `source`, confusing that label/identifier are references not targets in this case.

packages/myst-cli/src/process/mdast.ts
fwkoch: Add this as comment in code.

packages/myst-cli/src/transforms/images.ts
```diff
@@ -592,3 +593,50 @@ export async function transformWebp(
...
+export function transformPlaceholderImages(
+  mdast: Root,
+  opts?: { imageExtensions?: ImageExtensions[] },
+) {
+  const validExts = opts?.imageExtensions ?? DEFAULT_IMAGE_EXTENSIONS;
+  selectAll('container', mdast)
+    .filter((container: GenericNode) => container.kind === 'figure')
+    .forEach((figure: GenericNode) => {
+      const validContent = figure.children?.filter((child) => {
+        return child.type === 'code' || (isValidImageNode(child, validExts) && !child.placeholder);
```
fwkoch, line 626: Only check for the valid image.

packages/myst-directives/src/figure.ts
```diff
@@ -47,20 +47,51 @@ export const figureDirective: DirectiveSpec = {
...
-      alt: data.options?.alt as string,
-      width: data.options?.width as string,
-      height: data.options?.height as string,
-      align: data.options?.align as Image['align'],
-    };
-    const children: GenericNode[] = [img];
+    const url = data.arg as string;
+    const children: GenericNode[] = [];
+    // If figure arg starts with #, use embed node. Otherwise use image node.
+    if (url.startsWith('#')) {
```
fwkoch, line 67: Let's leave this logic in the image transform so it's in one place. The `remove-` options can just be tossed on the image nodes.

rowanc1:
AST nodes look good! Tested static outputs, tested web outputs. There need to be a few updates to the theme (for the source and thebe to show correctly, but they will be easy to do now!)

PR #480, opened 2023-07-12
✅ Check that node >=16

rowanc1:
Check that we are on a LTS version of node.


packages/mystmd/package.json
```diff
@@ -37,12 +37,12 @@
     "lint:format": "npm run copy:version; prettier --check \"src/**/*.ts\"",
     "test": "npm run link; npm run copy:version; vitest run",
     "test:watch": "npm run link; npm run copy:version; vitest watch",
-    "build:cli": "esbuild src/index.ts --bundle --outfile=dist/myst.cjs --platform=node --external:fsevents",
+    "build:cli": "esbuild src/index.ts --bundle --outfile=dist/myst.cjs --platform=node --external:fsevents --target=node14",
```
rowanc1, line 40: This ensures that we compile on node 14 and then ask people to upgrade.

PR #481, opened 2023-07-13
🤐 Use adm-zip to unzip templates, not unzipper

fwkoch:
This should address https://github.com/myst-templates/docx_default/issues/1

Same change we made here: https://github.com/curvenote/jats/pull/19 

---

Also this switches site template download from `git clone` to admzip. Before we needed a clone in order to pick up submodules. Now, though, the templates are bundled nicely and can simply be downloaded. Hopefully fixes: https://github.com/executablebooks/mystmd/issues/209

Note - it would be good to update the myst-templates API to have zip urls for site templates, not git, but for now, we construct zip from git.

### DIFF
--- a/.changeset/five-years-collect.md
+++ b/.changeset/five-years-collect.md
@@ -1,0 +1,5 @@
+---
+'myst-templates': patch
+---
+
+Download and unzip site templates, do not clone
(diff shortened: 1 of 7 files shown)


rowanc1:
@fwkoch can you confirm this also fixes the site clone, I am not seeing the code for that.

fwkoch:
Oops! You're right, added it now.

rowanc1:
This has broken site templates, they are not unzipped correctly and are nested into a folder.

PR #483, opened 2023-07-13
📺 Document videos and iframes

rowanc1:
Allow iframes to have captions as well, which turn them into figures.

### DIFF
--- a/packages/myst-spec-ext/src/types.ts
+++ b/packages/myst-spec-ext/src/types.ts
@@ -68,7 +68,7 @@ export type Iframe = {
   type: 'iframe';
   label?: string;
   identifier?: string;
-  src?: string;
+  src: string;
   width?: string;
   align?: Image['align'];
   class?: Image['class'];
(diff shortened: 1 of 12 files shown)


packages/myst-spec-ext/src/types.ts
```diff
@@ -64,6 +64,16 @@ export type Image = SpecImage & {
   height?: string;
 };
 
+export type Iframe = {
+  type: 'iframe';
+  label?: string;
+  identifier?: string;
+  src?: string;
```
fwkoch, line 71: This is functionally required, I think - but maybe there's a reason to not be required in the type.

fwkoch (approved):
Looks nice!

packages/myst-spec-ext/src/types.ts
rowanc1: Oh, yes, good point!

rowanc1:
```suggestion
  src: string;
```

PR #484, opened 2023-07-14
🚀 Typst export

rowanc1:
Convert markdown files into typst for PDF creation.

- [ ] Need better templating
- [ ] Remove unknown cross-references
- [ ] Tables
- [ ] Render

### DIFF
--- a/packages/myst-cli/src/build/build.ts
+++ b/packages/myst-cli/src/build/build.ts
@@ -15,6 +15,7 @@ export type BuildOpts = {
   docx?: boolean;
   pdf?: boolean;
   tex?: boolean;
+  typst?: boolean;
   xml?: boolean;
   md?: boolean;
   meca?: boolean;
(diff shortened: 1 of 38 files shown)


rowanc1:
@fwkoch could you take a look at this tomorrow? I mostly want to know if you have any tips for the exports!

packages/myst-cli/src/build/typst/single.ts
```diff
@@ -0,0 +1,303 @@
...
+      extraLinkTransformers,
+      simplifyOutputs: true,
+    },
+  );
+  writeBibtexFromCitationRenderers(
+    session,
+    path.join(path.dirname(templateOptions.output), DEFAULT_BIB_FILENAME),
+  );
+
+  const mystTemplate = new MystTemplate(session, {
```
fwkoch, line 144: This template stuff will be easier once there are actually `typst` templates. Right now there's a bunch of copied code that doesn't actually do anything I don't think.

packages/myst-cli/src/build/build.ts
```diff
@@ -26,29 +27,33 @@ export type BuildOpts = {
 };
 
 export function hasAnyExplicitExportFormat(opts: BuildOpts): boolean {
-  const { docx, pdf, tex, xml, md, meca } = opts;
-  return docx || pdf || tex || xml || md || meca || false;
+  const { docx, pdf, tex, typst, xml, md, meca } = opts;
```
fwkoch, line 30: We just need to refactor this boilerplate somehow to reduce repetition. Some way to specify things declaratively only once

packages/myst-cli/src/build/utils/collectExportOptions.ts
```diff
@@ -301,6 +306,18 @@ export async function collectExportOptions(
           )),
         );
       }
+      if (formats.includes(ExportFormats.typst)) {
+        fileExportOptionsList.push(
+          ...(await collectTexExportOptions(
```
fwkoch, line 311: Probably need to reframe this function as something like `collectThingsThatMayOrMayNotBeZippedExportOptions`

fwkoch:
In general, no tips - I think you are right to reuse the `tex` export since this has similar "zip or not" functionality. Beyond that though, I'd just have to dig in a bit and refactor away some of the repetition.

rowanc1:
Goign to break this into a few different PRs!

Issue #473, opened 2023-07-10
Change `thebe` frontmatter option to `jupyter`?


rowanc1:
I think that the key `jupyter` will be much clearer to folks than having `thebe` be the initial key. This would allow having a config like:

```yaml
jupyter: 'lite'
```

Which is probably clearer. We could use `thebe` as an alternative/alias which acts exactly the same way.

cc @stevejpurves

PR #495, opened 2023-07-15
🛠 Allow jupyter as alias for thebe

rowanc1:
- Alias for option `thebe` of `jupyter` #473
- Allows for { jupyter: lite }

This is a start at #463.

**Future work:**
- remove the `boolean | Object` from the binder and server options. We should be explicit in the CLI.
- Update the docs should happen in #463

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/validators.ts
+++ b/packages/myst-frontmatter/src/frontmatter/validators.ts
@@ -35,9 +35,9 @@ import type {
   TextRepresentation,
   Venue,
   Thebe,
-  ThebeBinderOptions,
-  ThebeServerOptions,
-  ThebeLocalOptions,
+  BinderHubOptions,
+  JupyterServerOptions,
+  JupyterLocalOptions,
 } from './types.js';
 
 export const SITE_FRONTMATTER_KEYS = [
@@ -69,9 +69,7 @@ export const PROJECT_FRONTMATTER_KEYS = [
   'math',
   'abbreviations',
   'exports',
-  'thebe',
-  'requirements',
-  'resources',
+  // Do not add any project specific keys here!
 ].concat(SITE_FRONTMATTER_KEYS);
 export const PAGE_FRONTMATTER_KEYS = [
   'kernelspec',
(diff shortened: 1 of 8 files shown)


packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -370,14 +378,14 @@ export function validateThebe(input: any, opts: ValidationOptions) {
     output.binder = validateBooleanOrObject(
       value.binder,
       incrementOptions('binder', opts),
-      validateThebeBinderOptions,
+      validateBinderHubOptions,
     );
   }
   if (defined(value.server)) {
     output.server = validateBooleanOrObject(
```
rowanc1, line 385: For these, I think we should be explicit in returning the object always, then the config is clear and there is less validation in the clients/theme.

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -449,10 +457,10 @@ export function validateThebeServerOptions(input: any, opts: ValidationOptions)
   return output;
 }
 
-export function validateThebeLocalOptions(input: any, opts: ValidationOptions) {
-  const value = validateObjectKeys(input, { optional: THEBE_LOCAL_OPTIONS_KEYS }, opts);
+export function validateJupyterLocalOptions(input: any, opts: ValidationOptions) {
+  const value = validateObjectKeys(input, { optional: JUPYTER_LOCAL_OPTIONS_KEYS }, opts);
   if (value === undefined) return undefined;
-  const output: ThebeLocalOptions = {};
+  const output: JupyterLocalOptions = {};
```
rowanc1, line 463: We might want to delete this fully.

packages/myst-frontmatter/tests/thebe.yml
```diff
@@ -43,3 +44,15 @@ cases:
           token: local-secret
           kernelName: local-kernel
           sessionName: local-session
+  - title: jupyter lite
```
rowanc1, line 47: We should add a few more cases here for the various validation.

rowanc1:
This is a little bit of a smaller change than I had initially imagined, and is mostly refactoring. I have now ensured that thebe/resources/requirements are only added to the project (before they were validated but not used on a page).

This also opens up `jupyter: 'lite'` and other simple config declarations, but does not change the underlying key (still thebe). This should make any downstream changes pretty simple.

stevejpurves:
ok, so after this change `jupyter` and `thebe` are interchangeable but the string based root value `lite` needs downstream code to change, correct?

PR #500, opened 2023-07-18
🐍 Add pypi deploy for myst

fwkoch:
This introduces a new package that allows deploying `myst` to PyPI, so we can install with:

```
pip install mystmd
```

In order for this to work, users must have node installed. If they do not, they get a warning, like:
image

If node is installed, the `myst` python library passes all commands/arguments directly through to the JS CLI:
image

### DIFF
--- a/.changeset/tiny-spoons-itch.md
+++ b/.changeset/tiny-spoons-itch.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Add pypi install instructions to update message
--- a/packages/myst-cli/src/session/session.ts
+++ b/packages/myst-cli/src/session/session.ts
@@ -23,21 +23,23 @@ export function logUpdateAvailable({
   current,
   latest,
   upgradeCommand,
+  pypiUpgradeCommand,
   twitter,
 }: {
   current: string;
   latest: string;
   upgradeCommand: string;
+  pypiUpgradeCommand?: string;
   twitter: string;
 }) {
   return boxen(
     `Update available! ${chalk.dim(`v${current}`)} ≫ ${chalk.green.bold(
       `v${latest}`,
-    )}\n\nRun \`${chalk.cyanBright.bold(
-      upgradeCommand,
-    )}\` to update.\n\nFollow ${chalk.yellowBright(
-      `@${twitter}`,
-    )} for updates!\nhttps://twitter.com/${twitter}`,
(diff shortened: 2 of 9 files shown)


packages/myst-cli/src/session/session.ts
```diff
@@ -23,21 +23,23 @@ export function logUpdateAvailable({
...
     `Update available! ${chalk.dim(`v${current}`)} ≫ ${chalk.green.bold(
       `v${latest}`,
-    )}\n\nRun \`${chalk.cyanBright.bold(
-      upgradeCommand,
-    )}\` to update.\n\nFollow ${chalk.yellowBright(
-      `@${twitter}`,
-    )} for updates!\nhttps://twitter.com/${twitter}`,
+    )}\n\nRun \`${chalk.cyanBright.bold(upgradeCommand)}\` to update.\n\n${
+      pypiUpgradeCommand
+        ? `(Or if using PyPI, \`${chalk.cyanBright.bold(pypiUpgradeCommand)}\`) \n\n`
```
fwkoch, line 40: We probably want both of these present so we do not confuse PyPI users with npm installation instructions.

We probably could figure out which the user needs by parsing the result of `which('myst')` but that feels like maybe overkill for the first pass.

packages/mystmd-py/scripts/pypi-deploy.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+cp ../mystmd/README.md .
```
fwkoch, line 2: This script should be run after releasing the JS mystmd library. It copies dependencies, updates the version in pypi to match npm, builds and deploys.

.changeset/tiny-spoons-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'myst-cli': patch
```
fwkoch, line 2: As far as I can tell, `changeset` only plays nicely with JS packages. I tried to add `mystmd-py` as a changeset package and it errored.

packages/mystmd-py/setup.cfg
```diff
@@ -0,0 +1,29 @@
+[metadata]
+name = mystmd
+version = 0.0.6
```
fwkoch, line 3: Once this PR is approved, we should do one more release to get this current with the npm version.

PR #504, opened 2023-07-25
🎑 Add thumbnail to project manifest

rowanc1:
This adds a thumbnail to the project manifest (if it doesn't already exist) using the index page thumbnail.


stevejpurves:
@rowanc1 how are our docs doing in terms of all the recent frontmatter changes? we need to do a pass and update them?

PR #507, opened 2023-07-31
🧮 Use katex for mathml rendering in jats

fwkoch:
This addresses the bullet point `Math should be parsed and exported to mml` on this issue: https://github.com/executablebooks/mystmd/issues/347

JATS export now uses mathml to represent equations / inline math, rather than leaving the raw latex math.


rowanc1 (commented):
Looks awesome! A few points/questions below.

Can you maintain the latex/cdata in the output using an alternatives? I think it is good to have both included.

Should we move `xmlns:mml="http://www.w3.org/1998/Math/MathML"` to the top of the document? Is it common to have that throughout?

rowanc1:
Tested locally, and JATS validation is working well!

PR #512, opened 2023-08-01
🐛 Update URLs for github.io domains

### LINKED ISSUE #511: Myst user site bug when project name has the word ".github.io" in it
Just a minor thing:

GitHub allows creating a [user site](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#types-of-github-pages-sites). From that docs: 

 > To publish a user site, you must create a repository owned by your personal account that's named username.github.io

That means my repo is

https://github.com/kolibril13/kolibril13.github.io

but in myst.yml this becomes 

https://github.com/kolibril13/kolibril13hub.io.git

because the first .git is cropped away by myst.

Same for the gh pages when initializing it with myst init --gh-pages

So the fix would be to just crop the .git part when it's at the end.

image

rowanc1:
Fixes #511

### DIFF
--- a/.changeset/good-hairs-drive.md
+++ b/.changeset/good-hairs-drive.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Fix github url for .github.io domains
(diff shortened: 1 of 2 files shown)


kolibril13:
love your super quick support! 
I've just tested renaming via hand in
https://github.com/kolibril13/kolibril13.github.io
but the page is still broken while the action runs without errors.

rowanc1:
I think that the base URL will need to change, as the main repo is not on a sub-path. I will open a PR on yours and see if we can fix it...!

rowanc1:
Ok, looks like it is working. I will merge this one in and get another PR going to flag that change automatically in the action (as well as add a comment in the deploy.yml).

rowanc1:
Almost got the other one under that time as well. :)

Thanks for the bug reports as always!

PR #517, opened 2023-08-02
Fix custom numref labels

psychemedia:
The reference label needs to be inside the `{numref}` backtick element.


rowanc1:
Thank you!

PR #523, opened 2023-08-02
📚 Use rst for python readme

fwkoch:
PyPI only nicely formats rst, so this change will make that page nicer.

Curently looks crummy:
image


rowanc1:
See #522. 🚀

PR #530, opened 2023-08-04
✍️ Abbreviations First Time Long (ftl)

Jan-David-Black:
Allowing to use the abbreviations plugin such that the first occurrence of an abbreviation is printed as follows:

```markdown
---
abbreviations_ftl: true
abbreviations: 
  MyST: Markedly Structured Text
---
Testing MyST does it render out long? Does MyST now?

or maybe now: MyST?
```

leads to the mdast:
```yaml
type: root
  children:
    - type: text
      value: 'Testing '
    - type: abbreviation
      title: Markedly Structured Text
      children:
        - type: text
          value: Markedly Structured Text (MyST)
    - type: text
      value: ' does it render out long? Does '
    - type: abbreviation
      title: Markedly Structured Text
      children:
        - type: text
          value: MyST
    - type: text
      value: ' now?'
    - type: text
      value: 'or maybe now: '
    - type: abbreviation
      title: Markedly Structured Text
      children:
      - type: text
        value: MyST
    - type: text
      value: '?'
```

Which basically ends up as:

> Testing [Markedly Structured Text (MyST)]() does it render out long? Does [MyST]() now?
>
>or maybe now: [MyST]()?


Jan-David-Black:
This tergets https://github.com/orgs/executablebooks/discussions/1056

rowanc1 (commented):
This looks awesome, thank you for implementing this! Are you primarily using this through the JS libraries? A few minor comments and then happy to bring this in!

I would be curious on your thoughts on how to expose this in the CLI or `myst.yml` config?

Jan-David-Black:
Hey @rowanc1,
this PR basically originates from my attempt to translate my thesis (from TeX) to MyST to see whether it is a viable option for my publishing needs going forward.[^1]
So it would be great if it was also use able for tex (e.g. Via the glossaries package or similar). So it might be nice to have some additional `key` field in the mdast, that can be used for that purpose, which would be basically just the short form. 

About your suggested changes: I completely agree and will implement them next time I work on this. 

About exposing the option I am very flexible and would be happy to comply to your conventions. We should however make sure to have reasonably sized names. `abbreviationsFirstTimeLong` seems a bit bulky. On the other hand I would like to make the association with `abbreviations` clear. I have been thinking about inserting it into the `abbreviations` config, but don't really see an elegant backward compatible option there. So: very open to suggestions. 

Sidenote: big compliment on MyST! 

[^1]: It would be quite amazing to have a tool to go from tex to myst. Maybe later...

rowanc1:
I have fixed up the linting and changed the name of the option. I will bring this in shortly, and we can figure out some of the other naming/config pieces in a different PR.

> It would be quite amazing to have a tool to go from tex to myst.

You can build your tex as "--md" which will convert it to MyST. This sill isn't perfect, but it does a pretty good job, and includes math macros if it finds them and most figures, tables, etc.

```shell
myst build main.tex --md
```

Thanks again for your help on this -- let me know if there is anything else we can do to make things easier to contribute?!

PR #535, opened 2023-08-08
🎓 Add rich affiliations to frontmatter

fwkoch:
This adds an affiliation object to frontmatter (instead of affiliation just being list of strings on each author).

Also, validating frontmatter, deduplicates affiliations across authors/pages/projects and normalizes them to a single list, referenced elsewhere by id.


packages/myst-templates/src/frontmatter.ts
```diff
@@ -65,24 +69,20 @@ function addIndicesToAuthors(
...
-
 export function extendFrontmatter(frontmatter: PageFrontmatter): RendererDoc {
   const datetime = frontmatter.date ? new Date(frontmatter.date) : new Date();
-  const affiliations = affiliationsFromAuthors(frontmatter.authors || []);
-  const collaborations = collaborationsFromAuthors(frontmatter.authors || []);
+  const affiliations =
+    frontmatter.affiliations
+      ?.filter((aff) => aff.id && !aff.collaboration)
+      .map((aff, index) => {
+        return { value: aff, ...indexAndLetter(index) };
```
fwkoch, line 78: These changes will need to be consumed in the tex templates themselves; now, affiliations are objects instead of strings.

rowanc1: There is also going to be a bit around @dressedfez's template that consumes the collaboration.

I think that they should be straight forward, however, we don't really have a way to version the templates which means this is a breaking change in that respect. Any better thoughts on how to handle it?

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -258,7 +266,11 @@ export function validateAndStashObject<T extends { id?: string; name?: string }>
       .digest('hex');
     warnOnDuplicate = false;
   }
-  if (!Object.keys(lookup).includes(value.id)) {
+  if (Object.keys(value).length === 1 && value.id) {
+    // This is a single object with an ID that may be defined later
+    // Only assign if it is not already defined
+    lookup[value.id] ??= value;
+  } else if (!Object.keys(lookup).includes(value.id)) {
```
rowanc1: This is the only major change, if there is an object `{ id: 'ubc' }` coming in, we don't want to overwrite anything, and in the initial stages of validation we also don't want to warn. This is a linking action.

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -305,7 +317,11 @@ export function validateAffiliation(input: any, opts: ValidationOptions) {
   if (typeof input === 'string') {
     input = { id: input, name: input };
   }
-  const value = validateObjectKeys(input, { optional: AFFILIATION_KEYS }, opts);
+  const value = validateObjectKeys(
+    input,
+    { optional: AFFILIATION_KEYS, alias: AFFILIATION_ALIASES },
```
rowanc1, line 322: Added some affiliation aliases (including to support qmd).

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -317,9 +333,10 @@ export function validateAffiliation(input: any, opts: ValidationOptions) {
   if (defined(value.institution)) {
     output.institution = validateString(value.institution, incrementOptions('institution', opts));
   }
-  if (!output.name && !output.institution) {
-    validationWarning('affiliation should include name or institution', opts);
-  }
+  // It is possible to have an ID only at this point
+  // if (!output.name && !output.institution) {
+  //   validationWarning('affiliation should include name or institution', opts);
+  // }
```
rowanc1: I removed this warning -- maybe there is a better place to put it, or perhaps if there is a single object with just ID, we skip it (that is better than removing it. I will push.).

rowanc1 (commented):
@fwkoch I added some docs and a few more tests. It would be great to have your check on the one piece of logic.

.changeset/cuddly-lies-think.md
```diff
@@ -0,0 +1,6 @@
+---
+'myst-frontmatter': patch
```
rowanc1, line 2:
```suggestion
'myst-frontmatter': minor
```

I think we could do this as a minor release for frontmatter.

packages/myst-templates/src/frontmatter.ts
dressedfez: I would be willing to introduce a versioning on my side. @rowanc1 which changes are necessary in the template?

rowanc1:
We are making the changes to all of the templates, including @dressedfez's. The site templates will also need to be updated before we release, that will happen in the next few hours!

PR #538, opened 2023-08-09
🚴‍♀️ Refactor Root --> GenericParent

rowanc1:
There are two goals for this:

(1) the types, especially in the theme almost always break
(2) `mdast` is deprecated as a package

This moves to using `GenericNode` in place of `Root`.

### DIFF
--- a/.changeset/giant-dingos-boil.md
+++ b/.changeset/giant-dingos-boil.md
@@ -1,0 +1,12 @@
+---
+'myst-transforms': patch
+'myst-to-docx': patch
+'myst-to-html': patch
+'myst-to-jats': patch
+'myst-common': patch
+'myst-parser': patch
+'myst-cli': patch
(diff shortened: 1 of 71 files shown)


fwkoch (approved):
Yeah, I think this is a great idea. Everywhere we use `Root` there are type problems and we need to cast values to `GenericNode` or `any` anyway.

My only musing is if would be worth making a `GenericRoot` just to be _slightly_ more explicit... something like:

```javascript
type GenericRoot = GenericParent & {
  type: 'root'
}
```

Either way though, good improvement as-is.

rowanc1:
Yeah, I think we should probably have that type in the future! A lot of these transforms don't actually require being at the root either, so I think they are actually more representative as just a node with children.

Still a few places that I couldn't get rid of `mdast`, but as you say -- it is an improvement!!

PR #546, opened 2023-08-11
🔧 Update generated affiliation ids to not use crypto

fwkoch:
`crypto` does not play nicely with the browser and we really do not need it here, since we are are only using it for generating a basic unique ID.

Instead, I am just using stringified JSON as the unique ID. This feels super silly - these IDs are really gross looking. However, they only exist in memory for string comparisons, so, in practice, as long as we don't start writing them to file I think it's ok.

### DIFF
--- a/packages/myst-frontmatter/tests/affiliations.yml
+++ b/packages/myst-frontmatter/tests/affiliations.yml
@@ -31,7 +31,7 @@ cases:
         - id: University
           name: University
     warnings: 1
-  - title: (Don't) Warn if affiliation has no name/institution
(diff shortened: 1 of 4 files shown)


packages/myst-frontmatter/tests/affiliations.yml
```diff
@@ -31,19 +31,19 @@ cases:
         - id: University
           name: University
     warnings: 1
-  - title: (Don't) Warn if affiliation has no name/institution
+  - title: Warn if affiliation has no name/institution
```
rowanc1, line 34: Thank you. :)

PR #554, opened 2023-08-11
📎 Handle base64 notebook cell attachments

fwkoch:
This PR changes how notebooks are parsed to MyST: Instead of building a single markdown document from the cells and calling `parseMyst` once, each cell is parsed individually, nested into `block` nodes (except frontmatter), and concatenated.

This allows us to run mdast transforms at a notebook-cell level (the only time we have access to cell `attachments`), which we use to replace `` image links with the base64 value directly. This base64 image is resolved later in `transformImages` (this functionality already existed prior to this PR).

Finally, we add one additional transform that trims base64 `urlSource` values on image nodes, so huge base64 payloads are not bundled into the mdast data.

Fixes: https://github.com/executablebooks/mystmd/issues/528

### DIFF
--- a/.changeset/angry-mirrors-draw.md
+++ b/.changeset/angry-mirrors-draw.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Handle notebook cell image attachments
--- a/.changeset/eight-moons-repeat.md
+++ b/.changeset/eight-moons-repeat.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Add transform to trim base64 urlSource values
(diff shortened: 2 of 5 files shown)


rowanc1:
This seems to work well! Testing out some earthcube notebooks which are quite large:

image

stevejpurves:
@fwkoch I take it this does not impact/change the final mdast structure for notebooks? i.e. do we end up with an additional wrapping block around cells in the tree?

fwkoch:
@stevejpurves No, there should be no impact in that regard.

Previously, this took the `output` node and replaced it with a single `image` or `text` nodes; now, if there are multiple outputs defined on the `output` node, it may be replaced with multiple `image`/`text` nodes. It does nothing to the wrapping block structure. Also - this change only affects tex, pdf, and docx exports; other exports (including the site build) keep the `output` node and pass it into their custom handlers. Hopefully that makes sense...?

PR #561, opened 2023-08-15
👩‍🎓 Extract abstract part for JATS frontmatter

### LINKED ISSUE #560: JATS Abstract
Currently we don't recognize the abstract `part` and pull it out of the JATS XML. We should use `extractPart` to pull out this content into the frontmatter.

fwkoch:
Blocks tagged with `part: 'abstract'` are now respected by JATS export and moved appropriately to `<article-meta><abstract>`.

Resolves #560

### DIFF
--- a/.changeset/chatty-forks-kick.md
+++ b/.changeset/chatty-forks-kick.md
@@ -1,0 +1,5 @@
+---
+'myst-to-jats': patch
+---
+
+Keep article-id/article-version in correct order for JATS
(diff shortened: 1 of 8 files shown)


rowanc1:
Nice, this is awesome.

rowanc1:
Looks like there is a required order to front-stub that currently fails.

image

The article-version has to be among the first elements.

rowanc1:
This isn't strictly part of this change, but it is now more evident and causes the agu-myst-lite template to fail!

Issue #563, opened 2023-08-19
imgur images


joamatab:
How can we fix imgur images?

![image](https://github.com/executablebooks/mystmd/assets/4514346/fcfd42a3-11d9-4ddc-a69c-dbbee854ad94)

fwkoch:
@joamatab Thanks for posting this! I poked around a bit, and it looks like imgur requires a `user-agent` header or else it just gives back an empty response. This should be a pretty trivial fix in mystmd - I'll get a fix out later today.

fwkoch:
This fix is now in the latest release on npm!

joamatab:
Awesome 😎 thank you

PR #565, opened 2023-08-21
🕵️‍♀️ Add User-Agent to image and link fetches

fwkoch:
This should address #563 - imgur was unhappy with no `User-Agent` header. Now these links seem to work fine:

image


stevejpurves (approved):
looks good! looking at MDN this is the right format and the minimum that we should include 👍🏼

PR #574, opened 2023-08-31
📡 enable custom binder providers in frontmatter

stevejpurves:
This opens up access to other providers (e.g. zenodo) beyond those supported in thebe 0.8.2 by default and is paired with changes to allow custom providers to be specified in thebe here https://github.com/executablebooks/thebe/pull/681

### DIFF
--- a/packages/myst-frontmatter/src/frontmatter/types.ts
+++ b/packages/myst-frontmatter/src/frontmatter/types.ts
@@ -72,16 +72,13 @@ export type Thebe = {
   local?: boolean | JupyterLocalOptions;
 };
 
-export enum BinderProviders {
-  git = 'git',
-  github = 'github',
-  gitlab = 'gitlab',
-}
+export type WellKnownRepoProviders = 'github' | 'gitlab' | 'git' | 'gist';
+export type BinderProviders = WellKnownRepoProviders | string;
 
 export type BinderHubOptions = {
   url?: string;
-  ref?: string; // org-name/repo-name
-  repo?: string; // valid git refs only?
+  ref?: string; // org-name/repo-name for WellKnownRepoProviders, url for 'meca', otherwise any string
+  repo?: string;
   provider?: BinderProviders;
 };
 
--- a/packages/myst-frontmatter/src/frontmatter/validators.ts
+++ b/packages/myst-frontmatter/src/frontmatter/validators.ts
@@ -672,22 +672,15 @@ export function validateBinderHubOptions(input: any, opts: ValidationOptions) {
       ...incrementOptions('provider', opts),
       regex: /.+/,
     });
-    if (output.provider?.match(/^(git|github|gitlab|gist)$/i)) {
-      if (defined(value.repo)) {
-        output.repo = validateString(value.repo, {
-          ...incrementOptions('repo', opts),
-          regex: GITHUB_USERNAME_REPO_REGEX,
-        });
-      }
-    } else {
-      // otherwise repo can be any value, validate as any non empty string
-      output.repo = validateString(value.repo, {
-        ...incrementOptions('repo', opts),
-        regex: /.+/,
-      });
-    }
+  }
+  if (defined(value.provider) && !output.provider?.match(/^(git|github|gitlab|gist)$/i)) {
+    // repo can be any value, but must be present -> validate as any non empty string
+    output.repo = validateString(value.repo, {
+      ...incrementOptions('repo', opts),
+      regex: /.+/,
+    });
   } else {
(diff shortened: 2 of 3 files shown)


stevejpurves:
ops! checking test failures 👀

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -667,21 +667,39 @@ export function validateBinderHubOptions(input: any, opts: ValidationOptions) {
...
     });
+    if (output.provider?.match(/^(git|github|gitlab|gist)$/i)) {
+      if (defined(value.repo)) {
+        output.repo = validateString(value.repo, {
+          ...incrementOptions('repo', opts),
+          regex: GITHUB_USERNAME_REPO_REGEX,
+        });
+      }
+    } else {
+      // otherwise provider can be any value, validate as any non empty string
```
fwkoch, line 683:
```suggestion
      // otherwise repo can be any value, validate as any non empty string
```

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -667,21 +667,39 @@ export function validateBinderHubOptions(input: any, opts: ValidationOptions) {
...
+      output.repo = validateString(value.repo, {
```
fwkoch, line 684: This doesn't check if `value.repo` is defined. Does that mean it's required in this case (so undefined gives an error?) or should this logic be moved around slightly?

packages/myst-frontmatter/src/frontmatter/validators.ts
```diff
@@ -667,21 +667,39 @@ export function validateBinderHubOptions(input: any, opts: ValidationOptions) {
...
+        ...incrementOptions('repo', opts),
+        regex: /.+/,
+      });
+    }
+  } else {
+    // assumes a WellKnownRepoProvider defaulting to 'github', validate repo as a git*/gist spec
+    if (defined(value.repo)) {
```
fwkoch, line 691: These six lines are duplicated above. Maybe another place we can rearrange logic and combine if statements... like
```
if (defined(value.provider)) {
  // validate any string for provider
}
if (defined(value.repo)) {
  if (output.provider && !output.provider.match(...) {
    // validate any string for repo
  } else {
    // validate github for repo
  }
}
```

I think the only change to the logic with this addresses my previous comment around undefined `repo`; not sure if that's right or wrong.

fwkoch (commented):
This makes sense - just a few comments about rearranging the logic slightly for readability and deduplication...

packages/myst-frontmatter/src/frontmatter/validators.ts
stevejpurves: yes it is required in this case

stevejpurves:
@fwkoch made some simpification. No rush to merge this until the theme stuff is ready, more changes might be coming

packages/myst-frontmatter/src/frontmatter/validators.ts
stevejpurves: did something like this

stevejpurves:
no further changes planned here @fwkoch so would like to get this one in - I think we've been over everything but if you could take a quick last look probably best 👍🏼

packages/myst-frontmatter/src/frontmatter/types.ts
```diff
@@ -72,16 +72,13 @@ export type Thebe = {
...
-  gitlab = 'gitlab',
-}
+export type WellKnownRepoProviders = 'github' | 'gitlab' | 'git' | 'gist';
+export type BinderProviders = WellKnownRepoProviders | string;
 
 export type BinderHubOptions = {
   url?: string;
-  ref?: string; // org-name/repo-name
-  repo?: string; // valid git refs only?
+  ref?: string; // org-name/repo-name for WellKnownRepoProviders, url for 'meca', otherwise any string
```
fwkoch, line 80: These comments don't quite seem to match the code where `ref` is validated as any string, regardless of provider, and `repo` is validated as either a github string or arbitrary string, depending.

However, the actual code functionality seems consistent with the previous behaviour (i.e. `ref` is still always validated as a simple string, `repo` has more complex logic) so I will go ahead and merge. Not going to change this comment in case there is something I am misunderstanding.

fwkoch (approved):

packages/myst-frontmatter/src/frontmatter/types.ts
stevejpurves: yes that comment is wrong, I'll remove on the next pass

PR #575, opened 2023-09-01
👩‍💻 Add html transform that combines related html nodes

fwkoch:
This is an attempt to address issues around persisting inline HTML in notebooks, raised here: https://github.com/executablebooks/jupyterlab-myst/issues/64

The parser breaks inline html up into individual tags per node; this transform reconstructs a single html node from these individual nodes as well as the markdown content between them (using myst-to-html). This means you can even include markdown inside html inline, e.g. something like `<i>[](#my-link)</i>` will resolve to an html ref.

Non-inline html is not broken up by the parser, so it will persist untransformed. Potentially we could process that html the same way we process the inline html allowing markdown inside, for example, an html table - but that is slightly out of scope here.


packages/myst-transforms/src/html.ts
```diff
@@ -69,6 +73,85 @@ export function htmlTransform(tree: GenericParent, opts?: HtmlTransformOptions)
...
+      hast: {
+        handlers: {
+          html: (h, node) => {
+            return fromHtml(node.value, { fragment: true }).children as ElementContent[];
+          },
+        },
+      },
+    },
+  );
+  // This would be good to sanitize, but the best solution requires jsdom, increasing build size by 50%...
```
fwkoch, line 97: @stevejpurves `dompurify` requires `jsdom` which bumps our `mystmd` bundle size from ~12mb -> ~17mb. Feels not great. We will need it on the other side, where this html is consumed (`jupyterlab-myst` or `myst-to-react`).

Another option is `sanitize-html`. This is a simpler library, but we need to explicitly decide which html tags/attrs to allow, and I don't think we have enough info for that yet. (For example, the defaults do not allow `button`, which is needed for this use case: https://github.com/executablebooks/jupyterlab-myst/issues/64https://github.com/executablebooks/jupyterlab-myst/issues/64)

fwkoch:
@agoose77 - this is the transform that Rowan mentioned in his comment here: https://github.com/executablebooks/jupyterlab-myst/pull/118#issue-1633283207 - it stitches together the split-up inline html (including any markdown content between the open and close tag) into a single html node. It works nicely at least with the example html from this issue: https://github.com/executablebooks/jupyterlab-myst/issues/64 as well as a few test cases. I'm sure there will be plenty of complicated html use cases which will require tweaking this, but hopefully it is an acceptable start.

PR #577, opened 2023-09-05
:1234: Support for scientific notation and decimal numbers in `si` role

jan-david-fischbach:
So far the `si` role only supported integer numbers. With this PR it becomes possible to use decimal and scientific "e" notation. Additionally a negative sign "-" at the start is also accepted

### DIFF
--- a/packages/myst-parser/tests/roles/si.yml
+++ b/packages/myst-parser/tests/roles/si.yml
@@ -50,3 +50,39 @@ cases:
                   unit: Hz
                   units: [hertz]
                   value: 1.345e10 Hz
+
+  - title: negative decimal number parses
+    markdown: '{si}`-1.345e10 <\hertz>`'
+    mdast:
+      type: root
+      children:
+        - type: paragraph
+          children:
+            - type: mystRole
+              name: si
+              value: -1.345e10 <\hertz>
+              children:
+                - type: si
+                  alt: hertz
+                  number: '-1.345e10'
+                  unit: Hz
+                  units: [hertz]
+                  value: -1.345e10 Hz
+
+  - title: no-exponent decimal number parses
+    markdown: '{si}`1.345 <\hertz>`'
+    mdast:
+      type: root
+      children:
+        - type: paragraph
+          children:
+            - type: mystRole
+              name: si
+              value: 1.345 <\hertz>
+              children:
+                - type: si
+                  alt: hertz
+                  number: '1.345'
+                  unit: Hz
+                  units: [hertz]
+                  value: 1.345 Hz
--- a/packages/myst-parser/tests/roles/si.yml
+++ b/packages/myst-parser/tests/roles/si.yml
@@ -32,3 +32,22 @@ cases:
                   unit: Hz
                   units: [hertz]
                   value: 100 Hz
+
+
+  - title: decimal number parses
+    markdown: '{si}`1.345e10 <\hertz>`'
+    mdast:
+      type: root
+      children:
+        - type: paragraph
+          children:
+            - type: mystRole
+              name: si
+              value: 1.345e10 <\hertz>
+              children:
+                - type: si
+                  alt: hertz
+                  number: '1.345e10'
+                  unit: Hz
+                  units: [hertz]
+                  value: 1.345e10 Hz
--- a/packages/myst-parser/tests/roles/si.yml
+++ b/packages/myst-parser/tests/roles/si.yml
@@ -69,7 +69,7 @@ cases:
                   units: [hertz]
                   value: -1.345e10 Hz
 
-  - title: no-exponent decimal number parses
+  - title: decimal number parses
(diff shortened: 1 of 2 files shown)


packages/myst-parser/tests/roles/si.yml
```diff
@@ -32,3 +32,22 @@ cases:
                   unit: Hz
                   units: [hertz]
                   value: 100 Hz
+
+
+  - title: decimal number parses
```
fwkoch, line 37: Can you add a couple more test cases (hopefully easy copy/paste)? At least one with a negative, maybe one with only a decimal, no exponent?

Hmm, I think `.345 <\hertz>` might fail still... That's totally fine with me, just another case to consider (maybe add a failing case for this one with `error: true` if you agree that failing is the right behaviour). Could also add a failing case for point-delimited numbers, eg. `1.000.000,05` 🤷‍♀️

packages/myst-roles/src/si.ts
```diff
@@ -10,11 +10,12 @@ export const siRole: RoleSpec = {
...
   run(data: RoleData): GenericNode[] {
     const value = data.body as string;
-    const match = value.match(/([0-9]+)\s?<([\\a-zA-Z\s]+)>/);
+    const match = value.match(/(-?\d+(,\d+)*(\.\d+(e\d+)?)?)\s?<([\\a-zA-Z\s]+)>/);
     if (!match) {
       return [{ type: 'si', error: true, value }];
     }
-    const [, number, commands] = match;
+    const number = match[1];
+    const commands = match[match.length-1];
```
fwkoch, line 18: 👍 This makes sense, necessary now that there are a bunch more groups in the regex.

fwkoch (changes requested):
@jan-david-fischbach Thanks, this is great! Do you mind running `prettier` and fixing up the linting error CI is catching (as well as adding a couple more tests from my other comment)? Hopefully that's all easy enough, let me know if I can help!

agoose77:
@fwkoch I updated this PR as I thought it would be a good place to start looking at roles. Would you be happy to give this another review?

I haven't given much thought to what the renderer should do with this AST yet, e.g. use KaTeX to render the math text.

agoose77:
@rowanc1 we should do something about this. Is there a reason to consider it `blocked`, or is it mergeable?

agoose77:
We triaged this PR as blocked during the recent weekly MyST Developers meeting. Although this functionality is useful, it will add complexity that we're not ready to support just yet. When the time comes, we think this would be a great candidate for a plugin, which will require some additional extensibility support.

stefanv:
@agoose77 I'm a bit confused about this decision: it looks as though `si` is already a role, and this change improves it at marginal complexity increase?

stefanv:
@jan-david-fischbach I'm looking into this functionality a bit, and wondered if you could share some usage examples. E.g., I can currently render all of the following using the existing `{si}` role (the exponent doesn't get rendered in any special way, though):

```
{si}``.342e45 <\hertz>``
```
```
{si}``-1,008,000.342e45 <\hertz>``
```

stefanv:
Here's an example that doesn't work (outside of the PR, which I haven't tried yet):
```
{si}``-1,008,000.342e45 <\meter\per\second>``
```

jan-david-fischbach:
Hi @stefanv,
I just realized, that seemingly I had forgotten about this PR entirely. I ended up encountering the bug again and (out of stupidity) opened a second PR (https://github.com/jupyter-book/mystmd/pull/2600) which actually got merged by now, leaving this PR in a bit of a weird state. I think we can close this PR and open a new one for handling mixtures of ',' and '.'
Further I would love support for 'qty', 'unit' etc. directly in math expressions.
Sorry for the mess.

stefanv:
No problem!

Next, I think we should document the functionality that is there, and make a list of functionality desired. Are you perhaps interested in helping with that?

The si2 plugin illustrates how this type of thing can be supported outside of mystmd itself. I think in the future we'd like to expand both plugin functionality, as well as establish a standard collection of plugins that ship with myst.

scimax:
Yes, it would be great to have some documentation, or at least a reference for the `si` role. It is listed in the reference, but the description is basically empty. I still don't understand, what the basic `si` role does or is supposed to do. I have the susppecion that something else didn't work in my configuration. I will boil it done to some MWE.

EDIT: Ok, so I was simply using it in the wrong way. so from testing it out, I understand now that you always need the brackets and the backslash. I think I saw it somewhere before, but didn't realize it. 

The best place for some minimal documentation would be in the [Roles References](https://mystmd.org/guide/roles#role-si) wouldn't it? The corresponding documentation therein should come from a `doc` line in [here](https://github.com/jupyter-book/mystmd/blob/6cc5e3ed9f3295753448f4d24f94b8a5c00b258c/packages/myst-roles/src/si.ts#L5), right? Probably a one-liner is easily done. But the overview of all commands should be placed somewhere as well. Any suggestions?

PR #578, opened 2023-09-05
🏷 Tag reduced outputs as data.type.output

fwkoch:
This PR fixes a bug introduced when we reduced `output` nodes for PDF output to `code` etc. where `remove-input`/`remove-output` flags stopped filtering correctly.

### DIFF
--- a/.changeset/smooth-pianos-reflect.md
+++ b/.changeset/smooth-pianos-reflect.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+🏷 Tag reduced outputs as data.type.output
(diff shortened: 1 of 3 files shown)


stevejpurves (approved):
looks good - if I understand it correctly, when building a PDF the nodes are flattened so checking `n.type` is fine, but when building a website, the nodes are not flattened as the embed is left if place and the relevant type information is in `n.data.type`

fwkoch:
@stevejpurves - you have the right idea, except it's the other way around. For PDFs (and tex and docx) we can't handle dynamic outputs, so we call the function `reduceOutputs`; this takes `output` nodes, and pulls out any `image` or `code` content that can be statically rendered by default. In these cases, `type: 'output'` is gone, so we need the extra `n.data.type` added (we do something similar to mark `blocks` as `notebook-code` vs. `notebook-content` here: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/process/notebook.ts#L29-L32)

For website builds, `output` nodes are left alone (since dynamic output is easier) - so here we can still check against `type: 'output'`.

(Also for what it's worth, none of this has to do with `embed` nodes; still no problem those are lifted out during processing.)

andry-tino:
Updated to 1.1.15 and now it works, thanks for the great work!

PR #586, opened 2023-09-07
📇 Support parsed author names and parse string names

fwkoch:
This introduces a new utility to parse `given`, `particle`, `family`, `suffix` from string name (maybe this utility should be split out into a separate library!?). The rules of this parsing attempt to follow bibtex patters as documented here: http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html#names - I had trouble finding a super-legitimate-feeling source of truth... But if we have a separate library, anyone could update it 🙂 

Anyway - on top of that, users may now specify frontmatter `name` as a string as before or as a `Name` object with the above fields, plus `display`. Then, on validation, this is normalized to `name: string` and `nameParsed: Name`. This means the display `name` is still unchanged, but libraries like `myst-to-jats` can access the structured parsed name.

### DIFF
--- a/packages/myst-frontmatter/src/utils/parseName.spec.ts
+++ b/packages/myst-frontmatter/src/utils/parseName.spec.ts
@@ -1,0 +1,94 @@
+import { describe, expect, test } from 'vitest';
+import type { Name } from '../frontmatter/types.js';
+import { parseName, renderName, startsWithUpperCase } from './parseName.js';
+
+describe('Test name parsing', () => {
+  test.each([
+    ['Abc', true],
+    ['ABC', true],
+    ['aBC', false],
+    ['abc', false],
+    ['àbc', false],
+    ['àBC', false],
+    ['Àbc', true],
+    ['1Abc', true],
+    ['1aBC', false],
+    ['123', true],
+  ])('test startsWithUpperCase: %s', async (word, upper) => {
+    expect(startsWithUpperCase(word)).toEqual(upper);
+  });
+  test.each<[string, Name, string?]>([
+    ['aa', { display: 'aa', family: 'aa' }, 'aa,'],
+    ['AA', { display: 'AA', family: 'AA' }, 'AA,'],
+    ['AA BB', { display: 'AA BB', family: 'BB', given: 'AA' }, 'BB, AA'],
+    [' AA \t BB  ', { display: ' AA \t BB  ', family: 'BB', given: 'AA' }, 'BB, AA'],
+    ['AA bb', { display: 'AA bb', family: 'bb', given: 'AA' }, 'bb, AA'],
+    ['AA bb CC', { display: 'AA bb CC', family: 'CC', given: 'AA', particle: 'bb' }, 'bb CC, AA'],
+    [
+      'AA bb CC dd EE',
+      { display: 'AA bb CC dd EE', family: 'EE', given: 'AA', particle: 'bb CC dd' },
+      'bb CC dd EE, AA',
+    ],
+    [
+      'AA 1B cc dd',
+      { display: 'AA 1B cc dd', family: 'dd', given: 'AA 1B', particle: 'cc' },
+      'cc dd, AA 1B',
+    ],
+    [
+      'AA 1b cc dd',
+      { display: 'AA 1b cc dd', family: 'dd', given: 'AA', particle: '1b cc' },
+      '1b cc dd, AA',
+    ],
+    [
+      'AA bb CC DD',
+      { display: 'AA bb CC DD', family: 'CC DD', given: 'AA', particle: 'bb' },
+      'bb CC DD, AA',
+    ],
+    [
+      'AA bb CC dd',
+      { display: 'AA bb CC dd', family: 'CC dd', given: 'AA', particle: 'bb' },
+      'bb CC dd, AA',
+    ],
+    ['bb CC, AA', { display: 'bb CC, AA', family: 'CC', given: 'AA', particle: 'bb' }, 'bb CC, AA'],
+    ['bb CC, aa', { display: 'bb CC, aa', family: 'CC', given: 'aa', particle: 'bb' }, 'bb CC, aa'],
+    [
+      'bb CC dd EE, AA',
+      { display: 'bb CC dd EE, AA', family: 'EE', given: 'AA', particle: 'bb CC dd' },
+    ],
+    ['CC dd EE, AA', { display: 'CC dd EE, AA', family: 'CC dd EE', given: 'AA' }],
+    ['CC dd EE,', { display: 'CC dd EE,', family: 'CC dd EE' }],
+    ['bb, AA', { display: 'bb, AA', family: 'bb', given: 'AA' }],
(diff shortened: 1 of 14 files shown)


rowanc1:
Did you take a look at this library by chance?

https://github.com/citation-js/name

packages/myst-frontmatter/src/utils/parseName.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    ['àBC', false],
+    ['Àbc', true],
+    ['1Abc', true],
+    ['1aBC', false],
+    ['123', true],
+  ])('test startsWithUpperCase: %s', async (word, upper) => {
+    expect(startsWithUpperCase(word)).toEqual(upper);
+  });
+  test.each<[string, Name, string?]>([
+    ['aa', { display: 'aa', family: 'aa' }, 'aa,'],
```
stevejpurves, line 21: it would be good to get some real names in the tests 😬 these character chucks are fine mechanically but it's difficult to see the need for a dropped particle in these examples

stevejpurves: here's an interesting test set thand to genai :p I think we want to limit to english and I don't think we can ignore dropped particle

```
* Simple Given and Family Name:

  * "John Doe"
  * "Alice Smith"
  * "Luis Garcia"

* Single Name:

  * "Madonna"
  * "Cher"
  * "Prince"

* Compound Given Names:

  * "Mary Jane Watson"
  * "Billy Bob Thornton"
  * "Jean Paul Gaultier"


* Particles (like 'de', 'von', 'van', 'di' often found in European names):

  * "Vincent van Gogh" (Given: Vincent, Particle: van, Family: Gogh)
  * "Ludwig von Beethoven" (Given: Ludwig, Particle: von, Family: Beethoven)
  * "Leonardo di Caprio" (Given: Leonardo, Particle: di, Family: Caprio)
  * "João de Silva" (Given: João, Particle: de, Family: Silva)

* Dropped Particles (which might not always be included, like 'le' or 'la'):

  * "Charles le Brun" or "Charles Brun" (Given: Charles, Dropped Particle: le, Family: Brun)
  * "Isabelle la Catholique" or "Isabelle Catholique" (Given: Isabelle, Dropped Particle: la, Family: Catholique)

Hyphenated Names:

  * "Jamie-Lee Curtis"
  * "Anne-Marie Duff"
  * "Jean-Luc Picard"

* Cultural Variants and Exceptions:

  * "Kim Jong Un" (In Korean names, 'Kim' is the family name)
  * "Xi Jinping" (In Chinese names, 'Xi' is the family name)
  * "Mohammed bin Salman" (Arabic, where 'bin' signifies 'son of')

* Names with Apostrophes or Other Punctuation:

  * "D'Arcy Wretzky"
  * "O'Connor"
  * "L'Engle"
  * "Jean-François"

* Titles and Suffixes (not asked for, but they can complicate parsing):

  * "Dr. Jane Doe"
  * "Sir Paul McCartney"
  * "John Smith Jr."

* Initials:

  * "J.K. Rowling"
  * "T.S. Eliot"
  * "F. Scott Fitzgerald"
```

packages/myst-frontmatter/tests/authors.yml
```diff
@@ -50,50 +70,167 @@ cases:
...
+            suffix: warnings,
+    normalized:
+      authors:
+        - name: Commas, here, are, great,
+          nameParsed:
+            display: Commas, here, are, great,
+            given: Commas,
+            particle: here,
+            family: are,
+            suffix: warnings,
```
stevejpurves, line 198: where do warnings come from?

stevejpurves: we are warning about the commas and there are 4 instances?

stevejpurves:
@rowanc1 thanks for the reference / link we should definitely draw from there or do you think that is usable as is? the lack of maintenance the citation-js stuff was a concern (even though that now it seems some maintenance has been happening recently on the main citation-js package).

rowanc1:
I haven't tried the package, but did look through the tests and source code. The name parts (input.js in that package) is pretty comprehensive and small from the looks of it. It is MIT licensed, so can reference if we use some parts, or the tests etc.

packages/myst-frontmatter/tests/authors.yml
fwkoch: Yes - The idea is if a comma ends up in any of the parsed parts, the name was maybe not parsed as expected.

fwkoch:
Thanks for the link to `@citation-js/name`. After looking carefully, I had a few concerns with switching to that package (besides lack of maintenance Steve mentioned). The test suite only has a single, basic case (https://github.com/citation-js/name/blob/master/test/parse/001.json - unless I totally missed something...?), there's some odd strictness around whitespace (e.g. requiring exactly one space after commas), and I think there are some fundamental bugs with string formatting (e.g. putting dropped particles before given names instead of after).

One thing in `citation-js/names` that is not here is `titles` (or `prefixes`); we simply parse those as part of the given names. When we add these, we could pull in the explicit list of titles from that library.

packages/myst-frontmatter/src/utils/parseName.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { describe, expect, test } from 'vitest';
+import type { Name } from '../frontmatter/types.js';
+import { parseName, renderName, startsWithUpperCase } from './parseName.js';
+
+describe('Test name parsing', () => {
```
stevejpurves: would be nice to also add a comment with the reference url here
http://maverick.inria.fr/~Xavier.Decoret/resources/xdkbibtex/bibtex_summary.html#names

packages/myst-frontmatter/tests/authors.yml
```diff
@@ -50,50 +70,167 @@ cases:
...
+            suffix: Jr.
+    normalized:
+      authors:
+        - name: Just A. Name
+          nameParsed:
+            display: Just A. Name
+            given: Little
+            particle: von
+            family: Junior
+            suffix: Jr.
```
stevejpurves, line 180: not clear on the meaning of this test? it leaves the normalized in some sort of mismatched state?

stevejpurves:
@fwkoch overall what was the reason for not handling a `dropping_particle`? Is that because it is difficult/impossible to reliably parse it out of the particle?

How does this PR align with other's handling?

packages/myst-frontmatter/src/utils/parseName.spec.ts
fwkoch: Url is commented in the relevant test yml file here: https://github.com/executablebooks/mystmd/pull/586/files#diff-d593203075583406317a2f9fd53086ff0dd32c771964d351e56195307e5e9510R2

fwkoch:
The parsed data structure is from CSL-JSON (citation style language: https://citationstyles.org/ - Schema: https://github.com/citation-style-language/schema/blob/e3ce254a72c4470a5ed3b9d23b428017d25674e9/schemas/input/csl-data.json#L463)

It seems to be the most common parsed format of names, used by citation-js, quarto, etc.

PR #587, opened 2023-09-11
✋🏽remove module overrides in build commands

stevejpurves:
As of typescript 5.2, errors will be thrown in cases where `module` and `moduleResolution` are mismatched in the ts compiler options. The error is:

```
option module must be set to node16 when option moduleresolution is set to node16
```

In mystmd, we have these two options aligned in the `packages/ts-config/src/base.json` but then override this in the `package.json` of each module with a `--module ES2015` this causes problems when building the cli and in particular if the mystmd package is being used as a submodule in a parent monorepo.

This change is definitely ok for usage in node, but I'm not yet clear on any impact of the change on wider usage in other consuming packages that target the browser. Any thoughts?


stevejpurves:
@rowanc1 seems like from recent PRs that setting `module: esnext, moduleResolution: node16` is the way to go and that `typescript@5.2.x` is happy with that? did you see these error messages on the way to making those changes in the recent executablebooks/markdown-it-dollarmath PRs?

rowanc1:
I haven't come across these errors yet -- I did need to change a few things to make the plugin work happy though. Let's leave this PR open for another few weeks to see where we are at. I don't think we have addressed the issue that you found yet...!

rowanc1:
Went a bit further than you @stevejpurves in pushing all of the options down to be the same in the `base.tsconfig.json`. This simplifies all of the build commands.

PR #592, opened 2023-09-13
Add support for glossary

### LINKED ISSUE #584: Node type "glossary" is not properly handled when exporting to PDF
When working on an MD file which uses the `:::{glossary}` environment, glossary terms are properly emitted when rendering HTML, but they generate an error when exporting to PDF.

Can be reproed in 1.1.15.

## Description
Create a simple MD file:

```
---
title: A simple page
authors:
    - name: me
      email: me@me.me
keywords: [example, stuff]
exports:
  - format: pdf
    template: arxiv_two_column
    output: exports/myfile.pdf
---

# Introduction
Nothing much to say. Just referencing this {term}`MDT` and that {term}`ADT`.

:::{glossary}
MDT
: My Dummy Term

ADT
: Another Dummy Term
:::
```

When creating a site using `myst start`, all is rendered fine.

However, when exporting to PDF: `myst build myfile.md --pdf`, an error is reported:

```
📬 Performing exports:
   myfile.md -> myfile.pdf
📖 Built index.md in 177 ms.
📖 Built myfile.md in 226 ms.
📚 Built 2 pages for export (including 2 dependencies) in 311 ms.
⛔️ Unhandled LaTeX conversion for node of "glossary"
📑 Exported TeX in 25 ms, copying to _build/temp/mystQb3TXh/myfile.tex
🖨  Rendering PDF (arxiv_two_column) to _build/temp/mystyfY6mn/myfile.pdf
📄 Exported PDF (arxiv_two_column) in 5.38 s, copying to exports/myfile.pdf
```

## Proposed solution
Glossary terms should be properly rendered in LaTeX. There are a variety of [rendering options](https://www.overleaf.com/learn/latex/Glossaries) to go for.

## Additional notes
The PDF is correctly rendered today and glossary terms simply skipped (words still rendered normally). This is good, meaning that the final PDF is still good, it does not look broken.

andry-tino:
This is an early review of the changes I made to add support to glossaries.

- Glossary definitions are added in the preamble regardless whether or not they will actually be used (room for improvement).
- Glossary references (cross-reference nodes) are rendered accordingly if a glossary definition is found.

### In progress
There is an issue. The rendered PDF will have the referred terms correctly rendered, however the final glossary, at the end of the page, will not be shown as explained at [Compiling the Glossary](https://www.overleaf.com/learn/latex/Glossaries) section. We need to trigger this sequence of build commands:

1. Invoke the PDF build command `pdflatex file.tex`
2. Invoke the glossary build command `makeglossaries file`, in the same folder
3. Invoke the PDF build command `pdflatex file.tex`

That is a problem I have not yet investigated how to solve.

### Note
The changes are not perfect and, since it is my first PR, I want to have some early feedback and guidance as soon as possible.

Fixes #584

### DIFF
--- a/packages/jtex/src/imports.ts
+++ b/packages/jtex/src/imports.ts
@@ -1,7 +1,7 @@
 import type { TemplateImports } from './types.js';
 import { writeTexLabelledComment } from 'myst-common';
 
-const commentLenth = 50;
+const commentLength = 50;
 
 export function createImportCommands(commands: Set<string>, existingPackages?: string[]): string[] {
   const sorted = [...commands].sort();
@@ -25,11 +25,11 @@ export function renderImports(
 ): string {
   if (!templateImports || typeof templateImports === 'string') return templateImports || '';
   const packages = new Set(templateImports.imports);
-  const imports = writeTexLabelledComment('imports', createImportCommands(packages, existingPackages), commentLenth);
-  const commands = writeTexLabelledComment('math commands', createMathCommands(templateImports.commands), commentLenth);
+  const imports = writeTexLabelledComment('imports', createImportCommands(packages, existingPackages), commentLength);
+  const commands = writeTexLabelledComment('math commands', createMathCommands(templateImports.commands), commentLength);
   const block = `${imports}${commands}`;
   if (!block) return '';
-  const percents = ''.padEnd(commentLenth, '%');
+  const percents = ''.padEnd(commentLength, '%');
   return `${percents}\n${block}${percents}`;
 }
 
--- a/packages/jtex/src/imports.ts
+++ b/packages/jtex/src/imports.ts
@@ -1,5 +1,5 @@
 import type { TemplateImports } from './types.js';
-import { label } from 'myst-common';
+import { writeTexLabelledComment } from 'myst-common';
 
 const commentLenth = 50;
 
@@ -25,8 +25,8 @@ export function renderImports(
 ): string {
   if (!templateImports || typeof templateImports === 'string') return templateImports || '';
   const packages = new Set(templateImports.imports);
-  const imports = label('imports', createImportCommands(packages, existingPackages), commentLenth);
-  const commands = label('math commands', createMathCommands(templateImports.commands), commentLenth);
+  const imports = writeTexLabelledComment('imports', createImportCommands(packages, existingPackages), commentLenth);
+  const commands = writeTexLabelledComment('math commands', createMathCommands(templateImports.commands), commentLenth);
   const block = `${imports}${commands}`;
   if (!block) return '';
   const percents = ''.padEnd(commentLenth, '%');
--- a/packages/jtex/src/imports.ts
+++ b/packages/jtex/src/imports.ts
@@ -1,28 +1,13 @@
-import type { GlossaryDirective, TemplateImports } from './types.js';
+import type { TemplateImports } from './types.js';
+import { label } from 'myst-common';
 
 const commentLenth = 50;
-
-function label(title: string, commands: string[]) {
-  if (!commands || commands?.length === 0) return '';
-  const len = (commentLenth - title.length - 4) / 2;
-  const start = ''.padEnd(Math.ceil(len), '%');
-  const end = ''.padEnd(Math.floor(len), '%');
-  const titleBlock = `${start}  ${title}  ${end}\n`;
-  return `${titleBlock}${commands.join('\n')}\n`;
-}
 
 export function createImportCommands(commands: Set<string>, existingPackages?: string[]): string[] {
   const sorted = [...commands].sort();
   const existingSet = new Set(existingPackages);
   const filtered = existingPackages ? sorted.filter((p) => !existingSet.has(p)) : sorted;
   return filtered.map((c) => `\\usepackage{${c}}`);
-}
-
-export function createGlossaryDirectives(directives: GlossaryDirective[]): string[] {
-  const usepackage = '\\usepackage{glossaries}';
-  const makeglossaries = '\\makeglossaries';
-  const entries = directives.map((entry) => `\\newglossaryentry{${entry.key}}{name=${entry.name},description={${entry.description}}}`);
-  return [usepackage, makeglossaries].concat(entries);
 }
 
 export function createMathCommands(plugins: Record<string, string>): string[] {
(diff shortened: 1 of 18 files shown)


packages/jtex/src/imports.ts
```diff
@@ -11,13 +11,20 @@ function label(title: string, commands: string[]) {
...
+export function createImportCommands(commands: Set<string>, existingPackages?: string[]): string[] {
   const sorted = [...commands].sort();
   const existingSet = new Set(existingPackages);
   const filtered = existingPackages ? sorted.filter((p) => !existingSet.has(p)) : sorted;
   return filtered.map((c) => `\\usepackage{${c}}`);
 }
 
+export function createGlossaryDirectives(directives: GlossaryDirective[]): string[] {
+  const usepackage = '\\usepackage{glossaries}';
+  const makeglossaries = '\\makeglossaries';
```
fwkoch, line 23: Does everything here need to be in the preamble with the imports? Or could we move `\makeglossaries`,`\newglossaryentry`, etc to `myst-to-tex` and just write it before everything else?

(Looking briefly at the docs, it appears `\makeglossaries` only needs to come before the first glossary definition? But I know LaTeX is quite picky...)

packages/jtex/src/imports.ts
```diff
@@ -27,6 +34,21 @@ export function createMathCommands(plugins: Record<string, string>): string[] {
...
 
+export function renderGlossaryImports(directives?: GlossaryDirective[]): string {
+  if (!directives) return '';
+  const block = label('glossary', createGlossaryDirectives(directives));
+  if (!block) return '';
+  const percents = ''.padEnd(commentLenth, '%');
+  return `${percents}\n${block}${percents}`;
+}
+
+export function renderGlossary(directives?: GlossaryDirective[]): string {
```
fwkoch, line 45: Since this is just appended to content, can it get moved to `myst-to-tex`?

packages/myst-cli/src/build/pdf/create.ts
```diff
@@ -178,11 +149,153 @@ export async function createPdfGivenTexExport(
...
+) {
+  const toc = tic();
+  let buildError = await runPdfBuildCommand(session, texFile, texLogFile, templateLogString, template, pdfBuild, buildPath);
+
+  if (buildError || !glossaries) {
+    return { buildError, toc };
+  }
+
+  // Glossaries require two more commands to be run
+  buildError = await runGlossariesBuildCommand(session, texFile, buildPath);
```
fwkoch, line 218: 🤔 Now that there are multiple commands that need to be run in a specific order, I'm wondering if some of this logic should move to jtex... Not great, though, because we probably don't want all the CLI logging stuff, etc in jtex?

I suppose for now, leaving all this here is fine; if more build commands get into the mix with subsequent features, we will probably need to refactor a bit.

fwkoch (commented):
Hey @andry-tino - thank you so much for the PR, this is awesome!

Just a couple early comments around the scope of the various packages... Mainly, it would be great to keep as much LaTeX micromanaging as possible scoped inside `myst-to-tex`; `jtex` is intended as a relatively lightweight wrapper to pull together LaTeX content and MyST templates into something that can be built into a PDF, with negligible additional LaTeX rendering.

To that end, is it possible to push all the `\makeglossaries`, `\printglossaries` etc into `myst-to-tex`, just prepending/appending to the rendered content? Then, in `jtex` the additional glossary import is just handled by the existing import renderer, and there is no more need to pass around glossary content...?

Other than that, I'm going to pull and test it out, then I'll probably leave another review with smaller comments!

andry-tino:
@fwkoch I applied your feedback :) Can you please give it another look :) Thanks!

packages/myst-to-tex/src/index.ts
```diff
@@ -301,13 +363,68 @@ const handlers: Record<string, Handler> = {
...
+    return `${percents}\n${block}${percents}`;
+  }
+
+  private createGlossaryDirectives(
+    glossaryDefinitions: Record<string, [DefinitionTerm, DefinitionDescription]>,
+  ): string[] {
+    const directives = Object.keys(glossaryDefinitions).map((k) => ({
+      key: k,
+      name: (glossaryDefinitions[k][0].children[0] as GenericNode).value || '',
+      description: (glossaryDefinitions[k][1].children[0] as GenericNode).value || '',
```
fwkoch, line 407: The values for `name` and `description` here need to be computed differently. These assume the `DefinitionTerm` and `DefinitionDescription` nodes each only have a single `text` child. However, if that's not the case it will break (I think this is a pretty realistic scenario, e.g. emphasis in a definition description).

The simplest option is to use `toText(glossaryDefinitions[k][1])` from `myst-common` - this will flatten all the content out to plain text.

Alternatively, if `newglossaryentry` allows nested tex for name/description, you could run `myst-to-tex` on the children... I suspect that's overkill though, and `toText` is best...?

fwkoch (changes requested):
Ok - love the changes you made since my last review! I've now run this myself a bit and only have a few more smaller comments, mostly inline.

In addition to those comments, two other things:
1. Can you please add some tests in `myst-to-tex` (don't worry about `myst-cli` and `jtex` - they are not well set up to extend tests easily, so no expectation for miracles there 🙂 )

    - You can add `yaml` test-cases here, with input `mdast` and expected `latex`: https://github.com/executablebooks/mystmd/blob/main/packages/myst-to-tex/tests/basic.yml
    - You could even extend the test cases to check `glossaryPreamble` here: https://github.com/executablebooks/mystmd/blob/main/packages/myst-to-tex/tests/run.spec.ts#L34 - something like
    
        ```typescript
        if (preamble) expect((file.result as LatexResult).glossaryPreamble).toEqual(preamble);
        ```

        And then you will also need to slightly extend the test case type to include an optional `preamble` field, which you can then just add to your `yaml` test cases.

2. Run `changeset` - https://github.com/changesets/changesets - this is how we manage our version bumps, releases, and release notes. Hopefully it's self explanatory - you will be prompted to select the packages you changed (should be jtex, myst-cli, myst-common, myst-to-tex), asked if it's major? no, minor? no, then just patch, and asked for a description of the change. This will produce a randomly name file in a `.changeset/` folder - then just commit that.

If you have questions about test, changeset, or any of my other comments, please let me know! More than happy to help.

packages/jtex/src/imports.ts
```diff
@@ -1,17 +1,9 @@
 import type { TemplateImports } from './types.js';
+import { writeTexLabelledComment } from 'myst-common';
 
 const commentLenth = 50;
```
rowanc1, line 4: Thanks @andry-tino -- just looking over this briefly, excited to get them in!

Saw this typo from before go to a few other files, would be great to clean it up, but not critical. :)

```suggestion
const commentLength = 50;
```

andry-tino:
@fwkoch I have added tests, let me know what you think. Since they had some special settings, I preferred creating a separate run file and yml file which will be used also for abbreviations in the next PR :)

rowanc1:
Two errors to get CI working:

Can you run lint in jtex:
image

In the `myst-cli` tests we are adding `glossaryPreamble` and that needs to be accounted for in a few places in the tests:
image

I think that you might have done this in a subsequent PR (?) should we rename `glossaryPreamble` to `preamble` now? Is that what you were thinking @andry-tino?

andry-tino:
@rowanc1 Thanks or pointing those, now I know better which commands I should run locally to be aware of lint issues and tests.

Regarding your last comment, yes, `glossaryPreamble` will be renamed, in later PRs, into `preamble`. We decided it is ok to leave it like that in this PR. The next PR we will address, abbreviations, will address this. :)

fwkoch (approved):
🚀 Nice, all the changes are looking great! Thanks for addressing the changes - just need to sort out the lingering error on the `lint` test, then we can merge.

packages/myst-cli/src/build/tex/single.spec.ts
```diff
@@ -42,6 +42,7 @@ describe('extractPart', () => {
         value: 'tagged content\n\nalso tagged content',
         imports: [],
         commands: {},
+        glossaryPreamble: "",
```
fwkoch, line 45: Probably the double quotes rather than single quotes here upsetting `lint`

rowanc1:
Nice! Super stoked about these PRs. Thanks for your help @andry-tino.

PR #601, opened 2023-09-15
🎆 Clean up extra image processing

fwkoch:
Thumbnails and webp-optimized images are only required for site builds. However, both of these were still processed in JATS/pdf/etc exports.

This PR cleans that up and turns off the webp/thumbnail processing where it is not needed.


stevejpurves (approved):
Looks good! and makes sense in the context of the additional/orphaned files we've seen in exported bundles!

PR #608, opened 2023-09-18
👯‍♀️ Add directive option `alias`

rowanc1:
You can now use `label` or `name` for all directives

### DIFF
--- a/packages/myst-parser/src/directives.ts
+++ b/packages/myst-parser/src/directives.ts
@@ -84,8 +84,29 @@ export function applyDirectives(tree: GenericParent, specs: DirectiveSpec[], vfi
         optionNodeLookup[optionNode.name] = optionNode;
       }
     });
+
+    // Deal with each option in the spec
     Object.entries(optionsSpec || {}).forEach(([optionName, optionSpec]) => {
-      const optionNode = optionNodeLookup[optionName];
+      let optionNameUsed = optionName;
+      let optionNode = optionNodeLookup[optionName];
+      // Replace alias options or warn on duplicates
+      optionSpec.alias?.forEach((alias) => {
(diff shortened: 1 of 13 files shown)


packages/myst-parser/src/directives.ts
```diff
@@ -84,8 +84,29 @@ export function applyDirectives(tree: GenericParent, specs: DirectiveSpec[], vfi
...
+        const aliasNode = optionNodeLookup[alias];
+        if (!aliasNode) return;
+        if (!optionNode && aliasNode) {
+          optionNode = aliasNode;
+          optionNameUsed = alias;
+          optionNodeLookup[optionName] = optionNode;
+        } else {
+          fileWarn(
+            vfile,
+            `option "${optionNameUsed}" used instead of "${alias}" for directive: ${name}`,
```
fwkoch, line 103: Nice - looks like this comes through if someone defines both default name and an alias (or even two aliases)

packages/myst-parser/tests/directives/directives.spec.ts
```diff
@@ -28,13 +30,18 @@ casesList.forEach(({ title, cases }) => {
...
+        const output = mystParse(markdown, { vfile });
         // Dont worry about position
         selectAll('[position]', output).forEach((node) => {
           delete node.position;
         });
         expect(output).toEqual(mdast);
+        if (vfile.messages.length !== warnings) {
+          console.log(vfile.messages);
+        }
+        expect(vfile.messages.length).toBe(warnings);
```
fwkoch, line 44: 🚀

fwkoch (approved):
Nice, looks great. Helps us normalize our language some but keep compatibility!

PR #609, opened 2023-09-19
Add support for proofs

andry-tino:
This change adds support for the following environments:

- `theorem`
- `corollary`
- `lemma`
- `proof`
- `proposition`
- `definition`
- `example`
- `remark`
- `axiom`
- `conjecture`
- `observation`

Notably, compared to what advertised in the documentation:

image

Not all environments will be supported. Those that were omitted in this change require more work.


andry-tino:
It would be best to wait for #592 and #606 to be done. I will then rename `glossaryPreamble` into `preamble` as in this change.

rowanc1 (commented):
I think the tests are failing for some other reason. I can take a look later today or tomorrow.

I left a few comments, moving the `hasProofs` flag into the `StateData` and then just lazy instantiation of that.

andry-tino:
@rowanc1 Thanks for the feedback, makes a lot of sense! :)

andry-tino:
About the tests @rowanc1... very strange, locally they are good! All of them pass...

rowanc1:
I think the jats-to-myst tests on CI were firing before the myst-to-tex was built, and we apparently use that in our tests. I added the package as a dev-dependency and turbo repo should now wait on it to be build before the tests run.

rowanc1:
Nice tests are green. ✅

Can you add a test for the added latex functionality. I am looking through the code now and I think we will want to refactor things a bit in the future so that the handers get added in at the CLI level, rather than in myst-to-tex. What that would mean is that we pass the handers from the extensions down, combining them with the default handlers [here](https://github.com/andry-tino/mystmd/blob/588_proofs/packages/myst-cli/src/build/tex/single.ts#L47), and the TexSerializers may have to have a different pattern where the plugin can define/add it's own preamble without changes in myst-to-tex. Maybe this is similar to adding an import, which any plugin can do -- we can add a preamble from any handler (it is a bit more complicated, as we don't want to add them twice, etc. which is probably why you did it the way you did it!).

I don't think those are critical changes to make now, but on my mind thinking about the plugin infrastructure lately in #617 and #636, I think that custom latex-handlers are coming up soon. Having a test will make sure that we don't loose the functionality when we do this refactor!

andry-tino:
Thanks @rowanc1. Sure I was planning to add tests of course, wanted to first have you guys have a first look at this. On it!

andry-tino:
@rowanc1 All done :)

packages/myst-to-tex/src/proof.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  preamble: string;
+
+  constructor() {
+    this.preamble = this.renderThmDefinitions();
+  }
+
+  private renderThmDefinitions(): string {
+    const definitions = [
+      '\\newtheorem{theorem}{Theorem}[section]',
```
fwkoch, line 88: Preamble always covers all theorem types - nice and easy. No need to worry about which ones to define for a specific document. I see how this could be made more customizable in the future, but for now, it's great 👍

packages/jats-to-myst/package.json
```diff
@@ -49,5 +49,8 @@
     "unist-util-remove": "^3.1.0",
     "vfile": "^5.0.0",
     "vfile-reporter": "^7.0.4"
+  },
+  "devDependencies": {
+    "myst-to-tex": "^1.0.7"
```
fwkoch, line 54: Nice catch - this is used in the tests, should be present here.

packages/myst-to-tex/src/proof.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export const proofHandler: Handler = (node, state) => {
+  state.usePackages('amsthm');
+
+  const p = node as ProofContainer;
+  const env = kindToEnvironment(p.kind);
+  if (!env) {
+    fileError(state.file, `Unhandled LaTeX proof environment "${p.kind}"`, {
+      node,
+      source: 'myst-to-tex',
```
fwkoch, line 44: Can you add `ruleId` here:
```typescript
ruleId: RuleId.texRenders
```

fwkoch (approved):
This is looking good to me! One tiny addition requested about `ruleId` but pretty trivial.

Thanks @andry-tino

packages/myst-to-tex/src/proof.ts
fwkoch:
```suggestion
      source: 'myst-to-tex',
      ruleId: RuleId.texRenders,
```

packages/myst-to-tex/src/proof.ts
```diff
@@ -0,0 +1,107 @@
+import { fileError, writeTexLabelledComment } from 'myst-common';
+import type { GenericNode } from 'myst-common';
```
fwkoch, line 2:
```suggestion
import type { GenericNode, RuleId } from 'myst-common';
```

packages/myst-to-tex/src/proof.ts
```diff
@@ -0,0 +1,108 @@
+import { fileError, writeTexLabelledComment } from 'myst-common';
+import type { GenericNode, RuleId } from 'myst-common';
```
fwkoch, line 2:
```suggestion
import type { GenericNode } from 'myst-common';
```

packages/myst-to-tex/src/proof.ts
```diff
@@ -0,0 +1,108 @@
+import { fileError, writeTexLabelledComment } from 'myst-common';
```
fwkoch, line 1:
```suggestion
import { fileError, writeTexLabelledComment, RuleId } from 'myst-common';
```

packages/myst-to-tex/src/proof.ts
andry-tino: Yep, have to say that was tricky to spot, thanks to @rowanc1 for helping me here <3

andry-tino:
@fwkoch added `ruleID` sorry for forgetting. We should make it mandatory :)

stevejpurves:
🥳 very cool to see this come in! 👏🏼

PR #612, opened 2023-09-20
＠Add pandoc-style citations

rowanc1:
Adds pandoc-style `@citations`:

- `[@author2016, pg. 23; @author2023]`
- `See @author2023`
- `In authors et al. [-@author2023]`

As part of this I have improved the column and line tracking in `myst-parser`.

image

The citations and roles are now correctly linked with column information:
image
This makes finding a particular citation much easier from the command line.

### DIFF
--- a/docs/citations.md
+++ b/docs/citations.md
@@ -6,8 +6,6 @@ thumbnail: ./thumbnails/citations.png
 
 Citations automatically show up in your site, including a references section at the bottom of the page. These citations are able to be clicked on to see more information, like the abstract. There are two different ways to add citations to your documents: (1) adding a markdown link to a [DOI](wiki:Digital_object_identifier); and (2) by adding a BibTeX file, which can be exported from any reference manager, and adding a `cite` role to your content.
 
-+++
-
 (doi-links)=
 
 ## Simple Referencing with a DOI Link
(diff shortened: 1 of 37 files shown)


packages/markdown-it-myst/src/citations.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export interface Citation {
+  label: string;
+  kind: CiteKind;
+  partial?: CitePartial;
+  prefix?: Token[];
+  suffix?: Token[];
+}
+
+export const citationsPlugin: PluginWithOptions = (md) => {
```
fwkoch, line 22: Didn't really review this in detail - trusting the tests!

packages/myst-cli/src/transforms/citations.ts
```diff
@@ -25,55 +24,42 @@ function pushCite(
...
 export function transformCitations(
-  log: Logger,
   mdast: GenericParent,
   renderer: CitationRenderer,
   references: Pick<References, 'cite'>,
-  file: string,
 ) {
-  // TODO: this can be simplified if typescript doesn't die on the parent
-  const citeGroups = selectAll('citeGroup', mdast) as CiteGroup[];
-  citeGroups.forEach((node) => {
```
fwkoch, line 63: I guess we had this before to pass `citeGroup` node `kind` to the `cite` children...? I don't see any equivalent now; am I missing something or do we just not need it? It shouldn't matter for the new `@cite` stuff, but does it matter for the old unchanged `cite` role code where `kind` is only on the `citeGroup`...?

fwkoch (approved):
Looks great - making it easier to cite stuff! Only had a couple minor questions.

docs/citations.md
```diff
@@ -24,13 +22,59 @@ which will insert the citation text in the correct format (e.g. adding an italic
...
+  - Prefix and suffix
+* - `@cockett2015 [pg. 22]`
+  - @cockett2015 [pg. 22]
+  - Suffix for narrative citations
+* - `@10.1093/nar/22.22.4673`
+  - @10.1093/nar/22.22.4673
+  - Citation using a DOI directly
+```
+
+Note that it is also possible to use the `@ref` syntax to link to other cross-references in addition to the link syntax.
```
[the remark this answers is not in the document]
rowanc1: This was an intentional feature, and the reverse does sort of work at the moment (e.g. link a DOI), but not fully with a citation-key as the link. This is parity with how quarto/pandoc work, which does really simplify the explanation of cross-references even further (although there are advantages to the link syntax in many places for rendering, e.g. in GitHub).

I think the docs are _ok_ for now, I wanted to go a bit slow with the update to phase it in, and then start to promote the `@`-syntax as general cross-reference down the road.

packages/myst-cli/src/transforms/citations.ts
rowanc1: I think that we previously only kept the "parenthetical" information on the cite-group, and now that is in both places, meaning that we don't have to first do the groups, then individual nodes (as well as skipping some).

This is both equivalent and a bit faster probably. Simpler from the code perspective as well.

rowanc1:
Thanks for the review @fwkoch -- I will add a few more tests for the base parser while I am in there -- improve the docs a bit and then call this good to go!

rowanc1:
@fwkoch I just went and deleted the docs for now. We can add this back in in the future!

PR #616, opened 2023-09-25
📏 Add enumerated `ruleId` to all errors/warnings in the redux store

fwkoch:
These `ruleId`s categorize all myst build errors in a structured way, allowing further sorting/reporting/etc of errors downstream.

In addition to adding `ruleId`s, this also cleans up many places where we were still `session.log`-ing errors instead of correctly tracking them through our `addWarningForFile` plumbing which adds the error to the redux store.


packages/myst-cli/src/config.ts
```diff
@@ -233,21 +249,23 @@ export function loadConfigAndValidateOrThrow(session: ISession, path: string) {
...
+  const conf = readConfig(session, vfile);
   const existingConf = selectors.selectLocalRawConfig(session.store.getState(), path);
   if (existingConf && JSON.stringify(conf) === JSON.stringify(existingConf)) {
     return;
   }
   session.store.dispatch(config.actions.receiveRawConfig({ path, file, ...conf }));
   const { site, project } = conf;
   if (site) {
-    validateSiteConfigAndSave(session, path, file, site);
+    validateSiteConfigAndSave(session, path, vfile, site);
```
rowanc1, line 262: Nice, it is good to see the full pass on error handling touch these again!!

packages/myst-common/src/ruleids.ts
```diff
@@ -0,0 +1,95 @@
+export enum RuleId {
+  // Frontmatter rules
+  validConfigStructure = 'valid-config-structure',
+  siteConfigExists = 'site-config-exists',
+  projectConfigExists = 'project-config-exists',
```
rowanc1, line 5: Very cool to see 80+ rules come into existence for this. Wow. :)

packages/myst-common/src/utils.ts
```diff
@@ -1,21 +1,25 @@
...
+import type { Position } from 'unist';
 import { map } from 'unist-util-map';
 import { customAlphabet } from 'nanoid';
 import type { Node, Parent, PhrasingContent } from 'myst-spec';
+import type { RuleId } from './ruleids.js';
 import type { AdmonitionKind, GenericNode, GenericParent } from './types.js';
 
 export type MessageInfo = {
-  node?: Node;
+  node?: Node | Position;
```
rowanc1, line 11: Nice, that is closer to `VFile` implementation.

rowanc1 (approved):
Really good to see the error functions getting some love. I am excited to see some of the documentation around the rule-ids in the future, but I think getting this in now is good to go, especially as other PRs are likely going to conflict with these changes!

packages/myst-common/src/ruleids.ts
fwkoch: Hah, yeah, I'm not quite sure what the correct granularity is for these guys, but for now I think this works!

Issue #168, opened 2023-01-29
Allow defining custom crossreference types


choldgraf:
Currently we support a number of standard cross-reference types like `figure`, `eq`, `doc`, etc. However there may be cases where people which to define their own **internal** cross reference types. This would allow them to group these references and treat subsets of them differently.

For example, consider an organization's documentation. You may want to define a collection of **teams** and **roles**. To reference these, you could use a generic term like `` {term}`My team` ``, but it would be more semantic and structured if you could instead do `` {team}`My team` `` and ``{role}`My role` `` (I know the "new style" markdown syntax would be different but you get the idea.

You could then use these labels for things like type-specific glossaries, e.g. the following could list a glossary of all team entries.

````
```{glossary}
:type: team
```
````

### Design idea

Could we reserve a special syntax for the "label anything" syntax? E.g.:

A normal label:

```
(foo)=
## My header
```

A label that attaches to a group:

```
(groupname::foo)=
## My header
```

**Comment by rowanc1** [comment]
Interesting, I think I am mostly following. Currently we have implemented the link syntax which allows you to target/reference any node. That is [described in this PR](https://github.com/executablebooks/myst-enhancement-proposals/pull/10).

As a silly example:
```
(my-paragraph)=
This is a paragraph that will recursively show a link to [](#my-paragraph)
```

![paragraph-recurrsion](https://user-images.githubusercontent.com/913249/215618738-350eab0c-fe48-4549-b175-efe5075b0662.gif)

If you can label anything, the additions that I see coming from this are the ability to count (e.g. "Team 1") or refer to these in a glossary (which doesn't yet exist in `mystjs`!). Am I missing other benefits?

From a task perspective, I need to (re-)familiarize myself with how sphinx does terms/glossaries, probably implement that and then come back to thinking how to extend to new roles/terms as you describe?!

choldgraf:
It's a good point that if you can label anything, you could always just come up with your own labeling structure to hint at collections. E.g., `(team:teamname)=` and then `[](#team:team-name)`. Maybe that's enough to de-prioritize this as a feature unless we can come up with other reasons that somebody would want to be able to quickly generate lists of the labels that fall under a particular group.

Also a meta note: don't treat these issues as a suggestion that we must prioritize them over everything else, I'm just opening up the ideas for discussion to see if they resonate with others!

PR #617, opened 2023-09-25
🔌 Support plugins for CLI (roles/directives only)

rowanc1:
This is a first draft of a plugin system for the mystmd CLI. It allows for new directives and roles to be created in javascript and run arbitrary code to pass into the renderer.

A few examples of things that this supports are:

- documenting directives programatically (all directives are now documented as well!)
- creating custom cross-reference roles (e.g. to a directive option, or a team member (#168 cc @choldgraf))
- loading/creating content based on a YAML file (etc.) (e.g. team member, project listings, etc.)

This PR does not implement some obvious next steps for adding custom "transforms" (which can be async actions that transform the document AST (query an API, take a screenshot, read from a DB)) and it does not yet implement rendering extensions (e.g. a custom LaTeX export node type, or Word, etc.). Both of those require a bit more work on firming up an API that we are happy with.

The PR also only works with local ESM Nodejs files (as opposed to CJS). A package manager and discovery mechanism similar to the templates API is likely another next step. The ESM requirement means changes to some base libraries which incorrectly support ESM/CJS, and, as is common in the Node ecosystem, moving to support ESM only:

- https://github.com/executablebooks/markdown-it-myst-extras/pull/2
- https://github.com/executablebooks/markdown-it-dollarmath/pull/11
- https://github.com/executablebooks/markdown-it-amsmath/pull/10

These PRs must be merged/released and updated here before any un-transformed ESM code can run (i.e. esbuild can figure it out, but NodeJS cannot at the moment).

## Docs

A few snippets from the docs:

image
image

cc @fperez @minrk @choldgraf who were helpful in prompting this last push! The directives/roles are now relatively easy to add, and there is scope to expand this rather quickly to support programatic documentation (e.g. papyri @Carreau), but that needs the `transform` extensions -- I will hopefully get a pass of that going soon and some example of integrating the two projects.


docs/directives.mjs
```diff
@@ -0,0 +1,154 @@
...
+  license: 'MIT',
+};
+
+/**
+ * @param {import('myst-common').OptionDefinition} option
+ */
+function type2string(option) {
+  if (option.type === 'string' || option.type === String) return 'string';
+  if (option.type === 'number' || option.type === Number) return 'number';
+  if (option.type === 'boolean' || option.type === Boolean) return 'boolean';
```
rowanc1, line 18: @fwkoch I extended this based on markdocs API which I quite liked the simplicity of. You can now say an option is of type string with 'String' instead of `ParsedENumType.string`. Very small change, but looks a bit better, backwards compatible as well.

docs/plugins.md
```diff
@@ -0,0 +1,80 @@
...
+transforms
+: These plugins transform the document source while it is rendered.
+: For example, add metadata or transform a link to an DOI.
+
+renderers
+: These plugins add handlers for various nodes when exporting to a specific format.
+: For example, do something special for node in HTML, React, Microsoft Word, or LaTeX.
+:::
+
+## Creating a Plugin
```
rowanc1, line 38: Need to do a bunch more documentation here, but it is a start!

docs/plugins.md
```diff
@@ -0,0 +1,80 @@
...
+
+To create a plugin, you will need a single Javascript file[^esm] that exports one or more of the objects above. For example, a simple directive that pulls a random image from [Unsplash](https://unsplash.com/) can be created with a single file that exports an `unsplash` directive.
+
+[^esm]: The format of the Javascript should be an ECMAScript modules, not CommonJS. This means it uses `import` statements rather than `require()` and is the most modern style of Javascript.
+
+:::{literalinclude} unsplash.mjs
+:filename: unsplash.mjs
+:caption: A plugin to add an `unsplash` directive that includes a beautiful, random picture based on a query string.
+:::
```
rowanc1, line 47: @nthiery actually using `literalinclude` -- very much enjoying it. :)

This PR is also prompted a bit from your latex requirements, hoping that having access to custom plugins will give you better stop-gap control over your documents while we also figure out the defaults!

packages/myst-cli/package.json
```diff
@@ -34,7 +34,7 @@
     "lint:format": "npm run copy:version; prettier --check \"src/**/*.ts\"",
     "test": "npm run copy:version; vitest run",
     "test:watch": "npm run copy:version; vitest watch",
-    "build:esm": "tsc --project ./tsconfig.json --module es2015 --outDir dist --declaration",
+    "build:esm": "tsc --project ./tsconfig.json --module esnext --outDir dist --declaration",
```
rowanc1, line 37: This change is necessary to support async imports.

packages/myst-cli/src/build/build.ts
```diff
@@ -126,6 +126,7 @@ function extToKind(ext: string): string {
 }
 
 export async function build(session: ISession, files: string[], opts: BuildOpts) {
+  await session.loadPlugins();
```
rowanc1, line 129: @fwkoch is there anywhere else that we would want to load the plugins? Right now this is in build and start.

packages/myst-cli/src/session/types.ts
```diff
@@ -7,6 +7,12 @@ import type { Store } from 'redux';
 
 import type { RootState } from '../store/index.js';
 import type { PreRendererData, RendererData, SingleCitationRenderer } from '../transforms/types.js';
+import type { DirectiveSpec, RoleSpec } from 'myst-common';
+
+export type MystPlugins = {
+  directives: DirectiveSpec[];
+  roles: RoleSpec[];
+};
```
rowanc1, line 15: I should move this to `myst-common` actually.

packages/myst-common/src/types.ts
```diff
@@ -78,7 +78,7 @@ export type DirectiveSpec = {
 
 export type RoleSpec = {
   name: string;
-  alias?: string | string[];
+  alias?: string[];
```
rowanc1, line 81: @fwkoch I removed the ambiguity here. I think this is better, we may need a directive-spec validator in the future that modifies this up front and accepts strings, etc.

packages/myst-directives/src/admonition.ts
```diff
@@ -4,6 +4,7 @@ import { ParseTypesEnum } from 'myst-common';
 
 export const admonitionDirective: DirectiveSpec = {
   name: 'admonition',
+  doc: 'Callouts, or "admonitions", highlight a particular block of text that exists slightly apart from the narrative of your page, such as a note or a warning. \n\n The admonition kind can be determined by the directive name used (e.g. `:::{tip}` or `:::{note}`).',
```
rowanc1, line 7: I have added a few places of better documentation to the various directives. I will continue improving that a bit in this PR!

rowanc1 (commented):
Some self-review and guide to the important parts of this diff.

packages/myst-cli/src/build/build.ts
fwkoch: Hmm, good question. I see two alternatives:

1. We could load `plugins` immediately when handling frontmatter (like at the end of validation or something) - this seems like the most mindless approach: To do anything, we must read frontmatter, so plugins will definitely be loaded. The drawback of this approach is we would do a bunch of unnecessary loading (e.g. `myst clean` reads frontmatter, really shouldn't need to load plugins).
2. Load plugins immediately before we do any processing at the `loadFile` level (I think this would be the place to do it - this is where we apply roles/directives, then transforms come immediately after)

The places you have it also work - they are just at a `cli`-centric level and maybe don't consider others using exported functions individually...

rowanc1: I have moved this over to `loadFile`. Great idea. :)

fwkoch (approved):
Awesome, love all the added declarative documentation, nice that implementing plugins was reasonably straightforward. A little bit of over-engineering early on paid off. 🙂

docs/directives.mjs
```diff
@@ -0,0 +1,154 @@
...
+ */
+const mystDirective = {
+  name: 'myst:directive',
+  arg: {
+    type: String,
+    required: true,
+  },
+  run(data, vfile) {
+    const name = data.arg;
+    const directive = defaultDirectives.find((d) => d.name === name);
```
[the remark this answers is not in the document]
rowanc1: I think that we actually want to pass some more information into the run function to allow something like this (as well as easier access to parsing, errors, etc.). The `vfile` alone doesn't cut it, and requires all sorts of other imports (at potentially other versions of the library).

packages/myst-cli/src/plugins.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { MystPlugin } from 'myst-common';
+
+export async function loadPlugins(session: ISession): Promise<MystPlugin> {
+  const config = selectCurrentProjectConfig(session.store.getState());
+
+  const plugins: MystPlugin = {
+    directives: [],
+    roles: [],
+  };
+  if (!config?.plugins || config.plugins.length === 0) {
```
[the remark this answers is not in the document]
rowanc1: yep, this is handled such that plugins only load once!

packages/myst-config/src/project/validators.ts
```diff
@@ -10,7 +10,7 @@ import {
 import type { ProjectConfig } from './types.js';
 
 const PROJECT_CONFIG_KEYS = {
-  optional: ['remote', 'index', 'exclude'].concat(PROJECT_FRONTMATTER_KEYS),
+  optional: ['remote', 'index', 'exclude', 'plugins'].concat(PROJECT_FRONTMATTER_KEYS),
```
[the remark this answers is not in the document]
rowanc1: That was my thought. I think that we want to keep this isolated as we build it out!

rowanc1:
Ok -- @fwkoch this is good enough for a first pass. 🚀

fperez:
🎉 I don't know enough TypeScript to meaningfully comment on the code, but I wanted to thank and commend the team for this progress! This is wonderful to see, and puts us on the path for the real promise of MyST being about combining the best experience of Markdown with the extensibility of ReST/Docutils/Pandoc, but with a modern architecture and toolchain. Thanks so much folks! This project makes me really happy, and I only wish I had more bandwidth to devote to it.

nthiery:
Yes! And similarly the extensibility of LaTeX with a modern reading experience.

fperez:
Fully agreed @nthiery - I see these tools as a viable replacement to LaTeX, reST/Sphinx and plain Markdown workflows for our community. We're not there yet but I see a clear path to that goal, and I think that goal is a worthwhile one. With the work this team is doing, I'm optimistic! :)

minrk:
This is great! If it helps, here's [one kind of sphinx extension](https://gist.github.com/minrk/2234020a647643ae88c63b20d3008e0b) that would block my migrating from sphinx-based jupyter-book to something based on mystmd.

It's a custom role `points` which tracks a sum of points (with some categories), and then a directive `pointreport` which renders the point totals (typically at the _top_ of the page with the points). So as long as:

1. roles can populate a per-document data structure, and
2. directives can read that data structure, and
3. the directive can read it _after_ all roles have populated it, regardless of order in document

I should be all set.

minrk:
Having tested it out, I can make this work for the most part, excluding item number 3, but I think that the not-yet-implemented transforms plugin API is probably where the data should be collected so that it's available at the right time, rather than in the roles themselves.

It just so happens that directives get evaluated before roles, so I can't use my previous strategy of collecting in roles and reporting in a directive, but I could do the _reverse_. There also doesn't appear to be a per-document event to indicate a render at which point the totals should be set to zero, so when I use `myst start`, each page render causes the point totals to increase. Not super relevant for a real site, but it is hitting me while I test.

In general, there seems to be something missing for 'stateful' plugins like this, but I suspect that transforms will be the way to collect this info at the right time (including the reset from zero, perhaps?). Possibly renderers as well.

Carreau:
Thanks, this is great ! I'm still looking how to integrate myst into papyri for rendering.
I'm still unfamiliar with Typescript, but I think this will gave me some (many?) of the hooks I need for papyri.

choldgraf:
Just voicing my support of this, I think it's a critical feature so I'm really happy to see a lot of voices providing their needs/experience/etc.

rowanc1:
Thanks all for the encouragement, really exciting to see these usecases @minrk, we are working through some of the implementation and choices here for the `transforms` which will be needed! See #636

As part of that I am translating your points example @minrk:
https://gist.github.com/rowanc1/7a7a40ffb0ead018e2b441977380aa62

A few questions on what to expose, and how to export up the plugin, but the basics are there and it can confidently add up points (and bonus points...!). This will also work for all of the exports (tex, docx, typst, jats, etc.), which is nice.

@Carreau -- let me know how I can help you get up and running with the myst theme, or if you want me to jump into your repos. I put a demo together here with a few videos:
https://github.com/rowanc1/myst-papyri

Issue #625, opened 2023-09-28
`Error: 'path' cannot be empty` when running `build`


fwkoch:
## Description

When running the `build` command, myst fails with the error `'path' cannot be empty`. This error is coming from attempting to set `VFile.path = undefined`. Digging a little deeper, this undefined value comes from here, where this selector is returning `undefined`: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/build/utils/collectExportOptions.ts#L273, which in turn is a result of these functions not being awaited on session startup: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/session/session.ts#L97

## Proposed solution

We need to ensure that the expected `session` setup is complete - which for now just means waiting for those load functions to finish.

andry-tino:
Unfortunately I can still repro this problem after merging main. When I run a command for building a PDF:

```
./packages/mystmd/dist/myst.cjs build /Users/me/myproj/myfile.md --pdf -o ./_build/attempts/f1.pdf
```

Which normally succeeded fine, I get this:

```
📬 Performing exports:
   ../../../myfile.md -> _build/attempts/f1.pdf
Error: `path` cannot be empty
```

PR #627, opened 2023-09-28
🐛 Fix bug with async project/site loading

fwkoch:
Addresses #625 - turns out the loading functions were erroneously marked as `async`. This pr just gets rid of that and gets rid of any `await`s elsewhere...


rowanc1:
Nice! Easy. :)

PR #636, opened 2023-10-02
🔌 Add transforms to plugin infrastructure

rowanc1:
This is an intial draft of getting transforms up and going as plugins.

At this point I have added a `stage` to the transform spec ('document' or 'project') which is when it will be run, i.e. directly after parsing or as part of the cross-linking between project documents.

There are a few things left to do:
- [ ] documentation
- [ ] some way of exposing options to various plugins, can defer that
- [ ] decide if we should export a "plugin" type with these options on it? That might be a better place for plugin options that can surface in a `myst.yml`, and would also be a place for the author name, etc. as well as having a type actually defined rather than the named exports.
- [ ] I have started a `utils` argument to the transform, I think this needs a bit of thought. Should it be an object, which we can expand over time?


fwkoch:
This is nice, very simple.

- `plugin` type in myst.yml would be good for options, etc. Not sure about author name specifically - This should remain definable in the plugin itself, so plugin authors (if they differ from project authors) may provide their name without required edits to all the consuming myst.yml.
- what is the scope of `utils`? Should this include anything a plugin may want to import... (feels like a bit too much)? Is there another way plugins can have their own requirements totally separate from anything used by myst...?
- Regarding stages - the existing myst transforms would require more ordering (or at least more than 2 stages). It's probably unrealistic to constrain plugins beyond what is required for basic processing.

I think the most interesting thing to do next would be implement all our `myst-transforms` with this interface (in the same way `myst-roles/directives` are already implemented with the plugin interface). This would expose constraints around stages, utils, anything else; if the plugin framework could support all existing transforms, it would be 🔥

fwkoch (approved):
As is, though, this is a great start, don't think we need to hesitate to land it.

PR #638, opened 2023-10-05
🏷 Tag abstract parts in JATS

rowanc1:
This adds tagging of abstract parts in JATS. Right now the `plain-language summary` and `keypoints` are added by default.


packages/myst-common/src/extractParts.ts
```diff
@@ -38,6 +54,7 @@ export function extractPart(tree: GenericParent, part: string): GenericParent |
         delete block.data.tags;
       }
     }
+    if (opts?.removePartData) delete block.data.part;
```
fwkoch, line 57: This doesn't delete the part data if it's in `tags` - is that an edge case we should keep consistent?

packages/myst-to-jats/src/index.ts
```diff
@@ -520,23 +561,29 @@ class JatsSerializer implements IJatsSerializer {
...
-          isNotebookArticleRep: this.data.isNotebookArticleRep,
-          slug: this.data.slug,
-          handlers: this.handlers,
-        });
-        abstractSerializer.render();
-        this.data.abstract = abstractSerializer.elements();
-      }
+      const abstractParts = opts.abstractParts ?? [{ part: 'abstract' }];
+      this.data.abstracts = abstractParts
+        .map((def) => renderAbstract(this.file, this.mdast, def, opts))
```
fwkoch, line 566: This seems fine, we are doing slightly more pre-rendering here, then just grab the outputs later. More consolidated than what we were doing with `getAbstract` before. 👍

packages/myst-to-jats/src/index.ts
```diff
@@ -520,23 +561,29 @@ class JatsSerializer implements IJatsSerializer {
...
+        .filter((e) => !!e) as Element[];
     }
+    this.data.acknowledgments = renderAcknowledgments(this.file, this.mdast, opts);
+    const backSections = opts?.backSections ?? [
+      {
+        part: ['data-availability', 'data_availability', 'availability'],
```
fwkoch, line 572: Should this just go in `myst-cli` too, with the various abstract sections? Thinking ahead to making this more extensible, seems like we should have the "myst defaults" in only one place (either here or myst-cli).

fwkoch (approved):
Nice, this looks like a good start, lifting out some jats extensibility.

packages/myst-common/src/extractParts.ts
rowanc1: Hm, yeah, this needs to be fixed. Will add a test. Thanks.

packages/myst-to-jats/src/index.ts
rowanc1: Yep, good point -- I will move it up to the CLI for now.

rowanc1 (commented):
Thanks for the review -- will get a few of those changes completed now.

PR #641, opened 2023-10-09
🎆 Add inline property to images in paragraphs

fwkoch:
For all images that are children of `paragraph` nodes, add `inline: true` property. This will happen when an image is declared inline, e.g.:

```md
Here is my image ![](image.png) - it's inline!
```

(note: this works well with the `myst-parser` transform that ensures standalone images are _not_ nested in paragraphs - https://github.com/executablebooks/mystmd/blob/main/packages/myst-parser/src/tokensToMyst.ts#L470)

### DIFF
--- a/.changeset/fifty-dolphins-sneeze.md
+++ b/.changeset/fifty-dolphins-sneeze.md
@@ -1,0 +1,7 @@
+---
+'myst-transforms': patch
+'myst-spec-ext': patch
+'myst-cli': patch
+---
+
+Add inline property to images in paragraphs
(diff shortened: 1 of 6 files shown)


fwkoch:
This PR along with https://github.com/executablebooks/myst-theme/pull/248 address https://github.com/executablebooks/myst-theme/pull/248

fwkoch:
We can likely just add this as a CSS rule in myst-theme without changing the myst data structure.

rowanc1:
Ultimately fixed here:
https://github.com/jupyter-book/myst-theme/pull/710

Issue #643, opened 2023-10-10
JATS export - Role with `&` breaks XML


rowanc1:
![image](https://github.com/executablebooks/mystmd/assets/913249/3293fb9b-9fe9-4ea0-85ad-e0ac9e9e8df6)

It should be replaced to an `&amp;`:

![image](https://github.com/executablebooks/mystmd/assets/913249/597161df-79af-4909-a040-fa5daf3443a7)

PR #645, opened 2023-10-10
🗝 Escape all attributes in js2xml for JATS

rowanc1:
See:
- #643
- #649

### DIFF
--- a/packages/myst-to-jats/src/frontmatter.ts
+++ b/packages/myst-to-jats/src/frontmatter.ts
@@ -167,7 +167,7 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
             attributes: {
               vocab: 'CRediT',
               'vocab-identifier': 'http://credit.niso.org/',
-              'vocab-term': escapeForXML(role),
+              'vocab-term': role,
             },
             elements: [{ type: 'text', text: role }],
           };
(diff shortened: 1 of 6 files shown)


packages/myst-to-jats/src/frontmatter.ts
```diff
@@ -166,7 +167,7 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
             attributes: {
               vocab: 'CRediT',
               'vocab-identifier': 'http://credit.niso.org/',
-              'vocab-term': `${role}`,
+              'vocab-term': escapeForXML(role),
```
rowanc1, line 170: @fwkoch is there a better place to do this so we don't run into it again?

fwkoch: Hmm, seems like the only other place, besides at this granular level would be at the xml serialization level. Looking at the docs, there doesn't seem to be any escaping built into `xml-js` but there are some hooks for processing functions... https://github.com/nashwaan/xml-js#options-for-custom-processing-functions - so maybe we could, like, escape _all_ attributes with an `attributeValueFn`...?

fwkoch: Not sure if there are implications of being too heavy handed there...? I guess we would mostly just need to be careful about not escaping twice (though our escape function is already idempotent I think, so maybe that's not a concern at all).

fwkoch: Seems to work ok with our existing test suite at least - see what you think, I just made a pr that merges into this branch: https://github.com/executablebooks/mystmd/pull/649

rowanc1: Nice, this looks great, thank you.

PR #646, opened 2023-10-10
🧑‍💻 Create flat XML by default in JATS export

rowanc1:
This adds new lines in the XML exports, but has zero indentation (previously this was 2 spaces).

### DIFF
--- a/packages/myst-to-jats/src/index.ts
+++ b/packages/myst-to-jats/src/index.ts
@@ -848,9 +848,14 @@ export function writeJats(file: VFile, content: ArticleContent, opts?: DocumentO
     : doc.body();
   const jats = js2xml(element, {
     compact: false,
-    spaces: opts?.spaces,
+    spaces: opts?.spaces === 'flat' ? 0 : opts?.spaces || 1,
   });
-  file.result = jats;
+  if (!opts?.spaces) {
+    // either `0` or `''`
+    file.result = jats.replace(/\n(\s*)</g, '\n<');
+  } else {
+    file.result = jats;
+  }
   return file;
 }
 
--- a/packages/myst-to-jats/src/index.ts
+++ b/packages/myst-to-jats/src/index.ts
@@ -848,6 +848,8 @@ export function writeJats(file: VFile, content: ArticleContent, opts?: DocumentO
     : doc.body();
   const jats = js2xml(element, {
     compact: false,
+    //  No way to write XML with new lines, but no indentation with js2xml.
+    // If you use 0 or '', you get a single line.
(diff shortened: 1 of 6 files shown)


packages/myst-to-jats/tests/basic.spec.ts
```diff
@@ -75,7 +75,14 @@ beforeEach(() => {
...
   test.each(cases.map((c): [string, TestCase] => [c.title, c]))('%s', async (_, { tree, jats }) => {
-    const pipe = unified().use(mystToJats, SourceFileKind.Article);
+    const pipe = unified().use(
+      mystToJats,
+      SourceFileKind.Article,
+      undefined,
+      undefined,
+      undefined,
+      { spaces: 'flat' },
+    );
```
rowanc1, line 85: @fwkoch, have run into this elsewhere - the API on this plugin is pretty hard to use - we should improve it at some point!

packages/myst-to-jats/src/index.ts
```diff
@@ -848,9 +848,14 @@ export function writeJats(file: VFile, content: ArticleContent, opts?: DocumentO
     : doc.body();
   const jats = js2xml(element, {
     compact: false,
-    spaces: opts?.spaces,
+    spaces: opts?.spaces === 'flat' ? 0 : opts?.spaces || 1,
```
fwkoch, line 851: Ah, I see - it looks like there's not actually a way to write xml with new lines but no indentation with `js2xml`. If you use `0` or `''`, you get a single line.

So we are deviating from the `js2xml` api by making `flat` a single line, and `0` multiple lines with no indentation. 👍

packages/myst-to-jats/tests/basic.spec.ts
fwkoch: Yes - I agree... I feel like there could be lots of cleanup making APIs consistent across MyST (and also we do a lot of wrapping functions in plugins, but then sometimes use the plugins and other times use the functions themselves... would be great to have a little more intentionality about all that as well...)

fwkoch (approved):
This looks fine - took me a few minutes to understand what exactly was going on... but the comment on `DocumentOptions` was quite clear, and that's what matters for using this going forward.

packages/myst-to-jats/src/index.ts
rowanc1: Yeah, code could have been clearer. I will add a comment!

rowanc1:
```suggestion
    //  No way to write XML with new lines, but no indentation with js2xml.
    // If you use 0 or '', you get a single line.
    spaces: opts?.spaces === 'flat' ? 0 : opts?.spaces || 1,
```

PR #652, opened 2023-10-11
🏷 Ability to customize IDs in myst-to-jats

rowanc1:
Still to do is to customize both the types of elements that are numbered as well as pass in the counter functions as options. Ideally through a plugin structure.


rowanc1:
I think these small changes we will bring in now, with @fwkoch having an eye on the affiliation IDs, which need to be done differently.

PR #654, opened 2023-10-11
🎓 JATS author/affiliation updates

fwkoch:
- [x] Separate author contrib-group from "other contributors" contrib-group
- [x] Explicitly list institutions under funding; do not reference them by id
- [x] Add doi institution id
- [x] Add open-funder-registry vocab for specific OFR doi

### DIFF
--- a/packages/myst-to-jats/src/frontmatter.ts
+++ b/packages/myst-to-jats/src/frontmatter.ts
@@ -213,16 +213,20 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
     }
     return { type: 'element', name: 'contrib', attributes, elements };
   };
-  const contribs = [
-    ...(frontmatter.authors ?? []).map((author): Element => {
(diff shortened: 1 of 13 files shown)


packages/myst-to-jats/src/frontmatter.ts
```diff
@@ -213,16 +213,86 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
...
+      name: 'institution-id',
+      attributes: { 'institution-id-type': 'ror' },
+      elements: [{ type: 'text', text: affiliation.ror }],
+    });
+  }
+  if (affiliation.doi) {
+    const doiAttrs: Record<string, string> = { 'institution-id-type': 'doi' };
+    if (affiliation.doi.includes('10.13039')) {
+      doiAttrs.vocab = 'open-funder-registry';
+    }
```
rowanc1, line 270: Maybe we can do a `doi.normalize` here and 'startsWith'?

PR #656, opened 2023-10-11
🌎 Add alternative urls to page frontmatter and JATS self-uri

fwkoch:
This also has a small tweak where departments are removed from JATS funding info.

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -90,7 +90,7 @@ The following table lists the available frontmatter fields, a brief description 
   - page only
 * - `alternatives`
   - alternative URLs and descriptions for the page content (see [](#alternatives))
-  - page only
+  - page & project
 * - `date`
   - a valid date formatted string
   - page can override project
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -88,6 +88,9 @@ The following table lists the available frontmatter fields, a brief description 
 * - `subtitle`
   - a string (max 500 chars)
   - page only
+* - `alternatives`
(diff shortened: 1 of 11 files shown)


docs/frontmatter.md
```diff
@@ -88,6 +88,9 @@ The following table lists the available frontmatter fields, a brief description
 * - `subtitle`
   - a string (max 500 chars)
   - page only
+* - `alternatives`
+  - alternative URLs and descriptions for the page content (see [](#alternatives))
+  - page only
```
rowanc1, line 93: Should this be page only? If we have a zenodo repository that is an alternate, that is for the project.

rowanc1 (approved):
Looks good, wonder if we open up to project as well?

docs/frontmatter.md
fwkoch: Cool, yeah this is an easy change. Should the project values read through to the page though? i.e. should the zenodo repository be an alternative for every single page that doesn't have a different alternative defined?

rowanc1: Maybe not? Certainly not in JATS. They seem like different things, one for the project and one for the page.

rowanc1: @fwkoch should we actually go the other way and only allow this on projects to start with? That seems a bit easier to handle. Right now the links from projects get lost when converting to JATS.

rowanc1 (changes requested):
Let's let this one sit for a bit, there might be other ways to do it in talking through with @fwkoch.

PR #657, opened 2023-10-11
🧪 Allow funding IDs to be numeric

rowanc1:


fwkoch (approved):
Cool, yeah this is nice. `coerceNumber` makes sense; `coerce` is too general since everything (objects, etc) can coerce to string.

Issue #597, opened 2023-09-14
myst-frontmatter: `funding.awards.recipients` should allow institutions in addition to people


fwkoch:
JATS allows award `recipients` to be institutions or people, only the `investigators` must be people. Currently `myst-frontmatter` validates `recipients` only as people, adding them to `contributors`. It would be nice to align with JATS and support institutions as well.

The implementation of this will be slightly complicated: Simply validating an object as one of two types is ok, but we can also reference an existing object by `id`, so we will have to keep track that this `id` may refer to a `contributor` or an `affiliation`... Should be easy enough, but I'm still keeping it out of the original funding PR for now: https://github.com/executablebooks/mystmd/pull/596

PR #659, opened 2023-10-12
🔬 Funding objects hold contributor/affiliation objects directly

fwkoch:
Previously `Contributors` and `Affiliations` could be defined and referenced from anywhere in the frontmatter, e.g. you could do this:
```yaml
author: jdoe
funding:
  investigator:
    id: jdoe
    name: John Doe
    orcid: 0000-0000-0000-0000
```
Or
```yaml
author:
  id: jdoe
  name: John Doe
  orcid: 0000-0000-0000-0000
funding:
  investigator: jdoe
```

Similarly, this could be done with affiliations defined at the top level or nested on contributors anywhere. While flexible, this resulted in some unexpected behaviour, like affiliations from a `funding` living alongside author affiliations at the front of the document...

This PR attempts to provide more sensible behaviour by constraining what is allowed in frontmatter a bit.

Now, funding investigators/recipients/sources may reference existing contributors/affiliations, defined under `authors`/`affiliations`, but then they are just filled in with a copy of the object. They may also define their own investigators/recipients/sources as objects, but these objects are not available elsewhere; they are only part of the funding object.

(So - in the previous yaml snippets, only the second will resolve as expected. In the first, the author `jdoe` will not resolve.)

As a small positive side effect, this addresses #597 - recipients may now be institutions.

### DIFF
--- a/.changeset/shiny-gorillas-drop.md
+++ b/.changeset/shiny-gorillas-drop.md
@@ -1,0 +1,6 @@
+---
+'myst-frontmatter': patch
+'myst-to-jats': patch
+---
+
+Funding objects hold contributor/affiliation objects directly
--- a/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
+++ b/packages/myst-frontmatter/src/frontmatter/frontmatter.spec.ts
@@ -690,7 +690,7 @@ describe('fillPageFrontmatter', () => {
             {
               awards: [
                 {
-                  investigators: ['jn'],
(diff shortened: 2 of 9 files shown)


fwkoch:
This PR is not an improvement to the data structures - it makes things more complex in the end. We should address the unexpected behaviour described above on the tex/jats rendering side, not on the data structure side.

Closing this PR, in favour of a simpler one that addresses rendering. I will also try to persist the fix for #597 which was added here.

PR #662, opened 2023-10-13
👩‍🔬 Only add affiliations associated with authors to JATS and myst-templates

fwkoch:
This addresses the issues in #659 without fundamental changes to frontmatter.

### DIFF
--- a/packages/myst-to-jats/src/frontmatter.ts
+++ b/packages/myst-to-jats/src/frontmatter.ts
@@ -149,7 +149,6 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
     const attributes: Record<string, any> = {};
     const elements: Element[] = [];
     if (type) attributes['contrib-type'] = type;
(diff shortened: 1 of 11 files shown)


packages/myst-to-jats/src/frontmatter.ts
```diff
@@ -217,16 +217,10 @@ export function getArticleAuthors(frontmatter: ProjectFrontmatter): Element[] {
...
-    return generateContrib(contributor);
-  });
   const contribGroups: Element[] = [];
   if (authorContribs.length) {
     contribGroups.push({ type: 'element', name: 'contrib-group', elements: authorContribs });
   }
-  if (otherContribs.length) {
-    contribGroups.push({ type: 'element', name: 'contrib-group', elements: otherContribs });
-  }
   return contribGroups;
```
rowanc1, line 224: Nice, can bring something else back when we do "editors" etc.

rowanc1:
Tested against both JATS and PDF, both look good!

PR #668, opened 2023-10-13
🏫 Simplify affiliation IDs for JATS

fwkoch:
This modifies affiliation ids that appear in JATS to be simpler and uniform, by default `aff-1`, `aff-2`, etc.

It also removes the attempted author/affiliation deduplication (which previously didn't work as intended anyway).
- For authors, they simply shouldn't be deduplicated - you want to know all the different authors on each subarticle. Since we removed author ids from JATS (#662) there is no longer any JATS validation worry.
- Regarding affiliations, we do not deduplicate these either: the implication of this is that identical affiliations in different articles/subarticles will appear separately in each. However, they will also be given a different id, so JATS will be happy. I _think_ it would be nicer if affiliations were compared across articles and deduplicated, but the current behaviour matches citations (identical citations appear in each article). Good enough for now.

Really, what we need to do is rely on the [JATS library](https://github.com/curvenote/jats) to build a JATS-like unist tree, then we can apply all these id changes as post-processing unist transforms, rather than finessing the inputs prior to JATS serialization...

### DIFF
--- a/.changeset/smooth-chairs-clap.md
+++ b/.changeset/smooth-chairs-clap.md
@@ -1,0 +1,5 @@
+---
+'myst-to-jats': patch
+---
+
+Simplify affiliation ids for jats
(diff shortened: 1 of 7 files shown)


rowanc1:
Nice, looking forward to improvements in the Jats library. :)

PR #679, opened 2023-10-16
🐛 Tex output serialize content has been deleted

rowanc1:


fwkoch:
"Better solution" was merged in #683 - still not perfect, there are some comments on that PR.

PR #680, opened 2023-10-16
👩‍💻 Improve HTML processing

rowanc1:
This improves a lot of the functionality around HTML processing to get it into the AST before any other transformations. We can now have things like `<figure><figcaption>The caption</figcaption></figure>` which will be processed into the correct container. This can also work with HTML links directly to that caption, which will be processed as normal.

The IDs are required to be on the `figure`, not the `image`, but that seems reasonable.

cc @dellaert who is motivated by having HTML that works in google collab, but also works with MyST for LaTeX export.

@fwkoch there are probably a number of other changes incoming for the HTML processing along these lines, but I think we can get them in iteratively.

Fixes:
- https://github.com/executablebooks/mystmd/issues/418

### DIFF
--- a/packages/myst-transforms/src/html.spec.ts
+++ b/packages/myst-transforms/src/html.spec.ts
@@ -336,4 +336,87 @@ describe('Test reconstructHtmlTransform', () => {
       ],
     });
   });
+  test('no paragraph when in a paragraph', async () => {
+    const mdast = {
+      type: 'root',
+      children: [
+        {
+          type: 'paragraph',
+          children: [
+            {
+              type: 'text',
+              value: 'See ',
+            },
+            {
+              type: 'html',
+              value: '<sup>',
+            },
+            {
+              type: 'text',
+              value: '[1]',
+            },
+            {
+              type: 'html',
+              value: '</sup>',
+            },
+            {
+              type: 'text',
+              value: '.',
+            },
+          ],
+        },
+      ],
+    };
+    reconstructHtmlTransform(mdast);
+    expect(mdast).toEqual({
+      type: 'root',
+      children: [
+        {
+          type: 'paragraph',
+          children: [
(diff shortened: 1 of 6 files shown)


packages/myst-transforms/src/html.spec.ts
```diff
@@ -162,4 +162,178 @@ describe('Test reconstructHtmlTransform', () => {
...
+          type: 'html',
+          value: '<img src="https://mystmd.org/logo.png"    />',
+        },
+        {
+          type: 'html',
+          value: '<hr>',
+        },
+        {
+          type: 'html',
+          value: '<br>',
```
rowanc1, line 183: Self-closing HTML values are now correctly added.

packages/myst-transforms/src/html.ts
```diff
@@ -66,6 +98,12 @@ export function htmlTransform(tree: GenericParent, opts?: HtmlTransformOptions)
     node.children = mdast.children as Parent[];
     visit(node, (n: any) => delete n.position);
   });
+  selectAll('paragraph > htmlParsed', tree).forEach((parsed) => {
+    const node = parsed as GenericParent;
+    if (node?.children?.length === 1 && node.children[0].type === 'paragraph') {
+      node.children = node.children[0].children as GenericNode[];
+    }
+  });
```
rowanc1, line 106: HTML is by default added into a paragraph, if it is already contained in one -- this strips it.

fwkoch (approved):
Nice - Any implications of pulling in the `reconstructHtmlPlugin` to the default behaviour? Probably note - it will just make the `htmlPlugin` better 🚀. Other than that, this PR adds: ✅ more custom handling of html, ✅ better self closing tags, and ✅ small paragraph nesting tweak. Looks good!

PR #684, opened 2023-10-19
🗣 Do not write implicit latex labels

rowanc1:
We show a warning in MyST if you reference an implicit heading. This removes them from the LaTeX export, which can't handle them and when writing multiple chapters will also warn on duplicate headers.

### DIFF
--- a/packages/myst-to-tex/src/index.ts
+++ b/packages/myst-to-tex/src/index.ts
@@ -15,7 +15,7 @@ import {
 } from './utils.js';
 import MATH_HANDLERS, { withRecursiveCommands } from './math.js';
 import { selectAll } from 'unist-util-select';
-import type { FootnoteDefinition } from 'myst-spec-ext';
+import type { FootnoteDefinition, Heading } from 'myst-spec-ext';
 import { transformLegends } from './legends.js';
 import { TexProofSerializer, proofHandler } from './proof.js';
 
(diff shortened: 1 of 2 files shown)


packages/myst-to-tex/src/index.ts
```diff
@@ -122,7 +122,7 @@ const handlers: Record<string, Handler> = {
     }
     state.renderChildren(node, true);
     state.write('}');
-    if (enumerated !== false && label) {
+    if (enumerated !== false && label && !node.implicit) {
```
fwkoch, line 125: Could this now lead to broken references in latex if something was using this implicit label?

I guess in that case there are warnings that say "you should not use implicit labels" - maybe that's enough of a breadcrumb to put the burden back on the user to fix up?

fwkoch (approved):

packages/myst-to-tex/src/index.ts
rowanc1: Yes, it should now become an error. I think that this is sort of ok, as you get (1) a myst error, and (2) in PDF build a latex error about cross references.

This is something that is also a problem in cross-references that work fine in a site, but not in isolation in a tex export.

I think this is an improvement for now, and there is at least a warning raised. I think we need another pass similar to JATS where we actually do reference matching and manipulation on the final tree (which may have more than one tex article, e.g. in a book).

PR #688, opened 2023-10-20
🌅 Trim option for PDFs images

rowanc1:


packages/myst-cli/src/utils/imagemagick.ts
```diff
@@ -135,7 +138,9 @@ export async function convertImageToWebp(
     return null;
   }
   const imageExt = path.extname(image).toLowerCase();
-  const allowedExtensions = ['.png', '.jpg', '.jpeg', '.tiff', '.gif', '.pdf'];
+  // A race condition may execute this before the PDF --> PNG conversion is complete
```
fwkoch, line 141: I thought this was handled by performing pdf -> png conversion as part of this function - https://github.com/executablebooks/mystmd/blob/3e2e930490c4ff802673983bac20099b9ecf8757/packages/myst-cli/src/utils/imagemagick.ts#L206

I would expect that line to ensure png conversion is complete before running the webp conversion 🤔

fwkoch: Putting that back in would mean we would also need the `-trim` command here: https://github.com/executablebooks/mystmd/blob/3e2e930490c4ff802673983bac20099b9ecf8757/packages/myst-cli/src/utils/imagemagick.ts#L189

fwkoch: Oh - I see from the changeset text, the problem is just overwriting with this command - so, maybe if we add `-trim` here, all will be fine. The png will be overwritten, but it will be overwritten with the correct image.

fwkoch (commented):
Adding options to these conversion functions makes a lot of sense, so we don't have to potentially hard-code many more picky image library options.

However, I think your race condition concerns may be avoidable - see my comment.

packages/myst-cli/src/utils/imagemagick.ts
rowanc1: I was pretty weirded out how we even get to having the PDF in the webp conversion part, and is hinting at a better way to do our image transformations, I think it could be that this is running in two articles and the conversions are running twice (expected) - and then somehow a third time when a webp-conversion is triggered.

I really don't think that the webp function should handle pdf --> png conversion at all. 🤔

Maybe safe for now to just put the `-trim` in this duplicate function.

rowanc1:
Ok, I have made this change a bit smaller based on feedback. The `-trim` call is now in the webp transform now as well, so they are consistent.

The image transforms certainly need some rework, but this is a worthwhile change to make for now!

packages/myst-cli/src/utils/imagemagick.ts
rowanc1: I have backed this out a bit now!

PR #689, opened 2023-10-20
👯‍♀️ Improve the `plural` logging function

rowanc1:


packages/myst-cli-utils/src/utils.ts
```diff
@@ -60,5 +62,8 @@ export function plural(f: string, count?: number | any[] | Record<any, any>): st
       : Array.isArray(count)
       ? count?.length
       : Object.keys(count ?? {}).length) ?? 0;
-  return f.replace('%s', String(num)).replace(/\(s\)/g, num === 1 ? '' : 's');
+  return f
+    .replace('%s', String(num))
+    .replace(/\(s\)/g, num === 1 ? '' : 's')
+    .replace(/\(([a-z0-9A-Z-]*)\|([a-z0-9A-Z-]*)\)/g, num === 1 ? '$1' : '$2');
```
fwkoch, line 68: You can replace line 67 and 68 with this:

```suggestion
    .replace(/\((([a-z0-9A-Z-]*)\|)?([a-z0-9A-Z-]*)\)/g, num === 1 ? '$2' : '$3');
```

In addition to only being one regex, it handles additional cases that were not covered before, e.g. `stitch(es)`.

packages/myst-cli-utils/src/utils.ts
```diff
@@ -60,5 +62,8 @@ export function plural(f: string, count?: number | any[] | Record<any, any>): st
       : Array.isArray(count)
       ? count?.length
       : Object.keys(count ?? {}).length) ?? 0;
-  return f.replace('%s', String(num)).replace(/\(s\)/g, num === 1 ? '' : 's');
+  return f
+    .replace('%s', String(num))
+    .replace(/\(s\)/g, num === 1 ? '' : 's')
```
fwkoch, line 67:
```suggestion
```

packages/myst-cli-utils/src/utils.ts
fwkoch: (Would be good to chuck some tests on this though...)

rowanc1: Time for some tests, yeah. :)

rowanc1: Brought in your changes and added tests!

fwkoch (approved):

PR #693, opened 2023-10-20
👨🏻‍🏫 Add proof alias (`label` <=> `name`)

rowanc1:


rowanc1:
Thought that I had got all of these!

PR #694, opened 2023-10-20
✍️ LaTeX/JATS improvements for Proofs

rowanc1:
This parses algorithm environments in latex and adds the serializers for JATS. There are also a number of latex parsing improvements.


packages/myst-spec-ext/src/types.ts
```diff
@@ -36,6 +37,18 @@ export type CaptionNumber = Parent & {
   enumerator: string;
 };
 
+/**
+ * Line is, e.g., a line in an algorithm and can be numbered as well as indented.
+ * Otherwise this works the same as a paragraph, ideally with tighter styling.
+ * The Line is used in Algorithms (e.g. when parsing from LaTeX)
+ */
+export type Line = Parent & { type: 'line'; indent?: number; enumerator?: string };
```
rowanc1, line 45: @fwkoch this is probably the most controversial change in the PR. I have added a new node of type `Line` which can have indentation and line numbers. This is to support the algorithm parsing, falling back to a div here is totally fine in the current theme.

packages/tex-to-myst/src/algorithms.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+  state.closeParagraph();
+}
+
+function numberParagraphsAsLines(node: GenericNode) {
+  const paragraphs = selectAll('paragraph', node) as Paragraph[];
+  paragraphs.forEach((p, i) => {
+    const l = p as unknown as Line;
+    l.type = 'line';
+    l.enumerator = String(i + 1);
```
fwkoch, line 53: This enumeration is at a granular level, without the customization allowed for other enumerators. Again, probably fine but it means the `enumerator` attribute works in different ways.

Maybe, though, having all the normal enumerator customization here is also the long-term goal.

packages/tex-to-myst/src/frontmatter.ts
```diff
@@ -281,6 +282,17 @@ const FRONTMATTER_HANDLERS: Record<string, Handler> = {
...
       .filter((k) => !!k);
   },
+  macro_newtheorem(node, state) {
+    // https://tex.stackexchange.com/questions/155710/understanding-the-arguments-in-newtheorem-e-g-newtheoremtheoremtheoremsec/155714#155714
+    const [, nameNode, x, labelNode, y] = node.args ?? [];
+    const name = texToText(nameNode);
+    const label = texToText(labelNode);
+    const countWith = texToText(x) || undefined;
+    const countAfter = texToText(y) || undefined;
+    state.data.theorems[name] = { label, countWith, countAfter };
```
fwkoch, line 292: I feel like I'm missing where these theorems stored on `data` are actually consumed... 🤔 Maybe they just aren't yet...?

fwkoch (approved):
I skimmed over a lot of the tex macro, etc, additions (looks like there is decent test coverage, but it doesn't quite get it all - not a big deal, tons of repetition...).

Overall, looks good, left my comments on `Line` but fine with me to go ahead here.

packages/myst-spec-ext/src/types.ts
rowanc1: I did start there, adding attributes to paragraph, and went back and forth. I wanted the JAST to have a "specific-use" identifier and the UI to have a special style, and this seemed special enough for its own type. A div is also probably a more appropriate fallback based on style than a paragraph in this case?

At some point we should do an assessment of these types and do a clean up. I think that we can collapse a few things like `admonitionTitle` and `proof` --> `statement` perhaps.

I changed this to `algorithmLine` based on your first point!

packages/tex-to-myst/src/frontmatter.ts
rowanc1: They are not yet, correct. When we start enumerating theorems in a custom way, they this count information will likely be used.

packages/tex-to-myst/src/algorithms.ts
rowanc1: I did start out with this being called "number", but also felt like we should use common terminology. Certainly a bit different than the other enumerators, with this one being much more granular.

rowanc1:
Thanks for the review @fwkoch, I am going to move ahead with this being a separate type, having this be different enough from a paragraph in the latex and jats implementations. Not totally happy with it, but also I think this is something we can back out of in the future when/if we get to a better solution.

PR #697, opened 2023-10-23
🧮 Add support for sub-equations

rowanc1:
This adds support for latex-parsing of sub-equations and subsequent export to JATS. The approach is to add a `mathGroup` node, and break up the align sub-equations (which are labeled `1a`, `1b`, etc.).

This breaks the alignment in the display equations, which is needed for JATS export as you cannot have one or more nested labels inside of a mathml block in JATS, and they should be a `disp-formula-group`.

In the future we can add better support for both modes with a custom renderer of `mathGroup` that prefers the aligned version, and stores it (currently) in an `equations` attribute. That theme work is not done yet.

### DIFF
--- a/packages/myst-transforms/src/enumerate.ts
+++ b/packages/myst-transforms/src/enumerate.ts
@@ -1,7 +1,7 @@
 import type { Plugin } from 'unified';
 import type { VFile } from 'vfile';
-import type { Container, CrossReference, Heading, Link, Math, Paragraph } from 'myst-spec';
-import type { Cite } from 'myst-spec-ext';
+import type { Container, CrossReference, Heading, Link, Paragraph } from 'myst-spec';
+import type { Cite, Math, MathGroup } from 'myst-spec-ext';
 import type { PhrasingContent } from 'mdast';
 import { visit } from 'unist-util-visit';
 import { select, selectAll } from 'unist-util-select';
@@ -38,6 +38,7 @@ function getDefaultNumberedReferenceLabel(kind: TargetKind | string) {
     case TargetKind.heading:
       return 'Section %s';
     case TargetKind.equation:
+    case TargetKind.subequation:
       return '(%s)';
     case TargetKind.figure:
       return 'Figure %s';
@@ -73,7 +74,7 @@ export enum ReferenceKind {
   eq = 'eq',
 }
 
-type TargetNodes = (Container | Math | Heading) & { html_id: string };
+type TargetNodes = (Container | Math | MathGroup | Heading) & { html_id: string };
 type IdentifierNodes = { type: string; identifier: string };
 
 type Target = {
@@ -83,6 +84,7 @@ type Target = {
 
 type TargetCounts = {
   heading?: (number | null)[];
+  subequation?: [number, number];
 } & Record<string, number>;
 
 export type StateOptions = {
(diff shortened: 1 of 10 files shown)


packages/tex-to-myst/src/math.ts
```diff
@@ -37,4 +39,33 @@ export const MATH_HANDLERS: Record<string, Handler> = {
...
+    state.closeParagraph();
+    state.openNode('mathGroup');
+    state.renderChildren(node);
+    const mathGroup = state.top();
+    transformSubEquations(mathGroup);
+    state.closeNode();
+  },
 };
+
+function transformSubEquations(node: GenericNode) {
```
rowanc1, line 52: I think this could be moved to a math transform in the future. Right now, the only thing that can create sub-equations is latex parsing, and this is targeted only at that!

packages/tex-to-myst/src/math.ts
```diff
@@ -37,4 +39,33 @@ export const MATH_HANDLERS: Record<string, Handler> = {
...
+  const mathNodes = selectAll('math', node) as Math[];
+  node.equations = mathNodes.map((math) => math.value);
+  const children = mathNodes
+    .map((math) => {
+      const tex = math.value.trim();
+      if (!tex.match(/^\\begin\{align\*?\}/)) return math;
+      return tex
+        .replace(/^\\begin\{align\*?\}/, '')
```
fwkoch, line 68: This seems good for now. Will there be problems if only some of the subequations are labeled in latex?

packages/myst-transforms/src/enumerate.ts
```diff
@@ -323,15 +328,31 @@ export class ReferenceState implements IReferenceState {
...
+      // This should be reset every time a mathGroup is encountered (see below)
+      this.targetCounts.subequation ??= [this.targetCounts.equation, 0];
+      this.targetCounts.subequation = [
+        this.targetCounts.subequation[0],
+        this.targetCounts.subequation[1] + 1,
+      ];
+      const sub = this.targetCounts.subequation[1];
+      // Will restart counting if there are more than 26 subequations
+      const letter = String.fromCharCode(((sub - 1) % 26) + 'a'.charCodeAt(0));
+      const enumerator = this.targetCounts.subequation[0] + letter;
```
fwkoch, line 346: These subequations always have sub-enumeration, i.e. there is no option to enumerate equations and subequations all sequentially. Seems like this is something we will want eventually. (And probably more complex than just frontmatter `numbering.subequation = false` - in case in a single paper you need both.)

This definitely seems fine for now though

packages/tex-to-myst/src/math.ts
fwkoch: I guess this would be something like extending the `labelMathNodes` functionality down to this transform... rather than just enumerating them all later.

fwkoch (approved):
This looks good - a few comments thinking ahead to when we are handling this case more generally, but for now with a smaller scope it's great.

packages/tex-to-myst/src/math.ts
rowanc1: The label would be consumed by the sub-equation scope, so if would still start with the `\begin{align}` here I think!

The labelMathNodes transform then does pull these labels out.

packages/myst-transforms/src/enumerate.ts
rowanc1: Yeah, we can also have align groups as `math` - and if we start transforming those into `mathGroups`, then we will want different numbering for each aligned equation.

Good enough for this pass though! :)

PR #702, opened 2023-10-24
🤐 Allow `zipcode` in frontmatter

rowanc1:


fwkoch (approved):
Could do `zip_code` too, hah, but yeah this is good.

rowanc1:
Added!

PR #703, opened 2023-10-24
🐛 LaTeX SI unit parsing bug

rowanc1:


fwkoch (approved):
Didn't go through all the logic here, but previously this resulted in `s ⁻¹ mm²` - so the fix works (and all the other tests still pass).

rowanc1:
Thanks! Yeah, this was an annoying one to track down. :)

We should likely do some work in the future to not make these unicode subscripts but actual children so that JATS is a bit better. This wasn't possible before some of the other theme changes.

PR #704, opened 2023-10-24
🏷 Improve JATS frontmatter export

rowanc1:
This adds a `license-p` paragraph for `CC-BY-4.0` output. Can add more in the future!

This also normalizes orcid output to be the URL.


fwkoch (approved):
Nice, good improvements!

PR #705, opened 2023-10-24
🏷 JATS math label location

rowanc1:
This move the `<label>` to the end of `disp-formula` which is closer to reading order and preferred by publishers.


fwkoch (approved):
👍

Issue #644, opened 2023-10-10
Config errors/warnings are logged twice


fwkoch:
Some of the config loading / validation logic is running twice when starting `myst` - this doesn't cause any problems to the build, but it does cause double logging of error/warning messages: 

![image](https://github.com/executablebooks/mystmd/assets/9453731/77284277-757a-4e24-986e-ca6eaddb6eb3)

Also noticed here: https://github.com/executablebooks/mystmd/pull/463#issuecomment-1742727797

fwkoch:
Resolved by https://github.com/executablebooks/mystmd/pull/708

PR #708, opened 2023-10-26
🐛 Do not reload configs or files unnecessarily

fwkoch:
When consuming file warnings there was a bug that sometimes made them disappear. You see the log message that the warning occurred, but the store no longer has any record of the warning when processing is complete.

This was caused by `getRawFrontmatterFromFile` always re-loading files, even if they were already loaded, which in the process clears all previous warnings. Simply not re-loading the file if it is already loaded prevents this bug.

### DIFF
--- a/packages/myst-cli/src/frontmatter.ts
+++ b/packages/myst-cli/src/frontmatter.ts
@@ -83,7 +83,7 @@ export function prepareToWrite(frontmatter: { license?: Licenses }) {
 
 export async function getRawFrontmatterFromFile(session: ISession, file: string) {
   const cache = castSession(session);
-  await loadFile(session, file);
+  if (!cache.$mdast[file]) await loadFile(session, file);
   const result = cache.$mdast[file];
   if (!result || !result.pre) return undefined;
   const vfile = new VFile();
--- a/packages/myst-cli/src/store/reducers.ts
+++ b/packages/myst-cli/src/store/reducers.ts
@@ -39,18 +39,13 @@ export const config = createSlice({
   } as {
     currentProjectPath: string | undefined;
     currentSitePath: string | undefined;
-    rawConfigs: Record<string, Record<string, any>>;
+    rawConfigs: Record<string, { raw: Record<string, any>; validated: Record<string, any> }>;
     projects: Record<string, ProjectConfig>;
     sites: Record<string, SiteConfig>;
     siteTemplateOptions: Record<string, SiteTemplateOptions>;
     filenames: Record<string, string>;
   },
   reducers: {
-    reload(state) {
-      state.rawConfigs = {};
-      state.projects = {};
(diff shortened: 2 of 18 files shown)


fwkoch:
The scope of this PR has expanded to address the repeated config loading (and validation error messages) from #644

packages/myst-cli/src/store/reducers.ts
```diff
@@ -39,18 +39,13 @@ export const config = createSlice({
...
     currentSitePath: string | undefined;
-    rawConfigs: Record<string, Record<string, any>>;
+    rawConfigs: Record<string, { raw: Record<string, any>; validated: Record<string, any> }>;
     projects: Record<string, ProjectConfig>;
     sites: Record<string, SiteConfig>;
     siteTemplateOptions: Record<string, SiteTemplateOptions>;
     filenames: Record<string, string>;
   },
   reducers: {
-    reload(state) {
```
fwkoch, line 49: Every time we `loadConfigs`, we compare the raw config to the cached raw config and, if it has changed, reload all the site/project configs.

This means we do not need to clear out the state like this; reloading will replace outdated configs.

packages/myst-cli/src/store/reducers.ts
```diff
@@ -39,18 +39,13 @@ export const config = createSlice({
   } as {
     currentProjectPath: string | undefined;
     currentSitePath: string | undefined;
-    rawConfigs: Record<string, Record<string, any>>;
+    rawConfigs: Record<string, { raw: Record<string, any>; validated: Record<string, any> }>;
```
fwkoch, line 42: Previously, the "raw config" that we saved in the store was a partially validated/coerced version of the actual config on disk. That meant to see if the config on disk changed, we had to re-validate before comparing.

Now, we still save the validated config (for using everywhere that we want a validated config), but we also save the totally raw, unvalidated config which we can compare to the file and check for changes without re-running any validation code.

packages/myst-cli/src/frontmatter.ts
```diff
@@ -83,7 +83,7 @@ export function prepareToWrite(frontmatter: { license?: Licenses }) {
 
 export async function getRawFrontmatterFromFile(session: ISession, file: string) {
   const cache = castSession(session);
-  await loadFile(session, file);
+  if (!cache.$mdast[file]) await loadFile(session, file);
```
fwkoch, line 86: This tiny change is all that was required to fix the bug addressed in the top PR comment 😆

packages/myst-cli/src/store/reducers.ts
rowanc1: Nice, this will be helpful in the future as well!

rowanc1:
Tested out, works well!

PR #715, opened 2023-10-30
🔧 Cache mdast at full file path

fwkoch:
There was some mismatch in `cache.$mdast` with saving relative paths, then accessing by resolved paths (e.g. https://github.com/executablebooks/mystmd/compare/feat/mdast-cache?expand=1#diff-e11e13e3b479fbfb25096e4e27367dfcc16a339b97d3cab1a4f94b1a769f599cL139)

This PR swaps to uniformly getting/setting these values with resolved paths.

### DIFF
--- a/.changeset/moody-walls-check.md
+++ b/.changeset/moody-walls-check.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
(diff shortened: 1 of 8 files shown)


rowanc1:
Looks good!

Issue #716, opened 2023-10-31
Allow directives to span multiple or all columns


RutgerK:
I was looking for a way to make a figure span the full with of a document when exporting to PDF and using a multi-column template (eg arxiv_two_column). @rowanc1 mentioned on the discord that this currently isn't fully implemented, but that some of the underlying code already does have a few things in place for this. See:
https://github.com/executablebooks/mystmd/blob/33140d21881349075c45a44b1bebf915db6c236d/packages/myst-to-tex/src/container.ts#L42

That code also hints that it's not only applicable to figures but also other directives like tables, code, math (and perhaps others like glossary/abbreviations?).

## Proposal

I've just started with MyST so I'm not sure how this should work from the user/directive point of view. I'm imagining assigning a property to the directive along the lines of `:class: full-width`, for example:

````
```{figure} ../document/figures/examplefig.png
:name: fig:examplefig
:class: full-width
A nice example figure! 
```
```` 

I guess that would make it available in the AST after parsing (?), and it's up to whatever "exporter" is used to utilize it accordingly. 

Documents (docx/pdf) often won't go beyond 2 columns, but maybe certain HTML or other formats would? If so, instead of a binary full-width or nothing something like a `:colspan: <n>` type property might also be considered. But that might complicate things quite a bit. 

An alternative implementation might be to have some container directive that nests others, analog to a `div` in HTML.

stevejpurves:
@rowanc1 could this also be added usefully to tables? I guess limited to list tables?

RutgerK:
> @rowanc1 could this also be added usefully to tables? I guess limited to list tables?

It already works on list tables, I tested that with the latest release. From my limited understanding of the changed code those also the only two directives for which this is the case.

rowanc1:
Yep! Should work for both tables and figures at the moment!

rowanc1:
Confirmed with @RutgerK that this works for their usecase!

PR #717, opened 2023-10-31
↔️ Full-width images in LaTeX

rowanc1:
Adding `class: full-width` will change a figure to the `figure*` environment that spans columns.

image

See #716

### DIFF
--- a/packages/myst-to-tex/src/container.ts
+++ b/packages/myst-to-tex/src/container.ts
@@ -11,6 +11,15 @@ export enum CaptionKind {
   table = 'table',
 }
 
+function getClasses(className?: string) {
+  const classes =
+    className
+      ?.split(' ')
+      .map((s) => s.trim().toLowerCase())
+      .filter((s) => !!s) ?? [];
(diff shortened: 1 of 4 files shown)


packages/myst-to-tex/src/container.ts
```diff
@@ -37,11 +46,13 @@ export function determineCaptionKind(node: GenericNode): CaptionKind | null {
 
 function nodeToCommand(node: Image | Table | Code | Math) {
   const kind = determineCaptionKind(node);
+  const classes = getClasses((node as any).class);
+  const fullWidth = classes.includes('full-width') || classes.includes('w-full');
   switch (kind) {
     case CaptionKind.fig:
-      return (node as any).fullpage ? 'figure*' : 'figure';
```
fwkoch, line 42: Ah, cool, so before this was just an arbitrary attribute that was not referred to anywhere else.

Now, it just takes advantage of the existing container `class` property.

fwkoch (approved):

Issue #719, opened 2023-11-03
Citation key with ending period should exclude the period


rowanc1:
![image](https://github.com/executablebooks/mystmd/assets/913249/bc7ed074-46d4-45bc-a995-55973c643840)

cc @alanlujan91

PR #720, opened 2023-11-03
🐛 Citations with trailing punctuation

rowanc1:
Can now also use the `@{Foo.}` syntax if you want to include trailing punctuation.

See #719


packages/markdown-it-myst/tests/citations.yml
```diff
@@ -0,0 +1,187 @@
...
+              kind: parenthetical
+              partial: year
+              prefix:
+                - content: see
+          - type: cite_group_close
+          - type: text
+            content: '.'
+      - type: paragraph_close
+  - title: Excludes ending punctuation
+    md: '...geophysical inversions @simpeg2015. However,'
```
rowanc1, line 99: @alanlujan91 does this fix what you expect?

packages/markdown-it-myst/src/citations.ts
```diff
@@ -22,7 +25,7 @@ export interface Citation {
 export const citationsPlugin: PluginWithOptions = (md) => {
   const regexes = {
     citation: /^([^^-]|[^^].+?)?(-)?@([\w][\w:.#$%&\-+?<>~/]*)(.+)?$/,
-    inText: /^@([\w][\w:.#$%&\-+?<>~/]*)(\s*)(\[)?/,
+    inText: /^@([\w|{][\w:.#$%&\-+?<>~/]*[\w|}])(\s*)(\[)?/,
```
fwkoch, line 28: Ah, ok, so this says the last character of the citation (besides whitespace) must be alphanumeric/underscore.

Definitely makes sense for punctuation, e.g. `@a&b.` leaving off the `.`. Something like `@a&b&` would leave of the last `&` - I don't know what sort of silly things are done with these citation keys, but I assume this is fine. Definitely an improvement to pulling in all punctuation to the key.

fwkoch: Seems like if we really wanted to handle a single character citation key, we could probably expand this regex to have 2 cases, something like this... (if we are actually making this change, we would probably need to update some of the match group indexing...)

```suggestion
    inText: /^@(([\w|{][\w:.#$%&\-+?<>~/]*[\w|}])|\w)(\s*)(\[)?/,
```

fwkoch (approved):
This is great, nice improvement to the tests too 😀

packages/markdown-it-myst/src/citations.ts
rowanc1: I think if there is a complex case like that suggesting `@{a&b&}` is better. I am updating to handle the single character -- using a non-capture group `(?:[\w])`.

rowanc1:
Thanks for pushing to fix the single citations, that is better.

PR #721, opened 2023-11-03
🎆Write images when serializing, not during mdast processing

fwkoch:
This addresses most of https://github.com/executablebooks/mystmd/issues/709

The image/output steps that write to disk are pulled out of the base `process/postProcessMdast` functions into a separate processing step called at final serialization for export / site build.


rowanc1:
This looks good. Tested out on multiple tex exports, site build, jats.

Issue #593, opened 2023-09-13
Myst Markdown table directive not supported


dressedfez:
## Description

According to https://myst-parser.readthedocs.io/en/latest/syntax/tables.html#table-with-captions should Myst Markdown
support the `table` directive. This is not the case as can be seen here in this [sandbox](https://mystmd.org/sandbox?tab=demo&myst=OgA6ADoAewB0AGEAYgBsAGUAfQAgAFQAYQBiAGwAZQAgAGMAYQBwAHQAaQBvAG4ACgA6AHcAaQBkAHQAaABzADoAIABhAHUAdABvAAoAOgBhAGwAaQBnAG4AOgAgAGMAZQBuAHQAZQByAAoACgB8ACAAZgBvAG8AIAB8ACAAYgBhAHIAIAB8AAoAfAAgAC0ALQAtACAAfAAgAC0ALQAtACAAfAAKAHwAIABiAGEAegAgAHwAIABiAGkAbQAgAHwACgA6ADoAOgA%3D).

## Additional notes

Since standard markdown tables do no support captions. This would be nice to have.

agoose77:
It's an unfortunate side-effect of having so much exciting activity in this space that it's not always clear where different projects begin and end. The `{table}` directive is a Python `myst-parser` feature, distinct from `mystmd` which pertains to the JS stack. 

What unifies the two endeavours is the [myst-spec](https://mystmd.org/spec/tables) which sets out the expected AST and source code. As such, the myst-spec is a subset of the features that e.g. myst-parser supports.

In time, we'll look to unify all of these by growing the spec. I can absolutely see how the table directive is useful. Once @rowanc1 takes a look at this, he might be able to expand on this :)

dressedfez:
If @rowanc1 could point me to the relevant classes, I could give it a try.

rowanc1:
Hi @dressedfez and @agoose77  -- this should be added in #726!

rowanc1:
Table directive should be coming in v1.1.27 later this week!

PR #726, opened 2023-11-03
🍽 Add table directive

rowanc1:
See #593

### DIFF
--- a/packages/myst-directives/src/table.ts
+++ b/packages/myst-directives/src/table.ts
@@ -1,5 +1,6 @@
 import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
 import { fileError, normalizeLabel, RuleId } from 'myst-common';
+import { select } from 'unist-util-select';
 import type { VFile } from 'vfile';
 
 export const tableDirective: DirectiveSpec = {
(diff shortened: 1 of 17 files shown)


packages/myst-directives/src/table.ts
```diff
@@ -2,6 +2,54 @@ import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
...
+  },
+  run(data: DirectiveData): GenericNode[] {
+    const children = [];
+    if (data.arg) {
+      children.push({
+        type: 'caption',
+        children: [{ type: 'paragraph', children: data.arg as GenericNode[] }],
+      });
+    }
+    children.push(...(data.body as GenericNode[]));
```
fwkoch, line 39: Could add validation here that children includes a `table` node? 🤷‍♀️ Trade-off between strictness now vs. unexpected behaviour later.

Looks like in other directives we err towards less strict, so maybe best to leave as-is.

fwkoch (approved):

packages/myst-directives/src/table.ts
rowanc1: I did think about that as well. Probably a good idea to add it and at least warn.

rowanc1:
@fwkoch to support better error messages I had to include `node` in the directive/role data. This allows us to raise specific errors about the directive.

This includes errors about options, which now point directly to the line where the error occurred:
image

fwkoch:
Cool, yeah this is great! Gives the `run` function more power (I don't think there's any drawback to that - for the most part, using validated options/args/body in the run function will be easier and preferable).

Hmm, you know, we also have the `validate` function which now gets the `node` as well... worth considering if separating `run` and `validate` is actually useful or if we should just keep the extra validation inside the `run` function as we do everywhere, including this PR. (And just delete `validate` entirely)

rowanc1:
I was thinking that as well, the only place we are using validate is in `list-table`, might be overkill. I suppose there may be a reason to validate without transforming? If you think we should remove this .. can we open another PR to do it?

Going to get this one in! 🚀

Issue #710, opened 2023-10-26
Markdown quotes


alanlujan91:
It would be helpful for myst to convert markdown quotes " to latex quotes ``.

rowanc1:
Thank you! We also likely want to turn on smart-quotes for the HTML outputs as well. There are a few tools I have found to do this:
- https://github.com/ngsctt/remark-smartypants
- https://unifiedjs.com/explore/package/retext-quotes/

rowanc1:
This should be fixed up in #727 and released next week!

PR #727, opened 2023-11-03
＂ Enable smartquotes

rowanc1:
See #710

### DIFF
--- a/packages/myst-parser/src/myst.ts
+++ b/packages/myst-parser/src/myst.ts
@@ -29,6 +29,7 @@ export const defaultOptions: Omit<AllOptions, 'vfile'> = {
     html: true,
   },
   extensions: {
+    smartquotes: true,
     colonFences: true,
     frontmatter: true,
     math: true,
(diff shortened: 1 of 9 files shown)


packages/myst-parser/src/myst.ts
```diff
@@ -59,7 +60,17 @@ function parseOptions(opts?: Options): AllOptions {
 export function createTokenizer(opts?: Options) {
   const parsedOpts = parseOptions(opts);
   const { extensions, markdownit } = parsedOpts;
-  const tokenizer = MarkdownIt(MARKDOWN_IT_CONFIG as any, markdownit);
+  const tokenizer = MarkdownIt(
+    {
+      ...MARKDOWN_IT_CONFIG,
+      options: {
+        ...MARKDOWN_IT_CONFIG.options,
+        typographer: extensions.smartquotes ? true : false,
```
fwkoch, line 68: Cool, so this is required to enable any of the typographic features - if it's `false` smartquotes don't get replaced because of this: https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/smartquotes.js#L190

packages/myst-parser/src/myst.ts
```diff
@@ -59,7 +60,17 @@ function parseOptions(opts?: Options): AllOptions {
...
+      },
+    } as any,
+    markdownit,
+  );
+  if (extensions.smartquotes) tokenizer.enable('smartquotes');
```
fwkoch, line 73: Then, once `typographer` is `true`, enabling `smartquotes` is still required to actually add the feature: https://github.com/markdown-it/markdown-it/blob/e476f78bc3ea3576beb61bdc94322d0a6b2d85cc/lib/parser_core.js#L19

fwkoch (approved):
This looks great - nice when features are already implemented upstream 😄 

Only comment is maybe we add a test to `myst-to-tex` that verifies smart quotes there, as that was the original issue?

packages/myst-parser/src/myst.ts
rowanc1: I beleive that turning on typographer is probably a performance optimization in the library. So only enabling it here when we need it seemed like a good idea?

rowanc1:
Will add a end-to-end test as suggested!

PR #730, opened 2023-11-07
🎆 Write thumbnails/banner back to redux store

fwkoch:
Thumbnail and banner paths in the site config were broken by this PR: #721 - The image transformations were moved later; we kept track of these changes in the file frontmatter, but we did not call `updateFileInfo` again with correct thumbnail/banner values.

Just updating file info again after image writing corrects the site config.


rowanc1:
Works great! :)

PR #734, opened 2023-11-07
🔧 Refactor myst-frontmatter

fwkoch:
Primarily, this PR cleans up `myst-frontmatter` from basically a single huge file with code+yaml tests, to granular folders for each frontmatter "feature" and tests in yaml as much as possible.

Among the giant diffs, a few small interesting changes:
- types/aliases are better shared between `myst-frontmatter` and `myst-config`
  - this led to a bug fix in `simple-validators` where aliases with invalid values were bypassing validation; now aliases with invalid values are just ignored (this makes it easier to reuse alias dictionaries across types}
- `postal_code` now takes advantage of the `coerceNumber` option on string validation
- tiny modifications to `thebe` validation - previously when there were errors, it could validate to an invalid object (then validating again would change it again...). Valid thebe values (including values that coerce with no errors) are unchanged.
- `coerce` option added to `validateList` - this reduces some boilerplate all over where we were coercing non-arrays to arrays before validating


rowanc1 (approved):
Haven't taken a close look through but talked through with @fwkoch. If you are happy -- go for the merge. The new changes sound great. 🚀

PR #736, opened 2023-11-09
Add `options` and `parts` to frontmatter

fwkoch:
- __options__ are template options for site config or page/project frontmatter; this (1) allows options to be provided in a centralized place, rather than on each export, and (2) gives an explicit place for these arbitrary options to be found
  - [x] TODO: tidy up warning messages for unused options, now that options may apply to different exports, so they may be unused on one place but used in another.
- __parts__ are document parts, previously only defined with block metadata `part: 'abstract'`, now you can put the abstract directly in the frontmatter. It is then inserted into the document as a hidden block, and transformed with the rest of the content.
  - [x] TODO: audit exports to make sure ~`hidden: true`~ `visibility: 'remove'` is respected on blocks - This is the case for exports defined in mystmd.
  - [x] For site builds, I added this pr: https://github.com/executablebooks/myst-theme/pull/259

### DIFF
--- a/packages/myst-config/src/site/types.ts
+++ b/packages/myst-config/src/site/types.ts
@@ -61,7 +61,7 @@ type ManifestProject = {
 } & ProjectFrontmatter;
 
 export type SiteManifest = SiteFrontmatter & {
-  myst: 'v1';
+  myst: string;
   id?: string;
   projects?: ManifestProject[];
   nav?: SiteNavItem[];
--- a/packages/myst-config/src/site/types.ts
+++ b/packages/myst-config/src/site/types.ts
@@ -22,8 +22,6 @@ export interface SiteAction {
   static?: boolean;
 }
 
-export type SiteTemplateOptions = Record<string, any>;
-
 export type SiteConfig = SiteFrontmatter & {
   projects?: SiteProject[];
   nav?: SiteNavItem[];
@@ -31,7 +29,7 @@ export type SiteConfig = SiteFrontmatter & {
   domains?: string[];
   favicon?: string;
   template?: string;
-} & SiteTemplateOptions;
+};
 
 type ManifestProjectItem = {
   title: string;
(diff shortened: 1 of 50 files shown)


packages/myst-config/src/site/types.ts
```diff
@@ -61,7 +61,7 @@ type ManifestProject = {
 } & ProjectFrontmatter;
 
 export type SiteManifest = SiteFrontmatter & {
-  myst: 'v1';
+  myst: string;
   id?: string;
   projects?: ManifestProject[];
   nav?: SiteNavItem[];
```
rowanc1, line 67: Should this type now have `options: SiteTemplateOptions`?

packages/myst-to-jats/src/types.ts
```diff
@@ -1,4 +1,4 @@
-import type { GenericNode, MessageInfo } from 'myst-common';
+import type { GenericNode, GenericParent, MessageInfo } from 'myst-common';
```
rowanc1, line 1:
```suggestion
import type { GenericNode, MessageInfo } from 'myst-common';
```

rowanc1 (commented):
Some comments!

rowanc1:
I think merging https://github.com/executablebooks/mystmd/pull/741 would be good. We need to by default turn the extracted abstracts, etc. to visible.

rowanc1:
Have tested parts and that works great! Still need to test out options.

packages/myst-config/src/site/types.ts
fwkoch: Hmm, we don't need to add it here as it comes in from `SiteFrontmatter` which now has `options?: Record<string, any>;` (and if you dig into `SiteTemplateOptions`, that type is actually just `Record<string, any>`).

I think maybe we should just delete `SiteTemplateOptions` as a type, since it only confuses the fact that it's a simple dict.

fwkoch: Oh, and we should remove the `& SiteTemplateOptions` here I guess!

stevejpurves:
@fwkoch do you have an example `myst.yml` file that you can post here showing the layout of parts and options after this PR?

PR #738, opened 2023-11-09
🐛 Turn off glossaries for rawTex output

rowanc1:
I have been working through exporting a book in latex and do not want each page to have the glossaries output. This turns it off when there is no template.


fwkoch (approved):
Easy!

PR #750, opened 2023-11-14
📦 Remove `JupyterLocalOptions`

rowanc1:


stevejpurves (approved):
yes I think it the left over from when other "local" options were removed

Issue #752, opened 2023-11-15
`myst-frontmatter` refactor bug - CJS/ESM problem


rowanc1:
I have been having some trouble with the frontmatter refactor @fwkoch in a Remix project that compiles to CJS. The `SITE_FRONTMATTER_KEYS` are undefined when they are imported in the project/validators. I think this is because of circular imports, with a few things importing from the main index file.

I couldn't nail it down easily though. I think it would be helpful to pull these `consts` into a single file that executes in order?

Both of these are undefined `console.log({SITE_FRONTMATTER_KEYS, FRONTMATTER_ALIASES})`:

![image](https://github.com/executablebooks/mystmd/assets/913249/8b028d28-de89-4b8d-b15d-fd92000e06b2)

PR #753, opened 2023-11-15
🐛 Move validations constants to single folder

rowanc1:
See #752


rowanc1:
Closing in favour of #754.

PR #759, opened 2023-11-17
⚙️ Add `settings` to project

rowanc1:
This adds a few known settings to the project/page, these are meant to change how parsing, plugins or transforms behave. The first settings I put in are around removing notebook stderr/stdout outputs from cells and removing matplotlib strings.

### DIFF
--- a/packages/myst-cli/src/transforms/outputs.ts
+++ b/packages/myst-cli/src/transforms/outputs.ts
@@ -129,7 +129,7 @@ export function transformFilterOutputStreams(
                 : 'Output contains matplotlib string',
               {
                 node: output,
-                note: `${content}\n   You can suppress this output by putting a semicolon after the final output in your cell or add "output_matplotlib_strings: remove" to the project settings.`,
+                note: `${content}\n   Fix: Put a semicolon on the last line of your cell or add "output_matplotlib_strings: remove" to the project settings.`,
               },
             );
           }
--- a/packages/myst-cli/src/transforms/outputs.ts
+++ b/packages/myst-cli/src/transforms/outputs.ts
@@ -53,7 +53,7 @@ export function stringIsMatplotlibOutput(value?: string): boolean {
   // We can add more when we find them...
   const match =
     value.match(/<(Figure|Text|matplotlib)(.*)at ([0-9a-z]+)>/) ||
-    value.match(/^<(Figure|Text|AxesSubplot)(.*)>$/) ||
+    value.match(/^<(Figure|Text|AxesSubplot|module)(.*)>$/) ||
     value.match(/^(Text)\((.*)\)$/);
   return !!match;
 }
@@ -68,10 +68,9 @@ export function transformFilterOutputStreams(
     output_matplotlib_strings: mpl = 'remove-warn',
   }: Pick<ProjectSettings, 'output_stdout' | 'output_stderr' | 'output_matplotlib_strings'> = {},
 ) {
-  if (stdout === 'show' && stderr === 'show') return;
   const blocks = selectAll('block', mdast) as GenericNode[];
   blocks.forEach((block) => {
-    const tags: string[] = Array.isArray(block.data.tags) ? block.data.tags : [];
+    const tags: string[] = Array.isArray(block.data?.tags) ? block.data.tags : [];
     const blockRemoveStderr = tags.includes('remove-stderr');
     const blockRemoveStdout = tags.includes('remove-stdout');
     const outputs = selectAll('output', block) as GenericNode[];
--- a/packages/myst-cli/src/transforms/outputs.ts
+++ b/packages/myst-cli/src/transforms/outputs.ts
@@ -11,7 +11,7 @@ import { remove } from 'unist-util-remove';
 import { selectAll } from 'unist-util-select';
 import type { VFile } from 'vfile';
 import type { IOutput, IStream } from '@jupyterlab/nbformat';
-import type { MinifiedContent, MinifiedOutput } from 'nbtx';
+import type { MinifiedContent, MinifiedOutput, MinifiedMimeOutput } from 'nbtx';
 import { ensureString, extFromMimeType, minifyCellOutput, walkOutputs } from 'nbtx';
 import { castSession } from '../session/cache.js';
 import type { ISession } from '../session/types.js';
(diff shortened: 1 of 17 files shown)


packages/myst-frontmatter/src/project/types.ts
```diff
@@ -5,6 +5,14 @@ import type { Numbering } from '../numbering/types.js';
 import type { SiteFrontmatter } from '../site/types.js';
 import type { Thebe } from '../thebe/types.js';
 
+type OutputRemovalOptions = 'show' | 'remove' | 'remove-warn' | 'remove-error' | 'warn' | 'error';
+
+export type ProjectSettings = {
+  output_stderr?: OutputRemovalOptions;
```
fwkoch, line 11: The attribute in jupyterbook is reversed `stderr_output` - any value to matching that? https://jupyterbook.org/en/stable/content/execute.html#dealing-with-code-that-produces-stderr (The `settings` key and everything is different, so not sure how much it matters)

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
...
+    const blockRemoveStderr = tags.includes('remove-stderr');
+    const blockRemoveStdout = tags.includes('remove-stdout');
+    const outputs = selectAll('output', block) as GenericNode[];
+    // There should be only one output in the block
+    outputs.forEach((output) => {
+      output.data = output.data.filter((data: IStream | MinifiedMimeOutput) => {
+        if (stderr !== 'show' && data.output_type === 'stream' && data.name === 'stderr') {
+          const doRemove = stderr.includes('remove') || blockRemoveStderr;
+          const doWarn = stderr.includes('warn');
+          const doError = stderr.includes('error');
```
fwkoch, line 84: It's a little confusing how all of these are dependent on the parts of the string value (i.e. it's not like an enum where you can swap out the actual values without changing the functionality) - especially since the values are defined all the way in `myst-frontmatter`... There's also repetition across these three filtering functions...

Potentially some of the logic could be pulled into a separate function... but hmf, it's probably fine as is, just takes a couple passes to read.

fwkoch (commented):
Looks good - my only concern is if matplotlib text output is actually getting removed by default, as intended. (also with cell-level `remove` tags)

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
...
+  {
+    output_stdout: stdout = 'show',
+    output_stderr: stderr = 'show',
+    output_matplotlib_strings: mpl = 'remove-warn',
+  }: Pick<ProjectSettings, 'output_stdout' | 'output_stderr' | 'output_matplotlib_strings'> = {},
+) {
+  if (stdout === 'show' && stderr === 'show') return;
+  const blocks = selectAll('block', mdast) as GenericNode[];
+  blocks.forEach((block) => {
+    const tags: string[] = Array.isArray(block.data.tags) ? block.data.tags : [];
```
fwkoch, line 74:
```suggestion
    const tags: string[] = Array.isArray(block.data?.tags) ? block.data.tags : [];
```

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
...
+    const blockRemoveStderr = tags.includes('remove-stderr');
+    const blockRemoveStdout = tags.includes('remove-stdout');
+    const outputs = selectAll('output', block) as GenericNode[];
+    // There should be only one output in the block
+    outputs.forEach((output) => {
+      output.data = output.data.filter((data: IStream | MinifiedMimeOutput) => {
+        if (stderr !== 'show' && data.output_type === 'stream' && data.name === 'stderr') {
```
fwkoch, line 81: Should this also capture `output_type === 'error'`?

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
   );
 }
 
+export function stringIsMatplotlibOutput(value?: string): boolean {
+  if (!value) return false;
+  // We can add more when we find them...
+  const match =
+    value.match(/<(Figure|Text|matplotlib)(.*)at ([0-9a-z]+)>/) ||
+    value.match(/^<(Figure|Text|AxesSubplot)(.*)>$/) ||
```
fwkoch, line 56:
```suggestion
    value.match(/^<(Figure|Text|AxesSubplot|module)(.*)>$/) ||
```

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
...
+export function transformFilterOutputStreams(
+  mdast: GenericParent,
+  vfile: VFile,
+  {
+    output_stdout: stdout = 'show',
+    output_stderr: stderr = 'show',
+    output_matplotlib_strings: mpl = 'remove-warn',
+  }: Pick<ProjectSettings, 'output_stdout' | 'output_stderr' | 'output_matplotlib_strings'> = {},
+) {
+  if (stdout === 'show' && stderr === 'show') return;
```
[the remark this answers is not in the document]
rowanc1: Nope, this is just a bug, thank you for catching it!

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -47,6 +48,99 @@ export async function transformOutputsToCache(
...
+  const blocks = selectAll('block', mdast) as GenericNode[];
+  blocks.forEach((block) => {
+    const tags: string[] = Array.isArray(block.data.tags) ? block.data.tags : [];
+    const blockRemoveStderr = tags.includes('remove-stderr');
```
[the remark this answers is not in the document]
rowanc1: Argh. Yep! Good catch, thank you. :)

packages/myst-frontmatter/src/project/types.ts
rowanc1: Added aliases!
It is the opposite in mystnb...!

packages/myst-cli/src/transforms/outputs.ts
rowanc1: I think we will want to do errors/exceptions in a new pass?

packages/myst-cli/src/transforms/outputs.ts
rowanc1: I am not sure if this one deserves to be an enum. I agree that the code is pretty repetitive, I think I am going to leave it until another pass in this area though.

I am going to start adding an enum for the known cell tags.

Issue #369, opened 2023-04-30
Use listings to highlight code


finsberg:
### Which area is this feature request for?

LaTeX Export

### Describe the feature you'd like to request

Thanks for a great tool. I really think this can transform the way we write scientific papers in the future. 

I am trying to write a document with some code blocks. This is not necessarily targeting any specific journals but it could for example be a text book about programming. In this case I would like to swap out the `verbatim` environment with listing, e.g as explained at the bottom of this page: https://www.overleaf.com/learn/latex/Code_listing

### Describe the solution you'd like

I would like to create a template where I can swap out the `verbatim` environment with `listing`. Potentially, this could also be an option in the frontmatter.  The template could then define styles for the different languages. 

I would guess that information about the language is parsed by myst (e.g here: https://github.com/executablebooks/mystjs/blob/20282a919e9eff1778c50e4fc39c3ac1f1e5fbed/packages/myst-directives/src/code.ts#L143-L156) and that the option of using `listing` rather than `verbatim` needs to somehow be implemented in https://github.com/executablebooks/mystjs/blob/20282a919e9eff1778c50e4fc39c3ac1f1e5fbed/packages/myst-to-tex/src/index.ts#L83-L88

I am happy to try to submit a PR, but would need to guidance in how this should be implemented.

### Describe alternatives you've considered

Consider the following example in markdown
````markdown
---
title: Code listing
date: 2023-04-30

exports:
  - format: pdf
    template: ./curvenote
    output: main.pdf
---

Example code with verbatim

```python
def print_hello():
    print("Hello")

x = 42
```

Text after code
````
Here I am using a local version of the curvenote template (https://github.com/myst-templates/curvenote.git). This produces the following output
<img width="631" alt="Screenshot 2023-04-30 at 16 52 41" src="https://user-images.githubusercontent.com/2010323/235359735-f0fad667-4135-4cf4-a8f8-6e60d31f5f0b.png">

However, I would like to get syntax highlighting and possible create a template that uses `listing` rather than `verbatim`.

One way to make it working (using the curvenote template) is to add the following lines at the end of `curvenote.sty`

```tex
\usepackage{listings}

\lstdefinestyle{myStyle}{
    belowcaptionskip=1\baselineskip,
    breaklines=true,
    frame=none,
    numbers=none,
    basicstyle=\footnotesize\ttfamily,
    keywordstyle=\bfseries\color{green!40!black},
    commentstyle=\itshape\color{purple!40!black},
[...]

PR #764, opened 2023-11-21
⚙️ Add settings for latex rendering

rowanc1:
- Introduce mystToTex settings
- Including minted, listings, or verbatim for code cells

See also #369

### DIFF
--- a/packages/myst-cli/src/build/tex/single.ts
+++ b/packages/myst-cli/src/build/tex/single.ts
@@ -52,6 +52,7 @@ export function mdastToTex(
     bibliography: templateYml?.style?.bibliography,
     printGlossaries,
     references,
+    ...frontmatter.settings?.myst_to_tex,
   });
   const result = pipe.runSync(mdast as any);
   const tex = pipe.stringify(result);
--- a/packages/myst-frontmatter/src/settings/validators.ts
+++ b/packages/myst-frontmatter/src/settings/validators.ts
@@ -1,0 +1,67 @@
+import type { ValidationOptions } from 'simple-validators';
+import { defined, incrementOptions, validateChoice, validateObjectKeys } from 'simple-validators';
+import type { ProjectSettings } from './types.js';
+import { validateMystToTexSettings } from './validatorsMystToTex.js';
+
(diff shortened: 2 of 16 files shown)


packages/myst-frontmatter/src/settings/validators.ts
```diff
@@ -0,0 +1,67 @@
...
+export const PROJECT_SETTINGS = [
+  'output_stderr',
+  'output_stdout',
+  'output_matplotlib_strings',
+  'myst_to_tex',
+];
+export const PROJECT_SETTINGS_ALIAS = {
+  stderr_output: 'output_stderr',
+  stdout_output: 'output_stdout',
+  mystToTex: 'myst_to_tex',
```
fwkoch, line 23: This could just be `tex` settings? Maybe we add that as an alias at least? Unless there's some other meaning for "tex settings" besides tex exports.

packages/myst-cli/src/build/tex/single.ts
```diff
@@ -52,6 +52,7 @@ export function mdastToTex(
     bibliography: templateYml?.style?.bibliography,
     printGlossaries,
     references,
+    ...frontmatter.settings?.myst_to_tex,
```
fwkoch, line 55: We need to be a little more careful with this - currently we only set `printGlossaries` to `true` in one place: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/build/tex/single.ts#L214 - a bunch of other places it must be false so we don't re-print a ton.

I think that line is where we want to consume `frontmatter.settings.myst_to_tex.printGlossaries` - everywhere else it still needs to be false.

But then we still need to unpack the other settings here...

fwkoch (commented):
Looks good - one issue around `printGlossaries` that needs some tidying I think.

packages/myst-frontmatter/src/settings/validators.ts
rowanc1: I have been wanting some settings for the texToMyst parser for a while, so that was leaving room for that. Will add the alias!

packages/myst-cli/src/build/tex/single.ts
rowanc1: Hm, ok. I think I am just going to take it out for now.

PR #766, opened 2023-11-21
🧮 Introduce math simplifications

rowanc1:
- subscript/superscript
- single latex symbols
- degrees


packages/myst-transforms/src/mathSimplifications.ts
```diff
@@ -0,0 +1,165 @@
...
+  const match = node.value.match(/^(-?[0-9.]+)$/);
+  if (!match) return false;
+  const text = match[1];
+  if (!text) return false;
+  (node as any).type = 'text';
+  node.value = text;
+  return true;
+}
+
+export function inlineMathSimplificationTransform(mdast: GenericParent) {
```
fwkoch, line 150: Ah, so this totally eliminates `inlineMath` nodes if they can be simply represented by symbols and basic markdown. If the inline math has any marginal complexity at all (interleaved sub/superscript, macros, unknown symbols, etc), it is just left alone and handled by whatever latex math renderer.

fwkoch (approved):
Looks good - seeing the tests is helpful.

It might be nice to have tiny docstrings to explain what cases each function captures? I was getting a little lost trying to parse regexes. But being able to trust the tests is also good.

rowanc1:
Thanks will add more docs!!

packages/myst-transforms/src/mathSimplifications.ts
rowanc1: Yeah, I think we can have some settings on this in the future. Most (I think all) of these are also captured on the way back to latex, however will be `\textsuperscript{th}` instead of `$^{th}`, which is actually correct!

PR #770, opened 2023-11-22
🎆 First pass at subfigures

fwkoch:
image

- Moves figure transformation to a single transform
- Figures-in-figures are subfigures (or "subcontainers")
- Enumeration is added for these subfigures


rowanc1:
Added a theme PR to do a few updates! This is looking great. Just a few requests on `figure-a` labels rather than `figure-sub-0` and turning those implicit.

PR #772, opened 2023-11-22
📮 Support multiple `articles` in a single export

fwkoch:
- `article` is now coerced to `articles` in export frontmatter
- Multiple articles are only valid for tex/pdf - other cases raise errors
- To support this, tex `preamble` must be kept as structured data and compiled at the end (similar to how tex `imports` are already handled)

### DIFF
--- a/.changeset/cold-pans-lay.md
+++ b/.changeset/cold-pans-lay.md
@@ -1,0 +1,5 @@
+---
+'myst-to-tex': patch
+---
+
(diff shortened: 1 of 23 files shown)


rowanc1:
This works great! Have added the first book template and it works well out of the gate. There needs to be another pass to maintain the titles and change the headings before exporting to tex, and having this be based on the table of contents (e.g. chapters).

Issue #773, opened 2023-11-22
Article toctree configuration


gouarin:
For now, the toctree is defined by the headings of an article and displayed in the right sidebar. The depth of this tree is fixed.

It would be interesting to be able to modify the behavior of the toctree view as it is done into sphinx. 
For example, we could define the keyword `maxdepth` which controls the depth of the tree.

rowanc1:
I am thinking that the site option should be prefixed with `outline_`. Here is an example of what I am thinking:

![outline_maxdepth](https://github.com/executablebooks/mystmd/assets/913249/2e629e8d-a9f5-46de-8acd-a523124f5f3c)

Let me know if this is what you were thinking as well!

rowanc1:
This has been released in v1.1.31 (and myst-theme v0.5.16). Maybe sure you run `myst clean --templates` and then you can add:

```yaml
site:
  options:
    outline_maxdepth: 2
```

And it should now work! Let us know how it goes.

PR #774, opened 2023-11-22
⚙️ Allow min/max/number on template types

rowanc1:
This adds the min/max/integer options to the template defs, for example in #773 we need an integer number with bounds.

### DIFF
--- a/packages/myst-templates/src/validators.spec.ts
+++ b/packages/myst-templates/src/validators.spec.ts
@@ -370,7 +370,6 @@ describe('validateTemplateOptionDefinition', () => {
           description: 'desc',
           required: true,
           default: 'a',
-          max_chars: 10,
           condition: {
             id: 'short_title',
             value: 'test',
@@ -385,7 +384,6 @@ describe('validateTemplateOptionDefinition', () => {
       description: 'desc',
       required: true,
       default: 'a',
-      max_chars: 10,
       condition: {
         id: 'short_title',
         value: 'test',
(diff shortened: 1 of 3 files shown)


packages/myst-templates/src/validators.ts
```diff
@@ -312,11 +315,42 @@ export function validateTemplateOptionDefinition(
...
-      ...incrementOptions('max_chars', opts),
-    });
+    if (output.type === 'string') {
+      output.max_chars = validateNumber(value.max_chars, {
+        min: 0,
+        integer: true,
+        ...incrementOptions('max_chars', opts),
+      });
+    } else {
+      validationError('type must be "string" to use "max_chars" option', opts);
```
rowanc1, line 325: @fwkoch this is true correct?

fwkoch: Maybe we add a similar constraint for `choices` only on `type: 'choice'`? (We could also just drop `type: 'choice'` in favour of `type: 'string'` with `choices` defined...)

packages/myst-templates/src/validators.spec.ts
```diff
@@ -370,7 +370,6 @@ describe('validateTemplateOptionDefinition', () => {
           description: 'desc',
           required: true,
           default: 'a',
-          max_chars: 10,
```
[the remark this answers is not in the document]
rowanc1: Added.

packages/myst-templates/src/validators.ts
```diff
@@ -312,11 +315,42 @@ export function validateTemplateOptionDefinition(
...
+    } else {
+      validationError('type must be "number" to use "min" option', opts);
+    }
+  }
+  if (defined(value.max)) {
+    if (output.type === 'number') {
+      output.max = validateNumber(value.max, {
+        ...incrementOptions('max', opts),
+        integer: output.integer,
+      });
```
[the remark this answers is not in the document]
rowanc1: Adding this:

```ts
if (defined(output.min) && defined(output.max) && output.max < output.min) {
    validationWarning('"min" and "max" options are flipped', opts);
    const [min, max] = [output.min, output.max];
    output.min = max;
    output.max = min;
  }
```

packages/myst-templates/src/validators.ts
rowanc1: Added.

PR #778, opened 2023-11-24
🎆 No subfigures option

fwkoch:
Recent release of subfigures caused some funny behaviour with notebook cell figures: the code became the "figure" and output was unexpectedly added to the "caption"

To fix that, this adds `output` as a figure-content type. However, that change turns notebook cell figures into subfigures (a) code and (b) output. Now there is an additional `no-subfigures` option - It can be used explicitly on `figure` directives to suppress subfigure behaviour, and it comes in implicitly on notebook cell figures to keep them a single figure with both code and output.

### DIFF
--- a/.changeset/lazy-sloths-grab.md
+++ b/.changeset/lazy-sloths-grab.md
@@ -1,0 +1,5 @@
+---
+'myst-transforms': patch
+---
+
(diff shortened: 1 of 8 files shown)


rowanc1:
Works great! Thank you!

PR #783, opened 2023-11-25
👩‍🎓 JATS citation / affiliation fixes

fwkoch:
This PR addresses some of the tasks here - https://github.com/executablebooks/mystmd/issues/347:

- IDs of affiliations needs to be normalized
  - This was caused by immutable objects returned by selectors - not the first time we will be bitten by that stealthy bug. Now, selectors return mutable copies, so hopefully this will not get us again 🤞
- parts need to be rendered after the ID normalization (citation labels are not being transformed)
  - now, `parts` are left in the tree for final myst-to-jats mdast transforms, then plucked off immediately before `render`
- multiple unused references show up in the final jats
  - now, only the used references are added to the xml, not all available citations

### DIFF
--- a/packages/myst-to-jats/src/transforms/blocks.ts
+++ b/packages/myst-to-jats/src/transforms/blocks.ts
@@ -1,20 +1,38 @@
 import type { Plugin } from 'unified';
 import type { Block } from 'myst-spec-ext';
-import type { GenericParent } from 'myst-common';
+import { selectBlockParts, type GenericParent } from 'myst-common';
 import { remove } from 'unist-util-remove';
 import { selectAll } from 'unist-util-select';
+import { ABSTRACT_PARTS, ACKNOWLEDGMENT_PARTS, type Options } from '../types.js';
+
+function blocksToKeep(tree: GenericParent, opts: Options) {
+  const keepParts: string[] = [...ACKNOWLEDGMENT_PARTS];
+  if (opts.extractAbstract) {
+    keepParts.push(...(opts.abstractParts?.map(({ part }) => part).flat() ?? ABSTRACT_PARTS));
+  }
+  if (opts.backSections) {
+    keepParts.push(...opts.backSections.map(({ part }) => part).flat());
+  }
+  return selectBlockParts(tree, keepParts);
+}
 
 /**
  * This transform does the following:
- * - Removes hidden or removed blocks from the tree
+ * - Removes hidden or removed blocks from the tree except those that are consumed as front/backmatter parts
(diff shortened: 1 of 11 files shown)


packages/myst-to-jats/src/transforms/blocks.ts
```diff
@@ -1,28 +1,52 @@
...
+  const doNotDelete = blocksToKeep(tree, opts);
   // This could also be an output, etc.
   (selectAll('[visibility=remove],[visibility=hide]', tree) as Block[]).forEach((node) => {
     if (node.visibility === 'remove' || node.visibility === 'hide') {
       (node as any).type = '__delete__';
     }
   });
+  // Blocks are converted to sections - avoid doing this for part blocks
+  doNotDelete.forEach((node) => {
+    node.type = 'block-part' as any;
```
fwkoch, line 35: This is a little sketchy - we want to include referenced `parts` in the transformations, but we do not want to include other `hidden` blocks.

So, first we find all the blocks that will be extracted as `parts` in `blocksToKeep`, then we make sure those are not marked for deletion here...

fwkoch: However, we must give them a type different from `'block'` since in another transform, all blocks are converted to "sections"

packages/myst-to-jats/src/transforms/blocks.ts
```diff
@@ -1,28 +1,52 @@
...
 
-export const blockPlugin: Plugin<[], GenericParent, GenericParent> = () => (tree) => {
-  blockTransform(tree);
+export const blockPlugin: Plugin<[Options], GenericParent, GenericParent> = (opts) => (tree) => {
+  blockTransform(tree, opts);
 };
+
+export function restoreBlockPartTypeTransform(tree: GenericParent) {
+  selectAll('block-part', tree).forEach((node) => {
+    node.type = 'block';
```
fwkoch, line 50: Finally, once transformations are done, we return them to their original `'block'` type, so the `extractParts` function will behave correctly. Pretty gross.

fwkoch: I am leaving this grossness in, however, since I think the real fix is bulid JATS in unist (there's already some utility for this in [jats-xml](https://github.com/curvenote/jats/blob/main/packages/jats-utils/src/utils.ts)) - then, we will not need to do these extra jats-specific transforms on the _input_ MyST AST; we instead can apply simpler, more sensible transforms to the _output_ JATS AST.

PR #790, opened 2023-11-30
🧮  Numbering improvements for custom containers

rowanc1:
This opens up the numbering dictionary to any item, and also allows figure directives (and notebook cells) to set the kind of a container directly. The two of these combined allow for numbering and referencing of custom container types.

### DIFF
--- a/.changeset/silent-mirrors-guess.md
+++ b/.changeset/silent-mirrors-guess.md
@@ -1,0 +1,7 @@
+---
+'myst-frontmatter': patch
+'myst-directives': patch
+'myst-transforms': patch
+---
+
+Allow new numbering options to filter thgouth. Add new kind option to figure directive
--- a/packages/myst-frontmatter/src/numbering/validators.ts
+++ b/packages/myst-frontmatter/src/numbering/validators.ts
@@ -3,39 +3,67 @@ import {
   defined,
   incrementOptions,
   validateBoolean,
-  validateObjectKeys,
+  validateKeys,
+  validateObject,
   validateString,
 } from 'simple-validators';
 import type { Numbering } from './types.js';
 
-export const NUMBERING_KEYS = [
-  'enumerator',
-  'figure',
-  'equation',
-  'table',
-  'code',
-  'heading_1',
-  'heading_2',
-  'heading_3',
-  'heading_4',
-  'heading_5',
-  'heading_6',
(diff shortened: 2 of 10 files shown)


.changeset/silent-mirrors-guess.md
```diff
@@ -0,0 +1,7 @@
+---
+'myst-frontmatter': patch
+'myst-directives': patch
+'myst-transforms': patch
+---
+
+Allow new numbering options to filter thgouth. Add new kind option to figure directive
```
rowanc1, line 7:
```suggestion
Allow new numbering options to filter through. Add new `kind` option to figure directive
```

packages/myst-frontmatter/src/numbering/validators.ts
```diff
@@ -3,39 +3,67 @@ import {
...
-    if (defined(value[key])) {
-      output[key] = validateBoolean(value[key], incrementOptions(key, opts));
-    }
-  });
+  if (defined(value.headings)) {
+    const headings = validateBoolean(value.headings, incrementOptions('headings', opts));
+    HEADING_KEYS.forEach((headingKey) => {
+      if (headings && !defined(value[headingKey])) {
+        // This will be validated next!
+        value[headingKey] = true;
```
fwkoch, line 55: 👍

packages/myst-frontmatter/src/numbering/validators.ts
```diff
@@ -3,39 +3,67 @@ import {
...
-  'heading_2',
-  'heading_3',
-  'heading_4',
-  'heading_5',
-  'heading_6',
-];
+export const NUMBERING_OPTIONS = ['enumerator', 'headings'];
+
+const HEADING_KEYS = ['heading_1', 'heading_2', 'heading_3', 'heading_4', 'heading_5', 'heading_6'];
+export const NUMBERING_KEYS = ['figure', 'equation', 'table', 'code', ...HEADING_KEYS];
```
fwkoch, line 15: And then I guess this just works with the `enumerate` transform, since it already increments count based on `kind`. Nice.

rowanc1: Yeah, I was impressed at how much of this already just worked and all we needed to do was to be less strict in validation!

PR #791, opened 2023-11-30
🗺 Put table legend at end

rowanc1:
This moves the legend to the end of the container.
Rather than directly after the caption.
This is helpful for table notes, etc.

Related to #778, which changes some of the logic there @fwkoch.

### DIFF
--- a/packages/myst-transforms/src/containers.ts
+++ b/packages/myst-transforms/src/containers.ts
@@ -194,10 +194,10 @@ export function containerChildrenTransform(tree: GenericParent, vfile: VFile) {
     }
     const children: GenericNode[] = [...subfigures];
     if (placeholderImage) children.push(placeholderImage);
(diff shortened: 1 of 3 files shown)


packages/myst-transforms/src/containers.ts
```diff
@@ -194,10 +194,10 @@ export function containerChildrenTransform(tree: GenericParent, vfile: VFile) {
     }
     const children: GenericNode[] = [...subfigures];
     if (placeholderImage) children.push(placeholderImage);
-    // Caption/legend are above tables and below all other figures
+    // Caption is above tables and below all other figures
     if (container.kind === 'table') {
-      if (legend) children.unshift(legend);
       if (caption) children.unshift(caption);
+      if (legend) children.push(legend);
```
fwkoch, line 200: 👍 Easy, totally makes sense.

PR #793, opened 2023-11-30
🐛  Fix Inline Expressions Render

sinanbekar:
This PR fixes `Unhandled LaTeX conversion for node of "inlineExpression"` error for PDF and LaTeX output converting from JupyterLab `ipynb` via `myst` command.

Added `inlineExpression` node handler to `myst-to-tex` and `myst-to-docx` packages.

It seems that the current `eval` transformer is working fine, except handling `html` MIME type.

Fixes https://github.com/executablebooks/jupyterlab-myst/issues/194

Manually tested using [examples/inline-interactivity.ipynb](https://github.com/executablebooks/jupyterlab-myst/blob/4684bc358f7d9be3fe62baccc862c35606c9c488/examples/inline-interactivity.ipynb)

### DIFF
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -173,11 +173,11 @@ export async function transformMdast(
 
   const pipe = unified()
     .use(reconstructHtmlPlugin) // We need to group and link the HTML first
-    .use(htmlPlugin, { htmlHandlers }) // Some of the HTML plugins need to operate on the transformed html, e.g. figure caption transforms
     .use(basicTransformationsPlugin, {
       parser: (content: string) => parseMyst(session, content, file),
     })
     .use(inlineExpressionsPlugin) // Happens before math and images!
+    .use(htmlPlugin, { htmlHandlers }) // Some of the HTML plugins need to operate on the transformed html, e.g. figure caption transforms
     .use(inlineMathSimplificationPlugin)
     .use(mathPlugin, { macros: frontmatter.math })
     .use(glossaryPlugin, { state }) // This should be before the enumerate plugins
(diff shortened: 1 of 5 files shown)


packages/myst-cli/src/process/mdast.ts
```diff
@@ -173,11 +173,11 @@ export async function transformMdast(
 
   const pipe = unified()
     .use(reconstructHtmlPlugin) // We need to group and link the HTML first
-    .use(htmlPlugin, { htmlHandlers }) // Some of the HTML plugins need to operate on the transformed html, e.g. figure caption transforms
     .use(basicTransformationsPlugin, {
       parser: (content: string) => parseMyst(session, content, file),
     })
     .use(inlineExpressionsPlugin) // Happens before math and images!
+    .use(htmlPlugin, { htmlHandlers }) // Some of the HTML plugins need to operate on the transformed html, e.g. figure caption transforms
```
rowanc1, line 180: Going to test the ordering of this as I have run into a couple issues with having HTML after the "basic transformations".  We could always (I think) move the inlineExpressions higher up.

rowanc1 (commented):
This is awesome, thank you so much @sinanbekar for the PR! I will take a test of this and aim to get it in and released shortly.

packages/myst-cli/src/process/mdast.ts
fwkoch: I reordered this so `inlineExpressionsPlugin` happens first, then `htmlPlugin`, then `basicTransformations` - I think that's less risky than moving `htmlPlugin` to after `basic`...

I also rendered the inline expression `value` when `children` were not available to render. I think this is preferable? It means the evaluated code shows up in the output document as opposed to nothing.

PR #796, opened 2023-12-04
👩‍🎓 Add typst export to CLI and support templating through jtex

fwkoch:
We should maybe consider renaming `jtex` -> `myst-render` or something? It is basically just an extension of `myst-templates` now, with an extra dependency or two.

This also has some improvements to bibtex rendering, which apply to tex and typst:
- when mystmd writes a new bibtex file, it attempts to only write the citations used in the article
- if no citations are present, this is passed to the template renderer, so we can avoid empty references sections

### DIFF
--- a/packages/myst-to-typst/tests/cite.yml
+++ b/packages/myst-to-typst/tests/cite.yml
@@ -23,4 +23,4 @@ cases:
                 - type: cite
                   label: jon85
     typst: |-
-      #cite("cockett2015", "jon85")
+      #cite(<cockett2015>, <jon85>)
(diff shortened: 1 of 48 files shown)


packages/myst-to-typst/tests/cite.yml
```diff
@@ -23,4 +23,4 @@ cases:
                 - type: cite
                   label: jon85
     typst: |-
-      #cite("cockett2015", "jon85")
+      #cite(<cockett2015>, <jon85>)
```
rowanc1, line 26: This should just be `@cockett2015 @jon85` without a group around it, there are no citation groups in typst as far as I can tell, and the cite function only take one arg.

rowanc1 (approved):
I think this is good to go in. @fwkoch can you update the citations as we touched that in this PR?

Issue #802, opened 2023-12-06
MECA exports should infer article if none is provided


fwkoch:
My expectation is that this simple export would work.

```yaml
export:
  - format: meca
```

It is currently raising an error: `TypeError: Cannot read properties of undefined (reading '0')`.

PR #803, opened 2023-12-07
🐛 Fix error when article is not defined on meca export

fwkoch:
Addresses #802 - this bug was introduced when multiple `articles` were allowed on an export.


packages/myst-cli/src/build/meca/index.ts
```diff
@@ -157,7 +157,7 @@ export async function runMecaExport(
   const toc = tic();
   const { output, articles } = exportOptions;
   // At this point, export options are resolved to contain zero or one articles
```
fwkoch, line 159: This comment says there may not be articles, and now the next line of code handles that case.

PR #817, opened 2023-12-19
🏗 Add watch to build command

fwkoch:
- You may now `myst build --watch` and built files will be live-reloaded
- Changing dependencies (i.e. included/embedded content) will cause dependent pages to reload (both for exports and site build)
- export format may be inferred from filename

### DIFF
--- a/packages/myst-frontmatter/src/exports/exports.yml
+++ b/packages/myst-frontmatter/src/exports/exports.yml
@@ -279,13 +279,13 @@ cases:
       exports:
         - format: pdf
           output: out.pdf
-  - title: export with only template validates
-    raw:
-      exports:
-        - template: template
-    normalized:
-      exports:
-        - template: template
+  # - title: export with only template validates
+  #   raw:
+  #     exports:
+  #       - template: template
+  #   normalized:
+  #     exports:
+  #       - template: template
   - title: unknown output errors
     raw:
       exports:
--- a/packages/myst-frontmatter/src/exports/exports.yml
+++ b/packages/myst-frontmatter/src/exports/exports.yml
@@ -256,3 +256,39 @@ cases:
           articles:
             - a.md
     errors: 3
+  - title: pdf filename coerces to format and output
+    raw:
+      exports:
+        - out.pdf
+    normalized:
+      exports:
+        - format: pdf
+          output: out.pdf
+  - title: pdf extension coerces to format
+    raw:
+      exports:
+        - .pdf
+    normalized:
+      exports:
+        - format: pdf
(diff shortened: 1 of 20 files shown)


packages/myst-frontmatter/src/exports/exports.yml
```diff
@@ -256,3 +256,39 @@ cases:
...
+      exports:
+        - format: pdf
+          output: out.pdf
+  # - title: export with only template validates
+  #   raw:
+  #     exports:
+  #       - template: template
+  #   normalized:
+  #     exports:
+  #       - template: template
```
stevejpurves, line 288: skipped test?

packages/myst-frontmatter/src/exports/validators.ts
```diff
@@ -12,14 +12,28 @@ import type { Export } from './types.js';
 import { ExportFormats } from './types.js';
 
 const EXPORT_KEY_OBJECT = {
-  required: ['format'],
-  optional: ['template', 'output', 'id', 'name', 'renderer', 'articles', 'sub_articles'],
+  required: [],
+  optional: ['format', 'template', 'output', 'id', 'name', 'renderer', 'articles', 'sub_articles'],
```
rowanc1, line 16: Is this a change for now? I thought this might be something we discussed, but aren't moving on quite yet.

rowanc1 (commented):
This works really well, just a question about the required format change and if we are doing that now?

packages/myst-frontmatter/src/exports/exports.yml
fwkoch: Yes - this is looking ahead to when we infer format from template (and therefore only template would be ok). Not implemented yet!

packages/myst-frontmatter/src/exports/validators.ts
fwkoch: I haven't moved on inferring `format` from `template` yet, but you can now infer `format` from `output` file extension.

So, `format` is still required on the export type: https://github.com/executablebooks/mystmd/pull/817/files#diff-15a4efc7ae10ddbda5633c4ef4ae20e03dcda80cc8c5662ae0bf6cfa3b2899a9R13 (no type errors means validation always adds `format`) - but it's not required on key validation since it may be filled in based on `output`. Does that make sense...?

fwkoch: When we do start inferring `format` from `template`, we will need to make `format` actually optional, even on the export type, since I doubt template loading should be part of validation...

PR #820, opened 2023-12-21
🎆 Support images in inline expressions and fix transform order

fwkoch:
Inline expression render fix here still didn't quite get transform order correct #793 - this fixes that up by running `blockMetadataTransform` before `inlineExpressionTransform`.

I also added `images` as a supported inline expression mimetype 🚀


fwkoch:
This needs to go in before release #818 - since inline expressions are a little broken on main.

PR #821, opened 2023-12-21
🗃 Remove file extensions from latex include

fwkoch:
Latex `include` commands should not include the file extensions: https://www.overleaf.com/learn/latex/Management_in_a_large_project Apparently some latex engines don't mind but some do! This PR simply removes the extensions.


stevejpurves (approved):
looks good!

Issue #816, opened 2023-12-19
Application Error when attempting thebe execution


ryanlovett:
We are not able to execute code-cell blocks with a local jupyter kernel.

## Description

We have a project with the following in `myst.yml`:
```yaml
project:
  thebe:
    server:
      url: http://localhost:8888
      token: test-secret
```

We invoke jupyter lab as
```bash
$ jupyter kernelspec list
Available kernels:
  ir         /path/to/miniforge3/share/jupyter/kernels/ir
  python3    /path/to/miniforge3/share/jupyter/kernels/python3
$ jupyter lab --NotebookApp.token=test-secret --NotebookApp.allow_origin='https://localhost:3000'
```

A file contains:
```yaml
:::{code-cell} python3
import numpy as np
mean = np.mean(np.random.normal(size=100))
print(f"{mean=}")
:::
```

I can run `myst start` and visit the server and load the example file. When I click on the power button:

![Screenshot 2023-12-18 at 5 27 30 PM](https://github.com/executablebooks/mystmd/assets/10248125/2137d3db-74b5-4554-a108-49fdf9548df0)

 to start the kernel, myst directs me to a page with:
```python
TypeError: Cannot read properties of undefined (reading 'reduce')
    at http://localhost:3000/myst_assets_folder/_shared/chunk-CDALZYYP.js:319:27535
    at Array.map (<anonymous>)
    at k5t (http://localhost:3000/myst_assets_folder/_shared/chunk-CDALZYYP.js:319:27053)
    at http://localhost:3000/myst_assets_folder/_shared/chunk-CDALZYYP.js:320:791
    at Ei (http://localhost:3000/myst_assets_folder/_shared/chunk-OXM6VUU4.js:8:24198)
    at yr (http://localhost:3000/myst_assets_folder/_shared/chunk-OXM6VUU4.js:8:42223)
    at http://localhost:3000/myst_assets_folder/_shared/chunk-OXM6VUU4.js:8:40571
    at Ca (http://localhost:3000/myst_assets_folder/_shared/chunk-OXM6VUU4.js:1:62053)
    at MessagePort.Ra (http://localhost:3000/myst_assets_folder/_shared/chunk-OXM6VUU4.js:1:62444)
```

I ran `myst clean --all` and replaced theme.server with theme.lite=true, and then ran`myst init`. I received a similar Application Error.

```bash
$ myst -v
v1.1.36
```

[...]

Issue #748, opened 2023-11-13
Markdown based notebooks fail to execute with `thebe`


stevejpurves:
In the image below, the left hand json is the mdast for a markdown based notebook (frontmatter.jupytext !== undefined) and the right hand side is the equivalent notebook in `ipynb`format.

![image](https://github.com/executablebooks/mystmd/assets/1473646/7b7900a0-6af3-4945-88cd-442a99dea82c)

The `ipynb` notebook renders and runs as expected with thebe in the `myst-book-theme`, however the `md` based notebook fails to execute fully, execution starts, completes (kernel returns results) but the busy spinner never stops and no output is shown.

This is because the `md` based notebook has no outputs specified in the cells `children`. 

The [notebook format](https://nbformat.readthedocs.io/en/latest/format_description.html#code-cells) specifies an `outputs` array within every code cell, even if that is empty. The parsing of the `ipynb` to include an output node in the tree with empty array in data, which is in line with that. We should make a fix so that the markdown based notebook also creates an empty output node by default.

This would fix this issue: https://github.com/executablebooks/thebe/issues/699

stevejpurves:
[this repo](https://github.com/cmarmo/test-thebe/) will reproduce the issue.
`jupytext` notebooks are flagged as Notebooks [here](https://github.com/executablebooks/mystmd/blob/35dbf47e19ab962c59b59cb3e3280458fb2ea05f/packages/myst-cli/src/process/mdast.ts#L232-L233) but `code-cell`s have probably already been parsed without adding output nodes by this point?
cc @fwkoch

stevejpurves:
If we decide that the mdast tree should not align for these two cases then we could update downstream code handle the no empty output case, but is seems like we should be trying to align the mdast output for these two cases?

stevejpurves:
this is now fixed

Issue #669, opened 2023-10-13
Incompatibility with jupytext for code-cells: language required


nthiery:
## Description

In MyST notebooks, Jupytext 1.15.2 writes code cells as:

    ```{code-cell}
    ```

It seems that the current myst expects the language to be given:

    ```{code-cell} python3
    ```

If not, it complains `required argument not provided for directive: code-cell`, and
the content of the cell is rendered as plain markdown.

Compare e.g.
- https://nicolas.thiery.name/Enseignement/Info111/Devoirs/Semaine5/tableaux-prise-en-main.html (built with JupyterBook)
- https://nicolas.thiery.name/Enseignement/Info111/myst/tableaux-prise-en-main (built with myst)

## Proposed solution

Like JupyterBook, don't require a language to be explicitly specified for code cells, and use the kernel spec from the notebook as default value.

fwkoch:
Ah, this make sense. MyST does add kernelspec languages to `code` blocks with this transform: https://github.com/executablebooks/mystmd/blob/27f48dba574f088e673cf2e3a6efb197209c7f36/packages/myst-cli/src/process/mdast.ts#L213

But, the `code-cell` directive is _strict_ in requiring an argument: https://github.com/executablebooks/mystmd/blob/27f48dba574f088e673cf2e3a6efb197209c7f36/packages/myst-directives/src/code.ts#L154 and fails before the MyST transform is able to take effect.

It's probably ok to just relax the latter strict requirement. We will still get a warning if no language is ever defined...

rowanc1:
@stevejpurves if you are in the `code-cell` today for improvements around jupytext (#748), can you also make this change?

nthiery:
Thanks everyone !

Occasion to wish the MyST team a Merry Christmas :-)

rowanc1:
Thanks @nthiery, happy holidays to you as well!!

PR #822, opened 2023-12-22
fixing execution in text based notebooks

stevejpurves:
should fix:
* #748 
* #816 
* https://github.com/executablebooks/thebe/issues/699
* #669 

See https://github.com/stevejpurves/text-based-myst-nbs for the test case I've been using - especailly note the thematic break in the text based notebook

### DIFF
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -65,6 +65,7 @@ import {
   transformImagesWithoutExt,
   transformImagesToDisk,
   transformFilterOutputStreams,
+  transformLiftCodeBlocksInJupytext,
 } from '../transforms/index.js';
 import type { ImageExtensions } from '../utils/resolveExtension.js';
 import { logMessagesFromVFile } from '../utils/logMessagesFromVFile.js';
@@ -221,6 +222,7 @@ export async function transformMdast(
   }
   // Combine file-specific citation renderers with project renderers from bib files
   const fileCitationRenderer = combineCitationRenderers(cache, ...rendererFiles);
+
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   await transformOutputsToCache(session, mdast, kind, { minifyMaxCharacters });
   transformCitations(mdast, fileCitationRenderer, references);
@@ -230,6 +232,8 @@ export async function transformMdast(
     .run(mdast, vfile);
   transformImagesToEmbed(mdast);
   transformImagesWithoutExt(session, mdast, file, { imageExtensions });
+  const isJupytext = frontmatter.kernelspec || frontmatter.jupytext;
+  if (isJupytext) transformLiftCodeBlocksInJupytext(mdast);
   const sha256 = selectors.selectFileInfo(store.getState(), file).sha256 as string;
   const useSlug = pageSlug !== index;
   const url = projectSlug
@@ -238,7 +242,7 @@ export async function transformMdast(
   const dataUrl = projectSlug ? `/${projectSlug}/${pageSlug}.json` : `/${pageSlug}.json`;
   updateFileInfoFromFrontmatter(session, file, frontmatter, url, dataUrl);
   const data: RendererData = {
-    kind: frontmatter.kernelspec || frontmatter.jupytext ? SourceFileKind.Notebook : kind,
+    kind: isJupytext ? SourceFileKind.Notebook : kind,
     file,
     location,
     sha256,
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -19,7 +19,7 @@ import { castSession } from '../session/cache.js';
 import type { ISession } from '../session/types.js';
 import { BASE64_HEADER_SPLIT } from '../transforms/images.js';
 import { parseMyst } from './myst.js';
-import { Code } from 'myst-spec-ext';
+import type { Code } from 'myst-spec-ext';
 
 function blockParent(cell: ICell, children: GenericNode[]) {
   const type = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -19,11 +19,7 @@ import { castSession } from '../session/cache.js';
 import type { ISession } from '../session/types.js';
 import { BASE64_HEADER_SPLIT } from '../transforms/images.js';
 import { parseMyst } from './myst.js';
-
-function createOutputDirective(): { myst: string; id: string } {
-  const id = nanoid();
-  return { myst: `\`\`\`{output}\n:id: ${id}\n\`\`\``, id };
-}
+import { Code } from 'myst-spec-ext';
 
 function blockParent(cell: ICell, children: GenericNode[]) {
   const type = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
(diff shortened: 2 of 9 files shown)


packages/myst-cli/src/process/notebook.ts
```diff
@@ -109,41 +103,41 @@ export async function processNotebook(
         return acc.concat(blockParent(cell, cellMdast.children));
       }
       if (cell.cell_type === CELL_TYPES.raw) {
-        const cellContent = `\`\`\`\n${ensureString(cell.source)}\n\`\`\``;
-        const cellMdast = parseMyst(session, cellContent, file);
-        return acc.concat(blockParent(cell, cellMdast.children));
+        const raw: Code = {
```
stevejpurves, line 106: changes up the process notebook code to just insert nodes, rather then writing the myst and parsing it.

packages/myst-cli/src/process/notebook.ts
```diff
@@ -109,41 +103,41 @@ export async function processNotebook(
...
         if (cell.outputs && (cell.outputs as IOutput[]).length > 0) {
           const minified: MinifiedOutput[] = await minifyCellOutput(
             cell.outputs as IOutput[],
             cache.$outputs,
             { computeHash, maxCharacters: opts?.minifyMaxCharacters },
           );
-          outputMap[id] = minified;
-        } else {
-          outputMap[id] = [];
+          output.data = minified;
```
stevejpurves, line 133: seems like this could be much simplier, like this... not clear on the need for the outputMap, unless cells were sharing outputs somehow? but that's not possible in a notebook?? 🤔

packages/myst-cli/src/transforms/code.ts
```diff
@@ -194,3 +194,45 @@ export function propagateBlockDataToCode(session: ISession, vfile: VFile, mdast:
...
+
+/**
+ * Because jupytext style notebooks can introduce code-cells in blocks
+ * after explicit thematic breaks, we need to lift them up to the parent
+ * to ensure that all executable code blocks are at the top level of
+ * the document.
+ *
+ * @param mdast
+ */
+export function transformLiftCodeBlocksInJupytext(mdast: GenericParent) {
```
stevejpurves, line 206: so this is a post process to prevent nesting of the code-cell in an extra block, when a thematic break is used in an `md` file. Doing this means the `code-cells` are all at root level, and more aligned with a notebook that originated from `ipynb`. 

This seems sensible from an mdast structure point of view, and right now is a pragmatic step as dependent code like the myst-theme expects executable cells to be top level children only, adding ui controls and hooking up compute to blocks( with code, output) at that level only. It's not setup to work with nested executable code-cells.

packages/myst-transforms/src/code.ts
```diff
@@ -13,6 +13,7 @@ type Options = {
 export function codeTransform(mdast: GenericParent, file: VFile, opts?: Options) {
   const code = selectAll('code', mdast) as Code[];
   code.forEach((node) => {
+    if (node.lang === '') return; // raw code cell passes through with no warning
```
stevejpurves, line 16: doing this means that notebook cells of `type=raw` do not cause missing language warnings on the console. This seems ok? unless there is another mechanism? e.g. set lang to `raw` when processing the notebook and test for `raw` here? i'm not sure what the namespace of possible `lang` strings are....?

packages/myst-cli/src/process/notebook.ts
```diff
@@ -109,41 +103,41 @@ export async function processNotebook(
         return acc.concat(blockParent(cell, cellMdast.children));
       }
       if (cell.cell_type === CELL_TYPES.raw) {
-        const cellContent = `\`\`\`\n${ensureString(cell.source)}\n\`\`\``;
-        const cellMdast = parseMyst(session, cellContent, file);
-        return acc.concat(blockParent(cell, cellMdast.children));
+        const raw: Code = {
+          type: 'code',
+          lang: '',
```
stevejpurves, line 108: raw cells have `lang=''` is there a better value?

packages/myst-directives/src/code.ts
```diff
@@ -154,7 +155,6 @@ export const codeCellDirective: DirectiveSpec = {
   name: 'code-cell',
   arg: {
     type: String,
-    required: true,
```
stevejpurves, line 157: this addresses #669

packages/myst-cli/src/transforms/code.ts
```diff
@@ -194,3 +194,45 @@ export function propagateBlockDataToCode(session: ISession, vfile: VFile, mdast:
...
+  const flattened = (mdast.children as GenericParent[]).reduce((acc, node) => {
+    if (node.type !== 'block' || !node.children) return [...acc, node];
+
+    const buriedCodeCells = selectAll('block:has(block)', node);
+    if (buriedCodeCells.length === 0) return [...acc, node];
```
fwkoch, line 211: This would basically happen anyway even without the early return, right? It would just go in a big circle reconstructing the block, ending up equivalent to how it started.

fwkoch: If we did that though, we would lose block metadata!

I guess as-is, we do lose any metadata on the top-level blocks (though the code/output blocks should keep their metadata since those blocks are pushed directly).

packages/myst-cli/src/transforms/code.ts
```diff
@@ -194,3 +194,45 @@ export function propagateBlockDataToCode(session: ISession, vfile: VFile, mdast:
...
+
+    const newBlocks: GenericParent[] = [{ type: 'block', children: [] }];
```
fwkoch, line 213: Maybe this should be, like
```suggestion
    const newBlocks: GenericParent[] = [{ ...node, children: [] }];
```

to maintain metadata. (it would then be lost on the subsequent broken up blocks - I guess we could also keep it on their but not sure that's best, e.g. duplicating `id`s etc)

packages/myst-transforms/src/code.ts
fwkoch: Yeah, I think this is ok... Needing to depend on an empty string is slightly magic - I think `text` would also be a sensible value for `lang`... but I don't know if that has different implications than a raw cell.

packages/myst-directives/src/code.ts
```diff
@@ -200,6 +200,22 @@ export const codeCellDirective: DirectiveSpec = {
...
+
+    const output = {
+      type: 'output',
+      id: nanoid(),
+      data: [],
+    };
+
+    const block = {
+      type: 'block',
+      meta: undefined, // do we need to attach metadata?
```
fwkoch, line 212: This will come later during transformation, when labels, etc, are lifted out of code up to parent blocks.

packages/myst-cli/src/process/mdast.ts
```diff
@@ -230,6 +232,8 @@ export async function transformMdast(
     .run(mdast, vfile);
   transformImagesToEmbed(mdast);
   transformImagesWithoutExt(session, mdast, file, { imageExtensions });
+  const isJupytext = frontmatter.kernelspec || frontmatter.jupytext;
+  if (isJupytext) transformLiftCodeBlocksInJupytext(mdast);
```
fwkoch, line 236: I think this is ok running here - The other block/code transformations happen pretty early, but I don't think there is anything in here that is order-dependent.

fwkoch (approved):

packages/myst-transforms/src/code.ts
fwkoch: Just landing this as-is - Hopefully the comment is enough to remind us why we have this check.

Issue #739, opened 2023-11-09
Improve  docs for working locally with a Jupyter Kernel


rowanc1:
It takes a while to understand how to work locally with a Jupyter Kernel:

-  🐛 `https` to `http` in  `NotebookApp.allow_origin='https://localhost:3000'`

We have tried to document this here:
https://mystmd.org/guide/integrating-jupyter#start-a-local-jupyter-server

However, it isn't actually isn't referenced in the main section:
https://mystmd.org/guide/integrating-jupyter#directly-connecting-to-a-jupyter-server

The bottom content could probably be moved up/combined.

cc @stevejpurves

PR #825, opened 2024-01-05
🐛 fixing jupyter docs bugs

stevejpurves:
fixing #739 and avoiding #816

### DIFF
--- a/.changeset/ten-feet-invent.md
+++ b/.changeset/ten-feet-invent.md
@@ -1,0 +1,10 @@
+---
+'mystmd': patch
+---
+
+Improvements to documentation on integrating jupyter:
+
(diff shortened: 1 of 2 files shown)


stevejpurves:
I added a changeset to the root package to reflect these doc changes, looks like that will cause a bump in `myst-cli` though due to the dependency being in the root `package.json`.

* is that dependency right or should it be removed? 
* do we typically changeset docs changes via the root package?

stevejpurves:
based on the discussion in #816 more improvement is needed to https://mystmd.org/guide/integrating-jupyter and https://mystmd.org/guide/interactive-notebooks.

perhaps aligning to some of the phrasing that `jupytext` uses will help and better introducing the text-based format that myst supports

PR #830, opened 2024-01-08
🧭 Order abbreviations by length

agoose77:
This might not be the right place, or the best way to do this, but this PR orders abbreviations defined in the frontmatter by length (then by locale). As such, we no longer need to warn about the ordering.

### DIFF
--- a/.changeset/thin-houses-provide.md
+++ b/.changeset/thin-houses-provide.md
@@ -1,0 +1,5 @@
+---
+'myst-transforms': minor
+---
+
+Sort (frontmatter) abbreviations definitions by length
--- a/docs/glossaries-and-terms.md
+++ b/docs/glossaries-and-terms.md
@@ -70,5 +70,5 @@ The abbreviations are case-sensitive and will replace all instances[^1] in your 
 
 :::{tip} Order of Abbreviations
 :class: dropdown
-The order of abbreviations in your frontmatter should specify the longer abbreviations first (e.g. `RHR`) and then the shorter abbreviations (e.g. `HR`). If you do this in the opposite order, you will have an abbreviation for R`HR`.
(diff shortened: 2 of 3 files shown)


packages/myst-cli/src/process/mdast.ts
```diff
@@ -172,6 +172,20 @@ export async function transformMdast(
   // This needs to come before basic transformations since it may add labels to blocks
   liftCodeMetadataToBlock(session, vfile, mdast);
 
+  // Order abbreviations by length (longest first)
+  let abbreviations: Record<string, string> | undefined;
```
rowanc1, line 176: Can you move this code to here:

https://github.com/executablebooks/mystjs/blob/main/packages/myst-transforms/src/abbreviations.ts#L28

Maybe refactoring that to another function? e.g. `sortAndFilterAbbreviations` or something?

rowanc1 (changes requested):
This looks awesome, thanks @agoose77. 🚀

Can we move the sort code down a level so it is closer to where we do some other filtering? This should then be able to be shared a bit easily as the CLI code isn't going to be easy to share, e.g. with jupyterlab-myst and others.

Could you also run `npm run changeset` and write a short message that bumps the `myst-transform` and `myst-cli` packages (the myst-cli one will likely be manual once the changes are pushed down). This helps keep track of which packages need bumping.

packages/myst-cli/src/process/mdast.ts
rowanc1: There is going to be a bit of reprocessing/sorting every time this is run, but I think that is probably fine.

agoose77:
@rowanc1 thanks! I realised that my understanding of what the transform did was not quite right; initially I thought it wasn't sensible to move the logic to myst-transforms.

As such, I've moved this to a simple in-line sort, making this PR very small!

rowanc1:
Thanks again @agoose77, excited to see these sorts of quality-of-life changes come to life! :)

PR #831, opened 2024-01-08
♹ Incremental `typst` improvements

fwkoch:
- code blocks no longer add `\` escapes. They also handle backticks correctly
- figures no longer include images from tables twice. Other formatting of the typst file is improved
- additional simplifications around typst serialization - explicitly add newlines as required, no longer use `closeBlock`

### DIFF
--- a/packages/myst-to-typst/src/index.ts
+++ b/packages/myst-to-typst/src/index.ts
@@ -238,9 +238,13 @@ const handlers: Record<string, Handler> = {
     const { width: nodeWidth, url: nodeSrc, align } = node;
     const src = nodeSrc;
     const width = getLatexImageWidth(nodeWidth);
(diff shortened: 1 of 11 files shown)


packages/myst-to-typst/src/index.ts
```diff
@@ -381,12 +394,6 @@ class TypstSerializer implements ITypstSerializer {
     this.trimEnd();
     this.file.result += ']';
   }
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  closeBlock(node: any) {
-    this.ensureNewLine(true);
-    this.file.result += '\n';
-  }
```
rowanc1, line 389: I think that we are loosing the block information between paragraphs with this change @fwkoch.

PR #841, opened 2024-01-11
📏 Incremental typst improvements

fwkoch:
This also adds html output support for static exports!


packages/myst-to-typst/src/container.ts
```diff
@@ -41,6 +41,7 @@ export const containerHandler: Handler = (node, state) => {
     });
     return;
   }
+  state.useMacro('#show figure: set block(breakable: true)');
```
rowanc1, line 44: Should this be in the theme? Can you explain what this is doing?

packages/myst-to-typst/src/container.ts
```diff
@@ -52,8 +53,6 @@ export const containerHandler: Handler = (node, state) => {
     (child: GenericNode) => child.type === 'image' || child.type === 'table',
   );
   if (!imagesAndTables || imagesAndTables.length !== 1) {
-    console.log(node.children.map((child: GenericNode) => child.type));
```
rowanc1, line 55: Was planning to hunt this down!

packages/myst-to-typst/src/container.ts
fwkoch: This allows figures to span multiple pages - https://github.com/typst/typst/pull/1121 - without it tables stop at the bottom of the page and all the remaining rows are mushed together.

I think moving it to theme is probably a good idea (I don't know why we would ever _not_ want this), but I'm still playing with table formatting some (`breakable` does not solve width issues...) so this may not be the final solution...

PR #842, opened 2024-01-11
🐛 Duplicate bibtex entries when uppercase

rowanc1:
There was a bug when `@ARTICLE` was used, it seems to work in both tex and typst, but we would render all citations when encountering it.


fwkoch (approved):
Yeah, the fallback to rendering all entries is pretty extreme. This makes sense.

PR #853, opened 2024-01-12
🔧 More small typst improvements

fwkoch:
- Citations: now narrative vs parenthetical is respected, and suffixes are supported
- Figures are fixed to allow images/tables plus other content - this allows jupyter notebook figures display both code and output image
- Nested definition lists indent correctly

### DIFF
--- a/packages/myst-to-typst/src/container.ts
+++ b/packages/myst-to-typst/src/container.ts
@@ -35,8 +35,8 @@ export function determineCaptionKind(node: GenericNode): CaptionKind | null {
 
 function renderFigureChild(node: GenericNode, state: ITypstSerializer) {
   const useBrackets = node.type !== 'image' && node.type !== 'table';
-  if (useBrackets) state.write('[');
-  state.write('\n');
+  if (useBrackets) state.write('[\n');
+  else state.write('\n  ');
   state.renderChildren({ children: [node] });
   if (useBrackets) state.write('\n]');
 }
--- a/packages/myst-to-typst/src/container.ts
+++ b/packages/myst-to-typst/src/container.ts
@@ -1,6 +1,6 @@
-import { fileError, fileWarn, type GenericNode } from 'myst-common';
+import { fileError, type GenericNode } from 'myst-common';
 import type { Image, Table, Code, Math } from 'myst-spec';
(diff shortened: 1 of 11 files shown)


packages/myst-to-typst/src/container.ts
```diff
@@ -41,40 +49,41 @@ export const containerHandler: Handler = (node, state) => {
     });
     return;
   }
-  state.useMacro('#show figure: set block(breakable: true)');
+
+  state.ensureNewLine();
+  state.write('#show figure: set block(breakable: true)');
```
fwkoch, line 54: I wrote this directly into the document - the `macro` was not importing across `.typ` files. Still maybe should just be on the template.

packages/myst-to-typst/src/table.ts
```diff
@@ -34,7 +34,9 @@ export const tableHandler: Handler = (node, state) => {
     return;
   }
   state.useMacro('#import "@preview/tablex:0.0.7": tablex, cellx');
-  state.write(`${command}(columns: ${columns}, header-rows: ${countHeaderRows(node)},\n`);
+  state.write(
+    `${command}(columns: ${columns}, header-rows: ${countHeaderRows(node)}, repeat-header: true,\n`,
```
fwkoch, line 38: `repeat-header` is documented in tablex, but it didn't seem to work super well. There is a note in the docs that behaviour may be somewhat erratic...

packages/myst-to-typst/src/container.ts
fwkoch: This also is a little overzealous - looks like a figure at the end of a page that could easily fit on the next page will remain in place and be broken...

packages/myst-to-typst/tests/cite.yml
```diff
@@ -9,7 +9,7 @@ cases:
             - type: cite
               label: cockett2015
     typst: |-
-      @cockett2015
+      #cite(<cockett2015>)
```
rowanc1, line 12: Can we only do this if necessary? i.e. leave it as a normal label for the most part and only change if we are overriding the default or something?

rowanc1: skipping for now!

Issue #550, opened 2023-08-11
Incorporate execution


akhmerov:
## Proposal

Right now including computation outputs requires creating `.ipynb` files that do contain these outputs. Storing `.ipynb` in git is a nuisance, especially with outputs. By comparison, the sphinx stack includes jupyter-cache that will interpret and execute all code cells in a document. Therefore, authoring experience would be much better if `myst` would integrate with jupyter-cache or provide utilities for execution through any other means.

rowanc1:
Thanks for all the issues @akhmerov, keep them coming. :)
This is absolutely on our roadmap for the project as we improve compatibility with JupyterBook.

agoose77:
Closed by #873!

PR #856, opened 2024-01-16
feat: add build-time execution

### LINKED ISSUE #839: Add support for local kernel execution
## Proposal
Right now, `myst` has no built-time execution support (e.g. to generate figures). We are in a strong position to implement cache-aware builds, and I don't think it will be too much work given the fact that we're just re-using the Jupyter APIs.

c.f. https://github.com/executablebooks/mystmd/issues/330

## Additional notes
- We may want a `myst execute` command to avoid building any render targets. Maybe just `myst build`?
- Add support for annotating documents as "un-cacheable" e.g. things that depend upon random inputs.
- Multiple cache backends? In future, could perhaps centralise a cache among users for e.g. a JupyterHub deployment that controls the execution environment

agoose77:
This PR is a jumping off point for execution work, and should make it possible to scope discussions to just interested parties.

- [x] Separate existing inline expression result population from rendering
- [x] Implement local on-disk caching (with a nice abstraction)
- [ ] Implement a CLI for execution w/o rendering
- [ ] Implement CLI options to control caching and whether to start existing an Jupyter server

Closes #839, #550

### DIFF
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -225,7 +225,10 @@ export async function transformMdast(
   // Combine file-specific citation renderers with project renderers from bib files
   const fileCitationRenderer = combineCitationRenderers(cache, ...rendererFiles);
 
-  const serverSettings = ServerConnection.makeSettings();
+  const serverSettings = ServerConnection.makeSettings({
+    baseUrl: process.env.JUPYTER_BASE_URL,
+    token: process.env.JUPYTER_TOKEN,
+  });
   const kernelManager = new KernelManager({ serverSettings });
   const sessionManager = new SessionManager({ kernelManager, serverSettings });
   await transformKernelExecution(sessionManager, mdast, frontmatter, file, true);
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -74,6 +74,8 @@ import { bibFilesInDir, selectFile } from './file.js';
 import { loadIntersphinx } from './intersphinx.js';
 import { frontmatterPartsTransform } from '../transforms/parts.js';
 import { parseMyst } from './myst.js';
+import { transformKernelExecution } from '../transforms/execute.js';
+import { ServerConnection, KernelManager, SessionManager } from '@jupyterlab/services';
 
 const LINKS_SELECTOR = 'link,card,linkBlock';
 
@@ -222,6 +224,11 @@ export async function transformMdast(
   }
   // Combine file-specific citation renderers with project renderers from bib files
   const fileCitationRenderer = combineCitationRenderers(cache, ...rendererFiles);
+
+  const serverSettings = ServerConnection.makeSettings();
+  const kernelManager = new KernelManager({ serverSettings });
+  const sessionManager = new SessionManager({ kernelManager, serverSettings });
+  await transformKernelExecution(sessionManager, mdast, frontmatter, file, true);
 
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   await transformOutputsToCache(session, mdast, kind, { minifyMaxCharacters });
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -65,7 +65,8 @@ import {
   transformImagesWithoutExt,
   transformImagesToDisk,
   transformFilterOutputStreams,
-  transformLiftCodeBlocksInJupytext, renderInlineExpressionsTransform
+  transformLiftCodeBlocksInJupytext,
+  renderInlineExpressionsTransform,
 } from '../transforms/index.js';
 import type { ImageExtensions } from '../utils/resolveExtension.js';
 import { logMessagesFromVFile } from '../utils/logMessagesFromVFile.js';
@@ -74,7 +75,12 @@ import { bibFilesInDir, selectFile } from './file.js';
 import { loadIntersphinx } from './intersphinx.js';
 import { frontmatterPartsTransform } from '../transforms/parts.js';
 import { parseMyst } from './myst.js';
-import { transformKernelExecution } from '../transforms/execute.js';
+import {
+  findExistingJupyterServer,
+  JupyterServerSettings,
+  launchJupyterServer,
+  transformKernelExecution,
+} from '../transforms/execute.js';
 import { ServerConnection, KernelManager, SessionManager } from '@jupyterlab/services';
 
 const LINKS_SELECTOR = 'link,card,linkBlock';
@@ -225,13 +231,7 @@ export async function transformMdast(
   // Combine file-specific citation renderers with project renderers from bib files
   const fileCitationRenderer = combineCitationRenderers(cache, ...rendererFiles);
 
-  const serverSettings = ServerConnection.makeSettings({
-    baseUrl: process.env.JUPYTER_BASE_URL,
-    token: process.env.JUPYTER_TOKEN,
-  });
-  const kernelManager = new KernelManager({ serverSettings });
-  const sessionManager = new SessionManager({ kernelManager, serverSettings });
-  await transformKernelExecution(session, sessionManager, mdast, frontmatter, false, vfile, false);
+  await transformKernelExecution(session, mdast, frontmatter, false, vfile, false);
 
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   await transformOutputsToCache(session, mdast, kind, { minifyMaxCharacters });
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -231,7 +231,7 @@ export async function transformMdast(
   });
   const kernelManager = new KernelManager({ serverSettings });
   const sessionManager = new SessionManager({ kernelManager, serverSettings });
-  await transformKernelExecution(sessionManager, mdast, frontmatter, file, true);
+  await transformKernelExecution(sessionManager, mdast, frontmatter, file, true, vfile);
 
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   await transformOutputsToCache(session, mdast, kind, { minifyMaxCharacters });
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -319,7 +319,7 @@ export async function postProcessMdast(
   });
   await pipe.run(mdast, vfile);
 
-  // Ensure there are keys on every node after post processing
+  // Ensure there are keys on every node after post-processing
   keysTransform(mdast);
   logMessagesFromVFile(session, fileState.file);
   logMessagesFromVFile(session, vfile);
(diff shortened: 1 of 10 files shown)


packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -37,7 +37,7 @@ export async function transformOutputsToCache(
 ) {
   const outputs = selectAll('output', mdast) as GenericNode[];
   // This happens sooner for notebooks
-  if (!outputs.length || kind !== SourceFileKind.Article) return;
```
agoose77, line 40: My current thinking is that we load the notebook as-is, and minify later during transformations in a single step. Is this going to have perf ramifications?

packages/myst-cli/src/transforms/inlineExpressions.ts
```diff
@@ -16,8 +16,8 @@ export interface IBaseExpressionResult {
 
 export interface IExpressionOutput extends IBaseExpressionResult {
   status: 'ok';
-  data: Record<string, string>;
-  metadata: Record<string, string>;
+  data: Record<string, any>;
+  metadata: Record<string, any>;
```
agoose77, line 20: MIME bundles should have untyped values, as AFAICT there's no requirement for the content to be a string.

packages/myst-cli/src/transforms/execute.ts
```diff
@@ -0,0 +1,212 @@
...
+        throw new Error();
+      } finally {
+        // Shutdown kernel
+        await kernel.shutdown();
+      }
+    }
+
+    // let rehydratedNodes: (Code | InlineExpression)[] | undefined;
+    // if (!ignoreCache) {
+    //   rehydratedNodes = getCachedExecution(codeOrEvalNodes);
```
agoose77, line 208: Haven't figured out the caching yet.

packages/myst-cli/src/transforms/execute.ts
```diff
@@ -0,0 +1,212 @@
...
+};
+
+type ICellBlock = GenericNode & {
+  children: (Code | ICellBlockOutput)[];
+};
+
+function isCellBlock(node: GenericNode): node is ICellBlock {
+  return node.type === 'block' && select('code', node) !== null && select('output', node) !== null;
+}
+
```
agoose77, line 127: Is there a better way to type this?

agoose77:
@fwkoch I've pinged you for review to hopefully course correct anything that's not looking like it's going in the right direction. 

At a high level, this is my thinking about the transforms / user flow:

1. We should prefer executed notebook outputs to pre-executed outputs
2. We should be able to skip execution e.g. if the cache is unchanged / user requests no execution

In getting the current WIP working, I switched from finding `Code` nodes to finding "notebook cell" blocks. This seems like a poor assumption on my part. Could you perhaps help me to understand why we have an `output` node rather than `output` being a property of `code` nodes? 

My plan is to slightly re-work inline expressions too, so that the "rendering" happens at a similar stage as the output rendering — inline expressions should likely be minified too.

agoose77:
At this stage, environment variables can be set to point to an existing Jupyter Server instance. In future, these will be configurable and optional (we could spin up our own server).

agoose77:
I've now unified the treatment of outputs and expression results such that the notebook processor just populates these from the notebook itself (e.g. metadata). A later stage would be required to minify these, and then "simplify" them for rendering.

@rowanc1 could you elaborate on the purpose of them minimisation? I assume it's to make manipulations of the AST less memory intensive?

packages/myst-cli/src/transforms/execute.ts
rowanc1: Things look good?!

packages/myst-cli/src/process/mdast.ts
```diff
@@ -223,6 +225,14 @@ export async function transformMdast(
   // Combine file-specific citation renderers with project renderers from bib files
   const fileCitationRenderer = combineCitationRenderers(cache, ...rendererFiles);
 
+  const serverSettings = ServerConnection.makeSettings({
+    baseUrl: process.env.JUPYTER_BASE_URL,
+    token: process.env.JUPYTER_TOKEN,
+  });
```
rowanc1, line 231: I would be curious on the vision here eventually on how to hook this up to the config.

rowanc1 (commented):
A few questions at this stage about where we are minifying outputs and rendering inline expressions/adding IDs etc.

Not sure if this is fully working yet, but looks like we have a cache, pulling out executable pieces, executing those against a kernel, and providing results.

Another thought as I am looking through this, I wonder if this would be good as a separate package? `myst-execute` maybe?

packages/myst-cli/src/transforms/outputs.ts
rowanc1: Yeah, I think that makes a lot of sense to minify later in the transforms. We need to do it before the image/math stuff, so likely `read -> parse -> optional execution -> minify -> other transforms`.

packages/myst-cli/src/transforms/execute.ts
rowanc1: @fwkoch did we have a `kind` on the block?

rowanc1:
The minimize outputs takes the encoded outputs (base64 images, html stuff, raw logs, etc.) and pulls them out of the mdast and writes them to disk. This is needed for writing latex, word, typst, etc. and for html is also more efficient for network requests and lazy loading of images in a long article.

We are also transforming the images into other formats (webp), which further improves load time. All of that needs them on disk.

packages/myst-cli/src/transforms/inlineExpressions.ts
```diff
@@ -84,31 +80,3 @@ function renderExpression(node: InlineExpression, file: VFile): StaticPhrasingCo
...
-  blocks.forEach((node) => {
-    const userExpressions = node.data?.[metadataSection] as IUserExpressionMetadata[];
-    const inlineNodes = selectAll('inlineExpression', node) as InlineExpression[];
-    inlineNodes.forEach((inlineExpression) => {
-      const data = findExpression(userExpressions, inlineExpression.value);
-      if (!data) return;
-      count += 1;
-      inlineExpression.identifier = `eval-${count}`;
-      inlineExpression.result = data.result;
-      inlineExpression.children = renderExpression(inlineExpression, file);
```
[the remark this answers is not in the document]
agoose77: Could you elaborate? My intention here is to move this to the end of the pipeline, rather than the front.

packages/myst-cli/src/process/notebook.ts
```diff
@@ -118,19 +140,14 @@ export async function processNotebook(
           value: ensureString(cell.source),
         };
 
-        const output: { type: 'output'; id: string; data: MinifiedOutput[] } = {
+        // Embed outputs into output block
+        const output: { type: 'output'; id: string; data: IOutput[] } = {
```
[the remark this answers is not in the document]
agoose77: Yes, I removed a guard on the minification transform.

packages/myst-cli/src/transforms/inlineExpressions.ts
```diff
@@ -84,31 +80,3 @@ function renderExpression(node: InlineExpression, file: VFile): StaticPhrasingCo
   }
   return [];
 }
-
-export function transformInlineExpressions(mdast: GenericParent, file: VFile) {
```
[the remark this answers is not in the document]
agoose77: Yes, it's pulled out into two functions now

packages/myst-cli/src/process/mdast.ts
```diff
@@ -309,7 +320,7 @@ export async function postProcessMdast(
   });
   await pipe.run(mdast, vfile);
 
-  // Ensure there are keys on every node after post processing
+  // Ensure there are keys on every node after post-processing
```
agoose77, line 323:
```suggestion
  // Ensure there are keys on every node after post processing
```

Revert comment change

packages/myst-cli/src/session/session.ts
```diff
@@ -157,4 +164,33 @@ export class Session implements ISession {
...
+      const kernelManager = new KernelManager({ serverSettings });
+      const manager = new SessionManager({ kernelManager, serverSettings });
+      // TODO: this is a race condition, even though we shouldn't hit if if this promise is actually awaited
+      this._jupyterSessionManager = manager;
+      return manager;
+    } catch {
+      this._jupyterSessionManager = undefined;
+      return undefined;
+    }
+  }
```
agoose77, line 195: This does not feel elegant to me, but I want to get "something" in place for us to talk about.

agoose77:
OK, at this point it seems like this is behaving as expected (though with no user configurability besides an ENV var). It will be good to discuss this on Monday :)

packages/myst-cli/src/transforms/execute.ts
fwkoch: We are tagging executable code blocks with `data.type = 'notebook-code'` - From notebooks - https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/process/notebook.ts#L24-L27 and from code cell directives - https://github.com/executablebooks/mystmd/blob/main/packages/myst-directives/src/code.ts#L215

However, I'm not sure we are relying on this field anywhere? In the myst-theme execution code at least, we find executable blocks in a similar way to your implementation: https://github.com/executablebooks/myst-theme/blob/855f29ec42a77deaa145481bfcab25b2d3c53757/packages/jupyter/src/execute/utils.ts#L8 The only difference is that function supports executable figures, i.e. `container` nodes with code/output children, in addition to blocks.

stevejpurves:
@agoose77 re: `outputs` are separate from `code` rather than properties I think stemmed from the `outputs` being embeddable as first-class entities, rather than a `code` cell that has it's source hidden. That thinking originally was also imprinted on the initial implementation, whether it's strictly essential or not.

stevejpurves:
I'd also just like to keep this visible: https://github.com/executablebooks/mystmd/blob/aa335d748edd4d636ad117a216ba78c1c1283e4f/packages/myst-cli/src/process/mdast.ts#L232

i.e. text-based notebooks are flagged as executable here and jupytext / myst text-based notebooks should probably be a test case for local execution too

packages/myst-cli/src/session/session.ts
```diff
@@ -157,4 +164,36 @@ export class Session implements ISession {
...
+              baseUrl: process.env.JUPYTER_BASE_URL,
+              token: process.env.JUPYTER_TOKEN,
+            });
+          }
+        },
+      );
+      const serverSettings = ServerConnection.makeSettings(partialServerSettings);
+      const kernelManager = new KernelManager({ serverSettings });
+      const manager = new SessionManager({ kernelManager, serverSettings });
+      // TODO: this is a race condition, even though we shouldn't hit if if this promise is actually awaited
```
stevejpurves, line 191: `await manager.ready` in turn does await the `kernelManager.ready` if that's what you mean

stevejpurves: callers of this function would need to know to `await (await jupyterSessionManager()).ready` unless you await `manager.ready` in here?

agoose77: In this instance it's that there's no logic to avoid creating two managers if one doesn't properly await the promise to `session.jupyterSessionManager()`. i.e., between the first await and finally setting `this._jupyterSessionManager`.

agoose77:
Closed in favour of #866 and #873.

PR #860, opened 2024-01-19
🤕 Transform heading depths

fwkoch:
- Add a new transform to normalized heading depths and warn on skipped depths, too deep headings
- Clarify behaviour around heading removal:
  - If there is no top-level h1, the first heading will still become the title
  - Title will only be stripped from mdast if it is top-level h1
  - Frontmatter has a flag to indicate if title is included in the mdast
- Respect page titles in multi-article export


packages/myst-cli/src/build/tex/single.ts
```diff
@@ -248,7 +253,12 @@ export async function localArticleToTexTemplated(
     const includeFilenames = results.map((result, ind) => {
       const includeFilename = `${name}-${content[ind]?.slug ?? ind}`;
       const includeFile = path.format({ dir, ext, base: `${includeFilename}${ext}` });
-      writeFileToFolder(includeFile, result.value);
+      let part = '';
+      const { title, content_includes_title } = content[ind]?.frontmatter ?? {};
+      if (title && !content_includes_title) {
+        part = `\\section{${title}}\n\n`;
```
fwkoch, line 259: I'm not sweating too much about duplicated code here - it is going to change almost immediately with (1) loading from a toc, where this `section`/`chapter` level is variable and (2) sub-article templating.

packages/myst-cli/src/process/file.ts
```diff
@@ -53,9 +56,10 @@ export async function loadFile(
         const { sha256, useCache } = checkCache(cache, content, file);
         if (useCache) break;
         const mdast = parseMyst(session, content, file);
+        const frontmatter = getPageFrontmatter(session, mdast, vfile);
```
fwkoch, line 59: Stripping frontmatter from a file now always happens here on load.

packages/myst-transforms/src/frontmatter.ts
```diff
@@ -20,7 +19,7 @@ type Options = {
 export function getFrontmatter(
   file: VFile,
   tree: GenericParent,
-  opts: Options = { removeYaml: true, removeHeading: true, propagateTargets: true },
+  opts: Options = { propagateTargets: true },
```
fwkoch, line 22: I took out some of these options at the code level - we were never passing in `false` anyway. Now, we _always_ remove yaml frontmatter block, and heading removal only happens if it is a top-level h1 and the user did not explicitly set `content_includes_title: true`.

rowanc1:
Have tested out web and other exports, works well! Thanks @fwkoch.

PR #863, opened 2024-01-22
fix: don't require MyST template when headless

agoose77:
`myst start --headless` shouldn't need to download the theme.

### DIFF
--- a/packages/myst-cli/src/build/site/start.ts
+++ b/packages/myst-cli/src/build/site/start.ts
@@ -108,8 +108,10 @@ export async function startServer(
   // Ensure we are on the latest version of the configs
   session.reload();
   warnOnHostEnvironmentVariable(session, opts);
-  const mystTemplate = await getMystTemplate(session, opts);
-  if (!opts.headless) await installSiteTemplate(session, mystTemplate);
+  if (!opts.headless) {
+    const mystTemplate = await getMystTemplate(session, opts);
+    await installSiteTemplate(session, mystTemplate);
+  }
   await buildSite(session, opts);
   const server = await startContentServer(session, opts);
   const { extraLinkTransformers, extraTransforms, defaultTemplate } = opts;


packages/myst-cli/src/build/site/start.ts
```diff
@@ -108,8 +108,10 @@ export async function startServer(
   // Ensure we are on the latest version of the configs
   session.reload();
   warnOnHostEnvironmentVariable(session, opts);
-  const mystTemplate = await getMystTemplate(session, opts);
-  if (!opts.headless) await installSiteTemplate(session, mystTemplate);
+  if (!opts.headless) {
+    const mystTemplate = await getMystTemplate(session, opts);
```
rowanc1, line 112: This is currently necessary to parse the options correctly. For example, the site template has `logo_dark` on it, which means moving files around and making that available etc. downstream. This also means not raising warnings on unknown options, etc.

I think that we need to split this into `getMystTemplateOptions` and `getMystTemplate`. The first one _should_ be an API call (or file system for a local template) that just loads up the `template.yml`.

Would be curious on @fwkoch's take as well.

fwkoch: Yeah - not much more to add. I'm not sure how much we gain by separating out API call for getting the options and (small?) download for getting the template. In that case we would need to be extra careful about keeping the API and each template package in sync (this would be nice 🙂 but I think currently the api only updates if the `myst-templates/templates` repo updates, not if a single template is updated).

---

Hmm, I think the problem here might be slightly different. When running `headless`, I suspect it's pretty likely the `template` value in frontmatter/myst.yml is actually just wrong; as a myst-theme developer, I'd think "I am running the theme separately, I don't need to also update the template to that path, just leave whatever default is already there." - In that case, we are downloading and validating against some arbitrary template, unrelated to the one running in the background.

If that's the case, then maybe themes should, like, serve up their own `template.yml` (well, `template.json`), then if we are running headlessly, the cli could just look up the template options from there...? This is a little funny... it feels like it's edging on a circular dependency (content and theme both need to know where the other is located, starting the content server before the theme would mean no template.yml to find) but maybe that's ok?

agoose77:
Closing this for now -- not clear on how to proceed but also need to be realistic about bandwidth! :)

PR #864, opened 2024-01-22
📦  Refactor to use `@citation-js/core`

agoose77:
This PR replaces the use of `citation-js` with `@citation-js/core`. With the CSL and bibtex plugins, the installed size is <1 MB.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,10 +8,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "@citation-js/core": "^0.6.8",
-        "@citation-js/plugin-csl": "^0.6.9"
-      },
       "devDependencies": {
         "@changesets/cli": "^2.26.1",
         "@types/node": "^20.2.5",
@@ -38,19 +34,6 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
-      "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
-      "dev": true,
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.0",
-        "@jridgewell/trace-mapping": "^0.3.9"
-      },
-      "engines": {
-        "node": ">=6.0.0"
-      }
-    },
     "node_modules/@babel/code-frame": {
       "version": "7.23.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.23.5.tgz",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -33,5 +33,9 @@
     "npm": ">=7.0.0",
     "node": ">=14.0.0"
   },
-  "packageManager": "npm@8.10.0"
+  "packageManager": "npm@8.10.0",
+  "dependencies": {
+    "@citation-js/core": "^0.6.8",
+    "@citation-js/plugin-csl": "^0.6.9"
+  }
```
rowanc1, line 40: We should only have to add these in the sub-package, not at the top!

rowanc1 (changes requested):
Nice, this looks good - I will do a quick test later today and then I think we can bring it in.

- [ ] fix up package.json in root
- [x] add changeset for citation-js-utils

package.json
agoose77: Wait, how did this happen?! I think my editor must have done this. I didn't notice the path when I was being lazy and using update for the add strategy. Yikes!

agoose77:
@rowanc1 I've run `prettier` in each package, and its changed some packages that I've not touched in this PR. Is there some kind of version dependency in prettier that's not captured by our development dependencies? It's interesting that the CI seems happy (i.e. my linting is not contra whatever version of prettier other people are running).

rowanc1:
Looks like the `package-lock` change is quite extensive, and changed prettier from `v3.0.0` --> `v3.2.4`, which is the root of these differences.

agoose77:
Ah, I was thinking about this the wrong way then. Fab, fixed.

rowanc1:
That is awesome, thanks for making the lock-diff much smaller!

Could you add a changeset for each package changed:

```
npm run changeset
```

and follow the instructions. This allows us to keep track of which packages need to be bumped and released next time we version.

agoose77:
> That is awesome, thanks for making the lock-diff much smaller!
> 
> Could you add a changeset for each package changed:
> 
> ```
> npm run changeset
> ```
> 
> and follow the instructions. This allows us to keep track of which packages need to be bumped and released next time we version.

Yes, sorry — been meaning to do that. Not sure why it slipped my mind to begin with :upside_down_face:

rowanc1 (approved):
Fell into the same problems, I put back in those changes and we can update the linting and merge it in!

package-lock.json
```diff
@@ -9768,9 +9619,9 @@
       }
     },
     "node_modules/prettier": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.0.0.tgz",
-      "integrity": "sha512-zBf5eHpwHOGPC47h0zrPyNn+eAEIdEzfywMoYn2XPi0P44Zp0tSq64rq0xAREh4auw2cJZHo9QUob+NqCQky4g==",
+      "version": "3.2.4",
```
rowanc1, line 9622: This is the root of the format changes.

package-lock.json
```diff
@@ -13357,7 +13188,9 @@
       "version": "1.0.2",
       "license": "MIT",
       "dependencies": {
-        "citation-js": "^0.6.3",
+        "@citation-js/core": "^0.7.6",
+        "@citation-js/plugin-bibtex": "^0.7.8",
+        "@citation-js/plugin-csl": "^0.7.6",
```
rowanc1, line 13193: @agoose77 I went to the next version, seems fine in testing!

PR #866, opened 2024-01-23
📦 Add myst-execute package

fwkoch:
This just adds a basic skeleton package for `myst-execute` - a sandbox for the jupyter execution work introduced in #856 

@agoose77 - hopefully this helps get you started. Let me know if you run into problems or need a hand with anything...

@rowanc1 - I published to npm, `ebp-bot` should have an invite to enable CI publishing.


packages/myst-execute/tests/run.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+casesList.forEach(({ title, cases }) => {
+  const filtered = cases.filter((c) => !only || c.title === only);
+  if (filtered.length === 0) return;
+  describe(title, () => {
+    test.each(filtered.map((c): [string, TestCase] => [c.title, c]))(
+      '%s',
+      (_, { before, after }) => {
+        executionTransform(before);
```
fwkoch, line 36: Maybe overly optimistic to assume we can run the tests in this way, including communication with a jupyter kernel. But maybe there's an easy way to make it work, or maybe we just move towards smaller unit tests 🤷‍♀️

packages/myst-execute/README.md
```diff
@@ -0,0 +1,3 @@
+# MyST Execute
+
+This library executes MyST content with Jupyter. It provides a transform that take a MyST tree and fills output nodes. It also provides caching for accessing code block outputs by ID.
```
fwkoch, line 3: This is super basic from our call this morning - please elaborate or rewrite however you see fit.

rowanc1:
`ebp-bot` has been added so CI + deployment should work for this!

packages/myst-execute/src/execute.ts
```diff
@@ -0,0 +1,340 @@
...
+    const thisResult = computedResult.shift();
+
+    if (isCellBlock(matchedNode)) {
+      // Pull out output to set data
+      const output = select('output', matchedNode) as unknown as { data: IOutput[] };
+      // Set the output array to empty if we don't have a result (e.g. due to a kernel error)
+      output.data = thisResult === undefined ? [] : (thisResult as IOutput[]);
+    } else if (isInlineExpression(matchedNode)) {
+      // Set data of expression to the result, or empty if we don't have one
+      matchedNode.result = // TODO: FIXME .data
```
agoose77, line 244: N.B. this should be `.data`, but will require a change to `myst-spec` too.

rowanc1: We could target that change in another PR? It will also require a change in `jupyterlab-myst` and `myst-theme`? This will also need to be backwards compatible in the theme to load from both places.

packages/myst-execute/src/types.ts
```diff
@@ -0,0 +1,26 @@
...
+   * Exception value
+   */
+  evalue: string;
+  /**
+   * Traceback
+   */
+  traceback: string[];
+}
+
+export type IExpressionResult = IExpressionError | IExpressionOutput;
```
agoose77, line 26: @rowanc1 not sure on where these types should go — they're very relevant to Jupyter, so putting them here makes sense. Equally, should we be able to talk about expression results _without_ needing to pull in execution? I suppose with type imports being compile-time only, it's not the end of the world. But, I'm wondering if we need a second package?

rowanc1: The `myst-common` package is the place for these:

https://github.com/executablebooks/mystmd/blob/main/packages/myst-common/src/types.ts

packages/myst-execute/src/execute.ts
```diff
@@ -0,0 +1,345 @@
...
+    } else {
+      // This should never happen
+      assert(false);
+    }
+  }
+}
+
+export type Options = {
+  cache: ICache<(IExpressionResult | IOutput[])[]>;
+  sessionFactory: () => Promise<SessionManager | undefined>;
```
agoose77, line 255: This maybe should return `SessionManager`, but I don't have a feel for MyST's error handling yet, so erring on the side of "report this in the transform".

rowanc1: Any error handling should be done through `fileError` and `fileWarning` and passing around a vfile.

Something like:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-cli/src/transforms/code.ts#L159

This will likely also be adding a new `RuleId` to keep track of these for later processing/ignoring/filtering.

packages/myst-common/src/index.ts
```diff
@@ -24,6 +24,7 @@ export {
   ParseTypesEnum,
   TargetKind,
 } from './types.js';
+export type { IExpressionResult, IExpressionError, IExpressionOutput } from './types.js';
```
rowanc1, line 27: @agoose77 can you add changesets for these as you are going! :)

agoose77: Yep! It's on my TODO, ha!

packages/mystmd-py/setup.cfg
```diff
@@ -25,6 +25,9 @@ keywords = markdown, latex, writing-software, scientific-visualization, pdf-gene
 packages = find:
 python_requires = >=3.6
 include_package_data = True
+install_requires =
+    jupyter-server
+    ipykernel
```
rowanc1, line 30: Can we hold off on these changes until it is in the CLI?

packages/myst-execute/src/execute.ts
```diff
@@ -0,0 +1,347 @@
...
+  });
+
+  const outputs: IOutput[] = [];
+  future.onIOPub = (msg) => {
+    // Only listen for replies to this execution
+    if (
+      (msg.parent_header as any).msg_id !== future.msg.header.msg_id ||
+      (msg.parent_header as any).msg_type !== 'execute_request'
+    ) {
+      console.debug('Ignoring IOPub reply');
```
rowanc1, line 45: Can we use the `session.log.debug` here? Or at the least pass in the `log` where we are piping things?

packages/myst-execute/src/execute.ts
```diff
@@ -0,0 +1,347 @@
...
+      const output = select('output', matchedNode) as unknown as { data: IOutput[] };
+      // Set the output array to empty if we don't have a result (e.g. due to a kernel error)
+      output.data = thisResult === undefined ? [] : (thisResult as IOutput[]);
+    } else if (isInlineExpression(matchedNode)) {
+      // Set data of expression to the result, or empty if we don't have one
+      matchedNode.result = // TODO: FIXME .data
+        thisResult === undefined ? undefined : (thisResult as unknown as Record<string, unknown>);
+    } else {
+      // This should never happen
+      assert(false);
```
rowanc1, line 247:
```suggestion
      throw new Error('Node must be either code block or inline expression.');
```

packages/myst-execute/src/manager.ts
```diff
@@ -0,0 +1,98 @@
+import type { ServerConnection } from '@jupyterlab/services';
+import which from 'which';
+import { spawn, spawnSync } from 'node:child_process';
```
rowanc1, line 3:
```suggestion
import { spawn } from 'node:child_process';
```

packages/myst-execute/tests/execute.yml
```diff
@@ -0,0 +1,200 @@
+title: executionTransform
```
rowanc1, line 1: Can we also add an ipynb test case? Maybe that is more of an end-to-end test later (e.g. if it requires the tree to be parsed)?

packages/myst-execute/tests/run.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Headers, Request, Response } from 'node-fetch';
+import type { IExpressionResult } from 'myst-common';
+import type { IOutput } from '@jupyterlab/nbformat';
+
+// fetch polyfill for node<18
+if (!globalThis.fetch) {
+  globalThis.fetch = nodeFetch as any;
+  globalThis.Headers = Headers as any;
+  globalThis.Request = Request as any;
+  globalThis.Response = Response as any;
```
rowanc1, line 19: Was this supposed to be in the CLI as well eventually to patch `fetch`?

packages/myst-execute/src/cache.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  get(key: string): T | undefined;
+
+  set(key: string, value: T): void;
+}
+
+/**
+ * An implementation of a basic cache
+ */
+export class LocalDiskCache<T> implements ICache<T> {
```
rowanc1, line 15: This is awesome, excited to see this expand. :)

rowanc1 (commented):
Taking another look through!

packages/mystmd-py/setup.cfg
rowanc1: I think if this comes out @agoose77, we can go ahead with a merge and then start working on the CLI pr for real.

rowanc1:
Looks like there is a merge conflict with the myst-common package and package.json after the release, we might need to rebase and just recommit the package-lock? Let me know if you need help!

packages/myst-execute/tests/run.spec.ts
agoose77: It is :)

rowanc1:
Thanks for fixing this up @agoose77 -- do you think it is a state that it is ready to come in?

agoose77:
Just want to tackle a few things RE cleanup first!

agoose77:
@rowanc1 I removed the logging statements that were expected anyway --- this is good to go imo.

fwkoch (commented):
Sorry to be slow hitting merge on this - it looks great! Nice to have a clean, separate package we can start pulling into the CLI.

PR #867, opened 2024-01-23
🕵️‍♀️ Support cross-references to file allowing labels on lifted headers

fwkoch:
By default, if no title is provided, MyST will lift the first H1 header out of the page and make it the title. However, if that header is labeled, the label is lost, e.g.

```markdown
(introduction)=
# Introduction To My Research

...
```

Anywhere that references `[](#introduction)` will be broken.

This PR fixes that, so now references to the label will become links to the page. To get here, it took a few steps:
1. processed pages now have `identifiers`
2. `ReferenceStateResolvers` can now get `fileTargets` in addition to "normal" node targets
3. Cross-references to these new file targets resolve into an internal `Link` node.

I also refactored the ReferenceState code a little bit... Previously there was a single page `ReferenceState` and a `MultiPageReferenceState`. Both had functionality to add targets and resolve targets; however, we should never actually add targets to a multi-page state - they are only on a single page. So now, there is a `ReferenceState` similar to before and a `MultiPageReferenceResolver` which only does the resolving part...

### DIFF
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -10,7 +10,7 @@ import {
   htmlPlugin,
   footnotesPlugin,
   ReferenceState,
-  MultiPageReferenceState,
+  MultiPageReferenceResolver,
   resolveReferencesTransform,
   mathPlugin,
   codePlugin,
@@ -85,13 +85,6 @@ const htmlHandlers = {
   },
 };
 
-export type PageReferenceStates = {
-  state: ReferenceState;
-  file: string;
-  url: string | null;
-  dataUrl: string | null;
-}[];
-
 export type TransformFn = (
   session: ISession,
   opts: Parameters<typeof transformMdast>[1],
@@ -133,6 +126,7 @@ export async function transformMdast(
     kind,
     frontmatter: preFrontmatter,
     location,
+    identifiers,
   } = cache.$getMdast(file)?.pre ?? {};
   if (!mdastPre || !kind || !location) throw new Error(`Expected mdast to be parsed for ${file}`);
   log.debug(`Processing "${file}"`);
@@ -159,7 +153,7 @@ export async function transformMdast(
   const references: References = {
     cite: { order: [], data: {} },
   };
-  const state = new ReferenceState({ numbering: frontmatter.numbering, file: vfile });
+  const state = new ReferenceState(file, { numbering: frontmatter.numbering, identifiers, vfile });
   cache.$internalReferences[file] = state;
   // Import additional content from mdast or other files
   frontmatterPartsTransform(session, file, mdast, frontmatter);
@@ -178,7 +172,7 @@ export async function transformMdast(
     })
     .use(inlineMathSimplificationPlugin)
     .use(mathPlugin, { macros: frontmatter.math })
-    .use(glossaryPlugin, { state }) // This should be before the enumerate plugins
+    .use(glossaryPlugin) // This should be before the enumerate plugins
     .use(abbreviationPlugin, { abbreviations: frontmatter.abbreviations })
     .use(enumerateTargetsPlugin, { state }) // This should be after math/container transforms
     .use(joinGatesPlugin);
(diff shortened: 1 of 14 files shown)


packages/myst-cli/src/process/mdast.ts
```diff
@@ -85,13 +85,6 @@ const htmlHandlers = {
   },
 };
 
-export type PageReferenceStates = {
-  state: ReferenceState;
-  file: string;
```
fwkoch, line 90: All of this stuff (file, url, etc) is now included directly on the `ReferenceState`, so we don't need another "state" class that includes "state" 🧹

packages/myst-cli/src/process/mdast.ts
```diff
@@ -178,7 +172,7 @@ export async function transformMdast(
     })
     .use(inlineMathSimplificationPlugin)
     .use(mathPlugin, { macros: frontmatter.math })
-    .use(glossaryPlugin, { state }) // This should be before the enumerate plugins
```
fwkoch, line 181: `state` was never actually used by this plugin 🧹

packages/myst-cli/src/process/mdast.ts
```diff
@@ -287,18 +281,18 @@ export async function postProcessMdast(
...
-    ? new MultiPageReferenceState(pageReferenceStates, file)
+    ? new MultiPageReferenceResolver(pageReferenceStates, file)
     : fileState;
   // NOTE: This is doing things in place, we should potentially make this a different state?
   const transformers = [
     ...(extraLinkTransformers || []),
     new StaticFileTransformer(session, file), // Links static files and internally linked files
   ];
-  linksTransform(mdast, state.file as VFile, {
+  linksTransform(mdast, state.vfile as VFile, {
```
fwkoch, line 291: Sometimes (even in this single function) `file` was the filename, sometimes `file` was the `VFile`... now we can just tell by the property name 🧹

packages/myst-cli/src/process/site.ts
```diff
@@ -139,17 +138,19 @@ export async function loadProject(
...
-      return {
-        state: cache.$internalReferences[page.file],
-        file: page.file,
-        url: selectedFile?.url ?? null,
-        dataUrl: selectedFile?.dataUrl ?? null,
-      };
+      const state = cache.$internalReferences[page.file];
+      if (state) {
+        const selectedFile = selectors.selectFileInfo(session.store.getState(), page.file);
+        if (selectedFile?.url) state.url = selectedFile.url;
```
fwkoch, line 146: (This is just consuming the change of these properties living directly on the `ReferenceState`, rather than a parent state that included them plus the state......)

packages/myst-transforms/src/enumerate.ts
```diff
@@ -229,29 +235,40 @@ export function formatHeadingEnumerator(counts: (number | null)[], prefix?: stri
   return out;
 }
 
-export interface IReferenceState {
-  file?: VFile;
-  initializeNumberedHeadingDepths: (tree: GenericParent) => void;
```
fwkoch, line 234: This interface is now only for "resolver" - so these initialize/add methods are only on `ReferenceState`.

packages/myst-transforms/src/enumerate.ts
```diff
@@ -279,8 +296,13 @@ export class ReferenceState implements IReferenceState {
         }
       }
     });
+    this.identifiers = opts?.identifiers ?? [];
```
fwkoch, line 299: `identifiers` are only added when `ReferenceState` is instantiated. This could change if we ever need another way to add these.

packages/myst-transforms/src/frontmatter.ts
```diff
@@ -64,6 +65,10 @@ export function getFrontmatter(
       // If this has a label what do we do? Add this label as a document reference
       (nextNode as any).type = '__delete__';
       frontmatter.content_includes_title = false;
+      if (nextNode.label) {
+        const { identifier } = normalizeLabel(nextNode.label) ?? {};
+        if (identifier) identifiers.push(identifier);
```
fwkoch, line 70: 🤔 Maybe we want to also push the html-id for the title, e.g. so `[](#introduction-to-my-research)` would also work...

fwkoch:
@agoose77 - I think you encountered this issue before, where an h1 label was broken?

rowanc1 (approved):
Have tested this locally and works well!

PR #870, opened 2024-01-24
🧹 Remove test-case patching in myst-parser tests

fwkoch:
This PR cleans up the myst-parser tests once we fix the source test data here: https://github.com/executablebooks/myst-spec/pull/55


fwkoch:
(Failure here is expected until the `myst-spec` PR lands)

PR #873, opened 2024-01-24
🪐 Allow for Jupyter execution (including inline) directly in mystmd CLI

agoose77:
This PR adds execution support to the mystmd CLI. A sketch at some docs below from @rowanc1!

- [ ] add health testing to Jupyter Server selection logic
- [ ] prewarm kernels?


packages/myst-cli/src/session/session.ts
```diff
@@ -19,6 +19,18 @@ import type { RootState } from '../store/reducers.js';
...
+import { findExistingJupyterServer, launchJupyterServer } from 'myst-execute';
+import { default as nodeFetch, Headers, Request, Response } from 'node-fetch';
+
+// fetch polyfill for node<18
+if (!globalThis.fetch) {
+  globalThis.fetch = nodeFetch as any;
+  globalThis.Headers = Headers as any;
+  globalThis.Request = Request as any;
+  globalThis.Response = Response as any;
+}
```
agoose77, line 33: This allows the `fetch` call in `@jupyterlab/services`

rowanc1:
Getting a test failure on my machine:

image

rowanc1:
Figured out my problem, there was a dead server hanging about that could not be removed. I went into the `runtime`  using `jupyter --paths` and deleted it manually.

As part of that testing I went through and changed a few things:

- improved some of the logging
- For errors: I made things that were happening (jupyter starting) be in `log` and things that were node/execution errors part of the `fileError`. The logs trigger immediately, while the ones in the vfile don't get logged for a while and are stored. I think this is a decent pattern, and a bit of a nuance on the advice I gave earlier.
- You now need to pass the `--execute` flag from the CLI, the default is `false`.
- You can clean the execution cache using `myst clean --execute`

Things I noticed:

- [ ] The jupyter server stays around and isn't shut down when running `myst build --execute`. Or `myst build --execute --html` this means that the process doesn't stop and just hangs. Need to investigate how to kill the server.
- [ ] I put the execute flag in, but this should maybe also be in the `myst.yml`?
- [ ] I don't think we are passing the minification flags around anymore? This needs to come back.

agoose77:
There's something wacky going on with Node - it refuses to quit even when the child process has been killed. There are tools to ask node why it's not quitting, but I don't have time right now to dig into it. It might be that we want to remove the use of closures and move this logic into a class. That might be less opaque to node, if refcycles are the problem.

packages/myst-cli/src/session/session.ts
```diff
@@ -134,6 +148,7 @@ export class Session implements ISession {
   publicPath(): string {
     return path.join(this.sitePath(), 'public');
   }
+
   _clones: ISession[] = [];
 
   clone(): Session {
```
rowanc1, line 154: We might need to put some logic into the clone session here, we don't want to be spanning a lot of jupyter servers to handle the results.

The session is cloned for each export, which goes through different transformation pipelines, but shouldn't create a new jupyter server.

packages/myst-cli/src/process/file.ts
```diff
@@ -50,7 +48,7 @@ export async function loadFile(
   file: string,
   projectPath?: string,
   extension?: '.md' | '.ipynb' | '.bib',
-  opts?: { minifyMaxCharacters?: number; preFrontmatter?: Record<string, any> },
```
rowanc1, line 53: I want to make sure this still works as expected. I don't now see where the minify code is called?

packages/myst-cli/src/session/session.ts
agoose77: I've resolved this, but it's not pretty. As previously discussed, I'm just avoiding putting anything in the public API that we don't desperately need

packages/myst-cli/src/process/mdast.ts
```diff
@@ -215,6 +219,19 @@ export async function transformMdast(
...
+      frontmatter: frontmatter,
+      ignoreCache: false,
+      errorIsFatal: false,
+      log: session.log,
+    });
+  }
+  transformRenderInlineExpressions(mdast, vfile);
+
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   await transformOutputsToCache(session, mdast, kind, { minifyMaxCharacters });
```
agoose77, line 236: @rowanc1 this function now maps `output.data` from raw MIME to minified.

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -37,7 +37,7 @@ export async function transformOutputsToCache(
 ) {
   const outputs = selectAll('output', mdast) as GenericNode[];
   // This happens sooner for notebooks
-  if (!outputs.length || kind !== SourceFileKind.Article) return;
```
agoose77, line 40: @rowanc1 this transform now does not ignore articles.

packages/myst-cli/src/process/file.ts
agoose77: See below for comments targe
tting the appropriate lines.

packages/myst-cli/src/process/mdast.ts
rowanc1: Awesome!!

packages/myst-cli/src/transforms/outputs.ts
rowanc1: Makes sense -- we are really blurring the lines now. :)

packages/myst-cli/src/process/mdast.ts
agoose77: Sorry @rowanc1, just to clarify -- the notebook ingest (`loadFile`) now loads the outputs in un-minified. We later re-write these to minified in the `transformOutputsToCache` on L236. So, I've not added any minification, I've just disabled the early-return that was previously there.

agoose77: @rowanc1 could you clarify your expectations? 

To my knowledge, the final AST _is_ minified, we just do it in a single place much later in the AST transform.

rowanc1: Ok, that is my expectation. So hopefully all is fine! I saw that we were _maybe_ not passing an option somewhere (in a previous diff). I will test it out on a bunch of content and see if there are any issues!

docs/execute-notebooks.md
```diff
@@ -0,0 +1,44 @@
...
+MyST performs execution of notebooks by communication with a Jupyter Server. Jupyter Server is distributed as a Python package, which can be installed from PyPI or conda-forge, e.g.
+```bash
+pip install jupyter-server
+```
+Jupyter Server is only responsible for orchestrating execution of your code. To actually perform execution, you must also install a kernel. For Python, this might be `ipykernel`, e.g.
+```bash
+pip install ipykernel
+```
+
+If Jupyter Server is installed and the `--execute` flag is passed to `myst start` or `myst build`, then MyST will attempt to find an healthy existing Jupyter Server. Internally, this is performed using `python -m jupyter_server list`. If no existing servers are found, then MyST will attempt to launch one using `python -m jupyter_server`.
```
rowanc1, line 37:
```suggestion
If Jupyter Server is installed and the `--execute` flag is passed to `myst start` or `myst build`, then MyST will attempt to find a healthy existing Jupyter Server. Internally, this is performed using `python -m jupyter_server list`. If no existing servers are found, then MyST will attempt to launch one using `python -m jupyter_server`.
```

agoose77:
@fwkoch I changed the Python package from using setuptools to using the Hatch build backend. The benefits of this are fewer config files, more declarative config, and the ability to use plugins to pull metadata from mystmd's `package.json`. Are you comfortable with this change?

package.json
```diff
@@ -13,7 +13,7 @@
     "test": "turbo run test",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "changeset": "changeset",
-    "version": "changeset version && npm install && cd packages/mystmd-py && bash scripts/bump-version.sh",
+    "version": "changeset version && npm install",
```
rowanc1, line 16: Oh cool, this is nice.

packages/myst-cli/src/process/site.ts
```diff
@@ -283,9 +287,7 @@ export async function processProject(
       // Load all citations (.bib)
       ...project.bibliography.map((path) => loadFile(session, path, siteProject.path, '.bib')),
       // Load all content (.md and .ipynb)
-      ...pages.map((page) =>
-        loadFile(session, page.file, siteProject.path, undefined, { minifyMaxCharacters }),
-      ),
+      ...pages.map((page) => loadFile(session, page.file, siteProject.path, undefined)),
```
rowanc1, line 290: I am still a bit concerned that we are no longer passing this option in at all? I think this actually might be breaking JATS - @fwkoch I would love another set of eyes here.

package.json
agoose77: To be clear here, the version is only updated at pypi deployment time (because it's taken from package.json, which is copied in the deploy bash scipt).

agoose77:
@rowanc1 I've updated the cache key struct to include the `raises-exception` tag that allows notebooks to ignore errors. I haven't included a cache-version in the key; my rationale is that this is implicitly part of the hash - if we change the structure, the hash will change.

fwkoch (approved):
Ok - putting my approval on this. PR looks great! Lots of tedium passing the `execute` flag around - maybe those CLI options could be a refactor target some time.

Regarding the `minifyMaxCharacter` stuff - I played around with static outputs and I think it's all fine...? During mdast processing we call `transformOutputsToCache` - this does the exact same minification we were also doing during notebook loading. So now (since minification was removed from `loadFile`) it just always happens in that transform instead of both places. I still have some vague worries that there was an edge case causing us to do this in two places, but no test coverage and I couldn't recreate it 🤷‍♀️ Good to land, and we can just keep an eye on it.

rowanc1 (approved):
Thanks @fwkoch for your review -- and @agoose77 for going back and forth on this. I am hugely excited to get this out and start testing it in the wild. I have been using it for the last several days and it is working well!

🚀

rowanc1:
Ah -- we didn't add changesets @agoose77! :)

PR #875, opened 2024-01-25
📑 Build multi-page exports from toc file

fwkoch:
The basic outcome of this PR is that multi-page exports (pdf/tex/typst) can now be structured from a `_toc.yml` file. To get there, I've made some finer-grained changes:

- In frontmatter `export` definitions, `articles` are now objects with `{ title, file, level }`. This structure is parallel to Of course, you can continue to specify articles as a list of filenames, as before.
  - [ ] Todo: `title` only applies to "articles" without a `file`, e.g. chapter names; if `file` is provided, `title` is ignored. Perhaps this title could replace the file title? (I'll make this a separate PR - we should maybe allow overriding of all page frontmatter...)
- Reading toc files, now has the option to support level -1 (parts) and level 0 (chapters) - this allows latex part/chapter to be separate from heading levels 1-6.
- In frontmatter, you may now specify `toc` file instead of articles. Then, the contents of that file will be converted to articles. If no toc/article is provided and the `export` is in your `myst.yml`, the export will follow the structure of the project, rather than just failing! (Exports defined on a MyST page without toc/article will still build a single page export from that page.)
- `jb-article` tocs (with top-level sections) are now supported.


packages/myst-cli/src/build/tex/single.ts
```diff
@@ -133,19 +153,25 @@ export async function localArticleToTexRaw(
...
+    let fileInd = 0;
     const { dir, name, ext } = path.parse(output);
-    const includeFileBases = results.map((result, ind) => {
-      const base = `${name}-${content[ind]?.slug ?? ind}${ext}`;
-      const includeFile = path.format({ dir, ext, base });
-      let part = '';
-      const { title, content_includes_title } = content[ind]?.frontmatter ?? {};
-      if (title && !content_includes_title) {
-        part = `\\section{${title}}\n\n`;
+    articles.forEach((article) => {
```
fwkoch, line 159: There is some nasty repetition among multipage exports for `raw tex`, `templated tex`, `raw typst`, `templated typst` that needs refactoring. I suspect this will continue to get worse as we start updating individual article frontmatter, etc. - eventually it will reach a breaking point. Not this PR though... 😅

packages/myst-cli/src/build/utils/collectExportOptions.ts
```diff
@@ -16,6 +23,30 @@ import { getDefaultExportFilename, getDefaultExportFolder } from './defaultNames
...
+  }
+}
+
+function assignArticlesFromTOC(session: ISession, exp: Export, tocPath: string) {
+  const allowLevelLessThanOne = [
+    ExportFormats.tex,
+    ExportFormats.pdf,
+    ExportFormats.pdftex,
+  ].includes(exp.format);
+  const proj = projectFromToc(session, tocPath, allowLevelLessThanOne ? -1 : 1);
```
fwkoch, line 46: tex/pdf exports may have part/chapter with levels -1 / 0 respectively. This also needs to take into account the template, though. These levels should only be used for book templates.

packages/myst-cli/src/build/utils/collectExportOptions.ts
```diff
@@ -388,5 +438,5 @@ export async function collectExportOptions(
       );
     }),
   );
-  return exportOptionsList;
+  return filterAndMakeUnique(exportOptionsList);
```
fwkoch, line 441: This makes export output names unique. Previously, we only did this per file. Now, though, it's fairly easy to get collisions across files if they specify the same toc. This eliminates those collisions.

(There are edge cases where collisions in a file _and_ across files are encountered. There, we run this function twice and will get file names like `my-output_0_0.pdf` - a bit of an ugly name, but I don't think it matters - names are still unique, and if you don't like them, simply define your own `output` value.)

packages/myst-cli/src/build/utils/localArticleExport.ts
```diff
@@ -33,12 +33,13 @@ async function runExportAndWatch(
...
   let results = await exportFn(session, $file, exportOptions, projectPath, clean);
   if (watch) {
+    const articleFiles = articlesWithFile(exportOptions.articles);
     const watchedFiles = new Set([
       $file,
-      ...exportOptions.articles,
-      ...exportOptions.articles
-        .map((article) => {
-          return selectors.selectFileDependencies(session.store.getState(), article);
+      ...articleFiles.map(({ file }) => file),
```
fwkoch, line 39: Since `toc` is converted to articles first, this watch function will not trigger on changes to the `toc`; only to the articles it includes.

(However, this isn't really a regression. Even if articles are defined directly, adding a new article before did not add that article to the watch. Picking up on changes to the article list / toc is a separate feature.)

packages/myst-transforms/src/headings.ts
```diff
@@ -4,27 +4,30 @@ import type { Heading } from 'myst-spec-ext';
...
+ * is greater than 6, a warning is also raised, and all values greater than
+ * 6 will be left at 6.
  */
 export async function headingDepthTransform(
   tree: GenericParent,
   vfile: VFile,
   opts?: HeadingDepthOptions,
 ) {
-  const titleDepth = opts?.titleDepth && opts.titleDepth > 0 ? opts.titleDepth : 0;
+  const firstDepth = opts?.firstDepth && opts.firstDepth > 0 ? opts.firstDepth : 1;
```
fwkoch, line 30: This heading now knows nothing about titles - previously there was some conflict between handing `title`/`content_includes_title` here and in `processMdast` where this transform was called. Now, all that logic is in `processMdast` and this transform only cares about the first (lowest) depth value.

rowanc1:
Tested out locally! Thanks for the docs improvements as well. Excited to get per-"part" templating for the multi-article export.

PR #879, opened 2024-01-30
🤖 Add build ci option to not write versions

fwkoch:
This adds a `--ci` flag to `myst build` - for now all it does is take the version comments out of jtex-rendered files.


rowanc1 (approved):
Nice -- looks like moving some things into option dicts. Feel free to bring this in whenever!

PR #880, opened 2024-01-30
🛑 Stop turning dois to links in typst

fwkoch:
DOI citations are now handled correctly in our typst exports so we no longer need to write them as links. Easy!


rowanc1 (approved):
Nice, this was done before we had bibtex output completed!

PR #887, opened 2024-02-06
🤖 Consolidate install steps for all github actions

fwkoch:
This is trying out a "composite action" so install steps are not duplicated across ci / release actions.

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -1,16 +1,11 @@
 name: MyST Install Action
-description: Checkout MyST and install Node, MyST, and other dependencies
+description: Install Node, MyST, and other dependencies
 inputs:
   node:
     required: true
 runs:
   using: 'composite'
   steps:
-    - name: Checkout Repo
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 2
-          submodules: recursive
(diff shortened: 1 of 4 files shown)


.github/actions/install/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    default: '18'
+runs:
+  using: 'composite'
+  steps:
+    - name: Install libxml
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libxml2-utils
+      shell: bash
+    - name: Setup Node.js 18.x
```
rowanc1, line 16:
```suggestion
    - name: Setup Node.js
```

rowanc1 (commented):
Looks great!

rowanc1:
I am consistently seeing timeout failures on the execute tests. Can we add a larger timeout flag, not sure if that is what is happening:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-execute/tests/run.spec.ts#L101

fwkoch:
Hmm, yeah, successful tests take right about the same amount of time as the timeout:
image

I'll try bumping the timeout, but I wonder if there's something we can do to speed up the jupyter loading (launch the server outside of running the tests or something...?).

PR #897, opened 2024-02-12
🔧  Prevent basic transforms from modifying headings by default

fwkoch:
Recently a new transform was added to `basicTransformations` which normalized heading depth to start at `firstDepth` and count sequentially from there (eliminating skipped depth levels in the source document, etc. - introduced here: https://github.com/executablebooks/mystmd/pull/860 and modified here: https://github.com/executablebooks/mystmd/pull/875/files#diff-8c1296487484c1e135623cfc74ae7da928a293e1b6e5078690cfc7a1fe01df09)

However, there was no way to disable this behaviour; running the basic transforms meant heading depths would get normalized.

Now, this transform _only_ runs if you explicitly provide `firstDepth`. `myst-cli` always provides a value for `firstDepth`, so default, opinionated MyST behaviour is unchanged. But if you use `headingDepthTransform` or `basicTransformations` directly, this functionality may be disabled.

### DIFF
--- a/.changeset/brown-numbers-wait.md
+++ b/.changeset/brown-numbers-wait.md
@@ -1,0 +1,5 @@
+---
+'myst-transforms': patch
+---
+
+Prevent basic transforms from modifying headings by default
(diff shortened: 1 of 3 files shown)


rowanc1:
Thanks @fwkoch -- this is great to leave the basic functionality not modifying the heading depth.

PR #899, opened 2024-02-12
👈 Cross reference the subEnumerator directly.

rowanc1:
Also adds support in latex for `subref`, `landscape` and improves `subfigure` support for multiple captions defined in latex.


fwkoch (approved):
This makes sense - gives `fillReferenceEnumerators` access to the target node so it can have more complicated logic about what the enumerator looks like.

PR #900, opened 2024-02-12
MAINT: update python release workflow

agoose77:
This PR:
- Updates the `mystmd` package metadata to provide a valid Git URL (only https scheme) for PyPI
- Updates the release workflow to use the new trusted publisher

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,13 @@ on:
   push:
     branches:
       - main
-
+  workflow_dispatch:
+    inputs:
+      forcePublishPyPI:
+        description: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -4,13 +4,21 @@ on:
   push:
     branches:
       - main
-
+  workflow_dispatch:
+    inputs:
+      forcePublishPyPI:
+        description: |
+          Force publish to PyPI, even if the NPM release has already occurred.i
```
rowanc1, line 11:
```suggestion
          Force publish to PyPI, even if the NPM release has already occurred.
```

PR #901, opened 2024-02-12
🛠 Fix deployment to PyPI

rowanc1:
Takes over from #900.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
   release-py:  
     name: Release Python
     runs-on: ubuntu-latest
-    if: ${{ needs.build-npm.outputs.published || inputs.forcePublishPyPI }}
+    if: ${{ needs.release-npm.outputs.published || inputs.forcePublishPyPI }}
     needs:
       - release-npm
     environment:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -4,13 +4,22 @@ on:
   push:
     branches:
       - main
```
rowanc1, line 6: @agoose77 if we want to keep testing here, we could change this to `pypi` and switch it back before squashing.

.github/workflows/release.yml
```diff
@@ -32,12 +41,35 @@ jobs:
...
+        working-directory: packages/mystmd-py
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-package-distributions
+          path: packages/mystmd-py/dist/
+  release-py:  
+    name: Release Python
+    runs-on: ubuntu-latest
+    if: ${{ needs.release-npm.outputs.published || inputs.forcePublishPyPI }}
```
rowanc1, line 60: Should this be `steps.changesets.outputs.published == 'true'` or does that not work across jobs?

PR #902, opened 2024-02-12
🛠 Fix deployment to PyPI

agoose77:
Takes over from #900 and #901

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
   release-py:  
     name: Release Python
     runs-on: ubuntu-latest
-    if: ${{ needs.release-npm.outputs.published || inputs.forcePublishPyPI }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Release
 on:
   push:
     branches:
-      - main
+      - pypi
```
rowanc1, line 6: Before merging in, change this back to main!

rowanc1 (changes requested):
Before we put this back..!

.github/workflows/release.yml
rowanc1: image

rowanc1, line 6:
```suggestion
      - main
```

rowanc1 (approved):

rowanc1:
I have also now deleted the `TWINE` secrets.

PR #906, opened 2024-02-13
﹫ Improve typst handling of `@` in text and code

rowanc1:


rowanc1:
Fixes up emails which are no longer treated as labels!

PR #911, opened 2024-02-14
🪆 Support `\input` in latex

rowanc1:
This supports the input macro for latex, and improves the include transform to be able to be recursive.

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -4,6 +4,7 @@ import { createHash } from 'node:crypto';
 import { tic } from 'myst-cli-utils';
 import { TexParser } from 'tex-to-myst';
 import { VFile } from 'vfile';
+import type { GenericParent } from 'myst-common';
 import { RuleId, toText } from 'myst-common';
 import { validatePageFrontmatter } from 'myst-frontmatter';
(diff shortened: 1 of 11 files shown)


packages/myst-cli/src/process/file.ts
```diff
@@ -112,6 +115,13 @@ export async function loadFile(
         const { sha256, useCache } = checkCache(cache, content, file);
         if (useCache) break;
         const tex = new TexParser(content, vfile);
+        await includeDirectiveTransform(tex.ast as GenericParent, vfile, {
```
rowanc1, line 118: @fwkoch this works, but might not be the best place for it. I think we probably want to improve the include directive to actually use the `loadFile` function so that it uses the cache and handles latex and ipynb, which it doesn't at the moment.

We also want to handle dependencies properly, and I don't know how that works!

rowanc1: Ok, I tried to make that happen all in one place, and it needs a bunch more love than I think I am willing to give it in this PR.

The load file needs to know if it is raw or not, and then _should_ use the caching layer, but if you have more than one layer of dependency then nothing updates quite right (in real time).

I am going to go for the merge, and welcome post-merge comments!

PR #912, opened 2024-02-14
🪵 Add programmatic `logs` to myst for inspection

rowanc1:
This adds a first pass at programatic logs, which can be used deterministically in CI systems (and other places). For example `myst build` will now tell you about the exports it has found and what it is building, and that can be accessed without parsing stdout.

image


fwkoch (approved):
Looks like a good starting point! Nice and simple and flexible, just log any json.

PR #913, opened 2024-02-15
🏴‍☠️ Cache DOIs and intersphinx results to file

fwkoch:
This adds a new `cache` directory to the `_build` folder, where web request results may be stored across sessions.

For now, this is storing:
- doi.org bibtex responses for each successful doi request
- intersphinx inventory files
- successful link check statuses


rowanc1 (approved):
Changes are awesome. Really so much faster!

choldgraf:
I didn't notice this - it is really cool!

PR #917, opened 2024-02-20
feat: lift ToC parsing into `myst-jb-toc`

agoose77:
This new package is very simple, but it means that we can access it from the JupyterLab extension.

I removed upgrade support, because the technical debt of handling the old ToC should probably not be something we inherit. I also added some simple tests.


packages/myst-cli/src/project/toToc.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import yaml from 'js-yaml';
 import { join, relative } from 'node:path';
 import { removeExtension } from '../utils/removeExtension.js';
-import type { JupyterBookChapter, TOC } from '../utils/toc.js';
+import type { JupyterBookChapter, TOC } from 'myst-jb-toc';
```
rowanc1, line 5: Thoughts on naming the package `myst-toc` or something? That might give us room to evolve this function (better support jb-article, and switch between these toc pieces pretty easily in the future).

.changeset/neat-cycles-walk.md
```diff
@@ -0,0 +1,6 @@
+---
+"myst-cli": minor
```
rowanc1, line 2: I think that we can consider most changes to the CLI a patch until it has backwards incompatible user-facing behaviour. We aren't really encouraging/supporting imports from the CLI at the moment. Thoughts?

```suggestion
"myst-cli": patch
```

packages/myst-cli/src/project/fromToc.ts
```diff
@@ -71,7 +96,9 @@ export function projectFromToc(
     throw new Error(`Could not find TOC "${filename}". Please create a '_toc.yml'.`);
   }
   const { dir, base } = parse(filename);
```
rowanc1, line 98:
```suggestion
  const { dir } = parse(filename);
```

packages/myst-cli/src/project/fromToc.ts
```diff
@@ -71,7 +96,9 @@ export function projectFromToc(
...
-  const toc = readTOC(session.log, { filename: base, path: dir });
+  const contents = fs.readFileSync(filename).toString();
+  const toc = parseTOC(contents);
```
rowanc1, line 100: Are we always going to pass validation at this point? (I think so?)

packages/myst-jb-toc/.gitignore
```diff
@@ -0,0 +1 @@
+licenses.json
```
rowanc1, line 1: Can probably bring this up to the base gitignore?

packages/myst-jb-toc/src/toc.ts
```diff
@@ -0,0 +1,48 @@
...
+export type TOC = {
+  format: string;
+  root: string;
+  sections?: JupyterBookChapter[];
+  chapters?: JupyterBookChapter[];
+  parts?: JupyterBookPart[];
+};
+
+export function parseTOC(contents: string): TOC {
+  const toc = yaml.load(contents) as any;
```
rowanc1, line 34: Maybe worth a specific error return here - passing on the yaml validation error if it exists, but giving some context to know that the yaml failed reading a toc?

rowanc1 (commented):
Some minor things I saw looking through, only things I think we need to address/comment on are:

1) name of the package (allowing us to have this toc package cover all of our toc use-cases into the future)
2) patch for cli, rather than minor?

packages/myst-cli/src/utils/toc.ts
```diff
@@ -1,105 +0,0 @@
...
-  };
-  return toc;
-}
-
-export function readTOC(log: Logger, opts?: TocOptions): TOC {
-  const filename = join(opts?.path || '.', opts?.filename || '_toc.yml');
-  const toc = yaml.load(fs.readFileSync(filename).toString()) as any;
-  if (Array.isArray(toc)) {
-    try {
-      const old = upgradeOldJupyterBookToc(toc);
```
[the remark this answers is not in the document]
agoose77: This is what i was referring to when I said that I removed upgrade support. I know these exist in the wild, but I reckon we could avoid taking in the technical debt as those books are likely old!

packages/myst-jb-toc/.gitignore
agoose77: Perhaps? I think I added this through copying another package - let me confirm whether it's necessary.

packages/myst-cli/src/project/toToc.ts
agoose77: My gut feeling is that we're moving towards not having a ToC as a separate part of MyST in the future development direction. If that's the case, then I think this ToC package will/should always be concerned with reading JB ToCs if needs be. Though, as JB v2 develops, we'd eventually probably move this package out into JB's repo.

.changeset/neat-cycles-walk.md
agoose77: As I removed some conversion logic, I made it a minor. Thus, I think the minor should stand unless I revert that change.

packages/myst-cli/src/project/toToc.ts
rowanc1: Ok, fair enough! Maybe we should call it `jupyterbook-toc`, as it is less to do with myst in that scenario.

.changeset/neat-cycles-walk.md
rowanc1: Can we revert the change please? The effort to support backwards compatibility in this case is minor (and already done) and we can just do it on the fly for these authors.

I think that if we can support jupyterbooks at various versions without too much effort we should, even if requires a bit more effort on our part.

packages/myst-cli/src/utils/toc.ts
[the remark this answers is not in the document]
rowanc1: See comment above, I think that we should maintain backwards compatibility as much as we can in general, unless that requires a huge amount of effort on our part (I think it is reasonable to auto-convert/upgrade and warn here).

packages/myst-jb-toc/.gitignore
rowanc1: Ah, yes, that is from `myst-frontmatter` (when SPDX licenses update, we generate that json file) -- should be deleted here!

packages/myst-jb-toc/src/toc.ts
agoose77: I'm not sure what the pattern for re-throwing an exception is for TS. Can you shed any light?

rowanc1: `throw new Error('Could not load TOC yaml')`

I usually print the initial error message (err.stack or trace?) to a debug log as well.

packages/myst-cli/src/project/fromToc.ts
```diff
@@ -16,39 +23,96 @@ import type {
...
+      filename,
+      `Table of Contents (ToC) file did not pass validation:\n - ${message}\n - An implicit ToC will be used instead\n`,
+      'error',
+      { ruleId: RuleId.validTocStructure },
+    );
+    return false;
+  }
+}
+
+function pagesFromSubtree(
```
agoose77, line 62: To my knowledge, this function `pagesFromSubtree` is mapping from a nested structure to a flat list of page/directory types. Is this understanding correct?

agoose77 (commented):
@fwkoch I'd benefit from your thoughts on the direction of this PR. The intention here is:

- Factor out ToC handling for re-use in JLab
- Add more robust upgrade / parsing for all the possible variants of JB ToCs
- Uses schemas to robustly define what is valid and what is not

I need to finish changes to `toToc`, but first it would be helpful to check my understanding here!

agoose77:
Closing this in favour of a smaller PR that just adds the new package.

PR #920, opened 2024-02-20
🫓 Flatten inlineCode (that may have children)

rowanc1:
When tex is being parsed, it is possible for inline code `\texttt{}` to have children. If all of those children are text, then the children are transformed into be a single node with `.value`.


fwkoch (approved):
Seems to pretty carefully apply to only a specific case, looks good!

PR #921, opened 2024-02-20
🏛 Support `\textrm{}` in latex

rowanc1:


packages/tex-to-myst/src/marks.ts
```diff
@@ -53,6 +54,11 @@ const TEXT_MARKS_HANDLERS: Record<string, Handler> = {
     state.data.openGroups.push('inlineCode');
     state.data.ignoreNextWhitespace = true;
   },
+  macro_textrm(node, state) {
+    state.openParagraph();
```
fwkoch, line 58: This only opens a paragraph if it's not already inside phrasing content 👍

However... looks like in tex this is often used inside math. Will that work here? I guess that will hopefully just need to be handled later in the math rendering.

fwkoch (approved):
This looks good - even if there is subtle weirdness around math, this is still an improvement when using this macro elsewhere!

packages/tex-to-myst/src/marks.ts
rowanc1: In a math context we take a different path and lift the original latex out, and don't process it (leaving that to katex/mathjax etc.).

PR #922, opened 2024-02-20
💬 Support quote environments in latex

rowanc1:
Support quote, displayquote and epigraph

### DIFF
--- a/packages/tex-to-myst/src/quotes.ts
+++ b/packages/tex-to-myst/src/quotes.ts
@@ -1,0 +1,35 @@
+import type { Handler } from './types.js';
+import { getArguments } from './utils.js';
+
+export const QUOTE_HANDLERS: Record<string, Handler> = {
+  env_quote(node, state) {
+    state.closeParagraph();
+    state.openNode('blockquote');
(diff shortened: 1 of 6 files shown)


packages/tex-to-myst/src/quotes.ts
```diff
@@ -0,0 +1,35 @@
...
+    state.renderChildren(node);
+    state.closeNode();
+  },
+  macro_epigraph(node, state) {
+    const [quote, author] = getArguments(node, 'group');
+    if (author) {
+      state.openNode('container', { kind: 'quote' });
+    }
+    state.openNode('blockquote');
+    state.openParagraph();
```
fwkoch, line 23: These `openParagraph`/`closeParagraph` calls aren't in the other handlers - maybe unnecessary? I think the behaviour is the same in each case (looking at test cases), so the differences in the code are a little misleading.

fwkoch (approved):
Looks nice! Just a minor nitpick-y comment.

rowanc1:
Tests were a timeout on SPDX. Merging!

PR #923, opened 2024-02-20
💭 Support verbatim comment environment in latex

rowanc1:


fwkoch (approved):
Nice! Just handling text inside comment environment.

PR #934, opened 2024-02-22
🪪 Add id to project frontmatter and populate on init

fwkoch:
Adding a unique `id` to project frontmatter gives us a way to consistently track the project across different contexts. By default, this is populated on `init` with `<org>/<repo>` if inside a github project and a `uuid` if not.

I also added brief documentation for the other project-only frontmatter fields 📚📚📚


rowanc1 (commented):
I think the default should probably just be the uuid!

PR #937, opened 2024-02-27
🎓 Citation improvements

fwkoch:
- Some dois do not have bibtex available from doi.org - we now give more useful warnings in this case
- Year is parsed from dates if they are defined as a string in bibtex (e.g. "August 1, 2012") instead of structured `year`/`day` values.
- Fix bug where all non-doi urls were deleted from citation html
- CLI warns on unknown citation labels, eg `@unknown`
- cite nodes now have enumerator value so we can switch to citing like `[1]`
- URLs are pulled from citation data to be available to the renderer

### DIFF
--- a/.changeset/five-countries-wait.md
+++ b/.changeset/five-countries-wait.md
@@ -1,0 +1,5 @@
+---
+'citation-js-utils': patch
+---
+
+Get year from citation issued literal if it is not parsed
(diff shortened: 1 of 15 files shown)


rowanc1 (approved):
Tested locally - this works so much better, thank you!!

PR #939, opened 2024-02-27
🥩 Add raw directive

fwkoch:
The `raw` directive lets you include raw latex in a myst file. The latex will be parsed using `tex-to-myst`:
````markdown
```{raw} latex
\section{Hello World!}
\textbf{Hello World!} Here's some latex!
\begin{equation}
\gamma^2+\theta^2=\omega^2
\end{equation}
Pretty cool!
```
````

You can also use the `raw` directive to add unparsed text:
````markdown
```{raw}
*Keep these astrisks!*
```
````

### DIFF
--- a/.changeset/new-walls-brake.md
+++ b/.changeset/new-walls-brake.md
@@ -1,0 +1,7 @@
+---
+'myst-directives': patch
+'myst-spec-ext': patch
+'myst-cli': patch
+---
+
+Add raw directive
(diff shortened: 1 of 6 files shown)


stevejpurves (approved):

fwkoch:
I think there's more to do on this, in terms of getting this into static exports correctly, supporting html and other raw types, maybe combining logic with `include` - since `raw` is similar to include, the content is just embedded in the page...

For now though, i'll land this first pass!

PR #940, opened 2024-02-28
🐍 Remove .py from linkify domains

fwkoch:
There are a ton of domains that `linkify` catches, including `py`... That means someone writing about their python `my_script.py` file gets linkified.

For now, I've addressed this by hard-coding excluded domains in myst parser. We may want to make this configurable at the MyST config level...?

I think hardcoding what is excluded is ok, since users can always explicitly make things a link. If something they want linkified is not working (because of our exclude list), they can just make it an explicit link. Potentially we even lock down the domain list way more... there are currently 1450 valid domains 😳 https://github.com/stephenmathieson/node-tlds/blob/master/index.json (1449 with `py` excluded).

### DIFF
--- a/packages/myst-parser/src/config.ts
+++ b/packages/myst-parser/src/config.ts
@@ -69,4 +69,4 @@ export const MARKDOWN_IT_CONFIG = {
 };
 
 // List of valid TLDs to exclude from linkify
-export const EXCLUDE_TLDS = ['py'];
+export const EXCLUDE_TLDS = ['py', 'md'];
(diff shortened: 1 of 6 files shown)


packages/myst-parser/src/config.ts
```diff
@@ -67,3 +67,6 @@ export const MARKDOWN_IT_CONFIG = {
     },
   },
 };
+
+// List of valid TLDs to exclude from linkify
+export const EXCLUDE_TLDS = ['py'];
```
rowanc1, line 72: We can also add md here!

rowanc1 (commented):
This looks great, maybe we can change this to be configured externally rather than bundling the domain list in the parser, but that can come in the future! We should also exclude .md

PR #942, opened 2024-02-28
🪵 Log missing citations in the correct place

fwkoch:
This PR #937 restored warnings for inline citations that could not be resolved - However, these warnings were added too early (before resolving these citation labels as normal cross-references) and, it turns out, there already was a warning in the correct place. It just didn't work because the `vfile` wasn't getting passed around correctly.

This PR removes the recently added warning and fixes the existing warning.


rowanc1:
🚀

Issue #951, opened 2024-03-04
🔢 Heading numbering options should apply before heading depth normalization


fwkoch:
## Description

You may enable and disable numbering for specific heading depths in frontmatter, e.g.

```yaml
numbering:
  heading_3: false
```

However, we recently introduced a transform to normalize heading depths (in addition to removing skipped depths, if title is in the frontmatter, this bumps them all down a level). Numbering is applied _after_ this transform is applied, resulting in numbering options applying to a different level than originally intended. 

## Proposed solution

We need to either move the heading normalization later or keep track of the original depths so numbering options can be applied against those.

fwkoch:
We have explicitly decided not to do this. `heading_3` applies to the 3rd level heading of the final MyST output.

The reasoning for this is to make multi-page numbering consistent - `heading_3` defined at a `myst.yml` level should have the same meaning as `heading_3` defined on a page.

PR #953, opened 2024-03-05
🔢 Expand numbering items to include template and start

fwkoch:
This updates numbering frontmatter so each field is an object with `enabled`, `template`, and `start`. `start` sets the first number for the corresponding kind and `template` sets the text when the item is referenced inline.
```yaml
numbering:
  figure:
    start: 3
    template: Fig. %s
```

This is backwards compatible with the previous numbering schema for enabling/disabling numbering, e.g. 
```yaml
numbering: true
```
and
```yaml
numbering:
  headings: true
  figure: false
```
are still valid.

More details in the docs updates in this PR!

### DIFF
--- a/packages/myst-transforms/src/enumerate.ts
+++ b/packages/myst-transforms/src/enumerate.ts
@@ -444,10 +444,6 @@ export class ReferenceState implements IReferenceStateResolver {
         target.node,
         copyNode(target.node as Heading).children as PhrasingContent[],
       );
-    } else if (target.kind === TargetKind.equation) {
-      // Equations are always numbered
-      const template = getReferenceTemplate(target, this.numbering, true, false);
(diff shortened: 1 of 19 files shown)


fwkoch:
I made a few more fixes to this so details with heading numberings work as expected. However, if you do anything with alternative numberings this bug is super obvious: #951 - need to address that asap.

packages/myst-transforms/src/enumerate.ts
```diff
@@ -247,6 +247,50 @@ export function formatHeadingEnumerator(counts: (number | null)[], prefix?: stri
   return out;
 }
 
+export function initializeTargetCounts(
```
stevejpurves, line 250: 👍 on centralizing this, sets us up for multi-article handling too

stevejpurves (approved):

PR #954, opened 2024-03-05
🎆 Expose max size for webp conversion

fwkoch:
By default we do not convert large images to webp so build time may be shorter. However, there are cases where we may want to convert all images, regardless of size (e.g. one-time publish actions). This PR exposes that functionality to pass in `maxSizeWebp` as an option to `buildSite`


stevejpurves (approved):
🚀

PR #956, opened 2024-03-05
ENH: add new sphinx-external-toc package

agoose77:
This PR adds a new `sphinx-external-toc` package. The intention is to replicate the functionality of the `sphinx-external-toc` Python package, for later plumbing into MyST-CLI and JupyterLab-MyST.

This PR:
- Adds support for upgrading legacy ToCs
- Adds support for book, article, and no-format ToCs
- Implements ToC validation using JSON Schema

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,8 @@
     "npm": ">=7.0.0",
     "node": ">=14.0.0"
   },
-  "packageManager": "npm@8.10.0"
+  "packageManager": "npm@8.10.0",
+  "dependencies": {
+    "json2ts": "^0.0.7"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -34,8 +34,5 @@
     "npm": ">=7.0.0",
     "node": ">=14.0.0"
   },
-  "packageManager": "npm@8.10.0",
-  "dependencies": {
-    "json2ts": "^0.0.7"
-  }
+  "packageManager": "npm@8.10.0"
 }
--- a/packages/sphinx-external-toc/src/toc.ts
+++ b/packages/sphinx-external-toc/src/toc.ts
@@ -4,6 +4,7 @@ import schema from './schema.json';
 import _Ajv from 'ajv';
 
 // Adjust types for ES module
+// @ts-ignore
 const Ajv = _Ajv as unknown as typeof _Ajv.default;
 
 // See https://executablebooks.org/en/latest/blog/2021-06-18-update-toc/
--- a/packages/sphinx-external-toc/src/toc.ts
+++ b/packages/sphinx-external-toc/src/toc.ts
@@ -1,7 +1,10 @@
 import yaml from 'js-yaml';
 import type { TOC, ArticleTOC, BookTOC, BasicTOC } from './types.js';
 import schema from './schema.json';
-import Ajv from 'ajv';
+import _Ajv from 'ajv';
+
+// Adjust types for ES module
+const Ajv = _Ajv as unknown as typeof _Ajv.default;
 
 // See https://executablebooks.org/en/latest/blog/2021-06-18-update-toc/
 // Implementation transpiled from https://github.com/executablebooks/sphinx-external-toc/blob/v1.0.1/sphinx_external_toc/tools.py#L277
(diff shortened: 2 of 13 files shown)


packages/sphinx-external-toc/src/toc.ts
```diff
@@ -0,0 +1,168 @@
...
+  if (Array.isArray(toc)) {
+    toc = upgradeOldJupyterBookTOC(toc);
+    console.log('upgraded', JSON.stringify(toc));
+    didUpgrade = true;
+  }
+
+  const ajv = new Ajv.default();
+  const validate = ajv.compile(schema);
+  if (!validate(toc)) {
+    throw new Error(
```
fwkoch, line 150: Worth adding `validate.errors` to this message? I think you were saying those errors were sometimes totally undecipherable because of all the complexity with unions, so probably not, I suppose.

packages/sphinx-external-toc/src/toc.ts
```diff
@@ -0,0 +1,168 @@
...
+  // Identify whether we have "sections" or "chapters" in our first key
+  // And pull them into "firstItems" if so
+  for (const key of ['sections', 'chapters']) {
+    if (key in tocUpdated) {
+      topItemsKey = key;
+      const items = tocUpdated[key] as Record<string, unknown>[];
+      delete tocUpdated[key];
+
+      if (!Array.isArray(items)) {
+        throw new Error();
```
fwkoch, line 27:
```suggestion
        throw new Error(`'${key}' in toc must be an array`);
```

package.json
```diff
@@ -33,5 +33,8 @@
     "npm": ">=7.0.0",
     "node": ">=14.0.0"
   },
-  "packageManager": "npm@8.10.0"
+  "packageManager": "npm@8.10.0",
+  "dependencies": {
+    "json2ts": "^0.0.7"
```
fwkoch, line 38: Should this be hanging around here?

fwkoch (approved):
Looking really nice! Didn't totally parse all the coercion logic, but it looks sensible. And the JSON validation from types is good - easy!

packages/sphinx-external-toc/src/toc.ts
agoose77: Fab, this one was left as a TODO :facepalm:

packages/sphinx-external-toc/src/toc.ts
agoose77: I'm not sure -- as you note, I found that the errors are fairly non-intuitive. I think for now it might be sufficient just to complain about an invalid TOC.

package.json
agoose77: It's mildly frustrating to me that the `npm add` command doesn't respect workspaces by default :shrug:

fwkoch:
I was just wondering if this should be a separate package on executablebooks, since it isn't really MyST related...?

agoose77:
That's a good question. I think for now we keep it here because it will exclusively be used by myst-cli. Soon it will end up in jupyterlab-myst / jupyter-book, and at that time we can pull it out once the organisational structure is clearer?

rowanc1:
@agoose77 looks like test are failing here. Are you working on that to get this ready to review?

agoose77:
> agoose77 looks like test are failing here. Are you working on that to get this ready to review?

Typical! I just rebased it, so I'll check out what's regressed. I *think* it's to do with AJV. c.f. https://github.com/ajv-validator/ajv/pull/2389

rowanc1:
I understand the TOC round-tripping issues a bit better. Thanks for the call @agoose77.

Plan we discussed:
- merge the first version of `sphinx-external-toc` when tests pass here (@agoose77 can advise)
- second PR (#957), which uses it and replaces the current reader is breaking because we are (potentially) round-tripping incorrectly. The current implementation is not as strict as the spec, and the writing can be wrong. We can either leave that for a bit, or have a fallback to the current implementation and a warning.
- We should have a simpler format for internally reasoning about tocs (probably unist based!), that requires more though, especially about numbering. That thinking/proposal can happen any time, maybe @agoose77 can have a go.

agoose77:
This is no longer necessary, as we are using our own TOC

PR #961, opened 2024-03-07
💬 Add `epigraph`, `pull-quote` and `blockquote` directive

agoose77:
The intentions of this PR are twofold:
1. Provide a learning curve for authoring directives in MyST
2. Implement support for epigraphs and pull-quotes

I'll try and document my thought processes as we go.

The merge order should be
1. [x] https://github.com/executablebooks/myst-spec/pull/64
2. [ ] https://github.com/executablebooks/mystmd/pull/961
3. [ ] https://github.com/executablebooks/myst-theme/pull/328


agoose77:
Here's what I wrote in Discord about my initial thoughts:

> I'm looking to implement support for epigraphs and pull-quotes. Adding a new quote directive (pull-quote) is fairly easy, but adding a new `epigraph` directive throws up some questions for me:
> 
> 1. We use transforms to apply custom syntax rules, e.g. GFM admonitions (blockquote -> admonition). Is this "safe" w.r.t third-party mdast utilities? i.e., if a third-party tool expects the non-MyST mdast, does the presence of our nodes in the tree in unexpected places cause issues? I assume that perhaps MDAST is built such that one _can_ extend it, so these tools should ignore unknown content?
> 2. For nodes with well-constrained children, should we lift the child into an attribute of the parent? e.g. for `epigraph`, there's an `attribution`. Sphinx seems to leave that as a node-type, added to children. Should we think about doing the same, or create a new `epigraph` MDAST node that has an `attribution` field? My feeling is strongly the former; we want to avoid duplicating existing MDAST nodes in favour of "extending" them through children. I also assume we shouldn't generally add attributes to existing MDAST nodes defined in https://github.com/syntax-tree/

agoose77:
I then noticed some prior art for epigraphs, [in `tex-to-myst`](https://github.com/executablebooks/mystmd/blob/76c4d7b961579e2d1a3464360424b977f062cebe/packages/tex-to-myst/tests/quote.yml#L34). So, it looks like we're re-using the `caption` node instead of `attribution` node. Hmm, OK, let's try that (see e3e8f30)

agoose77:
It looks like this falls foul of another transform, https://github.com/executablebooks/mystmd/blob/76c4d7b961579e2d1a3464360424b977f062cebe/packages/myst-transforms/src/containers.ts#L182, which seems to assume a closed set of container node types. It looks like the `tex-to-myst` epigraph conversion isn't anticipated here. I wonder why, I assume that this transform would apply after tex-to-myst (e.g. via a `raw`) directive.

agoose77:
The next step from here probably depends upon whether we want `blockquote` et al. to support attribution (captions). If we do (which feels like a commonmark regression), we can just make epigraph a synonym of blockquote, and implement the captioning using a transform. 

If not, then we need to adjust the container transform.

agoose77:
I think the simplest solution is to make all quotes support captions. Now, pull-quotes and epigraphs can be styled separately should we wish for them to be.

rowanc1 (changes requested):
Looking good. Have some comments about the change in functionality (I think we should keep it!), adding more tests, changesets, and some files that have been changed and probably don't need to be anymore (package.json, spec-ext).

On the changed functionality: This might have been hashed out in discord/convos, but I am a fan of the list item transform which this PR is planning to remove. One reason I like it is that the attribution doesn't require a new line, and in both cases we are interpreting the final markup using this transform, in one case to turn a paragraph into an attribution caption and in the other to turn a single list item into an attribution. I don't really see the difference, and it is nice that we can do this:

```
> We know what we are, but know not what we may be.
> - Hamlet act 4, Scene 5
```

Rather than:

```
> We know what we are, but know not what we may be.
>
> -- Hamlet act 4, Scene 5
```

I foresee challenges when people forget that there needs to be an extra new line, which means there is only one paragraph. Some of that can be addressed potentially if we take over the body content in the parsing of the epigraph/pull-quote directives (where we are explicitly expecting an attribution, and can change the parsing).

Regardless of the choice, this PR should update the docs here:

https://github.com/executablebooks/mystjs/blob/main/docs/typography.md#L112

## Testing
I would also love to see a few more tests around the regexp, expanding it to endash, I think there is a bug when it isn't the start of the line (e.g. `blah --- attribtuion`), and ensuring we return false if there is no actual attribution. It would also be good to add some tests for the directive parsing specifically.

There should also be some e2e test (or at least exploratory conformation) that this doesn't clash with the smart typography, which turns `--` into `–` (endash), in which case, expanding the regexp into `---` `\u2014-` meaning three dashes will always fail currently.

The directive docs/references should also be added to include the two new directives:

https://github.com/executablebooks/mystjs/blob/main/docs/directives.md#L2

Before merge, should add changesets!

rowanc1:
Somewhere along the way we have lost adding the class `quote` on the `figure[kind=quote]` node. Adding this back in makes the CSS work again:

image

I think this actually might be on the theme and putting the `figure.kind` in at the start of the classes. I will track it down.

package-lock.json
```diff
@@ -14612,6 +14612,11 @@
         "@types/katex": "^0.14.0"
       }
     },
+    "packages/myst-unist-types": {
+      "version": "0.0.0",
+      "extraneous": true,
+      "license": "MIT"
+    },
```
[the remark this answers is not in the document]
agoose77: Ugh, I explicitly unstaged this.

packages/myst-spec-ext/src/types.ts
```diff
@@ -242,3 +242,8 @@ export type Output = Node & {
   visibility?: Visibility;
   children?: (FlowContent | ListContent | PhrasingContent)[];
 };
+
+export type Epigraph = Node & {
```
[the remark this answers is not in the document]
agoose77: Yes, actually.

agoose77:
> On the changed functionality: This might have been hashed out in discord/convos, but I am a fan of the list item transform which this PR is planning to remove. One reason I like it is that the attribution doesn't require a new line, and in both cases we are interpreting the final markup using this transform, in one case to turn a paragraph into an attribution caption and in the other to turn a single list item into an attribution. I don't really see the difference, and it is nice that we can do this:

I see the convenience of using `- attribution` syntax for attributions. If it were just a matter of syntax, I'd be easily won in either direction. However, there are three factors that favour the `--` variant for me:[^1]

- Existing implementation: Sphinx, Asciidoctor(? https://github.com/arquillian/continuous-enterprise-development/blob/master/ch01.asciidoc), and others seem to use emdash/`--`. We are currently explicitly deviating from that
- Convention: non-parsed Markdown seems to reflect the common usage in written prose that prefers `--` or emdash `—`
- Fragility: it's much more common (assertion, no data) to see list items in quotes than `--` / `—` as trailing items. We would break these usages. Though, _single_ list items are probably not too common either.

> Some of that can be addressed potentially if we take over the body content in the parsing of the epigraph/pull-quote directives (where we are explicitly expecting an attribution, and can change the parsing).

I had this thought too. We could _also_ introduce an explicit role, e.g.
```markdown
> We know what we are, but know not what we may be.
> {attrib}`Hamlet act 4, Scene 5`
```

I'm not saying that I love that, just that it's a member of the phase space ;)

> Regardless of the choice, this PR should update the docs here:

Yep, TODO.


[^1]: I had an additional point in mind here, but stepped away from the keyboard and promptly forgot it. So, maybe there are four points!

packages/myst-transforms/src/blockquote.ts
```diff
@@ -1,23 +1,86 @@
...
+    }
+    // Otherwise, we create a container, and replace the blockquote with a container
+    // containing the blockquote
+    else {
+      const container = {
+        type: 'container',
+        kind: 'quote',
+        children: [quote],
+      };
+      if (maybeLiftAttribution(container as unknown as Container, quote)) {
```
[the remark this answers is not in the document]
agoose77: Yes

packages/myst-transforms/src/blockquote.ts
```diff
@@ -1,23 +1,86 @@
...
-import { selectAll } from 'unist-util-select';
-import type { GenericParent } from 'myst-common';
+import type { Blockquote, Caption, Container, Paragraph, Text } from 'myst-spec';
+import { selectAll, select, matches } from 'unist-util-select';
+import type { GenericParent, GenericNode } from 'myst-common';
+import { remove } from 'unist-util-remove';
+import { copyNode, liftChildren } from 'myst-common';
+import { visit } from 'unist-util-visit';
 
+const ATTRIBUTION_PATTERN = /(?:---?(?!-)|\u2014) +(.*)/;
```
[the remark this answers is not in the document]
agoose77: Oh, I'd thought that `.match` operates at the start of the line. Interesting!

packages/myst-transforms/src/blockquote.spec.ts
```diff
@@ -7,7 +7,7 @@ describe('Test blockquoteTransform', () => {
     const mdast = u('root', [
       u('blockquote', [
         u('paragraph', [u('text', 'We know what we are, but know not what we may be.')]),
-        u('list', { ordered: false }, [u('listItem', [u('text', 'Hamlet act 4, Scene 5')])]),
+        u('paragraph', [u('text', '-- Hamlet act 4, Scene 5')]),
       ]),
     ]);
     blockquoteTransform(mdast);
```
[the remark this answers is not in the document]
agoose77: I've added tests for all of these now (RE 2, I've added tests for `--`, `---`, and &mdash; )

packages/myst-transforms/src/containers.ts
```diff
@@ -115,6 +115,9 @@ export function containerChildrenTransform(tree: GenericParent, vfile: VFile) {
   const containers = selectAll('container', tree) as GenericParent[];
   // Process in reverse so subfigure processing persists
   containers.reverse().forEach((container) => {
+    if (container.kind === 'quote') {
+      return;
+    }
```
agoose77, line 120:
```suggestion
    // We don't need to process quotes
    if (container.kind === 'quote') {
      return;
    }
```

packages/myst-directives/src/quote.ts
```diff
@@ -0,0 +1,56 @@
+import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
+export const epigraphDirective: DirectiveSpec = {
+  name: 'epigraph',
+  doc: 'Inscriptions, or "epigraphs", provide a short quote or inscription at the beginning of a topic. They are usually pertinent to the subsequent content, either to set the theme, or establish a counter-example.',
+  body: {
+    type: 'myst',
+    doc: 'The body of the epigraph.',
+  },
```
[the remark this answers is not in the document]
agoose77: That's a great idea.

packages/myst-directives/src/quote.ts
```diff
@@ -0,0 +1,78 @@
...
+  run(data: DirectiveData): GenericNode[] {
+    const children: GenericNode[] = [];
+    if (data.body) {
+      children.push(...(data.body as GenericNode[]));
+    }
+
+    const container = {
+      type: 'container',
+      kind: 'quote',
+      class: 'epigraph',
```
rowanc1, line 29: This should pick up the class and id/label from the options!

packages/myst-transforms/src/blockquote.spec.ts
```diff
@@ -17,7 +78,24 @@ describe('Test blockquoteTransform', () => {
...
+          u('caption', [u('paragraph', [u('strong', [u('text', 'Hamlet act 4, Scene 5')])])]),
+        ]),
+      ]),
+    );
+  });
+  test('blockquote with empty attribution is unchanged', async () => {
+    const mdast = u('root', [
+      u('blockquote', [
+        u('paragraph', [u('text', 'We know what we are, but know not what we may be.')]),
+        u('paragraph', [u('text', '-- ')]),
```
rowanc1, line 90: Does this also work?

```suggestion
        u('paragraph', [u('text', '-- '), u('text', '')]),
```

packages/myst-transforms/src/blockquote.ts
```diff
@@ -1,23 +1,92 @@
...
-import { selectAll } from 'unist-util-select';
-import type { GenericParent } from 'myst-common';
+import type { Blockquote, Caption, Container, Paragraph, Text } from 'myst-spec';
+import { selectAll, select, matches } from 'unist-util-select';
+import type { GenericParent, GenericNode } from 'myst-common';
+import { remove } from 'unist-util-remove';
+import { copyNode, liftChildren } from 'myst-common';
+import { visit } from 'unist-util-visit';
 
+const ATTRIBUTION_PATTERN = /^(?:---?(?!-)|\u2014) +(.*)/;
```
rowanc1, line 9:
```suggestion
// Attributions start with 2 or 3 dashes (`--` or `---`) or an emdash (`—`)
const ATTRIBUTION_PATTERN = /^(?:---?(?!-)|\u2014) +(.*)/;
```

packages/myst-transforms/src/blockquote.ts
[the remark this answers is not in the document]
rowanc1: I am finding this difficult to read!

Can we simplify the nesting logic here? The `else` block can (I think) be removed? And maybe  an early return on not lifting the container.

packages/myst-transforms/src/blockquote.spec.ts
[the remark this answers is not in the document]
rowanc1: Thank you!

rowanc1 (changes requested):
I think only major comment left is picking up the options in the directives (and adding a test!).

Before merge:
- [x] We should also finalize/update the docs in `typography.md`.
- [x] changeset!

rowanc1:
On the functionality change, thanks for explaining your thinking - happy on my side to be taken in this direction. :)

We should make an explicit note in the docs updates to remember to add a newline.

Note that updating the docs will take a release, theme release, and then website release to consume the theme (in the live example).

packages/myst-directives/src/quote.ts
rowanc1: There are some other examples in directives of doing this, the label and identifier need to be made here.

packages/myst-transforms/src/blockquote.ts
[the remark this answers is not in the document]
agoose77: I am happy to rewrite this, but I think the current approach is somewhat necessary. Let me write out (in text) the logic as I understand it, and you can give a sense of your perspective:

1. In the `blockquoteTransform`, we have the following possible AST structures
   - A normal blockquote, from `> foo` markup.
   - An explicit `pull-quote` from `:::{pull-quote}` markup, seen as `container[kind=quote, class=pull-quote] > blockquote`
   - An explicit `epigraph` from `:::{pull-quote}` markup, seen as `container[kind=quote, class=epigraph] > blockquote`
   
   These latter two constructs are built by directives, and represent the fact that we can only add a "class" through the parent container. Thus, of the three kinds of quotes, only a subset (two) are pre-lifted into containers.
2. Through a node visitor, we look for blockquotes. If the _parent_ of the quote is a container, then we _know_ that we have either a pull-quote or an epigraph, though we do not make use of this knowledge. Otherwise, we have a normal quote. All three can have attributions; we don't restrict attributions to pull-quotes.
3. If we have a quote that is already a child of a container then, should we find a caption, we want to lift the caption into the parent.
4. If we have a quote that does _not_ have a parent container, we need to create the container and replace the existing quote with the container.

The "lifting" logic is encapsulated in `maybeLiftAttribution`, whilst `blockquoteTransform` handles the responsibility of creating a new container and replacing the quote.

I'm happy to remove the `else` and exploit the `return 'skip'` on L19 - this is a coding style that I've picked up from another project where explicit control flow was preferred. I tend to normally prefer the implicit style myself.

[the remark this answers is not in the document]
agoose77: @rowanc1 also, right now this transform is _not_ recursive, meaning pull-quotes inside pull-quotes won't have their quotes lifted. Do you think it should be?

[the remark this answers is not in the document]
rowanc1: Thanks for the comments and clarifications - we are trying to prefer early returns in the code-base so far. 

I think we can (likely) safely skip recursive quotes, certainly not as common as subfigures, subequations, etc. We can add it in when it becomes a feature request!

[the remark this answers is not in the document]
agoose77: OK, I already added recursion support before I saw your reply :monkey_face:. Hopefully it's not much more complex as a result.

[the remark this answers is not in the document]
rowanc1: Nice!

packages/myst-transforms/src/blockquote.spec.ts
agoose77: That _will_ leave an empty attribution. If we don't want this (and it's possible to produce via a parse), then should we e.g. check whether the `toString` (or whatever function we have) is non-empty?

packages/myst-directives/src/quote.ts
```diff
@@ -32,6 +32,7 @@ export const blockQuoteDirective: DirectiveSpec = {
     const container = {
       type: 'container',
       kind: 'quote',
+      label: data.options?.label as string | undefined,
```
rowanc1, line 35: Can you follow the pattern here:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-directives/src/figure.ts#L105

const { label, identifier } = normalizeLabel(data.options?.label as string | undefined) || {};

agoose77: Oh whoops yes, I was deep in `rg` when I added that particular string!

packages/myst-directives/src/blockquote.ts
```diff
@@ -0,0 +1,50 @@
+import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
+import { normalizeLabel } from 'myst-common';
+import type { Container } from 'myst-spec-ext';
+import classNames from 'classnames';
+
+export const blockQuoteDirective: DirectiveSpec = {
+  name: 'block-quote',
```
rowanc1, line 7: I think this should be `blockquote` (maybe aliased to `block-quote`?).

packages/myst-directives/src/blockquote.ts
```diff
@@ -0,0 +1,50 @@
...
+    const className = data.options?.class as string;
+    const container: Container = {
+      type: 'container',
+      kind: 'quote',
+      label,
+      identifier,
+      class: classNames({ [className]: className, [data.name]: data.name !== 'block-quote' }),
+      children: [
+        {
+          // @ts-expect-error: myst-spec needs updating to support blockquote
```
rowanc1, line 42: Is this handled in the myst-spec updates?

packages/myst-directives/src/blockquote.ts
```diff
@@ -0,0 +1,50 @@
...
+      type: String,
+      doc: `CSS classes to add to your block-quote. Special classes include:
+
+- \`pull-quote\`: used for block-quotes which should attract attention
+- \`epigraph\`: used for block-quotes that are usually found at the beginning of a document`,
+    },
+  },
+  body: {
+    type: 'myst',
+    doc: 'The body of the block-quote.',
```
rowanc1, line 25:
```suggestion
    doc: 'The body of the quote.',
```

as it might be a block/pull/epigraph

packages/myst-transforms/src/blockquote.spec.ts
rowanc1: I think that is fine and we don't have to catch that. I am not sure how that would be written in markdown anyways.

packages/myst-transforms/src/blockquote.ts
```diff
@@ -1,24 +1,115 @@
 import type { Plugin } from 'unified';
 import type { Blockquote, Caption, Container } from 'myst-spec';
-import { selectAll } from 'unist-util-select';
-import type { GenericParent } from 'myst-common';
+import { select, matches, selectAll } from 'unist-util-select';
```
rowanc1, line 3:
```suggestion
import { select, matches } from 'unist-util-select';
```

docs/typography.md
```diff
@@ -111,13 +111,24 @@ See [](#abbreviations) for more information.
...
+> -- Hamlet act 4, Scene 5
 ```
 
+MyST also supports pull-quotes via the `pull-quote` directive, which act as attention-grabbing visual elements.
+```{myst}
+:::{pull-quote}
+We know what we are
+:::
+
+They say the owl was a baker’s daughter. Lord, we know what we are but know not what we may be. God be at your table.
```
rowanc1, line 128: Reminds me that I don't know anything about Shakespeare.

.changeset/seven-roses-cough.md
```diff
@@ -0,0 +1,6 @@
+---
+"myst-directives": minor
+"myst-transforms": minor
+---
+
+Add support for epigraphs and pull-quotes
```
rowanc1, line 6:
```suggestion
Add support for epigraphs and pull-quotes using a new blockquote directive.
```

rowanc1 (commented):
Nice to see the consolidation into one directive. What do you think of naming this blockquote (no dash?). Can alias as well? Maybe I am too used to reading html source.

docs/typography.md
agoose77: I don't think he'll be too offended ;)

packages/myst-directives/src/blockquote.ts
agoose77: No! Actually now I understand where my confusion came from. Let me fix that up.

packages/myst-directives/src/blockquote.ts
agoose77: Hmm. I'm on the fence. I like that it reflects symmetry with `pull-quote`, but of course the internal AST node is `blockquote`. You make the call!

docs/directives.md
```diff
@@ -9,6 +9,9 @@ description: Code and code-blocks can be used to show programming languages.
 :::{myst:directive} bibliography
 :::
 
+:::{myst:directive} block-quote
```
agoose77, line 12:
```suggestion
:::{myst:directive} blockquote
```

agoose77:
@rowanc1 this is ready for handover!

rowanc1 (approved):

PR #970, opened 2024-03-11
📖 Contribute template docs

rowanc1:
@choldgraf following your lead this is some pointers to more tempalte docs and tutorials.

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -63,4 +63,5 @@ parts:
       - file: glossary
   - caption: Contribute
     chapters:
+      - file: contribute-templates
       - file: contribute-docs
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
This is great, I think it could be merged as-is, but if you have a moment it would be helpful to include a short explanation of what `jtex` is somewhere, since the links are taking people to the `jtex` documentation and it might be confusing to them to suddenly be in a new docs space. Either way it LGTM

rowanc1:
Roger that. I think we want to probably be calling this "templates" now or something less tied to latex. The `tex` part of `jtex` is confusing now that we are using the same libraries to render latex and typst using jinja.

Will update the text and merge!

rowanc1:
I went with removing the explicit reference to the (now outdated) library name and just explained that the current docs are aimed at latex templates. These will need to be improved as we add increasing support for typst etc.

PR #971, opened 2024-03-12
🙈 Ignore hidden directories from watch

rowanc1:
Currently all hidden directories are ignored by the project tree, which selects markdown files that are not in hidden folders or node_modules, etc.

However, these are still being watched, and that can lead to large performance hits if there are a lot of files in the hidden directories (e.g. pixi).

cc @roaldarbol

This is the previous output, in the this PR these files don't trigger changes:

image

### DIFF
--- a/.changeset/tidy-buttons-sparkle.md
+++ b/.changeset/tidy-buttons-sparkle.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Ignore hidden directories from watch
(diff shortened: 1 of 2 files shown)


fwkoch (approved):
Nice - these were already ignored in the project build too, so really no need to watch: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/utils/shouldIgnoreFile.ts#L5

roaldarbol:
@fwkoch I think it was only ignoring hidden *files*, not folders - which then contain lots of non-hidden files (see discussion on Discord). :-) Glad to see it merged!

rowanc1:
I think that function @fwkoch linked, it should really be named "`path`", as the "file" can be either folder or the file name. Regardless, this ignores them upstream in the watch, which is important for performance!

roaldarbol:
Ah, sorry - shouldn't be confusing things further then 😅

PR #987, opened 2024-03-14
📖 Contributing a new features to MyST (roles & transforms)

agoose77:
This PR will eventually add a new page to our documentation, that intends to act as a "ten minute guide to adding new features to MyST". It will overlap with the plugins page slightly, but I hope that it will bootstrap some of the contribution questions.


rowanc1 (commented):
Very excited about this - thanks for getting it started @agoose77!

docs/contribute-add-feature.md
```diff
@@ -365,11 +389,15 @@ export const wordCountPlugin: Plugin<[], GenericParent, GenericParent> =
 This pattern, of "filling-in" the children of an existing node using a transform, is commonly used in the MyST ecosystem.
 
 If we compiled MyST with `npm run build`, we'd notice that this transform never runs. Like our earlier `word-count` role, we need to include the `wordCountTransform` in the set of MyST transforms that run during building. We can do this by adding a new _export_ line in `packages/myst-transforms/src/index.ts`:
-```typescript
+
+:::{code-block} typescript
+:filename: packages/myst-transforms/src/index.ts
 export { wordCountTransform, wordCountTransform } from './word-count.js';
-```
+:::
```
rowanc1, line 396: I would suggest the `` ``` `` for anything that is actually code as some markdown formatters will change `_` and `*` aggressively.

agoose77: Could you elaborate? I want to keep the options, are you suggesting I use the fence-syntax with a directive?

agoose77:
@rowanc1 I think this is ready for a review pass. I'm sure it's not perfect, but I'd benefit from a fresh pair of eyes :pray:

docs/contribute-add-feature.md
rowanc1: You can use the options with a fence directive! There are some docs here:

https://mystmd.org/guide/syntax-overview#directives

image

agoose77: Ah cool, I'm familiar with that.

I'll make the change.

docs/contribute-add-feature.md
```diff
@@ -0,0 +1,460 @@
...
+:filename: packages/myst-roles/src/word-count.ts
+
+import type { RoleSpec, RoleData, GenericNode } from 'myst-common';
+
+export const wordCountRole: RoleSpec = {
+  name: 'word-count',
+  body: {
+    type: String,
+    required: true,
+  },
```
LecrisUT, line 78: These are default available fields right? Can you add an example with adding a new role option so that the API is a bit clearer?

docs/contribute-add-feature.md
```diff
@@ -0,0 +1,460 @@
...
+        {
+          type: 'wordCount',
+          value: data.body as string
+        }
+    ];
+  }
+};
+```
+
+Our new role is not-yet ready to be used. We next need to tell MyST that it should be included in the main program. To do this, first we must import the role in `packages/myst-roles/src/index.ts`
```
LecrisUT, line 122:
````suggestion
```

Up to this point, the role simply copies the contents of the body to an `AST` node. The rest of the implementation continues in ...

However, our new role is not-yet included in the main MyST program. To do this, first we must import the role in `packages/myst-roles/src/index.ts`
````

A short summary of what is implemented up to there and what is implemented in the next section.

LecrisUT (commented):
Do you think it's worth adding to this PR how to add a `frontmatter` option, or should it be added in another tutorial? The `frontmatter` option seems quite a bit more convoluted to document.

docs/contribute-add-feature.md
agoose77: I've edited this in favour of adding more explanation to the role

agoose77:
> Do you think it's worth adding to this PR how to add a `frontmatter` option, or should it be added in another tutorial? The `frontmatter` option seems quite a bit more convoluted to document.

Hmm, it's a tricky balance between covering "enough" and "too much". I want to keep this tutorial / guide small enough that someone could feel brave enough to get started. I think we might need more guides to help cover the wider challenges with contribution.

docs/contribute-add-feature.md
LecrisUT: Oh, I guess I'm confusing `directive` with `role`? I guess there are some `directive` that wrap the call of `role` in those cases :thinking:. For another documentation than.

LecrisUT:
> > Do you think it's worth adding to this PR how to add a `frontmatter` option, or should it be added in another tutorial? The `frontmatter` option seems quite a bit more convoluted to document.
> 
> Hmm, it's a tricky balance between covering "enough" and "too much". I want to keep this tutorial / guide small enough that someone could feel brave enough to get started. I think we might need more guides to help cover the wider challenges with contribution.

Indeed, I probably confused `role` with `directive`, so adding those parts to this documentation, would be a bit too much. Maybe the `directive` can be intuited a bit after this guide, but `frontmatter` seems to be more complicated, e.g. my attempts at navigating #979. Worth having an example for that imo.

rowanc1:
I have gone through this and updated a few things!

This is great @agoose77 and I think ready to be merged. I changed it so we transform the `wordCount` node to a text node so that the final picture is complete. There can be another tutorial in the future on how to add a new renderer.

PR #989, opened 2024-03-15
feat: add `keystrokes` role

agoose77:
This behaves like `<kbd>`, but allows us to implement something for LaTeX/Typst by using non-HTML

### DIFF
--- a/packages/myst-roles/src/index.ts
+++ b/packages/myst-roles/src/index.ts
@@ -13,6 +13,7 @@ import { smallcapsRole } from './smallcaps.js';
 import { subscriptRole } from './subscript.js';
 import { superscriptRole } from './superscript.js';
 import { underlineRole } from './underline.js';
+import { keystrokesRole } from './keystrokes.js';
 
 export const defaultRoles = [
   abbreviationRole,
@@ -30,6 +31,7 @@ export const defaultRoles = [
   subscriptRole,
   superscriptRole,
   underlineRole,
+  keystrokesRole,
 ];
 export { abbreviationRole } from './abbreviation.js';
 export { chemRole } from './chem.js';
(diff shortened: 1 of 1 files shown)


agoose77:
Should this be `keyboard` instead of `kbd`? I also note that there is an mdast package that creates a `kbd` node with multiple children, but my feeling is that we're not standardising against that, so it's fine to deviate.

agoose77:
Hmm, I wonder whether we should think about the ability to nest `kbd` and whether to support `samp`:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd#usage_notes

rowanc1:
I would prefer `keyboard` is there any precedence for the naming in unist? Keystrokes is the act of tracking what you are typing.

rowanc1 (commented):
Nice - love this. Other than the name choice (I think it is `keyboard`) this looks good to go -- easier than epigraphs! :)

agoose77:
I'm glad you picked up on the name. I had similar thoughts myself, but figured someone would point it out if it wasn't just me!

Do you have any feeling about nesting `<kbd>` elements, or supporting `<samp>`, as is described in the link above?

rowanc1:
I think that we should start simple and start with this PR. Potentially down the road we introduce new directives that are `command` or `menu` which are `kbd > kbd` and `kbd > samp` as well as some parsing logic to pull out Ctrl/shift (or even Mod, and change that based on the users platform in the renderer, etc.).

What that does mean is that `keyboard` type should be a parent, not a leaf:

```
- type: keyboard
  children:
     - type: text
       value: Ctrl + S
```

That allows us to expand in the future.

Can you add a test similar to:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-parser/tests/roles/subscript.yml#L1

.changeset/tasty-ducks-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"myst-roles": minor
+---
+
+Add `kbd`/`keystrokes` role
```
rowanc1, line 5:
```suggestion
Add `kbd`/`keyboard` role
```

PR #1010, opened 2024-03-20
:paintbrush: add `sidebar` directive

agoose77:
We already have a `margin` directive, this PR adds support for `sidebar`. I don't know whether we want to advertise this directive for core MyST (it might be better to stick to a reduced set of directives to allow us more control), but JB will certainly need it.

c.f. https://github.com/executablebooks/myst-theme/pull/341

### DIFF
--- a/.changeset/brave-comics-confess.md
+++ b/.changeset/brave-comics-confess.md
@@ -1,0 +1,5 @@
+---
+"myst-directives": minor
+---
+
+Add sidebar directive
--- a/docs/directives.md
+++ b/docs/directives.md
@@ -48,6 +48,9 @@ description: Code and code-blocks can be used to show programming languages.
 :::{myst:directive} margin
 :::
 
+:::{myst:directive} sidebar
+:::
+
 :::{myst:directive} math
 :::
 
--- a/packages/myst-directives/src/index.ts
+++ b/packages/myst-directives/src/index.ts
@@ -9,6 +9,7 @@ import { imageDirective } from './image.js';
 import { includeDirective } from './include.js';
 import { tableDirective, listTableDirective } from './table.js';
 import { marginDirective } from './margin.js';
+import { sidebarDirective } from './sidebar.js';
 import { glossaryDirective } from './glossary.js';
 import { mathDirective } from './math.js';
 import { mdastDirective } from './mdast.js';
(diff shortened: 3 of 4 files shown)


rowanc1:
Any possibility that this is an alias for margin? Or can we document what the differences are?

agoose77:
This is in a weird space, and I'd benefit from some input I think.

I don't think we have to implement every directive from JB in MyST if we feel that e.g. a directive was a misstep. However, we probably do need to try and provide compatibility for all of the JB features where possible. I think `sidebar` _might_ be one such case, but I'm not sure. 

If we're not just kicking the can down the road (which this PR does as-is, by implementing but not documenting `sidebar`), then it's a lot easier to reason about grid layouts than floating layouts. So, in the interest of robustness and simplicity, we drop `sidebar` in MyST (and in compatibility mode, it's just an `alias`).

@choldgraf thoughts welcome too.

rowanc1:
Ok - looking through the [jupyterbook docs](https://jupyterbook.org/en/stable/content/layout.html), I think these are just divs with classes from the AST perspective, that is: add new directive, but let's not add a new AST node though (and also remove the margin node which isn't working now anyways and is treated as a div). From the AST perspective this is content with style as a placement, not a new type of content node. (alternatively these are types of `asides` which has different semantic structure and classes?) Thoughts?

We also need to make some changes in the blocks in the theme to allow floating content, when there is a grid, you can't use float and we need to for the sidebar (and the margin). I think we still want to use the grid for placement when it is actually in the margin (for alignment etc. across different themes) and then transition to a right-floating div. I will review what you have already in the theme PR @agoose77.

agoose77:
> Ok - looking through the [jupyterbook docs](https://jupyterbook.org/en/stable/content/layout.html), I think these are just divs with classes from the AST perspective, that is: add new directive, but let's not add a new AST node though (and also remove the margin node which isn't working now anyways and is treated as a div). From the AST perspective this is content with style as a placement, not a new type of content node. (alternatively these are types of asides which has different semantic structure and classes?) Thoughts?

Oh, I really like that idea. My feeling is that an aside is semantically important, but I agree that we should probably lean into the `aside`-ness, rather than `margin`/`sidebar`. I'll implement it here, and rework https://github.com/executablebooks/myst-theme/pull/341

agoose77:
Replaced by https://github.com/executablebooks/mystmd/pull/1012

PR #1012, opened 2024-03-20
🖌 Add `aside` directive

agoose77:
This PR replaces #1010 with an `aside` directive, renaming `margin`, e.g.
```markdown
:::{aside}
This content is an aside / margin note
:::
```
There is also the `margin` alias, and `sidebar` which has a different `kind`.


docs/asides.md
```diff
@@ -0,0 +1,16 @@
...
+
+Asides provide an easy way to represent content that is only indirectly related to the article's main content. Where supported, MyST will attempt to display an `aside` _as close as possible_ but separately from the main article, such as in the side-margin. 
+
+````{myst}
+The main article contains main-article text!
+
+```{aside}
+This is an aside. It is not entirely relevant to the main article.
+```
+````
```
rowanc1, line 16: This one might be easier to start from the jupyter-books docs on this and demonstrate it in the page rather than in a demo component?

agoose77: I think an `aside` should be a MyST feature, but I would not be surprised if `aside` isn't very clear in the demo component. We can enable a grid layout on the MyST demo component if needs be.

Welcome your thoughts here if you think it would be better not to use the `myst` directive and instead include the `aside` in the main guide body. But, I think we can fix-up the demo component to make it possible to see, probably.

packages/myst-parser/tests/directives/aside.yml
```diff
@@ -0,0 +1,90 @@
+title: aside directive
+cases:
```
rowanc1, line 2: Thank you for the tests!

.changeset/brave-comics-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"myst-directives": minor
+---
+
+Add `aside` directive
```
rowanc1, line 5: 🚀

packages/myst-spec-ext/src/types.ts
```diff
@@ -242,3 +242,9 @@ export type Output = Node & {
   visibility?: Visibility;
   children?: (FlowContent | ListContent | PhrasingContent)[];
 };
+
+export type Aside = Node & {
+  type: 'aside';
+  kind?: 'sidebar';
```
rowanc1, line 248: missing margin!

PR #1013, opened 2024-03-20
fix: throw error if `typst` is not available

### LINKED ISSUE #962: Typst export fails silently if typst binary is not available
As the title says!

agoose77:
Closes #962 

I'm not sure if we have an internal policy for error handling. It seems fine to throw an error, but I'd welcome any feedback if we prefer log-and-return.

### DIFF
--- a/packages/myst-cli/src/build/typst.ts
+++ b/packages/myst-cli/src/build/typst.ts
@@ -57,6 +57,7 @@ export function isTypstAvailable() {
 
 export async function runTypstExecutable(session: ISession, typstFile: string) {
   if (!isTypstAvailable()) {
+    throw new Error(`typst CLI must be installed to build PDFs from typst`);
     session.log.debug('typst CLI must be installed to build PDFs from typst');
     return;
   }


packages/myst-cli/src/build/typst.ts
```diff
@@ -57,6 +57,7 @@ export function isTypstAvailable() {
 
 export async function runTypstExecutable(session: ISession, typstFile: string) {
   if (!isTypstAvailable()) {
+    throw new Error(`typst CLI must be installed to build PDFs from typst`);
     session.log.debug('typst CLI must be installed to build PDFs from typst');
```
rowanc1, line 61: I don't think we want to throw here because this is called within an export chain that may have other exports. @fwkoch might have other thoughts on if that function catches and will work gracefully?

I think just changing this to an error should be enough to get it to the console, and that should be enough to solve the near-term issue?

```suggestion
    session.log.error('The typst CLI must be installed to build PDFs with typst');
```

If that works, adding a changeset and getting this in is great. 🚀

PR #1020, opened 2024-03-21
📖 Embedding documentation

choldgraf:
This is a brief overview of "embedding" things with MyST. It's probably not complete, but I found it hard to discover this content and the sensible behavior so thought I'd write up my understanding of it right now.

A few questions:

1. I couldn't figure out how to embed content from a different page. Is that possible?
2. I wasn't sure whether it was best to use a dedicated page (so it's more discoverable) or use a sub-section of the "cross references" page, but this seemed to be the simplest approach. Is that OK?

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -24,6 +24,7 @@ parts:
       - file: blocks
       - file: diagrams
       - file: asides
+      - file: embed
       - file: dropdowns-cards-and-tabs
       - file: glossaries-and-terms
   - caption: Executable Content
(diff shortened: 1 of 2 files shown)


rowanc1:
You can use the same label syntax to pull in from any page. Not sure what didn't work?

choldgraf:
Huh weird, maybe I encountered this when I was trying to use the `{myst}` directive...that seemed to behave weirdly when it came to defining and referencing labels

choldgraf:
yeah I implicitly figured that one out when I was trying to use it in this PR haha, that's why I just ended up copy/pasting code blocks instead of using the directive. Is it documented anywhere?

PR #1022, opened 2024-03-22
:hammer: Add support for legacy glossary syntax

### LINKED ISSUE #618: Compatibility with RST Glossaries
We should add compatibility with Sphinx/RST glossaries (which are based around indentation). At the very least the compatibility warning should be improved with a link to how to do this in Sphinx.

See https://github.com/executablebooks/MyST-NB/issues/544#issuecomment-1734574664

agoose77:
Fixes #618 through an AST-level transformation. 

The Sphinx `{glossary}` node has the following ReST definition-list-based syntax
````
```{glossary}
term
  definition
  
  additional context

next-term
  next-definition
```
````

Whereas we require a Markdown definition list:
````
```{glossary}
term
:  definition
  
   additional context

next-term
:  next-definition
```
````

Whilst we have MyST Spec to define the nature of MyST, we have places where the spec is ill-defined. Glossaries are not yet mentioned in the set of AST directives, nor are definition lists. This PR allows users of the glossary transform to opt-in and upgrade them to deflists. I wonder whether this should be a separate transform that JupyterBook uses, rather than an option?

It's not foolproof to handle this at the AST level. My rationale is that this is a "backwards compatibility" feature, and so we expect that the input is already valid ReST.

### DIFF
--- a/packages/myst-transforms/src/glossary.ts
+++ b/packages/myst-transforms/src/glossary.ts
@@ -1,13 +1,86 @@
 import type { Plugin } from 'unified';
 import type { Node } from 'myst-spec';
 import type { VFile } from 'vfile';
-import { selectAll } from 'unist-util-select';
+import { selectAll, select } from 'unist-util-select';
 import type { GenericParent } from 'myst-common';
-import { normalizeLabel, toText, fileError, RuleId } from 'myst-common';
+import { normalizeLabel, toText, fileError, fileWarn, RuleId } from 'myst-common';
+import type { DefinitionTerm, DefinitionDescription, DefinitionList } from 'myst-spec-ext';
 
-export function glossaryTransform<T extends Node | GenericParent>(mdast: T, file: VFile) {
+export type Options = {
+  upgradeLegacySyntax?: boolean;
+};
+
+/**
+ * Lift out definitionTerms from top-level `TERM\nDEFINITION` text nodes
+ */
+function maybeLiftLegacyDefinitionTerm(node: GenericParent): GenericParent | undefined {
+  if (node.type !== 'paragraph') {
+    return undefined;
+  }
+  const firstChild = node.children[0];
+  if (firstChild?.type !== 'text') {
+    return undefined;
+  }
+  const value = firstChild.value as string;
+  const index = value.indexOf('\n');
+  if (index === -1) {
+    return undefined;
+  }
+  const term = value.slice(0, index);
+  const remainder = value.slice(index + 1, value.length);
+  firstChild.value = remainder;
+  return {
+    type: 'definitionTerm',
+    children: [{ type: 'text', value: term }],
+  };
+}
+
+export function glossaryTransform<T extends Node | GenericParent>(
+  mdast: T,
+  file: VFile,
+  opts: Options,
+) {
   const glossaries = selectAll('glossary', mdast) as GenericParent[];
   glossaries.forEach((glossary) => {
+    // Do we have a ReST-like glossary? If so, pull it out.
+    if (opts.upgradeLegacySyntax && !select('definitionList', glossary)) {
+      fileWarn(
+        file,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: packages/myst-transforms/src/glossary.ts [hunk 1]
```diff
@@ -1,13 +1,86 @@
...
 
-export function glossaryTransform<T extends Node | GenericParent>(mdast: T, file: VFile) {
+export type Options = {
+  upgradeLegacySyntax?: boolean;
+};
+
+/**
+ * Lift out definitionTerms from top-level `TERM\nDEFINITION` text nodes
+ */
+function maybeLiftLegacyDefinitionTerm(node: GenericParent): GenericParent | undefined {
```
agoose77, line 16: I know this pattern was confusing in the epigraphs PR.

It either identifiers `paragraph > text[value*=\n]` nodes (and then strips out the TERM, returning the definitionTerm node) or does not, acting then as a noop (returning undefined).

packages/myst-transforms/src/glossary.ts
```diff
@@ -1,13 +1,86 @@
...
+        type: 'definitionList',
+        children: [],
+      } as DefinitionList;
+
+      let termChildren: DefinitionDescription['children'] = [];
+
+      // Build up definitionList as alternating term-definition
+      for (const child of glossary.children) {
+        let definitionTerm: GenericParent | undefined;
+        if ((definitionTerm = maybeLiftLegacyDefinitionTerm(child as unknown as GenericParent))) {
```
agoose77, line 67: We could also just return the string of the term, and create the term node here.

packages/myst-transforms/src/glossary.ts
agoose77: It would be possible to write this operation as:
1. Lift `paragraph > text[value*=\n]` into [definitionTerm, paragraph]`
2. Lower `:not(definitionTerm)` into `definitionDescription` children

I think I prefer this but would be happy to rewrite if necessary :)

packages/myst-transforms/src/glossary.ts
```diff
@@ -1,13 +1,86 @@
...
+          // Push the term
+          definitionList.children.push(definitionTerm as DefinitionTerm);
+          // Push the yet-unfilled description
+          definitionList.children.push({
+            type: 'definitionDescription',
+            children: termChildren,
+          });
+        }
+        // Populate the current definition's children
+        termChildren.push(child);
```
agoose77, line 79: Note that this loses all children if we don't have a single term. I think that's fine - it wouldn't be a valid list.

rowanc1:
I have added and documented a project setting `legacy_glossary_syntax: true` so that this is accessible through the CLI. Using that flag and parsing an existing glossary like JupyterBook fails with no error message:

https://jupyterbook.org/en/stable/_sources/reference/glossary.md

It looks like they use 4 spaces (not sure if it matters), multiple line breaks, and multiple terms to a definition.
I think that the test cases need to be a bit more robust to these cases, possibly starting from the raw markdown rather than the mdast?

To one of your other comments @agoose77:
If a glossary can't get upgraded, I think we should _warn_ and revert to the original parsed content rather than just drop the internals of the glossary directive.

rowanc1 (changes requested):
This needs to be able to at least parse this page's source:

https://jupyterbook.org/en/stable/reference/glossary.html

Currently that isn't working. We might have to move this back from a transform to the base-level directive? Or switch out the directive parser that we use, and parse the markdown in chunks?

agoose77:
@rowanc1 yes, upon looking at this further, it looks like we'll need to do a proper line-by-line parse (which requires re-entrant MyST parsing). c.f. https://github.com/sphinx-doc/sphinx/blob/413e740dcadc342b34e9e0889fdebf7e2b01071f/sphinx/domains/std/__init__.py#L314

agoose77:
I'll therefore close this PR, as it will require re-writing with a different tack.

PR #1023, opened 2024-03-22
:books: Use title case in short titles

agoose77:


rowanc1:
Happy for an eager merge here. Small changes to titles look good!

agoose77:
Am out of office for a bit - summary, this PR uses & in place of 'and', and moves to titlecase!

PR #1024, opened 2024-03-22
📖 More instructions to embed nodes across documents

choldgraf:
This adds a few more instructions to embed:

- An example of cross-page embedding (now that @rowanc1 made me realize why it wasn't working before because `{myst}` doesn't track labels)
- An example of embedding via the `{figure}` directive (we should keep a running list of any directive that allows for embedding-like behavior.


rowanc1:
I added a link to an interactive figure, which I think shows off the power of this really well.

image

PR #1030, opened 2024-03-25
✨ Add `csv-table` directive

### LINKED ISSUE #893: Support for CSV table directive
Currently we only support `table` and `list-table`.

Sphinx Documentation:
- https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table, also [here](https://myst-parser.readthedocs.io/en/latest/syntax/tables.html#csv-tables).

agoose77:
Implements https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table-1, addressing need on #189 

- [ ] Support URL / file?
- [x] Support headings?
- [x] Handle invalid markup

See #893.

### DIFF
--- a/packages/myst-directives/src/table.ts
+++ b/packages/myst-directives/src/table.ts
@@ -1,6 +1,7 @@
-import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
+import type { DirectiveSpec, DirectiveData, DirectiveContext, GenericNode } from 'myst-common';
 import { fileError, normalizeLabel, RuleId } from 'myst-common';
 import type { VFile } from 'vfile';
+import { parse } from 'csv-parse/sync';
 
 export const tableDirective: DirectiveSpec = {
   name: 'table',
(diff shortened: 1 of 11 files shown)


packages/myst-transforms/src/table.ts
```diff
@@ -0,0 +1,85 @@
...
+      const container = {
+        type: 'container',
+        kind: 'table',
+        identifier: node.identifier,
+        label: node.label,
+        class: node.class,
+        children: [...node.children, table],
+      };
+      (node as GenericNode).type = '__lift__';
+      (node as GenericNode).children = [container];
```
agoose77, line 76: I've observed from @rowanc1's counter-edits to my PRs that we generally try to avoid replacing nodes in favour of nesting them. In my mind, this `csv-table` node contains unparsed markup and so is a bit unusual. If we want to keep the `csv-table` node, I'll need to figure out how to store the caption.

agoose77:
@rowanc1 this PR will probably be easier to audit #1091

rowanc1:
@agoose77 there needs to be some fixing of the git history now that #1091 is in! In this case, it is just the last commit.

agoose77:
@rowanc1 I've held off adding file/URL support for this pass - I don't think it should block this PR. I'll discuss that on Slack!

The only outstanding challenge is where to add tests ... I'm somewhat happy to skip tests here, as I think those are more end-to-end features that we might need to do another pass of down the road. That might also be a lazy developer's hat.

packages/myst-directives/src/table.ts
```diff
@@ -162,3 +170,137 @@ export const listTableDirective: DirectiveSpec = {
...
+          type: 'tableRow',
+          children: parsedRow.map((parsedCell) => ({
+            type: 'tableCell',
+            header: true,
+            children: parsedCell.children,
+          })),
+        })),
+      );
+    }
+    const cells = parseCSV(data.body as string, data.options, ctx);
```
rowanc1, line 272: Probably a try catch and errorFile on this if the csv is malformed!

rowanc1 (commented):
It would be great to have a test on the directive, and the error state of bad CSV. I can add that when I test if I get there first!

agoose77:
I've made the change to use `fileError`, but I want to dig in to examples of `fileError` that seem to pull the nodes for the options, rather than the directive itself, from the pre-AST. Not quite au-fait with that logic yet, but this is an example:
```typescript
select('mystDirectiveOption[name="tags"]', data.node) ?? data.node
```

rowanc1:
Providing a node with the option reports the correct line number of the option in the error, rather than the directive itself. It falls back to the directive in that case if it can't find the specific option. When we get column numbers and language server support, these will be correctly-placed red-underlines. :)

agoose77:
Oh yes, I'm following the use-case! What I meant was, I wasn't sure of the AST layout of these raw nodes. But, I think I've found it! I'll revise the change.

PR #1032, opened 2024-03-25
🧮 Fix math-spacing in paragraphs

rowanc1:
See #787


packages/myst-parser/src/tokensToMyst.ts
```diff
@@ -9,6 +9,22 @@ import { u } from 'unist-builder';
 import { MarkdownParseState, withoutTrailingNewline } from './fromMarkdown.js';
 import type { MdastOptions, TokenHandlerSpec } from './types.js';
 
+export function computeAmsmathTightness(
```
fwkoch, line 12: I think this is very similar to the other "compute tightness" function, but it's a little hard to be certain since the function is so complex. Can we just reuse that function and have the `+1` shift (with comment) in the ams math handler...?

packages/myst-transforms/src/math.ts
```diff
@@ -241,6 +241,30 @@ export function mathNestingTransform(
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   file: VFile,
 ) {
+  const paragraphs = selectAll('paragraph', tree) as GenericParent[];
+  paragraphs.forEach((paragraph) => {
+    if (paragraph.children.length === 1) return; // no need to traverse, the math node can never be tight!
+    paragraph.children.forEach((child, index) => {
```
fwkoch, line 247: So for this, math inside a paragraph is _always_ tight? Is there some redundancy with this and the parse-level logic? Like, if we get to this point, the node probably already has correct `tight` value since in the source the lines above/below aren't empty... but _maybe_ not and this cleans it up...?

packages/myst-directives/src/math.ts
```diff
@@ -15,13 +15,13 @@ export const mathDirective: DirectiveSpec = {
...
-        value: data.body as string,
-      },
-    ];
+    const math = {
+      type: 'math',
+      identifier,
+      label,
+      value: data.body as string,
+    } as GenericNode;
+    if (data.node.tight) math.tight = data.node.tight;
```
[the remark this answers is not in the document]
fwkoch: Should we also take `tight` as a directive argument?

packages/markdown-it-myst/src/directives.ts
```diff
@@ -51,6 +65,7 @@ function runDirectives(state: StateCore): boolean {
         directiveOpen.meta = {
           arg,
           options: simplifyDirectiveOptions(options),
+          tight: computeBlockTightness(state.src, token.map),
```
fwkoch, line 68: It feels slightly extreme to be passing the entire source content around to compute this... it seems like it _could_ be possible to do this by inspecting the previous/next tokens in `state.tokens` to look for blank lines? But that would probably require some dicey token stream logic... easier for now to just do things explicitly based on the content.

fwkoch (approved):
Looking ok - only a couple suggestions to tidy things up. Most of my comments just trying to get a full understanding of what's going on.

packages/myst-transforms/src/math.ts
fwkoch: Oh, this is to handle `dollar math` specifically. (AMS and directive math are handled at parse, dollar math is handled here) 👍

packages/myst-directives/src/math.ts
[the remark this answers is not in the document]
rowanc1: I think we can leave it off for now.

packages/myst-parser/src/tokensToMyst.ts
rowanc1: I started off that way, but opted to duplicate. If we change the underlying library for amsmath, this is the only place that should be impacted. I think this is appropriately guarded by tests, so am going to opt to leave as is!

packages/myst-transforms/src/math.ts
rowanc1: Adding a comment!

packages/markdown-it-myst/src/directives.ts
rowanc1: Yeah, I think that there are certainly better ways to do this. I will spend a bit of time on that now that we have tests...!

rowanc1: I think that we could look this up if we took over the block-level parsing. But currently we are not doing that for fences (`` ``` ``), only colon-fences (`:::`). That means that the previous token is a `paragraph_close` and there is no indication of blank lines (and it also doesn't have a map on it so we can't look to that to compare the line numbers).

I think that the current way of doing it is probably ok, I am going to cache the splitting of lines, which is the most time intensive part (I think).

rowanc1: Ok - I take it back, that is being shared in unexpected ways for the caching. Reverting the caching changes.

Issue #1021, opened 2024-03-21
Typst PDF building fails in a hard-to-identify way


choldgraf:
When you build a PDF with the `--typst` option, but it fails to build (for example, because `typst` isn't installed) then the logs still make it seem like the PDF was built even though it wasn't. You still see something like:

```
🖨 Rendering typst pdf to _build/pdf/report.pdf
```

It would be easier to debug if we had either a more prominent error message, or an error message that helped the user diagnose what was wrong.

### References 

Some discussion about this here:

- https://discordapp.com/channels/1083088970059096114/1177359316886503464/1220497756129591367

rowanc1:
This is fixed with #1033.

PR #1033, opened 2024-03-26
📤 Refactor exports for more flexibility

fwkoch:
Previously, when defining exports, you _had_ to specify format:
```
---
exports:
  - format: pdf
---
```

This is redundant with extensions...
```
---
exports:
  - format: pdf
    output: index.pdf
---
```

...and templates...
```
---
exports:
  - format: typst
    output: lapreprint-typst
---
```

There is also ambiguity around what `format: pdf` means, e.g. if I want a pdf built with `typst`, this failed since pdf was synonymous with `tex` build:
```
---
exports:
  - format: pdf
    output: lapreprint-typst
---
```

This PR refactors the export collection logic to be more human-friendly.
- Format can now be determined from `format`, `output`, or `template`, redundancy is not required
- `format: pdf` is more flexible and can result in a typst build if typst template or file extension are used
- Export may now specify `zip: true` - previously the only way to zip outputs was provide an `output` with `.zip` extension.

It also centralizes the logic into format-agnostic functions, so the `collectTexExportOptions`, `collectDocxExportOptions`, etc functions are deprecated. _Also_ previously there was some conflation between CLI-provided export options and file-provided export options... this has been separated a bit more cleanly, so all the CLI-provided options are handled in `build`, not in `collectExportOptions`...

👀 ALSO 👀 (Definitely not too much here. Definitely not.) This addresses #1021 by throwing an error when `typst` executable is not available, rather than silently carrying on and encountering other inexplicable errors...

### DIFF
--- a/packages/myst-cli/src/build/utils/collectExportOptions.ts
+++ b/packages/myst-cli/src/build/utils/collectExportOptions.ts
@@ -30,7 +30,7 @@ export const SOURCE_EXTENSIONS = ['.ipynb', '.md', '.tex'];
 
 type ResolvedArticles = Pick<Export, 'articles' | 'sub_articles'>;
 
-function resolveArticlesFromProject(
+export function resolveArticlesFromProject(
   exp: ExportWithFormat,
   proj: Omit<LocalProject, 'bibliography'>,
   vfile: VFile,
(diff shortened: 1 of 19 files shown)


packages/myst-cli/src/build/utils/collectExportOptions.ts
```diff
@@ -1,69 +1,307 @@
...
+/**
+ * Resolve format based on export format/template/output
+ *
+ * User-provided `format` is prioritized, then `template` type, then `output` file extension.
+ * (Validation ensures one of format/output/template is provided.)
+ *
+ * If `format` or `output` give PDF export, we look at the template kind and switch to `typst`
+ * in the case of a typst template.
+ */
+function resolveFormat(vfile: VFile, exp: Export): ExportFormats | undefined {
```
fwkoch, line 106: I need to add some unit tests at least on these `resolve*` functions before merging this...

Issue #833, opened 2024-01-08
Showing glossary terms with an apostrophe won't work


frenck:
## Description

Showing glossary terms that contain an apostrophe stopped working.

For example: `glossary.md`:
```md
:::{glossary}
Spook's influence
: Something
:::
```

When trying to reference it using:

```
{term}`Spook's influence`
```

Myst reports it as non-existing. This used to work before.

Asking around on Discord I was told to make an issue, as it "looks like a problem with converting to smart typography, which we did introduce a few months ago".

## Proposed solution

Hehe.. let's not ask for the captain obvious here :)

## Additional notes

![CleanShot 2024-01-08 at 21 49 49@2x](https://github.com/executablebooks/mystmd/assets/195327/a469f3e1-37b7-4bb0-a57d-fb247ac3b45a)

agoose77:
Yes, the reason that the reference breaks is because markdown-it (the underlying tokenizer) applies the smartquote transform, subtly changing the deflist identifier. #727 introduced this change.

Pinging @rowanc1 (PR author) for ideas here. My guess is that we might want to lift the smart-quotes to the MyST level so that we preserve more of the original markup.

rowanc1:
Fixed in #1034. Thank you for the bug report!

![image](https://github.com/executablebooks/mystmd/assets/913249/0bf138c3-b334-4d90-855c-18d03b60417f)

PR #1034, opened 2024-03-26
💬 Fix glossary terms with quotes in labels

rowanc1:
See #833


fwkoch (approved):
Seems fine! I don't see any big issue with making `identifier` normalization more normalized...

PR #1036, opened 2024-03-28
🏫 Add reviewers and editors to frontmatter

fwkoch:
This reuses the same `stash` infrastructure that as `funding` recipients/investigators.

On the final resolved frontmatter, `editors` and `reviewers` are lists of string ids which index into `contributors` and `authors` (each of those is still a full object).

(The actual editor/reviewer changes are pretty straightforward `frontmatter` updates - there was just another related edge case I ironed out to slightly complicate things...)

(This addresses a point in #1009)

### DIFF
--- a/packages/myst-frontmatter/src/contributors/contributors.yml
+++ b/packages/myst-frontmatter/src/contributors/contributors.yml
@@ -262,3 +262,37 @@ cases:
             literal: John Doe
             family: Doe
             given: John
+  - title: author as id reference resolves to object
+    raw:
+      reviewers:
+        - id: janedoe
+          name: Jane Doe
+          nameParsed:
+            literal: Jane Doe
+            family: Doe
+            given: Jane
+        - id: johndoe
+          name: John Doe
+          nameParsed:
+            literal: John Doe
+            family: Doe
+            given: John
+      author: janedoe
+    normalized:
+      reviewers:
+        - janedoe
+        - johndoe
+      authors:
+        - id: janedoe
+          name: Jane Doe
+          nameParsed:
+            literal: Jane Doe
+            family: Doe
+            given: Jane
+      contributors:
+        - id: johndoe
+          name: John Doe
+          nameParsed:
+            literal: John Doe
+            family: Doe
+            given: John
--- a/packages/myst-frontmatter/src/contributors/contributors.yml
+++ b/packages/myst-frontmatter/src/contributors/contributors.yml
@@ -166,3 +166,99 @@ cases:
             family: Author
             given: Test
     errors: 1
+  - title: editor/reviewer added to collaborators
+    raw:
+      reviewer: Jane Doe
+      editor: John Doe
+    normalized:
+      reviewers:
+        - Jane Doe
+      editors:
+        - John Doe
+      contributors:
+        - id: Jane Doe
+          name: Jane Doe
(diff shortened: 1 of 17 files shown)


packages/myst-frontmatter/src/contributors/contributors.yml
```diff
@@ -166,3 +166,133 @@ cases:
...
+            literal: Jane Doe
+            family: Doe
+            given: Jane
+        - id: johndoe
+          name: John Doe
+          nameParsed:
+            literal: John Doe
+            family: Doe
+            given: John
+  - title: author as id reference resolves to object
```
fwkoch, line 265: This edge case triggered some changes to the `stash` functionality. Previously, a warning was raised on author `janedoe`, referenced by id here. Even though `janedoe` is defined in `reviewers` and all the final frontmatter resolution was correct, initial validation still raised a warning since "janedoe" does not have a given name when parsed...

In this case, it's just an ID, not a name, so we don't actually want that warning, but we _would_ want the warning if it never resolved as an ID - then we'd have to assume it is in fact a name with no given name.

This means the `validateAndStash` functionality needs to suppress some warnings, but then maybe re-raise them later... Pretty tedious, and probably doesn't really matter. But now this tests pass as well as all the old stash-related tests.

packages/myst-frontmatter/src/utils/referenceStash.ts
```diff
@@ -64,19 +82,22 @@ export function validateAndStashObject<T extends { id?: string; name?: string }>
...
   });
+  stash[`${kind}Opts`] ??= {};
+  const optsLookup = stash[`${kind}Opts`] as Record<string, Options<T>>;
   if (typeof input === 'string' && Object.keys(lookup).includes(input)) {
     // Handle case where input is id and object already exists
     return input;
   }
-  const value = validateFn(input, opts);
+  // Suppress warnings that may occur if input is an id - these may be raised again later
+  const value = validateFn(input, { ...opts, suppressWarnings: typeof input === 'string' });
```
fwkoch, line 100: (Here's the warning suppression...)

packages/myst-frontmatter/src/utils/referenceStash.ts
```diff
@@ -93,14 +114,47 @@ export function validateAndStashObject<T extends { id?: string; name?: string }>
...
+export function finalizeStash(stash: ReferenceStash): {
+  affiliations?: (Affiliation & { id: string })[];
+  contributors?: (Contributor & { id: string })[];
+  authorIds?: string[];
+} {
+  return {
+    affiliations: stash.affiliations?.map((aff) => {
+      if (isStashPlaceholder(aff)) {
+        const { opts, validateFn } = stash.affiliationsOpts?.[aff.id] ?? {};
+        if (opts && validateFn) validateFn(aff, { ...opts, suppressErrors: true });
```
fwkoch, line 147: (...and here's where it comes back if the placeholder was never replaced by another object.)

rowanc1 (approved):
Worked through the demos locally! Works well. 🚀

fwkoch:
The changes I made to reduce unnecessary warnings still come up in cases where contributors are defined across project/page, so I'm going to back out those changes before I proceed. I'll just also remove the warning that was appearing about "no given name"

PR #1038, opened 2024-03-28
👯‍♀️ Add collaborations to Contributor type

fwkoch:
Now, in addition to being a `Person`, any `Contributor` in myst frontmatter may be a "collaboration" - that is, an `Affiliation` with `collaboration: true`. The logic to determine if a user-specified contributor is a person or a collaboration relies on checking the `collaboration` field - if `true`, it is a collaboration and validated as such; if `falsy` it is a person.

This addresses a point in #1009

### DIFF
--- a/packages/myst-frontmatter/src/contributors/authors.yml
+++ b/packages/myst-frontmatter/src/contributors/authors.yml
@@ -134,6 +134,33 @@ cases:
             family: Name
           email: example@example.com
           corresponding: false
+  - title: No email is corresponding if first is false
+    raw:
+      author:
+        - id: jn
+          name: Just A. Name
+          email: example@example.com
+          corresponding: false
+        - id: an
+          name: A. Nother Name
+          email: example@example.com
+    normalized:
+      authors:
+        - id: jn
+          name: Just A. Name
+          nameParsed:
+            literal: Just A. Name
+            given: Just A.
+            family: Name
+          email: example@example.com
+          corresponding: false
+        - id: an
+          name: A. Nother Name
+          nameParsed:
+            literal: A. Nother Name
+            given: A. Nother
+            family: Name
+          email: example@example.com
   - title: Respect corresponding flag
     raw:
       authors:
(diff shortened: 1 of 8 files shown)


packages/myst-frontmatter/src/contributors/validators.ts
```diff
@@ -127,16 +131,34 @@ export function validateName(input: any, opts: ValidationOptions) {
...
     input = stashPlaceholder(input);
   }
-  const value = validateObjectKeys(
-    input,
-    { optional: CONTRIBUTOR_KEYS, alias: CONTRIBUTOR_ALIASES },
-    opts,
-  );
+  const value = validateObjectKeys(input, { optional: PERSON_KEYS, alias: PERSON_ALIASES }, opts);
   if (value === undefined) return undefined;
+  if (inputAff && Object.keys(inputAff).length > Object.keys(value).length) {
```
fwkoch, line 156: This is a nice little check that says "Your input looks more like an affiliation than a person."

packages/myst-frontmatter/src/contributors/authors.yml
```diff
@@ -134,6 +134,33 @@ cases:
             family: Name
           email: example@example.com
           corresponding: false
+  - title: No email is corresponding if first is false
```
fwkoch, line 137: I noticed this behaviour and wanted to raise it:

Is it intentional that specifying `corresponding: false` on one author, prevents any other author from implicitly getting the `corresponding: true` flag?

I can see arguments either way:

(Yes, to the above question): the user knows about the `corresponding` flag, as they used it on one of their authors. If they did want a corresponding author, they would have known to add it.
(No, to the above question): we should always try to have a corresponding author where possible.

rowanc1: Lets make the change! And update this test case.

fwkoch: Updated!

PR #1043, opened 2024-04-01
🚦 set CONTENT_CDN_HOST in env of development server

stevejpurves:
These changes are to support the fix in https://github.com/executablebooks/myst-theme/pull/346 where url rewriting should honor the HOST variable used to start the `mystmd` development server.

This PR also changes the existing log messages to accurately report the usage of HOST rather than always citing `localhost`

### DIFF
--- a/.changeset/sharp-wasps-tap.md
+++ b/.changeset/sharp-wasps-tap.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Set `CONTENT_CDN_HOST` when `HOST` environment variable is specified and default to `localhost` otherwise. Also improve log messages when `HOST` is provided.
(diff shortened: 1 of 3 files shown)


fwkoch (approved):

rowanc1:
@stevejpurves how does this work with the `--keephost` option? I think we might need to change something around here:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-cli/src/build/site/start.ts#L85

The host variable is always set to localhost in that block to get around an earlier bug. Does the host variable set both the app and content servers? It looks like this is only targeting the content server?

rowanc1:
I am going to close this as fixed here:

https://github.com/jupyter-book/myst-theme/pull/468

Please open again if we need to address something here.

PR #1049, opened 2024-04-03
📥 Add downloads to page/project frontmatter and site config

fwkoch:
Previously, when you built a MyST site, all the `exports` defined on a page became **Downloads** and there was no other way to define additional or different downloads.

This PR adds `downloads` to page/project frontmatter as a place to specify (1) paths to local files and/or (2) export `id` values. Then, when you build your site, no longer do all the exports show up as downloads but only the specific `downloads` you specify.

This also makes improvements to site actions outlined here: https://github.com/executablebooks/mystmd/pull/1048

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -79,6 +79,12 @@ The following table lists the available frontmatter fields, a brief description 
 * - `description`
   - a string (max 500 chars)
   - page & project
+* - `exports`
+  - an export object, see [](#exports)
+  - page & project
+* - `downloads`
+  - a file or export ID, see [](#downloads)
+  - page & project
 * - `tags`
   - a list of strings
   - page only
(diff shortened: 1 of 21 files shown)


fwkoch:
TODO: Enable assigning a `title` to download, as an alternative to the file name. We do not currently allow this for page `exports`, so maybe it can be an enhancement in a separate PR? But we do allow it for site `actions` so there is some precedent.

stevejpurves:
Adding a title is defintiely a good idea, currently with just exports themes are left to guess at what to put on the UI (https://github.com/executablebooks/myst-theme/blob/ef36c695035ae6895e7d022c9d00c5016d6c6c8e/themes/article/app/components/Downloads.tsx#L4) and with arbitrary downloads adding the filename is not a good option. So a `title` will benefit the themes well in both cases 👍

docs/frontmatter.md
```diff
@@ -507,6 +513,45 @@ The date field is a string and should conform to a valid Javascript data format.
...
+* - `sub_articles`
+  - a list of strings - path(s) to sub-articles for `jats` export
+```
+
+(downloads)=
+
+## Downloads
+
+Downloads are files you want available on your MyST site. They may be defined at the project or the page level. In your frontmatter, `downloads` may include (a) relative paths to local files or (b) export `id` values to indicate you would like a specific export available as a download.
+
```
stevejpurves, line 554: @fwkoch do we need to update / expand the docs? what is the possible shape of the objects in the downloads list? I'm guessing in the below, and assuming that `static` and `internal` are inferred?

```
downloads:
  - url: ./manual.pdf
    title: Download Instructions
  - id: an-export-id-at-project-or-page-level
    title: Download This Article
  - url: https://policies.google.com/privacy
    title: Open Privacy Policy
    
```

packages/myst-config/src/site/types.ts
```diff
@@ -43,7 +50,7 @@ type ManifestProjectItem = {
   banner?: string | null;
   bannerOptimized?: string;
   tags?: string[];
-  exports?: Export[];
+  exports?: SiteExport[];
 };
```
stevejpurves, line 54: I get a type error in the theme when reconciling downloads 👇

image

downloads are also expected on pages? or is this reconcilation in the theme wrong? @fwkoch

fwkoch: Those `ManifestProjectItems` never actually have either exports or downloads. I removed `exports` from the type and deleted that line from `myst-theme`

docs/frontmatter.md
fwkoch: Updated the docs!

fwkoch:
One last thing this needs is differentiation between `downloads: undefined` and `downloads: []` - Pretty sure currently those both coerce to `[]` on the site config; we need to make sure we keep the differentiation.

PR #1052, opened 2024-04-03
🔭 Trim end of line only in directives (important for `code-blocks`)

rowanc1:
This code block now has indentation at the start. Previously it was trimmed.

````
```{code-block} c
:linenos: true
:lineno-start: 2
:emphasize-lines: 3
    for (int i = 0; i < 10; i++) {
        /* do something */
    }
```
````

### DIFF
--- a/.changeset/fuzzy-pots-think.md
+++ b/.changeset/fuzzy-pots-think.md
@@ -3,4 +3,4 @@
 "myst-parser": patch
 ---
 
-Code block only trims end, not start of lines.
+Only trim end of line for myst-directives, not both the start and end of lines. This is important for keeping indentation in code blocks.
--- a/.changeset/fuzzy-pots-think.md
+++ b/.changeset/fuzzy-pots-think.md
@@ -1,0 +1,6 @@
+---
+"markdown-it-myst": patch
+"myst-parser": patch
+---
+
+Code block only trims end, not start of lines.
--- a/packages/myst-parser/tests/myst.spec.ts
+++ b/packages/myst-parser/tests/myst.spec.ts
@@ -85,8 +85,10 @@ function stripPositions(tree: GenericParent) {
   return tree;
 }
 
-function removeTightAnnotations(tree: GenericParent) {
+function fixMystDirectives(tree: GenericParent) {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/myst-parser/tests/myst.spec.ts [hunk 1]
```diff
@@ -85,8 +85,10 @@ function stripPositions(tree: GenericParent) {
   return tree;
 }
 
-function removeTightAnnotations(tree: GenericParent) {
+function fixMystDirectives(tree: GenericParent) {
   selectAll('mystDirective', tree).forEach((node) => {
+    // Node markdown is trimmed
+    (node as any).value = (node as any).value?.trim();
```
fwkoch, line 91: Not totally clear why we need this, without digging into it...? Is there, like, some test-case round-tripping that does not work any more?

fwkoch (approved):
I think this change is fine: `trim()` on the directive content is problematic with leading spaces, using `trimEnd()` keeps leading spaces but makes the change slightly smaller since we are not entirely eliminating the `trim`.

My only suggestion is that the PR title and the changeset are misleading. This "trem end of line only" applies to _all_ directives, not just `code-block`.

packages/myst-parser/tests/myst.spec.ts
rowanc1: There are some of these defined in the myst-spec!

.changeset/fuzzy-pots-think.md
```diff
@@ -0,0 +1,6 @@
+---
+"markdown-it-myst": patch
+"myst-parser": patch
+---
+
+Code block only trims end, not start of lines.
```
rowanc1, line 6:
```suggestion
Only trim end of line for myst-directives, not both the start and end of lines. This is important for keeping indentation in code blocks.
```

packages/myst-parser/tests/myst.spec.ts
fwkoch: Yeah, ok - so the `trim()` originally came from respecting those tests, now we no longer totally align with them. (Well, there are already a few ways we don't align with them, this is just one more, not a big deal...)

PR #1073, opened 2024-04-08
:bug: Request DOI metadata as CSL-JSON after BibTeX

### LINKED ISSUE #1072: Fatal error when apparently invalid doi is encountered
## Description

The following markdown:

```
# DOI Errors

The following DOI causes the myst build to fail, even though it is a valid doi.

https://doi.org/10.2903/j.efsa.2019.5779

```

...causes the myst build to fail. 

## Proposed solution

(1) It is a valid DOI so there should be no problem parsing it (2) if it was an invalid doi then parsing should fail with a warning.

## Additional notes

```
myst -v
v1.1.51
```

agoose77:
The bug in #1072 appears to be caused by the key passed to `citation-js`:
```bibtex
@article{EFSA Panel on Nutrition, Novel Foods and Food Allergens (NDA)_Turck_Castenmiller_de Henauw_Hirsch‐Ernst_Kearney_Knutsen_Maciuk_Mangelsdorf_McArdle_et al._2019,
	title        = {Dietary reference values for chloride},
	author       = {EFSA Panel on Nutrition, Novel Foods and Food Allergens (NDA) and Turck, Dominique and Castenmiller, Jacqueline and de Henauw, Stefaan and Hirsch‐Ernst, Karen‐Ildico and Kearney, John and Knutsen, Helle Katrine and Maciuk, Alexandre and Mangelsdorf, Inge and McArdle, Harry J and Pelaez, Carmen and Pentieva, Kristina and Siani, Alfonso and Thies, Frank and Tsabouri, Sophia and Vinceti, Marco and Aggett, Peter and Fairweather‐Tait, Susan and Martin, Ambroise and Przyrembel, Hildegard and de Sesmaisons‐Lecarré, Agnès and Naska, Androniki},
	year         = 2019,
	month        = {Sep},
	journal      = {EFSA Journal},
	volume       = 17,
	number       = 9,
	doi          = {10.2903/j.efsa.2019.5779},
	issn         = {18314732, 18314732},
	url          = {https://data.europa.eu/doi/10.2903/j.efsa.2019.5779},
	place        = {IT},
	abstractnote = {This publication is linked to the following EFSA Supporting Publications articles: http://onlinelibrary.wiley.com/doi/10.2903/sp.efsa.2019.EN-1679/full, http://onlinelibrary.wiley.com/doi/10.2903/sp.efsa.2017.e15121/full This publication is linked to the following EFSA Journal article: http://onlinelibrary.wiley.com/doi/10.2903/j.efsa.2019.5778/full}
}
```

This is not valid BibTeX, and therefore does not parse. One solution is to request CSL-JSON via [content negotiation](https://citation.crosscite.org/docs.html). This trades one class of bugs (invalid BibTeX markup) for another (invalid CSL structures). I think this is a better part of the problem space to find oneself in; we can more easily attempt to fix-up bad CSL than we can bad BibTeX.

Indeed, it looks like using CSL-JSON from content-negotiation isn't foolproof either; some DOIs like [The Higgs Boson Paper](https://doi.org/10.1038/s41586-022-04892-x) report a `name` instead of a `literal`, or define a type of `journal-article` instead of `article-journal`. It looks like crossref is returning "extended citeproc": CrossRef/rest-api-doc#580 rather than true CSL-JSON.

Knowing that, we can probably do a reasonable job of coercing the result to the CSL-JSON that citation-js expects, but for now this PR does the minimum from testing the original DOI and the Higg's DOI.

I'm hoping that this change is _more_ robust than the BibTeX approach, but we will need to try more DOIs than I've thrown at it.

Fixes #1072


agoose77:
Good point! I'll rebase on main with the recommended changes.

packages/citation-js-utils/src/index.ts
```diff
@@ -189,15 +194,32 @@ export async function getCitations(bibtex: string): Promise<CitationRenderer> {
           },
           render(style?: CitationJSStyles) {
             return replaceUrlsWithAnchorElement(
-              cleanRef(cite.set(c).get({ ...defaultString, style: style ?? CitationJSStyles.apa })),
+              cleanRef(
+                cite.set(c).format('bibliography', {
```
agoose77, line 198: This (and the later usage) is an update to drop use of the deprecated `Cite.get` function.

packages/citation-js-utils/src/index.ts
```diff
@@ -171,12 +163,26 @@ export function firstNonDoiUrl(str?: string, doi?: string) {
...
+  }
+  return styleFormat;
+}
 
+export async function parseBibTeX(source: string): Promise<CSL[]> {
+  return (await Cite.async(source)).data;
+}
+
+export async function parseCSLJSON(source: object[]): Promise<CSL[]> {
+  return Promise.resolve(cleanCSL(source));
```
agoose77, line 179: This doesn't validate the CSL, but does try to "clean" it. Should we also use `Cite.async` here too?

packages/citation-js-utils/src/index.ts
```diff
@@ -171,12 +163,26 @@ export function firstNonDoiUrl(str?: string, doi?: string) {
...
+}
+
+export async function getCitationRenderers(data: CSL[]): Promise<CitationRenderer> {
```
agoose77, line 182: Now this function creates renderers for CSL (i.e. offloads parsing)

packages/myst-cli/src/transforms/doi.spec.ts
```diff
@@ -1,26 +1,44 @@
...
+    DOI: '10.1175/1520-0493(1972)100<0081:otaosh>2.3.co;2',
+    type: 'article-journal',
+    issue: '2',
+    issued: { 'date-parts': [[1972, 2]] },
+    page: '81-92',
+    publisher: 'American Meteorological Society',
+    title: 'On the Assessment of Surface Heat Flux and Evaporation Using Large-Scale Parameters',
+    volume: '100',
+  },
+];
```
agoose77, line 21: I simplified this to the subset of fields common to both resolvers. I think that's OK.

packages/myst-cli/src/transforms/dois.ts
```diff
@@ -13,8 +13,17 @@ import type { SingleCitationRenderer } from './types.js';
...
+const CSL_JSON_MIMETYPE = 'application/vnd.citationstyles.csl+json';
+const BIBTEX_MIMETYPE = 'application/x-bibtex';
+
+/**
+ * Build a path to the cache-file for the given DOI
+ *
+ * @param session: CLI session
+ * @param normalizedDoi: normalized DOI of the form `prefix/suffix`
+ */
+function doiCSLJSONCacheFile(session: ISession, normalizedDoi: string) {
```
agoose77, line 25: Now we cache the CSL-JSON, which I'm arguing is the "internal representation".

packages/myst-cli/src/transforms/dois.ts
```diff
@@ -99,14 +182,14 @@ export async function getCitation(
...
-  if (!bibtex) {
+  const data = await resolveDoiOrg(session, doiString, vfile, node);
+  if (!data) {
     const resolves = await doiOrgResolves(session, doiString);
     const normalizedDoi = doi.normalize(doiString);
     let message: string;
     let note: string | undefined;
     if (resolves) {
-      message = `No bibtex available from doi.org for doi:${normalizedDoi}`;
+      message = `No CSL JSON available from doi.org for doi:${normalizedDoi}`;
```
agoose77, line 192:
```suggestion
      message = `No CSL-JSON available from doi.org for doi:${normalizedDoi}`;
```

rowanc1:
Works well in web mode, cache seems good and doesn't conflict. From my testing there is a current bug in writing out the bibtex: `writeBibtexFromCitationRenderers` errors out when there is a CSL input [here](https://github.com/executablebooks/mystjs/blob/agoose77/fix-citation-parsing/packages/myst-cli/src/build/utils/bibtex.ts#L51).

image

I think this could/should be isolated into a test around falling back to CSL and then writing bibtex for the DOI of `10.2903/j.efsa.2019.5779`.

fwkoch (changes requested):
This is looking great - keeping bibtex as the default but adding a CSL fallback, explicitly making CSL the internal representation.

I don't have any big concerns with the changes to exported functions - there are a couple dependent packages that will need to update, but giving `citation-js-utils` a major bump makes that pretty clear.

I do think we need to fix the `writeBibtexFromCitationRenderers` functionality before releasing this - Probably for now just a simple function we can call if the bibtex entry is not found buried in the `cite._graph` - just take the CSL and return a bibtex string with title/author/year/journal/doi filled in.

agoose77:
Good spot on the BibTeX extraction; I'd assumed that we were using citation-js for this and hadn't noticed that instead we were pulling from the graph.

This PR now uses citation-js to build a BibTeX output. I'm assuming that we don't need to do anything to tweak the citation key, as it should match the `id` that citation-js reports (I _think_). We might need to be careful to ensure this.

In general, I think this PR represents a hardening of our BibTeX handling in addition to now providing a fallback for bad-BibTeX responses; we don't rely on implementation details from citation-js to store the original BibTeX, for example.

There _may_ be some non-round-tripping of certain BibTeX entries, but I think that should be very uncommon; BibTeX is a core plugin for citation-js.

packages/citation-js-utils/src/index.ts
```diff
@@ -171,12 +164,46 @@ export function firstNonDoiUrl(str?: string, doi?: string) {
   return matches.map((match) => match[0]).find((match) => !doi || !match.includes(doi));
 }
 
-export async function getCitations(bibtex: string): Promise<CitationRenderer> {
```
rowanc1, line 174: Changes look good, I will test again today!

@agoose77  -- can we get a shim on this that wraps that so this turns into a patch rather than a major release?

agoose77:
@fwkoch 
> So - in `parseBibTeX`, `Cite.async` gives us `bibtex` ->`CSL`, but here, it would just give us `CSL` -> `CSL` so maybe unnecessary and just "cleaning" is all that's required?
> 
> If all that is true, then I guess the only reason to use `Cite.async` is if doi.org gives an invalid CSL response...? Not totally out of the question I suppose... but then `Cite.async` is another mostly unnecessary API request?
> 
> Any other nuance to this decision?

Yes, even the BibTeX method doesn't need to be async. I suppose I was anticipating that we might want to support async resolvers in future. I'll move them to sync for now.

packages/myst-cli/src/transforms/dois.ts
```diff
@@ -182,19 +186,20 @@ export async function getCitation(
...
   if (!data) {
     const resolves = await doiOrgResolves(session, doiString);
     const normalizedDoi = doi.normalize(doiString);
     let message: string;
     let note: string | undefined;
     if (resolves) {
-      message = `No citation data available from doi.org for doi:${normalizedDoi}`;
-      note = `To resolve this error, visit ${doi.buildUrl(normalizedDoi)} and add citation info to .bib file`;
+      message = `Citation data from doi.org was not available or malformed for doi:${normalizedDoi}`;
+      note = `To resolve this error, visit ${doi.buildUrl(normalizedDoi)} and add citation info to local BibTeX file`;
```
rowanc1, line 197: One of the errors that I went through previously was that some DOIs are not "valid" without the `https://doi.org` on them, and so the double validation fails.

```suggestion
      note = `To resolve this error, visit ${doi.buildUrl(doiString)} and add citation info to local BibTeX file`;
```

fwkoch (approved):
Thanks for addressing the bibtex writing - using `citation-js` for that is much better. 🙂 

I noticed a couple more things when testing today:

First, I don't think this warning needs to be raised:
image
I don't think the user will care about implementation details of the underlying citation data format?

Also, totally non-existent citations were still causing a crash: 
image
This is related to `return undefined` vs `throwing an error`

I went ahead and fixed these up in a commit: https://github.com/executablebooks/mystmd/pull/1073/commits/559cbefe1c7ba6cd6abeaebd785a5def79c2910f

Hopefully that all looks ok? Otherwise, I'm happy with everything.

fwkoch:
One more bug I found on `tex` exports:
image

`cite` nodes resolved from CSL-JSON were not getting a `label`. Fixed here: https://github.com/executablebooks/mystmd/pull/1073/commits/ba9d0b925dfad749bd7f657c2a85653bb01e1b68

PR #1078, opened 2024-04-09
🐛 Set notebook path relative to session manager path

agoose77:
It turns out that the notebook path that is passed to the session manager should be given relative to the server base-path!

### DIFF
--- a/packages/myst-execute/tests/run.spec.ts
+++ b/packages/myst-execute/tests/run.spec.ts
@@ -78,7 +78,7 @@ casesList.forEach(({ title, cases }) => {
         file.path = path.join(__dirname, 'notebook.ipynb');
 
         await kernelExecutionTransform(before, file, {
-	      basePath: __dirname,
+          basePath: __dirname,
           sessionFactory: async () => await sessionManagerFactory.load(),
           cache: noOpCache,
           frontmatter: {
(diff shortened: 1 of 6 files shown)


packages/myst-execute/tests/run.spec.ts
```diff
@@ -78,6 +78,7 @@ casesList.forEach(({ title, cases }) => {
         file.path = path.join(__dirname, 'notebook.ipynb');
 
         await kernelExecutionTransform(before, file, {
+	  basePath: __dirname,
```
agoose77, line 81:
```suggestion
	      basePath: __dirname,
```

packages/myst-execute/tests/run.spec.ts
```diff
@@ -78,6 +78,7 @@ casesList.forEach(({ title, cases }) => {
         file.path = path.join(__dirname, 'notebook.ipynb');
 
         await kernelExecutionTransform(before, file, {
+	      basePath: __dirname,
           sessionFactory: async () => await sessionManagerFactory.load(),
```
rowanc1, line 82: I guess linting isn't on in the `*.spec.ts` files. GIthub puts 8 spaces on tabs... looks strange. :)
```suggestion
          basePath: __dirname,
          sessionFactory: async () => await sessionManagerFactory.load(),
```

Issue #797, opened 2023-12-05
Include latex snippets (tables)


alanlujan91:
## Proposal

A directive to explicitly include latex snippets such as tables. 

## Additional notes

Optional choice of direct latex in the directive or a reference to external file.

rowanc1:
Resolved in #1082.

Issue #548, opened 2023-08-11
Allow include directive work with notebooks


akhmerov:
## Proposal

Notebooks are a supported input format for myst. Trying to include a notebook using the `include` directive, however, only includes it as raw source. It would be more logical if notebooks were always included parsed.

rowanc1:
💯 ran into this the other day myself as well. Thank you for making an issue!!

rowanc1:
This has just been released with #1082.

PR #1082, opened 2024-04-10
🔄 Support recursive includes in myst and tex

fwkoch:
This adds more robust support for "recursive includes" - these are where one file `include`s another file which `include`s a third file, etc.

Additionally:
- Circular `include` dependencies error nicely (including printing the "`include`-stack") - no infinite loops:
  image
- `tex` source files with `\input` macros resolve and reload recursively as well

Resolves:
- #548
- #797


packages/myst-cli/src/transforms/include.ts
```diff
@@ -7,24 +7,64 @@ import type { VFile } from 'vfile';
...
+  };
+
+/**
+ * Return loadFile function
+ *
+ * Loaded file is added to original baseFile's dependencies.
+ */
+export const makeFileLoader = (session: ISession, baseFile: string) => (fullFile: string) => {
+  session.store.dispatch(
+    watch.actions.addLocalDependency({
```
fwkoch, line 45: These `include` dependencies _do not_ show up in the page json "dependency" field, but they are respected for live site reloading.

I think that's ok - the json "dependencies" are used for loading data from other pages; in the case of `include`, this data is actually moved to the page itself.

🤔 Maybe if we update `include` to allow including _partial_ snippets from md files we may need to load the rest of the data from that page...? (Similar to `embed` pulling in a single figure.) But for now, it's unnecessary.

rowanc1:
Works really well locally in both latex and markdown - and a mix as well, which is kinda cool:

image

packages/myst-cli/src/transforms/include.ts
rowanc1: Agree, the included files are a local convinience, not for how the final documents work.

rowanc1:
Added some docs, this is good to go in my opinion!

image

rowanc1 (approved):
Code looks good, fun to see the implementation. Nice work @fwkoch.

PR #1091, opened 2024-04-11
🪆 Add support for nested parsing

agoose77:
This is a simple PR to permit MyST directives to invoke the MyST parser.

I don't know whether we'll want to provide a more explicit set of parsing routines, e.g. `parseInline` and `parseBlock`. My hunch is that we can get to that fairly easily using transforms, so it's not necessary.

### DIFF
--- a/packages/myst-common/src/types.ts
+++ b/packages/myst-common/src/types.ts
@@ -85,7 +85,7 @@ export type RoleData = {
 };
 
 export type DirectiveContext = {
-  parseMyST: (source: string) => GenericParent;
+  parseMyST: (source: string, offset?: number) => GenericParent;
 };
 
 export type DirectiveSpec = {
--- a/packages/myst-common/src/types.ts
+++ b/packages/myst-common/src/types.ts
@@ -85,9 +85,8 @@ export type RoleData = {
 };
 
 export type DirectiveContext = {
-  parseMyST: (source: string) => GenericParent
+  parseMyST: (source: string) => GenericParent;
 };
-
 
 export type DirectiveSpec = {
   name: string;
--- a/packages/myst-common/src/types.ts
+++ b/packages/myst-common/src/types.ts
@@ -85,7 +85,7 @@ export type RoleData = {
 };
 
 export type DirectiveContext = {
-  parseMyST: (source: string, offset?: number) => GenericParent;
+  parseMyst: (source: string, offset?: number) => GenericParent;
 };
 
 export type DirectiveSpec = {
--- a/packages/myst-common/src/types.ts
+++ b/packages/myst-common/src/types.ts
@@ -84,6 +84,11 @@ export type RoleData = {
   body?: ParseTypes;
 };
 
+export type DirectiveContext = {
+  parseMyST: (source: string) => GenericParent
+};
+
+
 export type DirectiveSpec = {
   name: string;
   alias?: string[];
@@ -92,7 +97,7 @@ export type DirectiveSpec = {
   options?: Record<string, OptionDefinition>;
   body?: BodyDefinition;
   validate?: (data: DirectiveData, vfile: VFile) => DirectiveData;
-  run: (data: DirectiveData, vfile: VFile) => GenericNode[];
+  run: (data: DirectiveData, vfile: VFile, ctx: DirectiveContext) => GenericNode[];
 };
 
 export type RoleSpec = {
--- a/packages/myst-parser/src/directives.ts
+++ b/packages/myst-parser/src/directives.ts
@@ -14,7 +14,7 @@ import type { Directive } from 'myst-spec';
 
 type MystDirectiveNode = GenericNode & {
   name: string;
-}
+};
 /**
  * Apply directive `run()` methods to build directive ASTs.
  *
--- a/packages/myst-parser/src/directives.ts
+++ b/packages/myst-parser/src/directives.ts
@@ -15,11 +15,9 @@ type MystDirectiveNode = GenericNode & {
   name: string;
 };
 
-
 export type DirectiveContext = {
   parseMyST: (source: string) => GenericParent;
 };
-
 
 /**
  * Apply directive `run()` methods to build directive ASTs.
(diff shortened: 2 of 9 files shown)


packages/myst-parser/src/myst.ts
```diff
@@ -93,7 +93,9 @@ export function mystParse(content: string, opts?: Options) {
   const parsedOpts = parseOptions(opts);
   const tokenizer = createTokenizer(parsedOpts);
   const tree = tokensToMyst(content, tokenizer.parse(content, { vfile }), parsedOpts.mdast);
-  applyDirectives(tree, parsedOpts.directives, parsedOpts.vfile);
+  applyDirectives(tree, parsedOpts.directives, parsedOpts.vfile, {
+    parseMyST: (source: string) => mystParse(source, opts),
+  });
   applyRoles(tree, parsedOpts.roles, parsedOpts.vfile);
```
rowanc1, line 99: Do roles get the ctx as well?

packages/myst-parser/src/myst.ts
```diff
@@ -93,7 +93,9 @@ export function mystParse(content: string, opts?: Options) {
   const parsedOpts = parseOptions(opts);
   const tokenizer = createTokenizer(parsedOpts);
   const tree = tokensToMyst(content, tokenizer.parse(content, { vfile }), parsedOpts.mdast);
-  applyDirectives(tree, parsedOpts.directives, parsedOpts.vfile);
+  applyDirectives(tree, parsedOpts.directives, parsedOpts.vfile, {
+    parseMyST: (source: string) => mystParse(source, opts),
+  });
```
rowanc1, line 98: Can we have a way to fix line numbers from the re-entrant parsing?

This apply directives takes the function you have here, but wraps what it gives to each run in a special function. For example, this could be inside of the `applyDirectives` function before passing to the run function:

```typescript
const bodyOffset = directiveNode.pos[0]
const parseMystForDirective = (source: string, offset = 0) => {
   const children = mystParse(source)
  map(children, (node) => node.position.0 += offset + bodyOffset.)
  return children;
}
```

That way you can keep track of line numbers in the body, and then the line numbers get fixed up when the re-entrant parsing is complete.

rowanc1 (commented):
Would be great to track line numbers, this was simpler than I thought and I really like the approach you took!

packages/myst-parser/src/myst.ts
rowanc1: Maybe not based on your comments above, we can always add it if we need to later.

packages/myst-parser/src/myst.ts
agoose77: Ha, line numbers is something I was wondering about! You beat me to the punch :laughing:

packages/myst-common/src/types.ts
```diff
@@ -84,6 +84,10 @@ export type RoleData = {
   body?: ParseTypes;
 };
 
+export type DirectiveContext = {
+  parseMyST: (source: string, offset?: number) => GenericParent;
```
rowanc1, line 88:
```suggestion
  parseMyst: (source: string, offset?: number) => GenericParent;
```

This matches the current exported function name.

rowanc1 (commented):
This looks good from the code diff. I think your other PR builds upon it, and I can test there?

Maybe, is there a small generic test we can have for this functionality that ensures the line numbers are tested?

I think that we should change the capitalization on the `ctx` object to match the currently exported function: `parseMyst`.

packages/myst-common/src/types.ts
agoose77: Yeah, I think I've been internally thinking that we should rename our other types to `MyST` (I've had the same thought about DOI vs doi). However, happy to make that a future all-in-one pass if we get buy in.

packages/myst-parser/tests/recursiveParsing.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    const paragraphs = selectAll('paragraph', tree);
+    expect(paragraphs.length).toBe(1);
+    expect(paragraphs[0].position?.start.line).toBe(5);
+    expect(paragraphs[0].position?.end.line).toBe(5);
+    const text = selectAll('text', tree);
+    expect(text.length).toBe(2);
+    expect(text[0].position?.end.line).toBe(5);
+    expect(text[1].position?.end.line).toBe(5);
+    const subscript = selectAll('subscript', tree);
+    expect(subscript.length).toBe(1);
```
rowanc1, line 47: @agoose77 this currently fails, the recursive parsing will process all myst-roles and directives twice, and the result is not the same the second time around, which means there is no `roleBody` and that is required, so there is an error thrown.

We need to flag that the roles are directives have already been processed.

rowanc1: Ok, added a flag to only process once, this is good to go in!

packages/myst-parser/src/directives.ts
```diff
@@ -208,7 +208,7 @@ export function applyDirectives(
     node.children = run(data, vfile, {
       // Implement a parseMyst function that accepts _relative_ line numbers
       parseMyst: (source: string, offset: number = 0) =>
-        ctx.parseMyst(source, offset + node.pos[0]),
```
agoose77: Oof, looks like I took this verbatim without testing it. I hadn't actually gotten to validation yet.

packages/myst-parser/tests/recursiveParsing.spec.ts
agoose77: Oh, that didn't occur to me at all! Thanks for the rigorous testing on this, I think I needed another iteration by the looks of things.

rowanc1: It was close enough for me to get it over the line! :)

packages/myst-parser/src/directives.ts
rowanc1: That was a hard error. :)

The testing was helpful to pull out the other issue of recursive parsing, which I did not expect at all!

I also don't think this would have been easy to see in the next two PRs, so glad we have some test coverage on this at the parser level!

PR #1092, opened 2024-04-11
📑 Add support for Sphinx-style glossaries

agoose77:
Requires #1091

### DIFF
--- a/.changeset/mean-windows-hammer.md
+++ b/.changeset/mean-windows-hammer.md
@@ -1,0 +1,6 @@
+---
+"myst-common": minor
+"myst-parser": minor
+---
+
+Add support for nested parsing via a context object
--- a/.changeset/two-coins-invite.md
+++ b/.changeset/two-coins-invite.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Add legacy_glossary_syntax to project settings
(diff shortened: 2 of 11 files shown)


agoose77:
@rowanc1 to integrate the legacy parsing, we want to read the config option you defined in [633e158](https://github.com/executablebooks/mystmd/pull/1092/commits/633e1586b3b0e61606d5ebe227473983a4cc7cba). Right now we don't deal with that kind of context in our directives. Do you have any strong feelings about the direction to take here? I'm currently thinking that we need to customize the parser for each document.

rowanc1:
Can we pass this into the myst parse? It looks like options make it all the way through to the other PR?

We have access to the session here:
https://github.com/executablebooks/mystjs/blob/main/packages/myst-cli/src/process/myst.ts#L38

We could either swap out a directive completely, or pass the option into the parser? Or have an area of the options that is for directives?

```

  const parsed = mystParse(content, {
    markdownit: { linkify: true },
    directives: [
      cardDirective,
      gridDirective,
      reactiveDirective,
      proofDirective,
      ...exerciseDirectives,
      ...tabDirectives,
      ...(session.plugins?.directives ?? []),
    ],
    directiveOptions: {
        glossary: { parseLegacy: session.options.something },
    }
    extensions: {
      frontmatter: !opts?.ignoreFrontmatter,
    },
    roles: [reactiveRole, ...(session.plugins?.roles ?? [])],
    vfile,
  });
```

rowanc1:
@agoose77 there needs to be some fixing of the git history now that #1091 is in! Let me know if you need help!

PR #1093, opened 2024-04-11
🗺 Add `tectonic` support

markusritschel:
This PR includes [Tectonic](https://tectonic-typesetting.github.io/) as a LaTeX build engine instead of `latexmk`. 
However, I let `latexmk` in the code as a fallback option.
The warning in _packages/myst-cli/src/build/pdf/create.ts_ is extended with an availability check for `tectonic`.

See also the discussion [here](https://github.com/orgs/executablebooks/discussions/1154).

### DIFF
--- a/docs/creating-pdf-documents.md
+++ b/docs/creating-pdf-documents.md
@@ -91,7 +91,9 @@ Ensure that you download a full distribution with appropriate libraries installe
 
 #### Tectonic as a lightweight alternative
 As an alternative, you can also install [Tectonic](https://tectonic-typesetting.github.io/), "a modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive".<br>
-> "Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing." (excerpt from the website)
+> Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing.
+>
+> -- [Tectonic Website](https://tectonic-typesetting.github.io/)
 
 % Probably a note in the future about running this remotely?
 
--- a/docs/creating-pdf-documents.md
+++ b/docs/creating-pdf-documents.md
@@ -93,7 +93,7 @@ Ensure that you download a full distribution with appropriate libraries installe
 As an alternative, you can also install [Tectonic](https://tectonic-typesetting.github.io/), "a modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive".<br>
 > Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing.
 >
-> -- [Tectonic Website](https://tectonic-typesetting.github.io/)
+> – [Tectonic Website](https://tectonic-typesetting.github.io/)
 
 % Probably a note in the future about running this remotely?
 
--- a/docs/creating-pdf-documents.md
+++ b/docs/creating-pdf-documents.md
@@ -87,6 +87,12 @@ See the official documentation for installation instructions for $\LaTeX$ at:
 
 Ensure that you download a full distribution with appropriate libraries installed.
 
+(install-tectonic)=
+
+#### Tectonic as a lightweight alternative
+As an alternative, you can also install [Tectonic](https://tectonic-typesetting.github.io/), "a modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive".<br>
+> "Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing." (excerpt from the website)
+
 % Probably a note in the future about running this remotely?
 
 (rendering-pdfs-with-typst)=
--- a/packages/jtex/src/tex/export.ts
+++ b/packages/jtex/src/tex/export.ts
@@ -1,6 +1,15 @@
 import path from 'node:path';
 import type MystTemplate from 'myst-templates';
 import { createCommand } from '../utils.js';
+import which from 'which';
+
+export function isLatexmkAvailable() {
+  return which.sync('latexmk', { nothrow: true });
+}
+
+export function isTectonicAvailable() {
+  return which.sync('tectonic', { nothrow: true });
+}
 
 export function pdfTexExportCommand(
   texFile: string,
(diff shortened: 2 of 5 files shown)


packages/jtex/src/tex/export.ts
```diff
@@ -1,16 +1,29 @@
...
+
+export function isLatexmkAvailable() {
+  return which.sync('latexmk', { nothrow: true });
+}
+
+export function isTectonicAvailable() {
+  return which.sync('tectonic', { nothrow: true });
+}
 
 export function pdfTexExportCommand(
```
rowanc1, line 14: I think let's remove the duplication here by moving this command to `myst-cli/src/build/pdf`.

rowanc1 (commented):
Minor changes.

I think this is close, it would be wonderful if you could add a section in the docs.

I can aim to take a pass later today or tomorrow to clean anything up and get it in!

packages/jtex/src/tex/export.ts
markusritschel: Great.
The only thing regarding the duplicated functions is that I cannot import them from `myst-cli/src/build/pdf/utils.ts` to `jtex/src/tex/export.ts`.

rowanc1: I was thinking to move the whole file over, but I see that is a bit complex, and we probably want to do it for typst at the same time.

I am happy to come in after and do a quick refactor, I also want to talk to @fwkoch to get a sense check on moving those around.

Have you tested the functionality locally?

markusritschel:
Just added a paragraph for the documentation.

packages/jtex/src/tex/export.ts
markusritschel: Yes, it works on my local machine. Cloned the forked repo and followed your contribution guidelines for developers.

docs/creating-pdf-documents.md
```diff
@@ -87,6 +87,12 @@ See the official documentation for installation instructions for $\LaTeX$ at:
 
 Ensure that you download a full distribution with appropriate libraries installed.
 
+(install-tectonic)=
+
+#### Tectonic as a lightweight alternative
+As an alternative, you can also install [Tectonic](https://tectonic-typesetting.github.io/), "a modernized, complete, self-contained TeX/LaTeX engine, powered by XeTeX and TeXLive".<br>
+> "Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing." (excerpt from the website)
```
rowanc1, line 94: Thanks for the docs!!

```suggestion
> Tectonic automatically downloads support files so you don’t have to install a full LaTeX system in order to start using it. If you start using a new LaTeX package, Tectonic just pulls down the files it needs and continues processing.
>
> -- [Tectonic Website](https://tectonic-typesetting.github.io/)
```

rowanc1: Makes it look [like this](https://mystmd.org/sandbox?tab=demo&myst=PgAgAFQAZQBjAHQAbwBuAGkAYwAgAGEAdQB0AG8AbQBhAHQAaQBjAGEAbABsAHkAIABkAG8AdwBuAGwAbwBhAGQAcwAgAHMAdQBwAHAAbwByAHQAIABmAGkAbABlAHMAIABzAG8AIAB5AG8AdQAgAGQAbwBuABkgdAAgAGgAYQB2AGUAIAB0AG8AIABpAG4AcwB0AGEAbABsACAAYQAgAGYAdQBsAGwAIABMAGEAVABlAFgAIABzAHkAcwB0AGUAbQAgAGkAbgAgAG8AcgBkAGUAcgAgAHQAbwAgAHMAdABhAHIAdAAgAHUAcwBpAG4AZwAgAGkAdAAuACAASQBmACAAeQBvAHUAIABzAHQAYQByAHQAIAB1AHMAaQBuAGcAIABhACAAbgBlAHcAIABMAGEAVABlAFgAIABwAGEAYwBrAGEAZwBlACwAIABUAGUAYwB0AG8AbgBpAGMAIABqAHUAcwB0ACAAcAB1AGwAbABzACAAZABvAHcAbgAgAHQAaABlACAAZgBpAGwAZQBzACAAaQB0ACAAbgBlAGUAZABzACAAYQBuAGQAIABjAG8AbgB0AGkAbgB1AGUAcwAgAHAAcgBvAGMAZQBzAHMAaQBuAGcALgAKAD4ACgA%2BACAALQAtACAAWwBUAGUAYwB0AG8AbgBpAGMAIABXAGUAYgBzAGkAdABlAF0AKABoAHQAdABwAHMAOgAvAC8AdABlAGMAdABvAG4AaQBjAC0AdAB5AHAAZQBzAGUAdAB0AGkAbgBnAC4AZwBpAHQAaAB1AGIALgBpAG8ALwApAA%3D%3D)

packages/jtex/src/tex/export.ts
rowanc1: Fantastic - I will give it a go today or tomorrow and we can get this released early next week. Thanks again for your help in the PR, excited to try out tectonic and stop recommending 5gb downloads ... !

markusritschel: 😄 
My pleasure. Yes, it's still relatively unknown, but it's such a step forward for local LaTeX compiling.

fwkoch: Regarding the duplication - it would be nice to eliminate... that would mean "creating the command" (currently in `jtex`) and "invoking the command" (currently in `myst-cli`) in one single place. Definitely a smaller refactor to move "creation" to `myst-cli`, but it could also be interesting if part of the scope of `jtex` was to know how to create and invoke all the various rendering engines. MyST could then just say "jtex, build me a pdf using this template" hmm...

Anyway - the duplication here is relatively minor; I don't think we need to block this PR with a big refactor.

fwkoch (commented):
One thing I wanted to check on - if you have glossaries, the `latexmk` build runs twice with a `makeglossaries` command in between: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/build/pdf/create.ts#L218-L250

I think the `tectonic` build will do the same thing. Is this correct (i.e. is the `makeglossaries` workflow fundamental to latex, regardless of engine), or do we need to make changes deeper in the build invocation flow?

markusritschel:
Hi @fwkoch,
Tectonic should take care of it automatically. It usually figures out by itself how often it needs to run the build process. (One of its conveniences 😉).

rowanc1:
Update: I have taken a first look at this locally and am working through some of the error states and logging, which need to change a bit if tectonic does not succeed. I will continue to chip away at this during the week.

agoose77:
We triaged this PR as blocked during the recent weekly MyST Developers meeting. Although this functionality is useful, it will require work internally to handle the distinct flavour of log output. As such, we're holding off on pushing this forward until we have a few more cycles.

PR #1095, opened 2024-04-11
📖 Small updates to contributing and include docs

choldgraf:
I noticed that we were pulling in the `# Contributing` title markdown in our `{include}` statement for the contributing guide. I decided to try using the `start-after` argument to see if we can remove that content from the MyST site. In the process I also added an easier-to-find link to the reference docs for both `{embed}` and `{include}`.

### Question: Am I using `{include}` wrong?

I used the `start-after` argument similar to how it's used in Sphinx, and assumed it'd work since it is documented. However it's not cutting off the top part of the included page. I wonder if this is a bug or if I'm using it incorrectly. @rowanc1 or @fwkoch do you have guidance?

### Blockers

- [ ] https://github.com/executablebooks/mystmd/issues/1096

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -4,4 +4,5 @@ short_title: Contribution Guide
 ---
 
 ```{include} ../CONTRIBUTING.md
+:start-after: "<!-- MYST START -->"
 ```
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -3,6 +3,8 @@ title: Contribution Guide
 short_title: Contribution Guide
 ---
 
+% start-after doesn't currently work with {include} so this will just include the whole file
+% Tracking issue: https://github.com/executablebooks/mystmd/issues/1096
 ```{include} ../CONTRIBUTING.md
 :start-after: "<!-- MYST START -->"
 ```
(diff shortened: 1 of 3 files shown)


rowanc1:
That is currently only implemented for literal includes! Would welcome an issue to expand support!

Also thinking on what the syntax is for notebooks for a "start after" when it is json...?!

docs/contributing.md
```diff
@@ -4,4 +4,5 @@ short_title: Contribution Guide
 ---
 
 ```{include} ../CONTRIBUTING.md
+:start-after: "<!-- MYST START -->"
 ```
```
choldgraf, line 8:
````suggestion
% start-after doesn't currently work with {include} so this will just include the whole file
% Tracking issue: https://github.com/executablebooks/mystmd/issues/1096
```{include} ../CONTRIBUTING.md
:start-after: "<!-- MYST START -->"
```
````

Comment (choldgraf) [reply docs/contributing.md:8]: @rowanc1 what do you think about adding this comment for context and then just merging this one?

Comment (rowanc1) [reply docs/contributing.md:8]: Yep, that works!

File: docs/contributing.md [hunk 2]
```diff
@@ -3,5 +3,8 @@ title: Contribution Guide
 short_title: Contribution Guide
 ---
 
+% start-after doesn't currently work with {include} so this will just include the whole file
+% Tracking issue: https://github.com/executablebooks/mystmd/issues/1096
 ```{include} ../CONTRIBUTING.md
+:start-after: "<!-- MYST START -->"
```
rowanc1, line 9:
```suggestion
:start-after: <!-- MYST START -->
```

PR #1100, opened 2024-04-12
📓 Respect part aliases in parts frontmatter and extractPart function

fwkoch:
We have some "known parts" that we respect in frontmatter validation, e.g. `acknowledgments`, `abstract`. These also have aliases (in the same pattern as other frontmatter aliases), e.g. `acknowledgments` -> `acknowledgements`, `ack`.

However, these aliases were only respected for _top-level_ frontmatter keys; they were not respected when nested under `parts`, nor when called upon in `extractPart`.

Examples:
```
---
ack: Thank you!
---
```
would coerce to 
```
---
parts:
  acknowledgments: Thank you!
---
```
but
```
---
parts:
  ack: Thank you!
---
```
would remain unchanged during validation.

Also, `extractPart(tree, 'ack')` would not pick up on `part: acknowledgments`.

This PR updates frontmatter and `extractPart` to respect the aliases and coerce/extract consistently.

### DIFF
--- a/packages/myst-cli/src/transforms/dois.ts
+++ b/packages/myst-cli/src/transforms/dois.ts
@@ -183,7 +183,7 @@ export async function getCitation(
   session: ISession,
   vfile: VFile,
   doiString: string,
-  node: GenericNode,
+  node?: GenericNode,
 ): Promise<SingleCitationRenderer | null> {
   if (!doi.validate(doiString)) return null;
   const data = await resolveDoiOrg(session, doiString);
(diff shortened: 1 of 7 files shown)


packages/myst-cli/src/transforms/dois.ts
```diff
@@ -183,7 +183,7 @@ export async function getCitation(
   session: ISession,
   vfile: VFile,
   doiString: string,
-  node: GenericNode,
+  node?: GenericNode,
```
fwkoch, line 186: Super minor change - everywhere `node` is used inside this function it is optional, so there's no reason it cannot be optional here.

This is helpful if we want to use `getCitation` in a context outside mdast.

PR #1104, opened 2024-04-15
🚦 Change severity of the error messages

rowanc1:
This adds a `error_rules` to the project configuration, which can be used to disable logging rules in the CLI.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -70,4 +70,4 @@ project:
         - https://flaky-connection.com
 ```
 
-The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity will be adopted rather than the default log message severity. To find out the rule ID, run myst in debug mode to get the error (and optional key) printed to the console.
+The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity listed will be adopted rather than the default log message severity. The default severity for rules included in the list is `ignore`, which means that simply listing the rule IDs as strings will ignore those rules. To discover the rule ID, run myst in debug mode to get the error (and optional key) printed to the console. For example, the above configuration updates will no longer warn on `math-eqnarray-replaced` and will also ignore the two links when running `myst build --check-links --strict` in continuous integration.
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -52,3 +52,21 @@ beamer
 
     - `true`: Add `\begin{frame}` environment for each block, delimited by `+++`, and enable presentation outline with block metadata `+++ {"outline":true}`
     - `false` (default): No extra `\begin{frame}` environment will be used
+
+## Error Rules
+
+The `error_rules` list in the project configuration can be used to disable logging rules in the CLI:
+
+```{code-block} yaml
+:filename: myst.yml
+project:
+  error_rules:
+    - rule: math-eqnarray-replaced
+      severity: ignore
+    - rule: doi-check
+      severity: ignore
+      keys:
+        - bibtex_key
+```
+
+The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity will be adopted rather than the default log message severity.
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -55,7 +55,7 @@ beamer
 
 ## Error Rules
 
-The `error_rules` list in the project configuration can be used to disable logging rules in the CLI:
+The `error_rules` list in the project configuration can be used to disable or modify logging rules in the CLI:
 
 ```{code-block} yaml
 :filename: myst.yml
@@ -63,10 +63,11 @@ project:
   error_rules:
     - rule: math-eqnarray-replaced
       severity: ignore
-    - rule: doi-check
+    - rule: link-resolves
       severity: ignore
       keys:
-        - bibtex_key
+        - /known-internal-link
+        - https://flaky-connection.com
 ```
 
-The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity will be adopted rather than the default log message severity.
+The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity will be adopted rather than the default log message severity. To find out the rule ID, run myst in debug mode to get the error (and optional key) printed to the console.
--- a/packages/myst-cli/src/utils/addWarningForFile.ts
+++ b/packages/myst-cli/src/utils/addWarningForFile.ts
@@ -15,6 +15,8 @@ export function addWarningForFile(
     url?: string | null;
     position?: VFileMessage['position'];
     ruleId?: string | null;
+    /** This key can be combined with the ruleId to suppress a warning */
+    key?: string | null;
   },
 ) {
   const line = opts?.position?.start.line ? `:${opts?.position.start.line}` : '';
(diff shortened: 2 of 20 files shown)


packages/myst-config/src/errorRules/validators.ts
```diff
@@ -0,0 +1,63 @@
...
+  const value = validateObjectKeys(input, ERROR_RULE_KEY_OBJECT, {
+    ...opts,
+    suppressWarnings: true,
+    keepExtraKeys: true,
+  });
+  if (value === undefined) return undefined;
+  const id = validateString(value.id, incrementOptions('id', opts));
+  if (!id) return undefined;
+  const output: ErrorRule = { id };
+  if (defined(value.severity)) {
```
fwkoch, line 33: Feels a little ambiguous about what happens when only `id` is specified. Looks like it remains `undefined` - then, in `addWarningForFile` the default will be replaced by `undefined` and hit the `default` case in the switch statement... Hmm, then in the store, we will also get an undefined warning `kind`.

I think we need to make `severity` required on the type, and set a default here during validation?

fwkoch: That begs the question what the default is... If I put `error_rule: my-id` in my config... am I expecting that will suppress it...? Based on the name "error rule" I'd more expect the default behaviour to be "Error on rule my-id" - so the default `severity` is `'error'`?

But then what's the use-case? Are we intending this for error suppression as the default behaviour? If so, maybe we need to rethink `error_rule`...?

packages/myst-config/src/errorRules/validators.ts
```diff
@@ -0,0 +1,63 @@
...
+    const choice = validateChoice(value.severity, {
+      ...incrementOptions('severity', opts),
+      choices: ['ignore', 'warn', 'error'],
+    }) as ErrorRule['severity'];
+    if (choice) output.severity = choice;
+  }
+  // Put the unknown fields back on the object
```
fwkoch, line 40: My understanding is these unknown fields are intended to differentiate between multiple errors of the same `ruleId`? This feels maybe too flexible? To filter on these different entries, we are going to need to expand the line in `addWarningForFile`:
```typescript
const handler = config?.error_rules?.find((rule) => rule.id === opts.ruleId);
```
to something more like
```typescript
const handler = config?.error_rules?.find((rule) => {
  return rule.id === opts.ruleId && someOtherComparison(rule, opts);
});
```
If we allow arbitrary data, that `someOtherComparison` function becomes complicated and potentially rule-id-dependent. If instead we just have, say, `key`, then it's just an equality check. (Of course, we will still need rule-id-dependent key generation, but that can happen where the rule is actually getting violated, and by the time we get here, it's just a string.)

That also makes user feedback in the logging easier: "To suppress errors of this type, add `my-rule-id` to `error_rules`; to suppress this error only, add `key: my-rule-key`"

packages/myst-config/tests/examples.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const casesList = files
+  .map((file) => ({ name: file, data: fs.readFileSync(file).toString() }))
+  .map((file) => {
+    const tests = yaml.load(file.data) as TestFile;
+    tests.title = tests.title ?? file.name;
+    return tests;
+  });
+
+casesList.forEach(({ title, kind, cases }) => {
```
fwkoch, line 42: Oh brother, no nice yaml tests in this package yet... 😆 Thanks for adding!

packages/myst-cli/src/utils/addWarningForFile.ts
```diff
@@ -26,7 +27,18 @@ export function addWarningForFile(
...
+    const handler = config?.error_rules?.find((rule) => rule.id === opts.ruleId);
+    if (handler) {
+      if (handler.severity === 'ignore') {
+        session.log.debug(`${prefix}${formatted}`);
+        return;
+      }
+      severity = handler.severity as WarningKind;
+    }
+  }
+  switch (severity) {
```
fwkoch, line 41: After this switch statement, we need to add
```typescript
session.log.debug(`To suppress this message, add "${rule.id}" to "error_rules" in your project config`)
```

fwkoch: This make me think about my previous comment about default severity. It's easier to say "add X to error_rules" than to say "add an entry to error_rules with id X and severity info" - so maybe that means the default is `severity: info`. In that case, maybe it should be, like `error_exceptions`?

```yaml
error_exceptions:
  - my-id
```

sounds like we are excepting that error. But

```yaml
error_exceptions:
  - id: my-id
    severity: warn
```

also still makes sense - it's an exception because we are changing the severity.

packages/myst-config/src/errorRules/validators.ts
```diff
@@ -0,0 +1,63 @@
...
+    keepExtraKeys: true,
+  });
+  if (value === undefined) return undefined;
+  const id = validateString(value.id, incrementOptions('id', opts));
+  if (!id) return undefined;
+  const output: ErrorRule = { id };
+  if (defined(value.severity)) {
+    const choice = validateChoice(value.severity, {
+      ...incrementOptions('severity', opts),
+      choices: ['ignore', 'warn', 'error'],
```
fwkoch, line 36: 🤔 Would be nice if `validateChoice` had an option for `caseInsensitive`... (out of scope of this PR!)

packages/myst-cli/src/utils/addWarningForFile.ts
fwkoch: Also... maybe we don't want to tell the user how to suppress the message. That might make it too easy to abuse...? _But_ we definitely need to debug log the `ruleId`; that way, if they read the docs, they have all the info they need to suppress the message.

If we don't log the `ruleId` this whole feature feels inaccessible to users.

fwkoch (changes requested):
This is great! There's a couple necessary changes I think (log `ruleId`s to users; make sure `kind: undefined` doesn't make it to the warnings in the session store), as well as a bunch of other thoughts I had after marinating on this for a while.

packages/myst-config/src/errorRules/validators.ts
fwkoch: Being a little stricter on this would allow:

```yaml
error_rules:
  - rule: doi-check
    severity: ignore
    keys:
      - bibKey1
      - bibKey2
      - bibKey3
```

to coerce to:

```yaml
error_rules:
  - rule: doi-check
    severity: ignore
    key: bibKey1
  - rule: doi-check
    severity: ignore
    key: bibKey2
  - rule: doi-check
    severity: ignore
    key: bibKey3
```

Which I think is nice? Easy for the user, easy for the handling logic.

rowanc1:
@fwkoch thanks for the pair-programming to go through this. We addressed all the comments in the PR now I think, I have also gone through and updated all failing links.

docs/settings.md
```diff
@@ -52,3 +52,22 @@ beamer
...
+    - rule: math-eqnarray-replaced
+      severity: ignore
+    - rule: link-resolves
+      severity: ignore
+      keys:
+        - /known-internal-link
+        - https://flaky-connection.com
+```
+
+The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity listed will be adopted rather than the default log message severity. The default severity for rules included in the list is `ignore`, which means that simply listing the rule IDs as strings will ignore those rules. To discover the rule ID, run myst in debug mode to get the error (and optional key) printed to the console. For example, the above configuration updates will no longer warn on `math-eqnarray-replaced` and will also ignore the two links when running `myst build --check-links --strict` in continuous integration.
```
fwkoch, line 73: Updates to these docs are great 👍

fwkoch (approved):

PR #1119, opened 2024-04-17
🔌 Load plugins from all site projects, rather than just "current project"

fwkoch:
When loading plugins, we were previously only loading from `currentProjectConfig`. This PR updates the loading logic to also look at the `currentSiteConfig` and load all plugins from each of the site's projects.


stevejpurves (approved):
This change works well on a site with 2 sub projects, plugins in the first sub project are picked up correctly

PR #1121, opened 2024-04-17
Add execution to our documentation and test RTD previews

choldgraf:
This makes two improvements to our documentation:

1. Adds execution to our docs build with `--execute`. This way we can test execution-related changes in the docs.
	- @rowanc1 this needs to be activated in the live docs as well. As-is, this will only impact the docs PR previews.
2. Previews the docs build with ReadTheDocs to confirm that it works.
	- I'm adding a `robots.txt` file that I think will prevent indexing on RTD.

### DIFF
--- a/.github/robots.txt
+++ b/.github/robots.txt
@@ -1,0 +1,5 @@
+# This should disallow all search index crawling of the documentation.
+# we are using RTD *only* for PR previews, and not for hosting the actual docs.
+# So this should be added to the HTML output within readthedocs.
+User-agent: *
+Disallow: /
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,11 +8,13 @@ build:
   os: ubuntu-lts-latest
   tools:
     nodejs: "20"
+    python: "3.11"
   commands:
-    # Install and build our documentation
+    # Install dependencies and build MyST so we're using `main`
     - npm install
     - npm run build
     - npm run --workspace packages/mystmd link
+    - pip install -r docs/requirements.txt
     - myst --version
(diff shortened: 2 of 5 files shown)


.readthedocs.yaml
```diff
@@ -1,16 +1,25 @@
...
   tools:
     nodejs: "20"
   commands:
+    # Install and build our documentation
     - npm install
     - npm run build
     - npm run --workspace packages/mystmd link
     - myst --version
-    - cd docs && BASE_URL=${READTHEDOCS_CANONICAL_URL} myst build --html
+    - cd docs && BASE_URL=${READTHEDOCS_CANONICAL_URL} myst build --html --execute
```
rowanc1, line 17: We will also need to install the environment before here probably?

choldgraf: riiiight good point

.github/robots.txt
```diff
@@ -0,0 +1,5 @@
+# This should disallow all search index crawling of the documentation.
+# we are using RTD *only* for PR previews, and not for hosting the actual docs.
+# So this should be added to the HTML output within readthedocs.
+User-agent: *
+Disallow: /
```
rowanc1, line 5: It looks like we are generating this and RTD is serving this.

image

Also a bug.

docs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+jupyter
```
rowanc1, line 1: there was an environment.yml there, we should have one or the other, not fussed on which one!

choldgraf: oops - OK reverting to environment.yml

rowanc1:
Static builds on netlify are pretty straight forward. Getting the execution to run up there might be a bit harder, not really sure how much control we have!

PR #1125, opened 2024-04-17
📖 Add deployment docs on netlify

rowanc1:


choldgraf (approved):
This looks great, thanks for updating the deployment docs, it's nice to have another deploy target documented for folks!

PR #1128, opened 2024-04-18
🐛 Correctly parse tags in {code-cell}

rowanc1:
This now correctly parses tags for `{code-cell}` directives, and works for visibility. The theme still does not do `hide-` but `remove-input`/`remove-cell/`remove-output` now all work in markdown notebooks.

This means that things like this:

```
:::{code-cell}
:tags: remove-input
!myst build -h
:::
```

Now show the output nicely in the docs, with the `--execute` flag on. Super cool.

cc @choldgraf

### DIFF
--- a/.changeset/lucky-trees-doubt.md
+++ b/.changeset/lucky-trees-doubt.md
@@ -1,0 +1,5 @@
+---
+"myst-directives": patch
+---
+
+Correctly place tags in code-block
--- a/packages/myst-directives/src/code.spec.ts
+++ b/packages/myst-directives/src/code.spec.ts
@@ -1,7 +1,7 @@
 import { describe, expect, test } from 'vitest';
-import { getCodeBlockOptions } from './code.js';
+import { getCodeBlockOptions, parseTags } from './code.js';
 import { VFile } from 'vfile';
-import type { DirectiveData } from 'myst-common';
+import type { DirectiveData, GenericNode } from 'myst-common';
 
 function getCodeBlockOptionsWrap(options: DirectiveData['options'], vfile: VFile) {
   return getCodeBlockOptions({ name: '', options, node: {} as any }, vfile);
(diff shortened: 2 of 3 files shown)


choldgraf:
Ah nice, this confused me as well. I've added a quick example to document this so that others know how to control it. If that example looks correct to @rowanc1 then I say we merge!

docs/notebooks-with-markdown.md
```diff
@@ -50,6 +50,29 @@ phrase = f"{hello}, {there}!"
...
+````
+
+and results in the following:
+
+> ```{code-cell} python
+> :tags: "remove-input"
+> print("This will show output with no input!")
+> ```
+
+See [](#(notebooks:cell-visibility) for more information.
```
choldgraf, line 74:
```suggestion
See [](#notebooks:cell-visibility) for more information.
```

choldgraf (approved):
Looks good to me in the docs!

https://deploy-preview-1128--mystmd.netlify.app/notebooks-with-markdown#add-tags-to-code-cell-directives

One thing that's not clear to me (please update if you wanna quickly change this) is how you'd specify multiple tags. E.g. would it be `:tags: "one, two, three"`. Or would it be:

```
:tags:
  - one
  - two
  - three
```

?

I'll take your word for it on the tests :-D

choldgraf (dismissed):
Oops I spoke too soon, it looks like the cells on that page aren't being executed...could this be because `{code-cell}` is in a quote block?

https://deploy-preview-1128--mystmd.netlify.app/notebooks-with-markdown#add-tags-to-code-cell-directives

rowanc1:
They can be comma separated or parsed as yaml:

```
:tags: tag-1, tag-2
:tags: [tag-1]
```

Or

```
---
tags:
  - tag-1
  - tag-2
---
```

Right now I think that `"tag-with-quotes"` is not parsed correctly actually. Should add that case.

rowanc1:
This is a bit of a subtle change that was made is that single line `:key: values` are parsed as strings. Because we kept getting people erroring out on unexpected yaml stuff, especially in captions and labels with `#:` characters. I think we have docs on that!

I opted to support the current yaml parsing here to support jupyter-book out of the gate without changes. Maybe that should be under a flag?

choldgraf:
Let me know if the change in https://github.com/executablebooks/mystmd/pull/1128/commits/f396585a55eb52f7ced751849a827e012526fe9e looks correct and make edits as you wish!

choldgraf:
For some reason the execution is still not happening

rowanc1:
I am looking into that now, something to do with the blockquote.

rowanc1:
The bug is here, going to leave that one for a new day:
- #1134
I am going to un-nest this out of the block quote, which works and get this improvement in. :)

choldgraf:
yessss thanks

PR #1132, opened 2024-04-18
💅 🖼️ Add topic directive

agoose77:
This PR adds the RST `topic` directive, which behaves like an aside, with a stylised heading. 

Here's an example RST:
```rst
.. topic:: Bees Talk About Work Life Balance
  :class: 

   Bzzbzz bzz bzzbzz bzz bzzbzzbzzzzz!!!?? Bzz.
```

This generates the following HTML in Sphinx/docutils:
```html
<aside class="topic some-class">
  <p class="topic-title">About Bees</p>
  <p>Hello bees!</p>
</aside>
```

Our MDAST with this PR is
```json
{
  "type": "aside",
  "kind": "topic",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "text",
          "value": "Bees Talk About Work Life Balance"
        }
      ]
    },
    {
      "type": "paragraph",
      "children": [
        {
          "type": "text",
          "value": "Bzzbzz bzz bzzbzz bzz bzzbzzbzzzzz!!!?? Bzz."
        }
      ]
    }
  ],
  "class": "some-class"
}
```

I welcome any suggested changes here.

One extraneous change is the addition `class` and `label` (`identifier`) attributes to the `aside` AST node type. I _think_ it's our goal to add this to _all_ MyST nodes, right? Perhaps that would be done on the base `Node` type, rather than per-node like we currently do in `myst-spec`?


rowanc1:
In sphinx this looks like:

image

I think we should handle the header properly. I think that this should eventually be `title` as a node type, and be shared between admonitions, drop-downs, topics, exercises, etc. But for this PR we should adopt `admonitionTitle` and then follow up with fixing #1139.

As another bug, we are dropping other aside titles.

agoose77:
This PR avoids the new node type by assuming that the first child is the title. I don't hate that, but I recognise that we long-term want #1139

rowanc1:
Looking through this, I think we just want to merge with `aside` and expand to supporting optional titles in the others. Right now that is a bug, and then expanding that, both are the same anyways.

https://jupyterbook.org/en/stable/content/layout.html#sidebars-within-content

rowanc1:
Making it look like this:

image

I think there are some improvements to be made, but I think this is probably good enough for a first cut?

rowanc1:
Going to move ahead with this for now to queue up a release - we can always come back and make improvements.

PR #1142, opened 2024-04-19
🐛 Change CSV parser to use the browser

rowanc1:


rowanc1:
@agoose77 had trouble using this in the themes without this!

PR #1145, opened 2024-04-20
🐛 CSV Table directive parsing

rowanc1:
This fixes a failure I found in the live demo. @agoose77, you might have more insights as to why this happens?


agoose77:
It turns out that I read the documentation for `unist-util-select` wrong; I had thought it uses `undefined` for failing selectors, but it uses `null`.

agoose77:
Given that you opened this, and I found the same solution, I'm happy to merge.

PR #1149, opened 2024-04-22
:bug: Always bundle static links

### LINKED ISSUE #1135: `{download}` role for markdown source turns into a link
The download role turns into a link to the page, which is not the desired behaviour.

agoose77:
This fixes #1135 by scoping the URL rewriting to non-static links.

I think what was happening was that the URL was being re-written as an internal link, but the `static` attribute was causing MyST to try and resolve that re-written link as a static item.

This change now prevents rewriting for static items (we'll always want those in the `static` endpoint).

I didn't add a test yet - not sure where that should go, would appreciate some pointers.

### DIFF
--- a/packages/myst-cli/src/transforms/links.ts
+++ b/packages/myst-cli/src/transforms/links.ts
@@ -154,7 +154,8 @@ export class StaticFileTransformer implements LinkTransformer {
     const [linkFile, ...target] = linkFileWithTarget.split('#');
     const { url, title, dataUrl } =
       selectors.selectFileInfo(this.session.store.getState(), linkFile) || {};
-    if (url != null) {
+    // If the link is non-static, and can be resolved locally
+    if (url != null && !url.static) {
       // Replace relative file link with resolved site path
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/transforms/links.ts
```diff
@@ -154,7 +154,8 @@ export class StaticFileTransformer implements LinkTransformer {
     const [linkFile, ...target] = linkFileWithTarget.split('#');
     const { url, title, dataUrl } =
       selectors.selectFileInfo(this.session.store.getState(), linkFile) || {};
-    if (url != null) {
+    // If the link is non-static, and can be resolved locally
+    if (url != null && !url.static) {
```
rowanc1, line 158: URL here is a string? Maybe do you mean `link.static`?

```suggestion
    if (url != null && link.static !== true) {
```

agoose77: Yes ... how ... ?!

I've got a few PRs in flight, but I never recall writing `url.static` ha!

PR #1151, opened 2024-04-23
:ballot_box: Add new table-of-contents validator

### LINKED ISSUE #1102: Add a new table of contents format
## Proposal

Add a new table-of-contents format to live inside `myst.yml`. This will supersede the `_toc.yml` which is strongly constrained by the design of Sphinx (sphinx-external-toc).

## Additional notes
This format should be readable, easy to guess, and support styling markup (e.g. dropdown vs header)

agoose77:
In #1102, we identified the need for a new table of contents format.

This format must satisfy the following goals:
1. Easy to humanly read/write
2. Easy to programatically read/write
3. Expressive enough to represent complex trees

This PR is a WIP in that direction. The TOC schema can be build using 
```bash
npm run build:schema --workspace=packages/myst-toc
```
and then used with <https://www.jsonschemavalidator.net/> (or the CLI equivalent) to verify a TOC JSON blob.

Here's an example "complex" TOC:
```yaml
  - title: Main
    file: main.md
  - title: Overview
    children:
      - file: overview-1.md
      - file: overview-2.md
  - url: https://google.com
  - file: getting-started.md
    children:
      - file: getting-started-part-1.md
      - file: getting-started-part-2.md
```

There are notably two kinds of "subtrees" in MyST:
1. Non-hyperlink categories (grouping)
2. Categories with index pages (Sphinx-like)

This TOC supports both; a bare `children` entry requires a `title`, whereas a file/URL entry does not. 

@rowanc1 also mentioned the ability to choose between drop-downs and fixed-headings. This TOC format has a `class` string field that we could use to support that. 

We should iterate on this as necessary!

Closes #1102

### DIFF
--- a/packages/myst-toc/tests/test.yml
+++ b/packages/myst-toc/tests/test.yml
@@ -7,36 +7,44 @@ cases:
     content:
       - file: foo.md 
         title: Foo!
+  - title: Single file with part passes
+    content:
+      - file: foo.md 
+        part: abstract
+  - title: Single file with hidden passes
+    content:
+      - file: foo.md 
+        hidden: true
+  - title: Single file with invalid hidden fails
+    content:
+      - file: foo.md 
+        hidden: "yes"
+    throws: 'The given contents do not form a valid TOC'
   - title: Single file with children passes
     content:
       - file: foo.md 
-        children:
-          - file: bar.md
-  - title: Single file with title and children passes
-    content:
-      - file: foo.md 
-        title: Foo
         children:
           - file: bar.md
 
   - title: Single URL passes
     content:
       - url: https://bar.com/foo.md 
-  - title: Single URL with title passes
-    content:
-      - url: https://bar.com/foo.md 
-        title: Foo!
   - title: Single URL with children passes
     content:
       - url: foo.md 
         children:
           - url: https://bar.com/foo.md
-  - title: Single URL with title and children passes
+
+  - title: Single pattern passes
     content:
-      - url: https://bar.com/foo.md 
-        title: Foo
+      - pattern: foo/*.md 
+  - title: Single pattern with children passes fails
+    content:
+      - pattern: foo/*.md 
         children:
-          - url: https://bar.com/bar.md
+          - url: https://bar.com/foo.md
+    throws: 'The given contents do not form a valid TOC'
+
 
   - title: Single parent passes
     content:
--- a/packages/myst-toc/tests/test.yml
+++ b/packages/myst-toc/tests/test.yml
@@ -1,0 +1,63 @@
+title: Table of Contents
+cases:
+  - title: Single file passes
+    content:
+      - file: foo.md 
+  - title: Single file with title passes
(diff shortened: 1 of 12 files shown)


rowanc1:
Looking good, some thoughts below as I go through this.

Some of the things we can maybe choose to add:

- `id`: be able to reference a part of the toc
- `numbered`: (maybe `numbering`?) We have had a lot of requests to have full-book numbering rather than per page. This is a step in that direction, I think that `numbered` would be validated to `numbering` as well (defined in frontmatter already, important for `figure: start: 5` and `enumerator: 1.%s`), which has the enumerator to be used on the page.
- `part`? Something to indicate that this is a chapter, appendix, or dedication, etc. I think this is different than a class, which is about visual styles. This would be helpful for placing in templates.

Options like `glob` can be read at validation time, and coerced into this format. Also implicit folders, similar to the default behaviour now.

Is there some way to apply default classes? i.e. a top-level `kind: book-sections`? Or do we want people to write `class: heading` on each? Maybe this is also validation/coersion to this format? I am not sure what the ajv validation is used for in that case - if the user is never interacting with this format?

agoose77:
- We don't need a root document; we can use the _first_ document.
- Part should be singular, we can pluralise it later if needs be
- `numbering` needed, should inherit
- ID needed
- Normalise `pattern` entry type to array of files, don't use internally after normalisation.
- `hidden` needed, should inherit
- `class` and `classes` are plural vs singular

agoose77:
@rowanc1 I've updated this PR to add `id`, `hidden`, `part` fields to all entries. We _don't_ have `class` on all entries; right now it's limited to parents. Is that reasonable, or do we have need of it on every entry?

agoose77:
@rowanc1 the schema generator has a dependency on NodeJS>=18 due to `structuredClone`. What are your thoughts here; it's only a development dependency, we probably want to try and patch it right?

rowanc1:
I think class makes sense on every node? Thinking about that `beta-feature` example with a custom icon in the TOC which exists only on a child.

rowanc1:
I will take it for a spin this afternoon and get it in and released!

packages/myst-toc/tests/test.yml
```diff
@@ -0,0 +1,69 @@
...
+        class: beta
+  - title: Single file with hidden passes
+    content:
+      - file: foo.md 
+        hidden: true
+  - title: Single file with invalid hidden fails
+    content:
+      - file: foo.md 
+        hidden: "yes"
+    throws: 'The given contents do not form a valid TOC'
```
rowanc1, line 26: It would be nice to get nicer errors here to show to users. I suspect when we integrate this into myst-frontmatter, we will use the `simple-validators` as that does now, which will allow us to have coercion and better validation messages that work with where we are at now.

rowanc1 (approved):
The types look good here. I think things will change a bit when we integrate it with `myst-frontmatter`, but that can happen in the next iteration.

PR #1153, opened 2024-04-23
🧹 Improve MyST CLI interfaces for export

fwkoch:
This is not a pretty PR with a large diff, but it's scope is relatively small and works to make things nicer for developers in a few ways:

1. `Options` types are cleaned up so they have useful names and are appropriately reused - no more amalgamated `Options` types that serve 3 purposes.
2. Export functions that are not used anywhere in the myst ecosystem  (but are still used externally) are moved to a `legacy` file
3. `myst-cli` exports some reusable cli interfaces, so parallel tools that consume myst can stay in-sync with available options.


packages/myst-cli/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './build/index.js';
+export * from './cli/index.js';
```
rowanc1, line 2: Nice -- good to see these being made more reusable. 🚀

packages/myst-cli/src/build/legacy.ts
```diff
@@ -0,0 +1,385 @@
...
+  watch?: boolean;
+  throwOnFailure?: boolean;
+  ci?: boolean;
+  renderer?: RendererFn;
+};
+
+// These "local article" export functions are not used across the MyST ecosystem.
+// However, they are still maintained as exports for external libraries.
+
+export async function localArticleToWord(
```
fwkoch, line 42: Add deprecated annotation to these as well.

rowanc1 (approved):
Spent a few hours going through this with @fwkoch in person. And have tested init/build/start in myst and no problems found.

I think the only thing to add is the `@deprecated` label on the legacy functions.

packages/myst-cli/src/index.ts
rowanc1: It is really good to see the options actually correct as well.

PR #1154, opened 2024-04-23
🎨 Support theme variables for static site exports

rowanc1:
See executablebooks/myst-theme#370

### DIFF
--- a/packages/myst-cli/src/build/html/index.ts
+++ b/packages/myst-cli/src/build/html/index.ts
@@ -42,7 +42,7 @@ export async function currentSiteRoutes(
           };
         }),
         // Download other assets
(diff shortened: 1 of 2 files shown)


rowanc1:
This is a very baby step towards custom CSS, and starting with dynamic CSS file that is based on the site config.

packages/myst-cli/src/build/html/index.ts
```diff
@@ -42,7 +42,7 @@ export async function currentSiteRoutes(
           };
         }),
         // Download other assets
-        ...['favicon.ico', 'robots.txt'].map((asset) => ({
+        ...['favicon.ico', 'robots.txt', 'myst-theme.css'].map((asset) => ({
```
rowanc1, line 45: These should eventually be handled by the theme.

PR #1156, opened 2024-04-24
🧮 Frontmatter in `include` files removed, math/abbreviations maintained

fwkoch:
Files loaded in the `include` directive previously retained their frontmatter in the `mdast` tree, resulting in an unwanted `yaml` code block. Additionally, everything in the `include` frontmatter was ignored (mostly we do want to ignore it, but for `.tex` source files, `include`s may have math macros, etc, that we need to inherit).

Now, we reuse the file loading functions from `loadFile` in the `include` directive. These functions remove and give access to file frontmatter. That means no more yaml blocks and we can inherit `math` and `abbreviations` into the parent file's frontmatter.

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -4,7 +4,9 @@ import { createHash } from 'node:crypto';
 import { tic } from 'myst-cli-utils';
 import { TexParser } from 'tex-to-myst';
 import { VFile } from 'vfile';
+import type { GenericParent } from 'myst-common';
 import { RuleId, toText } from 'myst-common';
+import type { PageFrontmatter } from 'myst-frontmatter';
 import { validatePageFrontmatter } from 'myst-frontmatter';
 import { SourceFileKind } from 'myst-spec-ext';
 import { frontmatterValidationOpts, getPageFrontmatter } from '../frontmatter.js';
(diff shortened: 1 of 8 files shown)


fwkoch:
This begins to address the "latex subfile" task here: https://github.com/econ-ark/REMARK/issues/152

fwkoch:
Also, as a side effect, this addresses the first half of https://github.com/executablebooks/mystmd/issues/1152 - now, when providing a non-existent file at the CLI, you are told it does not exist. (Complaints about the extension will only surface if the file _exists_ but has a bad extension. Anyway - still more to do on that one...)

packages/myst-transforms/src/frontmatter.ts
```diff
@@ -70,7 +77,7 @@ export function getFrontmatter(
   if (nextNodeIsH1 && !titleNull) {
     const title = toText(nextNode.children);
     // Only remove the title if it is the same
-    if (frontmatter.title && frontmatter.title === title) {
+    if (frontmatter.title && frontmatter.title === title && !opts.keepTitleNode) {
```
fwkoch, line 80: I think an option like this existed before, and I got rid of it a little hastily... Now, we can once again leave the title node in the tree. Useful for `include` files where we almost certainly do not want to remove headings (frontmatter title for those files is totally lost).

packages/myst-cli/src/process/file.ts
```diff
@@ -63,82 +138,34 @@ export async function loadFile(
   }
   // ensure forward slashes and not windows backslashes
   location = location.replaceAll('\\', '/');
-  const vfile = new VFile();
-  vfile.path = file;
 
   try {
+    const content = fs.readFileSync(file).toString();
```
fwkoch, line 143: Error early on `file does not exist`.

rowanc1:
Tested locally, works well!

PR #1162, opened 2024-04-25
:bucket: Bump minimum NodeJS version

agoose77:


rowanc1:
There is another place in the session that compares npm and node versions. On my phone at the moment. I think also the install node pages might need an update?

agoose77:
Thanks, found the remaining items I *think* :crossed_fingers:

rowanc1:
A few more places:
- https://github.com/executablebooks/mystmd/blob/main/packages/mystmd-py/src/mystmd_py/main.py#L17
- https://github.com/executablebooks/mystmd/blob/main/packages/mystmd-py/src/mystmd_py/main.py#L30
- https://github.com/executablebooks/mystmd/blob/main/packages/mystmd-py/src/mystmd_py/main.py#L32
- https://github.com/executablebooks/mystmd/blob/main/docs/installing-prerequisites.md#L12

agoose77:
Next commit should've done those I think.

rowanc1:
Ok, I think this is good - a changeset of a `minor` for the CLI?

If we turn this into a warning rather than an exit, maybe that is better, but then we would need to test in 16 and not sure how to do that if we can't build in it.

agoose77:
Yeah I think we just make a minor bump. Now's the time to make these kind of changes vs in a year's time (i.e., we can move a bit fast and break some things).

rowanc1:
Ok - can you add the changeset! :)

packages/mystmd-py/src/mystmd_py/main.py
```diff
@@ -27,9 +27,9 @@ def main():
         raise SystemExit(f"MyST could not determine the version of Node.js: {_version}")
 
     major_version = int(major_version_match[1])
-    if not (major_version in {16, 18} or major_version >= 20):
+    if not (major_version in {18, 20} or major_version >= 22):
```
rowanc1, line 30: 22 is fine?

```suggestion
    if not (major_version in {18, 20, 22} or major_version > 22):
```

agoose77: Yep, just preserving the existing logic!

rowanc1:
Works well locally:

image

Issue #1161, opened 2024-04-25
Html to Latex regression 1.1.56


dellaert:
I noticed a regression when upgrading from 1.1.55 to 1.1.56, in that now the HTML output in notebooks no longer renders.

$ myst -v
v1.1.56

## Description

`myst build --tex` now throws these errors whereas in 1.1.55 this worked fine
```
⛔️ Unhandled LaTeX conversion for node of "div"
⛔️ Unhandled LaTeX conversion for node of "span"
```

The HTML conversion was put in I think by @rowanc1 to support html output from code in roboticsbook.org, in the conversion to latex. But I guess it broke here :-)

## Proposed solution

 I'll revert to 1.1.55 for now. I believe @rowanc1 will probably know exactlhy what happened and how to fix.

agoose77:
@dellaert Rowan will likely know the context of this, but could you share a reproducer so that I might take a look?

rowanc1:
We did change the HTML parsing to now preserve divs, but potentially there is something else going on. I will get the latex parsing back up and working with these changes shortly!

dellaert:
Thanks @rowanc1 
@agoose77 A full repro is hard for me but an example is https://github.com/gtbook/robotics/blob/main/S31_vacuum_state.ipynb. The output of `pretty`. (Rowan has access to a repo in which the entire book is rendered, which he helped set up).

rowanc1:
I will take a look now @dellaert!

rowanc1:
This will be released tomorrow!

PR #1163, opened 2024-04-25
🧱 Basic support for div/span in typst and LaTeX

rowanc1:
See #1161

### DIFF
--- a/.changeset/heavy-flies-dress.md
+++ b/.changeset/heavy-flies-dress.md
@@ -3,4 +3,4 @@
 "myst-to-tex": patch
 ---
 
-Support dvi/span in typst/tex
+Support `div` and `span` in typst and tex
--- a/.changeset/heavy-flies-dress.md
+++ b/.changeset/heavy-flies-dress.md
@@ -1,0 +1,6 @@
+---
+"myst-to-typst": patch
+"myst-to-tex": patch
+---
+
+Support dvi/span in typst/tex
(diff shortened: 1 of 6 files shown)


packages/myst-to-typst/src/index.ts
```diff
@@ -341,6 +348,12 @@ const handlers: Record<string, Handler> = {
   //     );
   //   }
   // },
+  div(node, state) {
+    state.renderChildren(node, 1);
+  },
+  span(node, state) {
+    state.renderChildren(node, 0, { trimEnd: false });
```
fwkoch, line 355: I was wondering the justification for `trimEnd: false` but `trailingNewLines: 0`... I think, though, this implies we are in a paragraph (so no new lines) and the intended trailing whitespace is just spaces (which we want to keep).

This feels slightly risky - we have less control over the `typst` output in cases where we get strange `myst` - but then again, adding something like `trailingSpaces` feels maybe too micromanage-y.

Probably fine as is... I suspect the use-cases for `span` are pretty narrow.

rowanc1: Yeah, I think that the short-term way to go is to have a `renderInline` that does the children, without the line ending stuff.

.changeset/heavy-flies-dress.md
```diff
@@ -0,0 +1,6 @@
+---
+"myst-to-typst": patch
+"myst-to-tex": patch
+---
+
+Support dvi/span in typst/tex
```
rowanc1, line 6:
```suggestion
Support `div` and `span` in typst and tex
```

Issue #1026, opened 2024-03-22
Parse markdown code-cell outputs as part of the document AST


choldgraf:
Jupyter notebook cells can produce text output that is specific to markdown (`text/markdown`). We should parse this output along with the rest of the markdown in the notebook. This would allow for **programmatic generation of notebook content**, and along with the `{embed}` directive could allow you to nicely stitch together MyST content via any Jupyter Kernel.

### An example from Jupyter Book 1

For example:

<img src="https://github.com/executablebooks/mystmd/assets/1839645/1bdd6f58-d437-4c63-bdd5-03c173e2b3d9" width=300 />

As an example, [the MyST-nb Sphinx extension documents this functionality here](https://myst-nb.readthedocs.io/en/latest/render/format_code_cells.html#markdown).

### An example from our docs where this would be useful

In the admonitions docs here:

https://github.com/executablebooks/mystmd/blob/c0f51ee51890158081736722cb01bcf510699786/docs/admonitions.md?plain=1#L34-L98

We have a big list of all the different types of admonitions for demonstration. It's just boilerplate myst repeated over and over. If we parsed Markdown outputs as MyST, we could replace it with something like:

``````python
inner = []
for admonition in ["note", "tip", "warning"]:
	s = """
	````{tab-item}
    ```{%s}
    This is an %s admonition. 
    ```
    ````
    """ % admonition
	inner.append(s)
template = """
`````{tab-set}
%s
````` % "\n".join(inner)
"""
``````

### Related issues

- https://github.com/jupyter-book/mystmd/issues/1633 is about generating AST directly rather than `text/markdown`

HelgeGehring:
continuing #1559 here:

An example for this would be the following:

- Right now I have a nice notebook which analyzes exactly one physical structure.
- The analysis results in several graphs / tables / images / ..., each generated by a code cell

My problem now is, that I have 2-20 different structures. For each I get graphs based on the same code.
Right now the only solution is to put them all below each other, leading to a quite long doc.

Better would be to have
- tabs in which I can select which structure I want to see

and/or

- a table which has only the main results (maybe broken down in a number or two), but on click on a row, I see a detailed analysis for that structure.


A simplified example might be statistics about bank accounts (not scientific, but less abstract).
- It's simple to write a notebook which produces graphs / tables, ... based on an account number
[...]

PR #1167, opened 2024-04-29
Demonstrate markdown outputs and plotly support

choldgraf:
Per a question from @rowanc1 - here are two features missing that I'd need before deploying mystmd at 2i2c.org/kpis.

1. ~Plotly support. Plotly requires its own JS bundle and can be hard to get properly. JupyterBook [recommends that you manually link the require.js package](https://jupyterbook.org/en/stable/interactive/interactive.html#plotly) but I don't believe this is possible in mystmd yet.~ This is possible after all! I debugged it and learned that I needed to explicitly call `fig.show`. I've updated the docs now :-)
2. Markdown outputs parsed as myst. Currently the markdown outputs aren't parsed as anything (though weirdly, they _are_ in JupyterLab MyST).

Both of these can be demoed in this PR at the docs preview link below :-) 

[Deploy preview link](https://deploy-preview-1167--mystmd.netlify.app/)

CleanShot 2024-04-29 at 16 12 30@2x


agoose77:
Requires fixing https://github.com/executablebooks/mystmd/issues/1026

choldgraf:
Closing this PR because it's not actionable. Here's the issue that will need to be resolved in the future:

- #1026

PR #1172, opened 2024-04-30
⚙️  Use simple validators in TOC

agoose77:
As we want to integrate TOC validation with our config system, and raise informative errors, this PR drops schema generation in favour of explicit imperative validation (until we get around to moving to a schema/ORM based approach).


packages/myst-toc/package.json
```diff
@@ -20,20 +20,19 @@
...
   "scripts": {
-    "clean": "rimraf dist ./src/schema.json",
+    "clean": "rimraf dist",
     "lint": "eslint \"src/**/!(*.spec).ts\" -c ./.eslintrc.cjs",
     "lint:format": "npx prettier --check \"src/**/*.ts\"",
-    "test": "npm-run-all build:schema && vitest run",
+    "test": "vitest run",
     "test:watch": "vitest watch",
     "build:esm": "tsc",
-    "build:schema": "npx ts-json-schema-generator --path src/types.ts --type TOC -o ./src/schema.json",
```
rowanc1, line 29: I feel like we could keep this around, even if we aren't using it?

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+  if (!outputEntry) {
+    return undefined;
+  }
+
+  return outputEntry as ParentEntry;
+}
+
+export function validateEntry(entry: any, opts: ValidationOptions): Entry | undefined {
+  if (defined(entry.file)) {
+    return validateFileEntry(entry, opts);
```
rowanc1, line 172: I think there might be some way to pass the context in for the opts and say that we are parsing this as a file entry?

packages/myst-toc/tests/examples.spec.ts
```diff
@@ -1,49 +1,121 @@
-import { describe, test, expect } from 'vitest';
-import fs from 'node:fs';
-import path from 'node:path';
-import yaml from 'js-yaml';
```
rowanc1, line 4: Would love to see these remain as yaml tests if possible, much easier to read/change/update, not a big deal though.

rowanc1 (commented):
Verbosity 🚀. @fwkoch I think your review here would be helpful.

packages/myst-toc/package.json
agoose77: Sure but we don't publish it yet. I'm thinking let's add it later?

packages/myst-toc/src/toc.ts
agoose77: Should we not just export `validateFileEntry`?

packages/myst-toc/tests/examples.spec.ts
agoose77: I usually feel that way. However, here I like the ease of parametrisation and the complexity (match error vs match result etc) is such that I think code is more maintainable.

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+  validateList,
+  validateNumber,
+  validateObjectKeys,
+  validateString,
+  validationError,
+} from 'simple-validators';
+
+const COMMON_ENTRY_KEYS = ['title', 'hidden', 'numbering', 'id', 'part', 'class'];
+
+function validateCommonEntry(entry: any, opts: ValidationOptions): Entry | undefined {
```
fwkoch, line 18:
```suggestion
function validateCommonEntry(entry: Record<string, any>, opts: ValidationOptions): Entry {
```
Just for slight type safety/accuracy

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+    opts,
+  );
+  if (!outputEntry) {
+    return undefined;
+  }
+
+  outputEntry.file = validateString(outputEntry.file, incrementOptions('file', opts));
+
+  outputEntry = validateCommonEntry(outputEntry, opts);
+  if (!outputEntry) {
```
fwkoch, line 62: Don't think this is possible with how `validateCommonEntry` is defined (see the corresponding type updates I made).

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+      optional: [...COMMON_ENTRY_KEYS, 'children'],
+    },
+    opts,
+  );
+  if (!outputEntry) {
+    return undefined;
+  }
+
+  outputEntry.file = validateString(outputEntry.file, incrementOptions('file', opts));
+
```
fwkoch, line 60:
```suggestion
if (!outputEntry.file) return undefined
```

We should have this here - if `file` is defined but invalid we want to short-circuit, not return ann invalid `FileEntry`.

fwkoch: Usually how I do this, so types are a little more helpful, is have an intermediate object:

- input `entry: any`
- then, like `const entryAsObject: Record<string, any>` as the result of `validateObjectKeys`
- finally `const outputEntry: FileEntry` as the final output

With that, we no longer have to cast `as FileEntry` at the end of the function, and instead get nice type checks throughout. Hopefully that makes sense...

Something like this: https://github.com/executablebooks/mystmd/blob/main/packages/myst-config/src/site/validators.ts#L51-L62 - which has `input`, intermediate `value`, and final, typed `output`.

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+      required: ['url'],
+      optional: [...COMMON_ENTRY_KEYS, 'children'],
+    },
+    opts,
+  );
+  if (!outputEntry) {
+    return undefined;
+  }
+
+  outputEntry.url = validateString(outputEntry.url, incrementOptions('url', opts));
```
fwkoch, line 90: `validateURLEntry`/`validatePatternEntry`/`validateParentEntry` all follow the same pattern as `validateFileEntry`, so my comments there apply to these as well.

packages/myst-toc/src/toc.ts
```diff
@@ -1,21 +1,193 @@
...
+  outputEntry = validateCommonEntry(outputEntry, opts);
+  if (!outputEntry) {
+    return undefined;
+  }
+
+  return outputEntry as ParentEntry;
+}
+
+export function validateEntry(entry: any, opts: ValidationOptions): Entry | undefined {
+  if (defined(entry.file)) {
```
fwkoch, line 171: This should have a `validateObject(entry, opts)` at the top, so we don't get errors on `entry.file`, etc.

fwkoch (changes requested):
Looks nice - I suggested some changes to tighten up the types so we don't accidentally let errors sneak through. Let me know if you have questions...

packages/myst-toc/src/toc.ts
agoose77: Interesting - I left this as `| undefined` to give us future scope to change that ... but, it's actually not public so there's no need!

agoose77:
@fwkoch I've made some edits following the suggestions that you and @rowanc1 made! Would you mind another glance and then merging if happy?

packages/myst-toc/package.json
```diff
@@ -20,20 +20,19 @@
...
   },
   "bugs": {
     "url": "https://github.com/executablebooks/mystmd/issues"
   },
   "dependencies": {
-    "ajv": "^8.12.0"
+    "simple-validators": "^1.0.4"
   },
   "devDependencies": {
     "ts-json-schema-generator": "^1.5.0"
```
rowanc1, line 38: We should remove this if we are not building the schema!

fwkoch (approved):

PR #1177, opened 2024-05-01
✏️ Rename Toc -> TOC

agoose77:
This is a clean-up PR, before we start doing surgery on TOC handling.

### DIFF
--- a/.changeset/wet-beds-change.md
+++ b/.changeset/wet-beds-change.md
@@ -1,0 +1,7 @@
+---
+"myst-common": minor
+"myst-cli": minor
+"mystmd": patch
+---
+
+Rename Toc to TOC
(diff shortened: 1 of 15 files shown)


.changeset/wet-beds-change.md
```diff
@@ -0,0 +1,7 @@
+---
+"myst-common": minor
+"myst-cli": minor
```
fwkoch, line 3: I think this is the most notable part of this PR - making a breaking change for some naming consistency.

fwkoch (approved):
I don't feel super strongly about this - the naming consistency is nice; the breaking change is not so nice. However, if we are making these changes _sometime_, sooner is better than later!

agoose77:
Thanks Franklin! I'm taking that as merge approval :D

rowanc1:
@agoose77 bug I found:

```
myst init --write-toc
```

no longer works.

Will make a patch.

PR #1182, opened 2024-05-07
🐛 `--write-toc` case bug

rowanc1:
Fixes case bug introduced in #1177


packages/mystmd/src/options.ts
```diff
@@ -8,7 +8,9 @@ export function makeWriteTOCOption() {
   return new Option(
     '--write-toc',
     'Generate editable _toc.yml file for project if it does not exist',
-  ).default(false);
+  )
+    .default(false)
+    .implies({ writeTOC: true });
```
agoose77, line 13: I learned something today :D

PR #1183, opened 2024-05-07
📐 Replace xref text when using angle brackets

fwkoch:
If you define a cross-project reference like so `[](xref:key#identifier)`, it will be filled in with relevant text from the target. However, like this `<xref:key#identifier>`, the text remains "xref:key#identifier"

With this PR, that default text is removed, allowing relevant text to be filled later.


rowanc1:
The format text happened before the transform, it can now happen after. The formatter add places for line breaks in long URLs, and the text no longer matches.

PR #1185, opened 2024-05-07
Document link option with card

### LINKED ISSUE #1184: docs on linkable `card` are misleading
The docs here: https://mystmd.org/guide/dropdowns-cards-and-tabs#cards

mentions a `link` argument, but provides no example and the `card` directive reference subsection does not list card as an argument or option. Also `link` appears to actually be an option

choldgraf:
This documents the `:link:` option for cards and adds it to the short reference below.

I tried to use a reference to the card directive like `` {myst:directive}`card` ``, but that didn't resolve for some reason. Maybe that's why the reference is manually included in the page?

Either way:

- closes #1184

### DIFF
--- a/docs/dropdowns-cards-and-tabs.md
+++ b/docs/dropdowns-cards-and-tabs.md
@@ -41,7 +41,7 @@ Card content
 ```
 ````
 
-You can also add a `link` argument to the card, which will allow you to make the entire card clickable:
+You can also add a `link` option to the card, which will allow you to make the entire card clickable:
 
 ````{myst}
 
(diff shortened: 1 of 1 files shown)


docs/dropdowns-cards-and-tabs.md
```diff
@@ -41,13 +41,16 @@ Card content
 ```
 ````
 
-You can also add a `link` argument to the card, which will allow you to make the entire card clickable.
+You can also add a `link` argument to the card, which will allow you to make the entire card clickable:
```
rowanc1, line 44:
```suggestion
You can also add a `link` option to the card, which will allow you to make the entire card clickable:
```

rowanc1:
I think that technically the carg/grid/tabs are "extensions" and aren't loaded currently. This is a great improvement, thanks @choldgraf.

PR #1191, opened 2024-05-09
📺 Prefer video in web build if available

rowanc1:
This is possible to use the `image.*` as the src, if that is the case, an mp4 will be used by default for the web, but png for the static exports.


choldgraf:
Is that documented?! That is cool

rowanc1:
Yes! Actually! 🚀 

https://mystmd.org/guide/figures#multiple-images

PR #1194, opened 2024-05-09
[FIX] 🧊 don't overwrite `node.data`, append new fields

stevejpurves:
In certain transforms, the `node.data` object was being completely replaced. This behaviour prevents upstream plugins (e.g. user myst plugins in individual projects) from including additional data on a node.

My understanding of the role of the data field is to enable a "userdata" type store where arbitrary fields could be added, it's purpose mainly to enable meaningful storage of data that would otherwise be buried in a node's children, body or value, etc...

This PR fixes a specific issue in the `doi` transform that was overwriting user data. It also includes similar changes in other places to preserve any existing `node.data` where similar assignments were being made, although I don't have a failure case to test for those changes and they need review.

### DIFF
--- a/packages/myst-transforms/src/links/doi.ts
+++ b/packages/myst-transforms/src/links/doi.ts
@@ -30,7 +30,7 @@ export class DOITransformer implements LinkTransformer {
       return false;
     }
     link.url = doiUrl;
-    link.data = { doi: doi.normalize(doiUrl) };
+    link.data = { ...link.data, doi: doi.normalize(doiUrl) };
     link.internal = false;
     updateLinkTextIfEmpty(link, '');
     return true;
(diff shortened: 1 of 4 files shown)


packages/myst-transforms/src/links/doi.ts
```diff
@@ -30,7 +30,7 @@ export class DOITransformer implements LinkTransformer {
       return false;
     }
     link.url = doiUrl;
-    link.data = { doi: doi.normalize(doiUrl) };
+    link.data = { ...link.data, doi: doi.normalize(doiUrl) };
```
stevejpurves, line 33: This was the change that fixed my immediate upstream issue, and has been tested. The `doi` field is added alongside other userdata.

rowanc1:
Thank you!

PR #1197, opened 2024-05-10
🔗 Support project cross references in `embed`

fwkoch:
Previously, project cross references were enabled with `[](xref:key#identifier)`. Now, you can also embed this content with

```markdown
:::{embed} xref:key#identifier
:::
```

This content gets modified to indicate that the source location has a different base URL, so links/xrefs/etc resolve correctly.

This PR includes a number of other related improvements:
- project cross references allow cite style, e.g. `@xref:key#identifier`
  - This required some reordering and reorganization of link/cross-reference resolution
- slightly better caching - centralized logic, `maxAge` - this de-risks the reordered link resolution a bit (we run the link transforms more often, but results are more reliably cached)
- local cross references with page, e.g. `[](page#identifier)` now resolve to the same type of cross references as identifier-only, e.g. `[](#identifier)`
- embed nodes with page also resolve, e.g.
  ```markdown
  :::{embed} page#identifier
  :::
  ```
  Previously those just did not work at all.


rowanc1:
Testing now @choldgraf ... hopefully getting this out this eve! 🚀

PR #1198, opened 2024-05-10
🦁 Add ROR hover links

rowanc1:
https://github.com/executablebooks/mystmd/assets/913249/260e7642-fdb3-4ae7-9026-a5ca53c23e07

### DIFF
--- a/packages/myst-cli/src/transforms/ror.ts
+++ b/packages/myst-cli/src/transforms/ror.ts
@@ -1,0 +1,133 @@
+import fs from 'node:fs';
+import { join } from 'node:path';
+import type { Link } from 'myst-spec';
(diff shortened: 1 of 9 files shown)


packages/myst-cli/src/transforms/ror.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (!response || !response.ok) {
+    session.log.debug(`ror.org fetch failed for ${url}`);
+    return undefined;
+  }
+  const data = (await response.json()) as RORResponse;
+  return data;
+}
+
+/**
+ * Fetch CSL-JSON formatted metadata for the given doi.org DOI
```
rowanc1, line 54:
```suggestion
 * Fetch ROR JSON formatted metadata for the given ID
```

PR #1213, opened 2024-05-17
:abc: Update Jupyter case and typos

agoose77:
This small PR just fixes some typos in our documentation.


agoose77:
I'm merging this as it's evidently "correct", and I know that we're short on bandwidth for review right at this moment in time.

PR #1215, opened 2024-05-17
✨ Add `extends` field to myst.yml for composing multiple yamls

fwkoch:
This PR allows users to have project frontmatter defined in multiple files (all of the same `myst.yml` format), then the main `myst.yml` can bring in those other files under the top level `extends` key. A small example has been added to the docs.

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -55,7 +55,7 @@ project:
 ```
 
 (composing-myst-yml)=
-:::{note} Composing multiple `.yml` files
+#### Composing multiple `.yml` files
 
 You may separate your frontmatter into multiple, composable files. To reference other files from your main `myst.yml` file, use the `extends` key with relative path(s) to the other configuration files:
 
@@ -71,7 +71,6 @@ extends:
 Each of these files listed under `extends` must contain valid `myst.yml` structure with `version: 1` and `site` or `project` keys. They may also have additional files listed under `extends`.
 
 Composing files together this way allows you to have a single source of truth for project frontmatter that may be reused across multiple projects, for example math macros or funding information.
-:::
 
 +++
 
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -54,6 +54,25 @@ project:
   open_access: true
 ```
 
+(composing-myst-yml)=
+:::{note} Composing multiple `.yml` files
+
+You may separate your frontmatter into multiple, composable files. To reference other files from your main `myst.yml` file, use the `extends` key with relative path(s) to the other configuration files:
+
+```yaml
+version: 1
+site: ...
+project: ...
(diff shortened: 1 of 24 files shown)


stevejpurves (changes requested):
looks good! would perhaps just add some tests around existing Page->Project frontmatter inheritance, since the function that was doing that previously looks like it has been made generic? and is now doing Project->Project as well as Page->Project filling...

stevejpurves (approved):

docs/frontmatter.md
```diff
@@ -54,6 +54,25 @@ project:
   open_access: true
 ```
 
+(composing-myst-yml)=
+:::{note} Composing multiple `.yml` files
```
choldgraf, line 58: If the TOC PR gets in before this one is finished:

- https://github.com/executablebooks/mystmd/pull/1188

I'd recommend using that as an example here. I think a common use-case will be to have a separate `toc.yml` file and use this feature to bring it in (for communities that have a _lot_ of TOC content and thus don't want to have to scroll through a huge list of pages in their config file)

choldgraf, line 58: Could this be a subsection instead of a note? That way it is easier to discover in the table of contents

docs/frontmatter.md
```diff
@@ -54,6 +54,25 @@ project:
...
+
+You may separate your frontmatter into multiple, composable files. To reference other files from your main `myst.yml` file, use the `extends` key with relative path(s) to the other configuration files:
+
+```yaml
+version: 1
+site: ...
+project: ...
+extends:
+  - ../macros.yml
```
choldgraf, line 67: Does this work for externally-defined `.yml` files? (e.g. in a different github repository?)

If not, then would it be possible to download external files locally _as part of the build process_ in a way that would work as expected here?

agoose77:
I was _just_ thinking of looking at this in the coming week. How exciting!!!

docs/frontmatter.md
fwkoch: Currently the files must be available locally, from a relative path at build-time (they _can_ be outside the MyST project).

I agree we should be able to also put a url here, e.g.:
```yaml
extends:
  - https://example.com/my-other-myst.yml
```

We get into a discussion of this over here: https://github.com/executablebooks/mystmd/issues/336

But I think the url download/caching is a separate (smaller) PR.

docs/frontmatter.md
fwkoch: That makes sense - I'll update, depending on the order these come in...

I also suspect we will be scattering this all around the docs. E.g. where we talk about math macros, we can say "reference external macros by..." and where we talk about the toc, "reference external toc by..." etc. So hopefully this one little section won't be the only mention in the docs, long term.

PR #1218, opened 2024-05-20
🧮 Give title/description to math macros

fwkoch:
You may now define `title/description` for math macros in frontmatter, e.g.
```yaml
math:
  '\aMat':
    title: Custom Macro
    description: Macro I made up only for the sake of this example
    macro: '\mathrm{a}'
```

Defining a simple command:macro pair is also still valid:
```yaml
math:
  '\aMat': '\mathrm{a}'
```

This PR also updates the typst exports to resolve recursive math commands prior to filling them in to the document. Latex renderers handle recursive commands just fine, but typst does not so we must manually resolve them early.

Finally, this makes one small change to `simple-validators` where `validateObject` no longer validates arrays (these still have `typeof 'object'` so they passed before). Only allowing actual `Record<string, any>`-style objects has always been the expected behaviour, so I consider this a bug fix as opposed to a breaking change.


packages/myst-to-typst/src/math.ts
```diff
@@ -1,44 +1,57 @@
...
-    const key = texToTypst(k);
-    if (plugins[key]) return;
-    pluginsList.forEach(([, value]) => {
-      if (value.includes(key)) addedPlugins[key] = texToTypst(v);
-    });
-  });
-  const newPlugins = { ...addedPlugins, ...plugins };
-  if (Object.keys(addedPlugins).length === 0) return newPlugins;
-  return withRecursiveCommands(state, newPlugins);
+export function resolveRecursiveCommands(plugins: MathPlugins): MathPlugins {
```
rowanc1, line 35: The otehr way to go is to actually write these out to typst?

packages/mystmd/tests/math-macros/outputs/index.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD with MathML3 v1.3 20210610//EN" "http://jats.nlm.nih.gov/publishing/1.3/JATS-archivearticle1-3-mathml3.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ali="http://www.niso.org/schemas/ali/1.0/" dtd-version="1.3" xml:lang="en" id="index">
+<front>
+<article-meta>
+<title-group>
+<article-title>Testing Math Plugins</article-title>
```
rowanc1, line 7: Are the JATS macros denormalized now? i.e. the latex and mathml both do not have macros?

rowanc1 (commented):
Looking through, looks good overall, and excited to see these improvements!

packages/myst-to-typst/src/math.ts
rowanc1: In talking with @fwkoch this is dependent on the order, and that isn't something we want to figure out now.

Issue #1220, opened 2024-05-21
Embedded notebook content is no longer executable


fwkoch:
## Description

Previously, embedding notebook output using the `embed` or `figure` directive allowed execution directly on the page where embedded.

In the latest release, this interactivity was accidentally disabled: https://github.com/executablebooks/mystmd/releases/tag/mystmd%401.2.2 - Updates to embedding for cross-project references slightly changed the document structure, but the theme was reliant on the old structure for enabling execution.

## Proposed solution

It should be simple to restore the old structure (this means removing an extra, invisible `block` around the embedded content).

choldgraf:
- closed by https://github.com/executablebooks/mystmd/pull/1221

PR #1221, opened 2024-05-21
🐛 Fix embed mdast structure for executable content

fwkoch:
The latest release disabled execution of embedded jupyter outputs. This was caused by a small structural change to the mdast - an extra `block` node. This PR removes that extra node (and adds a test for the correct mdast structure!).

This addresses #1220


stevejpurves:
look good @fwkoch! great to have tests around this too

PR #1223, opened 2024-05-22
🆙 Upgrade Jupyter Books in `myst init`

### LINKED ISSUE #618: Compatibility with RST Glossaries

agoose77:
With the goal of running Jupyter Book on top of MyST, we need the ability to upgrade a book's configuration and TOC to `myst.yml`.

This PR adds a Jupyter-Book upgrade path when:
1. `myst init` is invoked and
2. the current directory does not contain a `myst.yml` and
3. the current directory contains a `_config.yml`.

This PR depends upon #1188. 

I've introduced `zod` for this PR, with the intention that we'll slowly phase out `simple-validators`. There's a challenge with lazy inference and circular references detailed here: https://github.com/colinhacks/zod/issues/3510

- [x] Do not clobber existing `myst.yml`
- [x] Upgrade glossaries
- [x] Add unit tests 
- [x] Add try-catch for error handling
- [x] Make renaming atomic.

Closes #618


stevejpurves:
Should we really begin phase out `simple-validators`? there is a significant cost in doing so...
and it's better to not to start a change on that scale unless we intend to follow through. Using `zod` in isolation for an upgrade command seems ok though, and a chance to see how it might be applied.

agoose77:
@stevejpurves I agree. My suggestion is that we evaluate `zod` here, and look to make better use of it if it transpires that it's a good fit internally.

choldgraf:
@agoose77 if the goal is to use this to learn about `zod` and whether it's a good idea for future adoption, I'd recommend:

- Opening an issue like 'evaluate the use of `zod` in the myst CLI' and link to the module where we've trialed it here
- Add the questions that you'd want to answer to decide whether it should be adopted more broadly
- Define a date by which you want to revisit the issue to make a decision
- Add a comment in the codebase that references that issue and notes the date to look at it

That way you can structure the information gathering rather than have it as a one-off tool that might confuse us in N months when we've forgotten about this conversation

choldgraf:
One more recommendation here (or in a follow up pr if it's out of scope)

If we detect a config.yml file, I assume we want to note the upgrade command. But for a while, I think we should also note that people can run the v1 version of jupyter book by installing the version ==1. Alternatively what do you think about making a pypi package like jupyter-book-sphinx and pinning that to the 1.x series (similar to nbclassic)

agoose77:
I tried this on CMIP6:
- `toc` was placed in the root indentation
- `toc` missed the `README`.

agoose77:
@fwkoch your thoughts on this would be welcome. I've decided that instead of a `myst upgrade` command, this should just be `myst init` when we detect a Jupyter Book. This has some nice ergonomics, mainly

1. `myst`/`myst --init` directly upgrades un-ported Jupyter Books (nice short UX)
2. Repeated `myst`/`myst --init` does not, because `myst.yml` exists
3. `myst upgrade` is reserved for actual MyST-specific scoped upgrades

I tested the WIP on The Turing Way, and it seems to handle it fairly well.

agoose77:
The line between "upgrading" a JB and upgrading MyST syntax is a little blurry. My take is that the JB upgrade is special, and we should treat it as such -- users will have an "old" project with known legacy aspects, vs users with one year old MyST projects that use some slightly deprecated syntax.

agoose77:
I need some help setting up unit tests for this (I am hoping someone else will make my life easier...)!

agoose77:
@rowanc1 I've rebased this now :)

rowanc1:
image

agoose77:
For tests, we can just test the glossary `syntax.ts`

packages/myst-cli/src/init/jupyter-book/syntax.ts
```diff
@@ -0,0 +1,230 @@
...
+  for (const transform of [upgradeGlossary]) {
+    const nextLines = await transform(documentLines);
+    didUpgrade = didUpgrade || nextLines !== undefined;
+    documentLines = nextLines ?? documentLines;
+  }
+
+  return didUpgrade ? documentLines : undefined;
+}
+
+async function upgradeGlossary(documentLines: string[]): Promise<string[] | undefined> {
```
rowanc1, line 132: Can we add a unit test around this.

agoose77:
Upgrade admonition cases too!!

packages/myst-cli/src/init/jupyter-book/syntax.ts
```diff
@@ -0,0 +1,230 @@
...
+async function upgradeContent(documentLines: string[]): Promise<string[] | undefined> {
+  let didUpgrade = false;
+
+  for (const transform of [upgradeGlossary]) {
+    const nextLines = await transform(documentLines);
+    didUpgrade = didUpgrade || nextLines !== undefined;
+    documentLines = nextLines ?? documentLines;
+  }
+
+  return didUpgrade ? documentLines : undefined;
```
rowanc1, line 129: Should we do `:::{Note}` here? Can kick that down the line.

agoose77: Let's kick that down the road?

rowanc1:
I think that there should be a confirmation step before we upgrade from jupyterbook as well @agoose77, and a little explaination of what it will do (we will upgrade glossaries, import information from the toc, etc.)

packages/myst-cli/src/init/jupyter-book/syntax.ts
```diff
@@ -0,0 +1,279 @@
...
+
+      // Will there be following terms?
+      const isFinalEntry = i === entries.length - 1;
+      if (!isFinalEntry) {
+        newLines.push('');
+      }
+    }
+    const nodeSpan = { start: node.position?.start?.line, stop: node.position?.end?.line };
+    const spanLength = nodeSpan.stop! - nodeSpan.start! - 1;
+    documentLines.splice(nodeSpan.start! + editOffset, spanLength, ...newLines);
```
agoose77, line 267: Are these `!` assertions reasonable?

packages/myst-cli/src/init/jupyter-book/syntax.ts
```diff
@@ -0,0 +1,279 @@
...
+  const mdast = mystParse(data);
+
+  const caseInsenstivePattern = new RegExp(admonitionPattern.source, admonitionPattern.flags + 'i');
+  const directiveNodes = selectAll('mystDirective', mdast);
+  const mixedCaseAdmonitions = directiveNodes.filter((item) => {
+    const name = (item as any).name as string;
+    return name.match(caseInsenstivePattern) && !name.match(admonitionPattern);
+  });
+  mixedCaseAdmonitions.forEach((node) => {
+    const start = node.position!.start.line;
```
agoose77, line 145: Is this `!` assertion reasonable?

agoose77:
@rowanc1 this is ready to merge imo.

packages/myst-cli/src/init/jupyter-book/syntax.ts
rowanc1: I think that it should be fine in this case, I think in general an early return of:

```
if (!node.position) return
```

Might be safer in all cases. If this works for you though (e.g. in the turing way), then it should be fine.

agoose77: I think it's safe as we're using a non-transformed pass, and we're looking for `mystDirective` nodes which are from the parse phase.

packages/myst-cli/src/init/jupyter-book/syntax.ts
rowanc1: Similar to the note above, I would bail if there are not positions. Not sure if that is possible in this case though. Given that there are unit tests now, this should be safe!

packages/myst-cli/src/init/jupyter-book/syntax.ts
rowanc1: Awesome. Will look at this and merge today or tomorrow!

packages/myst-cli/src/init/jupyter-book/syntax.yml
```diff
@@ -0,0 +1,145 @@
...
+
+      bar
+      : bar def
+
+      biff
+      : baff def
+
+      baff
+      : {term}`biff`
+
```
rowanc1, line 65: This is great.

rowanc1:
Final things:

- [ ] make sure it fails gracefully (try/catch)
- [ ] make it more permissive

agoose77:
@rowanc1 I've now finished this PR and tested on TTW and https://github.com/astropy/ccd-reduction-and-photometry-guide

The fallback logic will follow the "generate default config" path, meaning that if users fix their configuration, they'll then need to delete the `myst.yml`.

We could change it to _always_ try and upgrade if there's a `_config.yml`, but I think that will also be confusing (it will ignore myst.yml)

agoose77:
As per our meeting, I'll merge this in the next hour if you don't have any problems!

PR #1229, opened 2024-05-23
:sparkle: Read configuration for execution from `myst.yml`

### LINKED ISSUE #1214: Read Jupyter Server information from `project.jupyter`
## Proposal
We already read this configuration for Thebe, so let's also use it for local execution where it is given.

agoose77:
Fixes #1214

### DIFF
--- a/.changeset/long-jobs-listen.md
+++ b/.changeset/long-jobs-listen.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": minor
+---
+
+Read execution config from myst.yml
--- a/docs/integrating-jupyter.md
+++ b/docs/integrating-jupyter.md
@@ -287,7 +287,7 @@ Add the specific list options for custom wheel paths, etc.
 In addition to how you might normally start a JupyterLab session, it's necessary to provide two additional command line options, as follows.
 
 ```{code} bash
-jupyter lab --NotebookApp.token=<your-secret-token> --NotebookApp.allow_origin='http://localhost:3000'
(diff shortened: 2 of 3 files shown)


stevejpurves:
I am wondering if it is the best thing for the `myst.yml#thebe` config to be used as the first source of the execute configuration?

My thinking is that if I have a manuscript or book ready for publication, or perhap I already have v0.0.666 published and I am making a change to a notebook to add or fix something. Then my myst.yml is set up for production... I do want to re-execute locally before I publish v0.0.667 but I don't want to have to do that via say a public binder (slow) and I just want to hit my local server. I don't really want to have to change my `myst.yml` as that is all set up for prod, and maybe i'll forget to change it back...

So TLDR, should `JUPYTER_BASE_URL` take precedence over `configServer` allowing me to point to a local config, without messing with production `myst.yml`?

that seems to make sense to me, but maybe I am missing a counter use case that you have in mind @agoose77

agoose77:
> So TLDR, should `JUPYTER_BASE_URL` take precedence over `configServer` allowing me to point to a local config, without messing with production `myst.yml`?

Yes, and that _was_ my original intention. I must've forgotten to invert the order!

The more I think on it, there are a lot of ways we could write this. Perhaps we hold this PR for now, and `--execute` becomes a multi-modal flag e.g.
- `--execute` (try `jupyter`, allow env)
- `--execute=config` (use `jupyter.server`)
- `--execute=http://(...):8888?token=(...)` (use URL)
- `--execute=env` (use env)

I'm not sure on the best solution, so I'll close this _for now_

PR #1251, opened 2024-05-28
🌎 Fetch config extensions from URL

fwkoch:
`extends` pr was added here to allow one myst.yml to extend another: https://github.com/executablebooks/mystmd/pull/1215

In that PR and in this issue https://github.com/executablebooks/mystmd/issues/336 the next step was allowing remote urls for the extend content so centralized data may be reused across users/projects. This PR adds that remote resolution step!

Biggest thing about this pr is going from sync to async at the config loading level; this means many functions are newly awaited... big diff minor impact.

### DIFF
--- a/packages/myst-cli/src/config.ts
+++ b/packages/myst-cli/src/config.ts
@@ -385,7 +385,7 @@ async function resolveProjectConfigPaths(
   if (projectConfig.exclude) {
     resolvedFields.exclude = await Promise.all(
       projectConfig.exclude.map(async (file) => {
-        const resolved = await resolutionFn(session, path, file, false);
+        const resolved = await resolutionFn(session, path, file, { allowNotExist: true });
         return resolved;
       }),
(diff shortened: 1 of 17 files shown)


packages/myst-cli/src/config.ts
```diff
@@ -247,21 +264,40 @@ export function loadConfig(session: ISession, path: string, opts?: { reloadProje
...
   relativePath: string,
-  checkExists = true,
+  opts?: {
+    allowNotExist?: boolean;
+    allowRemote?: boolean;
+  },
 ) {
-  let message: string;
+  let message: string | undefined;
+  if (opts?.allowRemote && isUrl(relativePath)) {
```
fwkoch, line 277: 👀 This is the "interesting" part of the PR. Everything else is `async/await`...

fwkoch:
Add note to docs for this!

PR #1254, opened 2024-05-29
🚨 Surface warnings during toc file extension resolution

fwkoch:
Old jupyterbook-style `_toc.yml` files allowed (required...?) filenames without extensions. We are hoping to encourage explicit filenames with extensions in the new `myst.yml` toc.

This PR raises warnings/errors during toc resolution; there are 4 distinct errors:
- toc entry has an extension but does not exist
- toc entry has no extension and cannot be resolved to a file that exists
- toc entry has no extension and resolves to a single file (this is just a warning, but still surfaced to the user)
- toc entry has no extension and resolves to multiple files

This PR addresses https://github.com/executablebooks/mystmd/issues/1248

Note - legacy jupyterbook toc behavior and warnings/errors are unchanged.

### DIFF
--- a/.changeset/large-ravens-shop.md
+++ b/.changeset/large-ravens-shop.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Surface warnings during tocfile extension resolution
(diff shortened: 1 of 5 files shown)


agoose77:
LGTM!

PR #1257, opened 2024-05-29
:chair: Update `table` docs

### LINKED ISSUE #1241: Automatic table numbering
Currently it seems that tables aren't automatically numbered. I've found examples where numbering is specified in `myst.toml`:
```
# See docs at: https://mystmd.org/guide/frontmatter
version: 1
project:
  ...
  numbering:
    table: true
```

## Proposal
Like figures, having table numbering would be great!

roaldarbol:
This PR does two small things: 

1. Documents that `:label:` is needed for table numbering.
2. Introduces import table from file via `{include}`.

I've added an HTML table file too.
Closes #1241.

### DIFF
--- a/docs/table-from-file.html
+++ b/docs/table-from-file.html
@@ -1,0 +1,38 @@
+<table>
+    <tr>
+       <th rowspan="2">Projection</th>
+       <th colspan="3" align="center">Area in square miles</th>
+    </tr>
+    <tr>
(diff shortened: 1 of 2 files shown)


roaldarbol:
I think the relevant error is:
```
⚠️  tables.md:135 Cross reference target was not found: include
```

PR #1258, opened 2024-05-29
:bug: Support `raises-exception` in cell execution

agoose77:
This was already supposed to be implemented, but a silly mistake prevented it from working.

### DIFF
--- a/docs/execute-notebooks.md
+++ b/docs/execute-notebooks.md
@@ -28,6 +28,20 @@ The following computational content will be executed:
 In order to execute your MyST content, you must install a Jupyter Server and the kernel needed to execute your code (e.g., the [IPython kernel](https://ipython.readthedocs.io/en/stable/), the [Xeus Python kernel](https://github.com/jupyter-xeus/xeus-python), or the [IRKernel](https://irkernel.github.io/).)
 :::
 
+
+## Deadling with errors
+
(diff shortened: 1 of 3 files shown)


agoose77:
I think the tests are flaking at the moment.

choldgraf:
Looks good, I recommend documenting this as well so that it is clear this is intended behavior

agoose77:
> Looks good, I recommend documenting this as well so that it is clear this is intended behavior

Already on it ;)

docs/execute-notebooks.md
```diff
@@ -28,6 +28,20 @@ The following computational content will be executed:
 In order to execute your MyST content, you must install a Jupyter Server and the kernel needed to execute your code (e.g., the [IPython kernel](https://ipython.readthedocs.io/en/stable/), the [Xeus Python kernel](https://github.com/jupyter-xeus/xeus-python), or the [IRKernel](https://irkernel.github.io/).)
 :::
 
+
+## Deadling with errors
```
fwkoch, line 32:
```suggestion
## Dealing with errors
```

fwkoch (approved):
Nice, this makes sense, just a fix to correctly handle the tag.

choldgraf:
I did a quick pass on the docs to make it a bit clearer, I think it's GTG 👍

rowanc1:
Re flaky tests:

image

DOI was down yesterday and the site is still taking a long time on may requests. I think we just wait it out for now.

PR #1261, opened 2024-05-30
:wrench: Match YAML section keys with comments

agoose77:
This PR fixes a bug with TOC writing, if the `toc:` entry contains a trailing comment.

### DIFF
--- a/packages/myst-cli/src/project/utils.yml
+++ b/packages/myst-cli/src/project/utils.yml
@@ -9,7 +9,7 @@ cases:
         - foo
       section:
         body: "hi"
-    result: 
+    result:
       start: 1
       stop: 1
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/project/utils.yml
```diff
@@ -31,8 +31,8 @@ cases:
       foo: bar
       bar:
         - baz
-        - foo    
-    result: 
+        - foo
+    result:
```
rowanc1, line 35: Are the whitespace changes intentional?

agoose77: Yes, I ran `prettier`.

agoose77:
Merging, as I assume you're happy! :)

Issue #1244, opened 2024-05-28
`pattern` toc entries same default ordering as implicit project ordering


fwkoch:
If you have a glob `pattern` in your project `toc`, files will be ordered with default filesystem ordering. This needs to be replaced with the same ordering we do for implicit project builds, e.g. prioritize index files, order by alpha numeric, put folders at the end, etc.

fwkoch:
This is done but contentious! See #1323 for further discussion

Issue #1243, opened 2024-05-28
New TOC should be normaliseable to a TOC without pattern entries


agoose77:
## Description
The new TOC implements a pattern entry. This is really just shorthand for enumerating files, and we should not have to reason about patterns in our TOC handling logic

## Proposed solution
Normalise a TOC to a TOC without patterns

fwkoch:
There is now a step in the project processing that converts a toc into a toc without patterns. This isn't really visible or available externally, but I don't think it needs to be...? It's just an implementation detail that simplifies subsequent toc handling.

PR #1266, opened 2024-06-03
🔧 Resolve toc patterns in intentional order

fwkoch:
The new project `toc`s introduced support for `pattern` entries, where users may specify a glob pattern as part of their tree. However, these were just getting resolved to a flat list of files, ordered by default system ordering.

This PR adds much more explicit handling of patterns. `glob` is still used to collect the files, but then we:
- keep folder nesting
- order within each folder as: index, remaining files in alpha order, then subfolders in alpha ordrer
- deduplicate files in the tree, so if you specify a `file` that is also part of a `pattern`, it will not show up in the resolved pattern tree
- respect `exclude` from project frontmatter

These changes mean pattern resolution is similar to the implicit project resolution, with a few small differences:
- patterns do _not_ automatically ignore files like `_build`, `node_modules`, `.*` - they can still be ignored with `exclude`
- index file is handled at a per-folder level (maybe we want this behavior for implicit resolution...?)
- deduplication of files is not required on implicit resolution since files are never encountered twice there.

This addresses #1244 - it also addresses #1243 in the process, since we resolve patterns to create a new tree without patterns.


fwkoch:
Chatted through this with @rowanc1 - merging without official review.

This also fixes up the test failure that is currently plaguing all builds.

PR #1271, opened 2024-06-05
🦶🎵 Footnotes that are reused should have the same number

rowanc1:


rowanc1:
@fwkoch there looks to be a failing e2e test that is dependent on the deployed myst documentation remaining constant. Do you think that you could relax the constraints of the test in a separate PR?

PR #1285, opened 2024-06-11
🎓 Improvements to handling remote doi citations

fwkoch:
Recently doi.org started returning 403s when building myst sites on CI - this means citations from DOI links do not resolve.

This PR makes a few changes to help get around this issue:
- First, in addition to fetching dois from doi.org, we also try handle.net, the underlying service that serves dois. This doesn't work for everything, e.g. short dois do not work, but it is an easy way to provide a fallback for many cases.
- Also, before fetching dois from the web, myst will now look at bibtex files to see if there are any matching dois on existing references.
- Finally, you may now call `myst build --doi-bib` to write all doi citations to a bib file. Users may build their site locally, then invoke this command and use the resulting bib file for CI builds.


agoose77:
Making CI more stable is always a wonderful PR to see, thanks @fwkoch.

Something to think about here -- have we given much thought into using CSL-JSON as our canonical bib format, rather than using lossy Bibtex? It wouldn't likely be in this PR that we made that change, though.

fwkoch:
@agoose77 - ah, yeah, that would be nice, allow reading CSL-JSON from file. That would help out for this issue: https://github.com/executablebooks/mystmd/issues/1284

We would still be in trouble for latex export with bibtex, but in the case of writing the DOI to file for more stable consumption during site builds, CSL-JSON would be nice.

I agree this is a separate PR - this would get into the "supported myst file extensions" that `loadFile` supports. I guess add `'.json'` as an allowed extension and try to load as CSL-JSON, and quietly ignore if it ends up being other json? Seems relatively straightforward to add.

rowanc1:
Reviewed this with @fwkoch now, other than a few docs I think this is good to go. 🚀

rowanc1:
This is currently impacting all scipy proceedings builds for the automated checking where doi.org is not accepting requests from github actions, see:

- https://github.com/scipy-conference/scipy_proceedings/issues/935

docs/citations.md
```diff
@@ -39,6 +39,16 @@ For DOIs with multiple slashes in the identifier you also have to use the full h
...
 
+### Writing DOIs to BibTeX
+
+If you encounter problems fetching DOIs from `https://doi.org`, for example the downloaded citation does not include all the data you expect or requests to `https://doi.org` are failing on an automated continuous integration platform, you may write your DOI citations to file using the MyST command
+
+```bash
+myst build --doi-bib
+```
+
+This will generate a BibTeX file `myst.doi.bib` which you may then rename, edit, and save to your project. On subsequent builds, the DOIs will load be loaded from this file rather than fetched remotely.
```
rowanc1, line 50:
```suggestion
This will generate a BibTeX file `myst.doi.bib` which you may then rename, edit, and save to your project. On subsequent builds, the DOIs will be loaded from this file rather than fetched remotely.
```

choldgraf:
It's great to see this quick turnaround, good job folks :-)

PR #1287, opened 2024-06-11
:hammer_and_wrench: Combine frontmatter from notebook to support different kernels

### LINKED ISSUE #1283: Allow --execute to detect kernelspec in ipynb
## Proposal

Executing notebooks is a really great feature, thanks! Following these docs (https://mystmd.org/guide/execute-notebooks#how-myst-executes-your-code) I was hoping to be able to point at different kernels for different notebooks. 

Currently (myst v1.2.5) it seems the kernel in which myst is installed is used rather the using whatever is in the notebook metadata? 

A given `notebook.ipynb` has metadata pointing to a specific kernel. But it appears that `myst start --execute` uses whatever the current python environment is rather than checking available kernels or inspecting notebook metatadata.
```
 "metadata": {
  "kernelspec": {
   "display_name": "hypercoast",
   "language": "python",
   "name": "hypercoast"
  },
```

```
# jupyter kernelspec list
Available kernels:
  python3         /Users/scotthenderson/miniforge3/envs/mystmd/share/jupyter/kernels/python3
  hypercoast      /Users/scotthenderson/Library/Jupyter/kernels/hypercoast
  xarray-tests    /Users/scotthenderson/Library/Jupyter/kernels/xarray-tests
```

## Additional notes

`jupyter nbconvert` uses `kernelspec` in `notebook.ipynb` by default, or you can override it: 

`jupyter nbconvert --execute --ExecutePreprocessor.kernel_name=xarray-tests --to html myfigure.ipynb`

see https://nbconvert.readthedocs.io/en/latest/execute_api.html?highlight=kernel_name#execution-arguments-traitlets

agoose77:
Fixes #1283 by merging notebook metadata (frontmatter) with cell-defined metadata

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -7,7 +7,7 @@ import { VFile } from 'vfile';
 import type { GenericParent } from 'myst-common';
 import { RuleId, toText } from 'myst-common';
 import type { PageFrontmatter } from 'myst-frontmatter';
-import { validatePageFrontmatter } from 'myst-frontmatter';
+import { validatePageFrontmatter, fillProjectFrontmatter } from 'myst-frontmatter';
 import { SourceFileKind } from 'myst-spec-ext';
 import { frontmatterValidationOpts, getPageFrontmatter } from '../frontmatter.js';
 import type { ISession, ISessionWithCache } from '../session/types.js';
@@ -73,7 +73,11 @@ export async function loadNotebookFile(
     opts?.preFrontmatter,
     opts?.keepTitleNode,
   );
-  const frontmatter = fillProjectFrontmatter(cellFrontmatter, nbFrontmatter, { /** validation opts */ };
+  const frontmatter = fillProjectFrontmatter(
+    cellFrontmatter,
+    nbFrontmatter,
+    frontmatterValidationOpts(vfile),
+  );
   return { kind: SourceFileKind.Notebook, mdast, frontmatter, identifiers };
 }
 
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -73,7 +73,7 @@ export async function loadNotebookFile(
     opts?.preFrontmatter,
     opts?.keepTitleNode,
   );
-  const frontmatter = { ...nbFrontmatter, ...cellFrontmatter };
+  const frontmatter = fillProjectFrontmatter(cellFrontmatter, nbFrontmatter, { /** validation opts */ };
   return { kind: SourceFileKind.Notebook, mdast, frontmatter, identifiers };
 }
 
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -18,7 +18,7 @@ import { logMessagesFromVFile } from '../utils/logging.js';
 import { addWarningForFile } from '../utils/addWarningForFile.js';
 import { loadBibTeXCitationRenderers } from './citations.js';
 import { parseMyst } from './myst.js';
-import { processNotebook } from './notebook.js';
+import { processNotebookFull } from './notebook.js';
 import { selectors } from '../store/index.js';
 
 type LoadFileOptions = { preFrontmatter?: Record<string, any>; keepTitleNode?: boolean };
@@ -65,14 +65,15 @@ export async function loadNotebookFile(
 ): Promise<LoadFileResult> {
   const vfile = new VFile();
   vfile.path = file;
-  const mdast = await processNotebook(session, file, content);
-  const { frontmatter, identifiers } = getPageFrontmatter(
+  const { mdast, frontmatter: nbFrontmatter } = await processNotebookFull(session, file, content);
+  const { frontmatter: cellFrontmatter, identifiers } = getPageFrontmatter(
     session,
     mdast,
     vfile,
     opts?.preFrontmatter,
     opts?.keepTitleNode,
   );
+  const frontmatter = { ...nbFrontmatter, ...cellFrontmatter };
   return { kind: SourceFileKind.Notebook, mdast, frontmatter, identifiers };
 }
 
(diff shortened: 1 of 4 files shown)


packages/myst-cli/src/process/mdast.ts
```diff
@@ -215,6 +215,7 @@ export async function transformMdast(
 
   if (execute) {
     const cachePath = path.join(session.buildPath(), 'execute');
+    console.log(JSON.stringify(frontmatter, null, 2));
```
rowanc1, line 218:
```suggestion
```

rowanc1 (changes requested):
Not sure if this is in progress @agoose77, but if it is ready, can you add a changeset and remove the log message?

packages/myst-cli/src/process/notebook.ts
```diff
@@ -70,13 +74,27 @@ export async function processNotebook(
...
+): Promise<{ mdast: GenericParent; frontmatter: PageFrontmatter }> {
   const { log } = session;
   const { metadata, cells } = JSON.parse(content) as INotebookContent;
   // notebook will be empty, use generateNotebookChildren, generateNotebookOrder here if we want to populate those
 
   const language = metadata?.kernelspec?.language ?? 'python';
   log.debug(`Processing Notebook: "${file}"`);
 
+  // Load frontmatter from notebook metadata
+  const vfile = new VFile();
```
rowanc1, line 94: Here we are going to loose all error messages as this isn't referenced anywhere. Is that intended? If not, we should pass the vfile in?

packages/myst-cli/src/process/mdast.ts
agoose77: :face_holding_back_tears: 

I officially cannot multitask.

packages/myst-cli/src/process/notebook.ts
agoose77: Hmm, I was thinking that the notebook metadata isn't strictly frontmatter, so it might not always validate. But, if you think that's a strong assumption, then yes we should. Defer to your judgement.

agoose77: There's no test scaffolding right now, does @fwkoch want to weigh in?

packages/myst-cli/src/process/notebook.ts
```diff
@@ -70,13 +74,27 @@ export async function processNotebook(
...
+  vfile.path = file;
+  const frontmatter = validatePageFrontmatter(metadata ?? {}, frontmatterValidationOpts(vfile));
```
stevejpurves, line 96: @agoose77 what exactly ends up in this object here? is it all notebook metadata or will the frontmatter be pruned down to known fields only?

stevejpurves: just checking through the validate function I *think* that is what will happen, i.e. weare only allowing known fields into the processed notebook json, rather than the whole metadata...

agoose77: Yes, the `frontmatter` result should be the validated object - we should be dropping unknown fields.

packages/myst-cli/src/process/mdast.ts
rowanc1: Thank you! Can you also add a changeset! :)

packages/myst-cli/src/process/file.ts
```diff
@@ -65,14 +65,15 @@ export async function loadNotebookFile(
...
-  const { frontmatter, identifiers } = getPageFrontmatter(
+  const { mdast, frontmatter: nbFrontmatter } = await processNotebookFull(session, file, content);
+  const { frontmatter: cellFrontmatter, identifiers } = getPageFrontmatter(
     session,
     mdast,
     vfile,
     opts?.preFrontmatter,
     opts?.keepTitleNode,
   );
+  const frontmatter = { ...nbFrontmatter, ...cellFrontmatter };
```
rowanc1, line 76: @fwkoch is there a better way to merge this?

fwkoch: Could use: 
```suggestion
  const frontmatter = fillProjectFrontmatter(cellFrontmatter, nbFrontmatter, { /** validation opts */ };
```

That should work fine with `PageFrontmatter` too. Main difference is that it combines some fields if they are defined in both objects, rather than just overriding.

packages/myst-cli/src/process/notebook.ts
fwkoch: Regarding tests: for the complicated parts with no test scaffolding, I've been making end-to-end tests in `packages/mystmd/tests` - make a mini project that includes your feature, define the command to run, and expect some output. Seems like that could be possible here?

Regarding the extra keys and extra errors: you could run
```typescript
const filteredMetadata = filterKeys(metadata ?? {}, PAGE_FRONTMATTER_KEYS)
```
before running
```typescript
validatePageFrontmatter(filteredMetadata, ...)
```
then we would get real errors but none of the extra key warnings.

packages/myst-cli/src/process/file.ts
agoose77: *That* is what I was hoping for. Thanks @fwkoch

packages/myst-cli/src/process/notebook.ts
agoose77: @fwkoch I added this and extended it to accept `FRONTMATTER_ALIASES` which are aliases of `PAGE_FRONTMATTER_KEYS`. Does that seem reasonable?

packages/myst-cli/src/process/notebook.ts
```diff
@@ -143,5 +177,8 @@ export async function processNotebook(
     Promise.resolve([] as GenericNode[]),
   );
 
-  return { type: 'root', children: items };
+  logMessagesFromVFile(session, vfile);
```
agoose77, line 180: Actually, let's log these errors; we do care about validation after we've filtered the keys.

packages/myst-cli/src/process/notebook.ts
fwkoch: Oh, I see you did exactly this in a later commit😆

fwkoch (approved):
Thanks for addressing the comments, this looks great!

PR #1289, opened 2024-06-12
🏷 Add tags to project frontmatter

rowanc1:
@agoose77 @jmunroe as we were talking through this today in the pythia meetup, this brings tags up to the top level to support some of the work in Pythia, especially around galleries. This is necessary to support some of the work in collecting centralized resources.

Previously we could add tags to pages, but not projects. This opens up the possibility to tag a project and is more in line with how pythia, scipy-proceedings, and curvenote are dealing with projects that have this metadata at the project level.

### DIFF
--- a/.changeset/seven-kangaroos-matter.md
+++ b/.changeset/seven-kangaroos-matter.md
@@ -1,0 +1,5 @@
+---
+"myst-frontmatter": patch
+---
+
+Add tags to project frontmatter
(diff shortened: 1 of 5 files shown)


fwkoch (commented):
This looks pretty straightforward. Two questions:

- What is the difference between `tags` and `keywords`? I noticed `keywords` isn't documented in `frontmatter.md` - maybe we should add that and just have a one-sentence description of each?
- `'tags'` is not in `USE_PROJECT_FALLBACK` - this means pages _do not_ inherit project tags if the page doesn't define any tags. Is this the behavior we want? I assume so, but that change is also easy to overlook when adding new fields...

rowanc1:
Yes, tags should be similar to how title works (they exist on both project and page, but are not inherited).

Will document keywords, and add descriptions.

PR #1290, opened 2024-06-12
🔧 Error if toc index file has children

fwkoch:
Currently there is a constraint that the first toc entry must be a file; this becomes the project's `index` file. However, if this entry has a `file` and `children`, the children were previously just ignored. This PR adds an error if children are present on the index file.

### DIFF
--- a/.changeset/clean-chicken-own.md
+++ b/.changeset/clean-chicken-own.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Error if toc index file has children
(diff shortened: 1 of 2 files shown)


rowanc1 (approved):
Thanks - I ran into this in testing!

nthiery:
Thank you! That error message was useful to me this very morning :-)

rowanc1:
Glad you are using the new format @nthiery! Keep the reports coming in!

PR #1292, opened 2024-06-12
🌄 Fix unsplash images

rowanc1:
Currently all unsplash images are broken, this pins them to images that I have taken..

I have not put them in the repo so that we can also use them in the examples. Not sure on the best way to do this.

### DIFF
--- a/docs/cross-references.md
+++ b/docs/cross-references.md
@@ -159,7 +159,7 @@ There are many ways that you can label a target, and the sections below describe
 Targets are custom anchors that you can refer to elsewhere, for example, a figure, section, table, program, or proof. To be referenced, they must have a `label`/`identifier` pair [in the AST](xref:spec#association). These can be created by setting the `label` option in many directives. For example, to label and reference a figure, use the following syntax:
 
 ````{myst}
-```{figure} https://source.unsplash.com/random/500x200/?mountain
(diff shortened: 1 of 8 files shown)


docs/cross-references.md
```diff
@@ -159,7 +159,7 @@ There are many ways that you can label a target, and the sections below describe
 Targets are custom anchors that you can refer to elsewhere, for example, a figure, section, table, program, or proof. To be referenced, they must have a `label`/`identifier` pair [in the AST](xref:spec#association). These can be created by setting the `label` option in many directives. For example, to label and reference a figure, use the following syntax:
 
 ````{myst}
-```{figure} https://source.unsplash.com/random/500x200/?mountain
+```{figure} https://github.com/rowanc1/pics/blob/main/mountains.png?raw=true
```
fwkoch, line 162: Is `raw=true` required for these to work inside the `myst` directives?

fwkoch (approved):
Super annoying that unsplash is no longer working. This seems fine - I don't quite get what you mean about "also using in examples" - Is there a problem with putting images in this repo?

docs/cross-references.md
rowanc1: Yes, there is a transform that we run which isn't run in the demos.

Issue #1301, opened 2024-06-12
Cross references for static sites


rowanc1:
Currently the `xref:` does not work for deployed github-pages sites. This is because we are not rewriting the `myst.xref.json` to respect the basepath and there is an extra `/content/` in the paths.

PR #1302, opened 2024-06-12
🐛 Cross references for static sites

rowanc1:
See #1301

### DIFF
--- a/packages/myst-cli/src/build/html/index.ts
+++ b/packages/myst-cli/src/build/html/index.ts
@@ -1,6 +1,7 @@
 import fs from 'fs-extra';
 import path from 'node:path';
 import { writeFileToFolder } from 'myst-cli-utils';
+import type { MystXRefs } from 'myst-transforms';
 import type { ISession } from '../../session/types.js';
 import type { SiteManifestOptions } from '../site/manifest.js';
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/build/html/index.ts
```diff
@@ -154,10 +154,15 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
...
   fs.copySync(path.join(session.sitePath(), 'objects.inv'), path.join(htmlDir, 'objects.inv'));
-  fs.copySync(
-    path.join(session.sitePath(), 'myst.xref.json'),
-    path.join(htmlDir, 'myst.xref.json'),
-  );
+
+  // NOTE: HTML static output needs to patch the contents, this is done on the fly by the server
+  const xrefs = JSON.parse(
+    fs.readFileSync(path.join(session.sitePath(), 'myst.xref.json')).toString(),
+  ) as any;
```
fwkoch, line 161: This should have an actual type (will need the import too `import type { MystXRefs } from 'myst-transforms'`);


```suggestion
  ) as MystXRefs;
```

fwkoch (commented):
This seems ok - nice, narrowly scoped fix.

PR #1306, opened 2024-06-13
🌎 Allow remote files for template/site options

fwkoch:
Previously, site template options of type "file" (e.g. `logo`) had to be local, existing paths. Now, they may be remote URLs, which are downloaded and cached to file first.

This partially addresses https://github.com/executablebooks/mystmd/issues/1296 (logo and logo_dark should be good; favicon is currently not working at all, I'll fix in a separate pr).

### DIFF
--- a/packages/myst-cli/src/config.ts
+++ b/packages/myst-cli/src/config.ts
@@ -275,12 +275,12 @@ export async function resolveToAbsolute(
 ) {
   let message: string | undefined;
   if (opts?.allowRemote && isUrl(relativePath)) {
-    const cacheFilename = `config-item-${computeHash(relativePath)}${extname(relativePath)}`;
+    const cacheFilename = `config-item-${computeHash(relativePath)}${extname(new URL(relativePath).pathname)}`;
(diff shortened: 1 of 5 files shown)


packages/myst-cli/src/config.ts
```diff
@@ -275,12 +275,12 @@ export async function resolveToAbsolute(
 ) {
   let message: string | undefined;
   if (opts?.allowRemote && isUrl(relativePath)) {
-    const cacheFilename = `config-item-${computeHash(relativePath)}${extname(relativePath)}`;
+    const cacheFilename = `config-item-${computeHash(relativePath)}${extname(new URL(relativePath).pathname)}`;
```
fwkoch, line 278: This allows query params, e.g. `logo: https://github.com/rowanc1/pics/blob/main/grapes-wide.png?raw=true`. However, the extensions still needs to be in the URL. (We could optimize slightly by getting extension from content type header...)

rowanc1:
This is working great!

Issue #1296, opened 2024-06-12
Logo and favicon should be able to be remote


rowanc1:
Currently it isn't possible to use remote images for the favicons and logos, that is quite important for shared configuration files.

The important things to be able to be remote are:

- favicon
- logo
- logo_dark

jnywong:
Are we waiting on the favicon to close this? Looking to see if I can merge projectpythia-mystmd/radar-cookbook#2

rowanc1:
I think that is good to go now @jnywong! The favicon will start to work in the next few days, but need not block that work.

fwkoch:
Looking into favicon - there is discrepancy handling this in `myst-cli` vs `myst-theme`:

- In `myst-cli`, `favicon` is a property directly on `SiteConfig`: https://github.com/executablebooks/mystmd/blob/main/packages/myst-config/src/site/types.ts#L31
- In `myst-theme`, `favicon` is handled as a site template option: https://github.com/executablebooks/myst-theme/blob/main/themes/article/template.yml#L35 & https://github.com/executablebooks/myst-theme/blob/main/packages/common/src/utils.ts#L133

So, if you just chuck `favicon` under `site.options` it works fine, including URLs, with the most recent release of myst.

This should be tidied up to either (1) eliminate `favicon` from `SiteConfig` and require all themes to bring in as an option, if desired, or (2) change things on the theme side to respect the top-level `favicon`, not in options. I'm not super opinionated here - I feel like (1) maybe makes slightly more sense, though: Given a theme that does not allow users to customize favicon, it's better to have a warning message on an invalid option than allow users to set favicon in the config that will be totally ignored.

agoose77:
@fwkoch how does MyST handle file-paths in the theme options for favicons? Do we do any prep work to copy the favicon to the static paths?

fwkoch:
This is where all the theme `file` options get downloaded (if remote) and copied to `static` folder: https://github.com/executablebooks/mystmd/blob/main/packages/myst-cli/src/build/site/manifest.ts#L406

That covers `site.options.favicon` if it's defined in the theme.

fwkoch:
Regarding `SiteConfig.favicon`, that will be obsolete if we go through with the changes discussed here: https://github.com/executablebooks/mystmd/issues/1308#issuecomment-2174233550

That field will just _be_ the template option and no longer be defined in the `SiteConfig` type.

fwkoch:
I am going to close this issue:

[...]

PR #1309, opened 2024-06-13
🤝 Combine site options when using config extend

fwkoch:
Previously, if `options` were defined in `myst.yml#site` and in an `extend` config file - the `extend` fields would be overridden. Now, they are combined, as we already do with project frontmatter. This is helpful as we are making site options more reusable: #1296

### DIFF
--- a/.changeset/cold-games-warn.md
+++ b/.changeset/cold-games-warn.md
@@ -1,0 +1,6 @@
+---
+'myst-frontmatter': patch
+'myst-cli': patch
+---
+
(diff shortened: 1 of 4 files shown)


agoose77:
Hmm, does this mean it is not always possible to undo what is set in the inherited configuration?

fwkoch:
You would have to set a different value to override. To just delete an option, it could be `my_option: null`, but in practice that might not work, like, if there are issues with `null` vs `undefined`.

Issue #1304, opened 2024-06-13
Add `_build` to the `.gitignore` on `myst init`


rowanc1:
This is something that I have run into quite a few times, and was a sticking point in some demos that we ran into now at the Pythia hackathon.

cc @jmunroe @jnywong @agoose77

jnywong:
🎉

PR #1314, opened 2024-06-13
:zap: Write initial `.gitignore` in `myst init`

agoose77:
This idea came out of the Pythia Hackathon.

Fixes: #1304

### DIFF
--- a/packages/myst-cli/src/build/init.ts
+++ b/packages/myst-cli/src/build/init.ts
@@ -97,11 +97,11 @@ export async function init(session: ISession, opts: InitOptions) {
       session.log.info(`💾 Writing default .gitignore`);
       await fs.promises.writeFile('.gitignore', GITIGNORE);
     } else {
-	    // Parse the existing gitignore
+      // Parse the existing gitignore
       const contents = await fs.promises.readFile('.gitignore', { encoding: 'utf-8' });
       const lines = contents.split(/\r\n|\r|\n/);
       // Do we have mention of `/?_build`?
-      if (lines.some(line => /^\/?_build([#\/].*)?$/.test(line))) {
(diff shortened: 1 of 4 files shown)


packages/myst-cli/src/build/init.ts
```diff
@@ -69,6 +90,17 @@ export async function init(session: ISession, opts: InitOptions) {
   if (!project && !site && !writeTOC) {
     session.log.info(WELCOME());
   }
+
+  // Assume we're working in the cwd
+  if (await checkFolderIsGit()) {
+    if (await gitignoreExists()) {
+      session.log.info(`✅ .gitignore already exists`);
```
rowanc1, line 97: I think we should add `_build` to it if it doesn't exist!

packages/myst-cli/src/build/init.ts
```diff
@@ -38,6 +38,17 @@ const SITE_CONFIG = `site:
   domains: []
 `;
 
+const GITIGNORE = `/_build/
+
+# Python development
+.ipynb_checkpoints/
+__pycache__/
```
agoose77, line 45: This is the simplest list I could think of for Python-based projects. We could even drop the Python entries, and just ignore `_build`?

rowanc1: Yeah, I think we should be super simple. Just `_build`!

packages/myst-cli/src/build/init.ts
```diff
@@ -38,6 +38,17 @@ const SITE_CONFIG = `site:
...
+.venv
+
+# MacOSX
+.DS_Store				    
+`;
```
agoose77, line 50:
```suggestion
const GITIGNORE = `/_build/`;
```

fwkoch (approved):
Tested this out and seems to work great!

Issue #1318, opened 2024-06-14
During pdf builds custom bibtex tags are not respected


stevejpurves:
## Description

If an author has their own `bib` file,  `articles.bib`, in which they are using a different tag style e.g. `jones-etal-2014` on their bibtex entries as they prefer `[@jones-etal-2014]` in their manuscript.

Upon building a latex pdf, a `main.bib` file is produced in the `_build/` folder that contains (enriched?) version of the bibtex entries, but with normalized tags e.g. `jonesAdam2014`, yet the tags used in the manuscript in the `_build` folder have not been normalized to match.

## Proposed solution
Either normalize everything, or don't normalize tags at all and respect origin tags used by the author in their bib file?

fwkoch:
This is fixed by https://github.com/executablebooks/mystmd/pull/1322

Updating `citation-js` gives us a nice config option that prevents this implicit label behavior.

fwkoch:
Released with v1.2.9!

PR #1322, opened 2024-06-15
🎓 Configure citation-js to keep original citation labels

fwkoch:
This addresses #1318

By default `citation-js` uses computed labels when exporting bibtex. However, there is a config setting to persist original id as label. This setting also requires an update to `citation-js` version to work correctly.


packages/citation-js-utils/src/index.ts
```diff
@@ -1,11 +1,15 @@
...
 import { doi as doiUtils } from 'doi-utils';
 import { clean as cleanCSL } from '@citation-js/core/lib/plugins/input/csl.js';
 import sanitizeHtml from 'sanitize-html';
 
 import '@citation-js/plugin-bibtex';
 import '@citation-js/plugin-csl';
 
+const config = plugins.config.get('@bibtex');
+config.format.useIdAsLabel = true;
+config.format.checkLabel = false;
```
fwkoch, line 11: If `checkLabel` is not explicitly set to `false`, punctuation is removed, causing the case from #1318 to still fail.

packages/mystmd/tests/dois/outputs/dois.bib
```diff
@@ -60,7 +60,7 @@ @article{Koch_2012
 
 @article{EFSA2019Dietary,
 	address = {IT},
-	author = {{EFSA Panel on Nutrition, Novel Foods and Food Allergens (NDA)} and Turck, Dominique and Castenmiller, Jacqueline and de Henauw, Stefaan and HirschErnst, KarenIldico and Kearney, John and Knutsen, Helle Katrine and Maciuk, Alexandre and Mangelsdorf, Inge and McArdle, Harry J and Pelaez, Carmen and Pentieva, Kristina and Siani, Alfonso and Thies, Frank and Tsabouri, Sophia and Vinceti, Marco and Aggett, Peter and FairweatherTait, Susan and Martin, Ambroise and Przyrembel, Hildegard and de SesmaisonsLecarr{\' e}, Agn{\` e}s and Naska, Androniki},
+	author = {{EFSA Panel on Nutrition, Novel Foods and Food Allergens (NDA)} and Turck, Dominique and Castenmiller, Jacqueline and de Henauw, Stefaan and HirschErnst, KarenIldico and Kearney, John and Knutsen, Helle Katrine and Maciuk, Alexandre and Mangelsdorf, Inge and McArdle, Harry J and Pelaez, Carmen and Pentieva, Kristina and Siani, Alfonso and Thies, Frank and Tsabouri, Sophia and Vinceti, Marco and Aggett, Peter and FairweatherTait, Susan and Martin, Ambroise and Przyrembel, Hildegard and de SesmaisonsLecarr{\' e}, Agn{\` e}s and Naska, Androniki},
```
fwkoch, line 63: Since we bumped the `citation-js` version, this must just be a small side effect; different treatment of whitespace. For what it's worth, the new space is a normal space, whereas the old "space" was a weird null character. So I think this is an improvement.

packages/citation-js-utils/src/index.ts
```diff
@@ -252,13 +263,20 @@ export async function getCitationRenderers(data: CSL[]): Promise<CitationRendere
...
-            const bibtexObjects = cite.set(c).format('bibtex', { format: 'object' }) as {
-              label: string;
-            }[];
-            return bibtexObjects[0]?.label;
+            if (getLabel(cite, c).startsWith('temp_id_')) {
+              config.format.useIdAsLabel = false;
+            }
+            const label = getLabel(cite, c);
+            config.format.useIdAsLabel = true;
+            return label;
```
agoose77, line 271: Perhaps we should instead be normalising ingested entries so that we don't have to configure plugin config in a contextual manner?

fwkoch: Yeah - I think you are right about getting rid of the contextual configuration.

We just need our own id generation for cases where the citation doesn't have an id.

fwkoch: Ok - I just pulled out the `formatLabel` function from `citation-js` and used that in these cases, instead of the contextual configuration. Much better I think.

rowanc1 (approved):

PR #1329, opened 2024-06-18
:computer: Add support for IPC-backed plugins

### LINKED ISSUE #1298: Python Plugins for MyST

agoose77:
This PR fixes #1298 in a different way to #1325 by adding a new "executable" type of plugin (the approach taken in https://gist.github.com/agoose77/66700b2a172b4c11d4801f742534e4f7). Unlike #1325, this PR adds support for Python via an IPC mechanism (stdout, stdin).

#1325 and this PR are _not_ mutually exclusive; #1325 adds support for an in-process Python executor, which supports sharing stateful proxies between JS and Python. This PR, meanwhile, is fully decoupled and allows for further languages. Note that, unlike #1325, we do not vendor Python, so it is possible that users may define plugins whose environment is not sufficiently reproducible and thus see build errors.

> [!NOTE]
> #1325 will need to be rebased and updated to use this new mechanism for declaring plugin _types_.

To implement such a plugin, the executable needs to support the following invocation modes:
- `proc`
  Return the "spec" of the plugin, i.e. the implementation-free `plugin` block
- `proc --directive <DIRECTIVE>`
  Execute the implementation of a directive `<DIRECTIVE>`
- `proc --role <ROLE>`
  Execute the implementation of a role `<DIRECTIVE>`
- `proc --transform <TRANSFORM>`
  Execute the implementation of a transform `<DIRECTIVE>`
whereby the auxillary data for these executions is provided via `stdin`, the results read out from `stdout`, and error/debug information read from `stderr`.

Right now, I didn't add the `validate` method. We could; it's not much extra work. But do we need it? An example of a directive and transform is given in the docs of this PR.

> [!CAUTION]
> Syntax Changes from this PR in `myst.yml`:
> <table><tr><td>
>
> ```yaml
> project:
>   plugins:
>     - test.mjs
> ```
>  
> </td><td>
> 
> ```yaml
> project:
>   plugins:
>     - type: javascript
>       path: test.mjs
>     - type: executable
>       path: python.py
> ```
> </td></tr></table>
> N.b. this change is opt-in; existing `plugin` declarations will work just fine.

### DIFF
--- a/docs/executable-plugins.md
+++ b/docs/executable-plugins.md
@@ -17,14 +17,18 @@ First, we'll declare the plugin _specification_ that allows MyST to discover the
 :::{literalinclude} unsplash.py
 :caption: a plugin to add an `unsplash-py` directive that includes a beautiful, random picture based on a query string.
 :::
-this code should be referenced from your `myst.yml` under the `projects.plugins`:
+this file should be executable, e.g.
+```{code} shell
+chmod +x ./unsplash.py
+```
+and should be referenced from your `myst.yml` under the `projects.plugins`:
 
 ```{code} yaml
 :filename: myst.yml
 project:
   plugins:
     - type: executable
-      path: unsplash.mjs
+      path: unsplash.py
 ```
 
 then start or build your document using `myst start` or `myst build`, and you will see that the plugin is loaded.
--- a/docs/executable-plugins.md
+++ b/docs/executable-plugins.md
@@ -1,0 +1,88 @@
+---
+title: Executable Plugins
+description: Plugins built as stand-alone applications can be written in languages such as Python, and may be more familiar to some developers.
+---
+
+MyST is able to invoke plugins written in different languages through standard IO protocols. Executable MyST plugins are treated as a [](wiki:Black_box), whereby MyST only sees the data it passes to the plugin, and the response from the plugin itself.
+
+
+## Defining a new directive
+
+:::{note}
+There are many different ways to create an executable plugin. Here we'll use Python to implement an `unsplash-py` directive, but any programming language that can read and write from stdin, stdout, and stderr, and define a commandline interface would work.
+:::
+
+First, we'll declare the plugin _specification_ that allows MyST to discover the directives, transforms, and/or roles that the plugin implements. This specification looks very similar to [the definition of a JavaScript plugin](javascript-plugins.md#unsplash-js-source), except the implementation logic (e.g. the directive `run` method) is not defined.
+
+:::{literalinclude} unsplash.py
+:caption: a plugin to add an `unsplash-py` directive that includes a beautiful, random picture based on a query string.
+:::
+this code should be referenced from your `myst.yml` under the `projects.plugins`:
+
+```{code} yaml
+:filename: myst.yml
+project:
+  plugins:
+    - type: executable
+      path: unsplash.mjs
+```
+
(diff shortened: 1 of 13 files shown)


rowanc1:
Can we auto detect that in the transform and expand `plugin.py` to the dict?

e.g.

```
project:
  plugins:
    - test.mjs
    - plugin.py
```

Which is basically what you wrote, but slightly nicer UX / overhead for the user. :)

rowanc1:
Note that the way you wrote it should *also* work, as I see now the javascript could also be an executable.

agoose77:
@rowanc1 yes, I think the forward-looking preferred syntax would be the verbose kind. For single strings, we should only support known extensions but IPC plugins don't have a specific extension. If we added support for Python as in #1325, then we'd be able to write
```yaml
  plugins:
    - foo.mjs
    - bar.py
```

docs/executable-plugins.md
```diff
@@ -0,0 +1,88 @@
...
+:caption: a plugin to add an `unsplash-py` directive that includes a beautiful, random picture based on a query string.
+:::
+this code should be referenced from your `myst.yml` under the `projects.plugins`:
+
+```{code} yaml
+:filename: myst.yml
+project:
+  plugins:
+    - type: executable
+      path: unsplash.mjs
```
rowanc1, line 27: Should this be py?

docs/javascript-plugins.md
```diff
@@ -0,0 +1,122 @@
...
+:::
+
+This code should be referenced from your `myst.yml` under the `projects.plugins`:
+
+```{code} yaml
+:filename: myst.yml
+project:
+  plugins:
+    - type: javascript
+      path: unsplash.mjs
```
rowanc1, line 26: Is it possible to leave things as:

```suggestion
    - unsplash.mjs
```

And it gets transformed?

rowanc1: Perhaps having a small comment about that? Maybe in tabs?

rowanc1 (commented):
Minor docs comments. @agoose77 I think if this is tested it should come in!

I haven't tested on my own.

docs/executable-plugins.md
agoose77: Haha yes, I noticed that the other day. About to board a flight - could you tweak?

docs/javascript-plugins.md
agoose77: It is. I suppose we can keep JS plugins as the "preferred route" and error if the ext isn't mjs.

agoose77:
I've tested this on Pythia, and it works as expected!

PR #1330, opened 2024-06-18
:bug: Address rule-id typo

### LINKED ISSUE #1327: Typo in rule ID for container children
See https://github.com/executablebooks/mystmd/blob/6326cf5965ec96461a3c4b80e3b5c53c1f0cdaa6/packages/myst-common/src/ruleids.ts#L104

agoose77:
Fixes #1327

### DIFF
--- a/.changeset/rich-vans-admire.md
+++ b/.changeset/rich-vans-admire.md
@@ -1,0 +1,5 @@
+---
+"myst-common": patch
+---
+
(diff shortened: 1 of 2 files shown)


fwkoch (approved):
Nice, enum makes this easy - don't think there's any problems here - this rule shouldn't be in anyone's `myst.yml` or anything.

Issue #1332, opened 2024-06-18
HTML parsing breaks easily with whitespace


fwkoch:
## Description

Given this simple HTML:
```html
<table><tr><th>Heading</th></tr>

<tr><td>Data</td></tr></table>
```
MyST fails. Rather than creating a single `table`, it creates a `table` for the first line and a `tableRow` for the second line. MyST attempts to handle this with the `reconstructHtml` transform - https://github.com/executablebooks/mystmd/blob/main/packages/myst-transforms/src/html.ts#L243 - but that function makes _major_ assumptions about the shape of the HTML which are easily violated.

## Proposed solution

We probably need an actual HTML parser in the `reconstructHtml` transform so we know if the block of HTML is entirely closed or if it still has open elements.

fwkoch:
There should be an easy first pass on this that just combines sequential HTML nodes together if they are only separated by whitespace...

fwkoch:
I can't actually reproduce this issue in any other case besides blank whitespace interruptions, so maybe we don't need more complicated html parsing... Once #1333 lands, I'll consider this closed - can always re-open if we encounter more issues!

fwkoch:
Ok - it's still pretty easy to get html to fail, for example:
```
<b>blah

</b>
```
doesn't work - it just loses the bold tag. However, fixing that requires a deeper change to the actual html parsing. This issue just handles correctly parsed html that is split across mdast nodes...

fwkoch:
The fix to resolve the specific example in the original issue has been released, so I am going to consider this closed.

As my previous comment suggests, there is potentially more work to be done for smarter html parsing, but that can wait until more real examples come up.

PR #1333, opened 2024-06-18
❕ Combine html nodes split by whitespace in reconstructHtml transform

fwkoch:
Addresses #1332


fwkoch:
Discussed with @rowanc1 - this doesn't solve all html issues, but is an incremental improvement. See #1332 for more.

PR #1334, opened 2024-06-19
✂️ Remove doi requests to handle.net

fwkoch:
Previously we encountered 403 responses when hitting doi.org. Adding a request to the separate handle.net service _seemed_ to help with that. Now, though, we are getting 429 responses instead and the requests to handle.net definitely _do not_ help.

This PR just takes out those requests, reducing the total number of requests required to resolve a doi and speeding up the build.

### DIFF
--- a/.changeset/chilled-ants-sin.md
+++ b/.changeset/chilled-ants-sin.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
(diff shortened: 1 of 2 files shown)


fwkoch:
Partially addresses https://github.com/executablebooks/mystmd/issues/1335

PR #1339, opened 2024-06-20
🚦 Share doi request limiter across session

fwkoch:
This partially addresses https://github.com/executablebooks/mystmd/issues/1335 - now, the rate limiting previously implemented at a page level (and therefore easily exceeded during a build with multiple pages) is shared across the entire session.


fwkoch:
Discussed / tested with @rowanc1

PR #1363, opened 2024-06-27
⚫️ Ignore `dot` top level domain

rowanc1:
Have come across this in a few scipy proceedings and our linker is turning `numpy.dot` for example into a link.

### DIFF
--- a/packages/myst-parser/src/config.ts
+++ b/packages/myst-parser/src/config.ts
@@ -69,4 +69,4 @@ export const MARKDOWN_IT_CONFIG = {
 };
 
 // List of valid TLDs to exclude from linkify
-export const EXCLUDE_TLDS = ['py', 'md'];
+export const EXCLUDE_TLDS = ['py', 'md', 'dot'];
--- a/.changeset/dry-grapes-smash.md
+++ b/.changeset/dry-grapes-smash.md
@@ -1,0 +1,5 @@
+---
+"myst-parser": patch
+---
+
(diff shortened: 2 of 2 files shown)


agoose77:
Long term I would be in favour of only detecting URIs with schemes on them, e.g.

- google.com -> `google.com`
- https://google.com -> `<https://google.com>`

Do you see that being possible?

fwkoch:
Regarding schemes, if we want to do that, it's an option on `LinkifyIt`: https://github.com/markdown-it/linkify-it?#new-linkifyitschemas-options - It won't be totally straightforward, since we do not run `LinkifyIt` directly, it's just included with `MarkdownIt`, but I suspect we can figure out a way to configure it, as we did here: https://github.com/executablebooks/mystmd/blob/main/packages/myst-parser/src/myst.ts#L76

packages/myst-parser/src/config.ts
```diff
@@ -69,4 +69,4 @@ export const MARKDOWN_IT_CONFIG = {
 };
 
 // List of valid TLDs to exclude from linkify
-export const EXCLUDE_TLDS = ['py', 'md'];
+export const EXCLUDE_TLDS = ['py', 'md', 'dot'];
```
fwkoch, line 72:
```suggestion
export const EXCLUDE_TLDS = ['py', 'md', 'dot', 'next'];
```

`.next` is another one from scipy content...

PR #1369, opened 2024-07-02
🪣 Remove specific reference to `eq` role in general docs

agoose77:
image

We have a section on equation targets, but it's not really up-to-date; now we prefer using general Markdown links for references unless specific text is needed.

### DIFF
--- a/docs/cross-references.md
+++ b/docs/cross-references.md
@@ -152,11 +152,15 @@ For example, in this reference syntax:
 ```
 
 The text `my-targets-label` is the label for the target.
-There are many ways that you can label a target, and the sections below describe the most common approaches.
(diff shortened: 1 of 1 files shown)


docs/cross-references.md
```diff
@@ -152,11 +152,15 @@ For example, in this reference syntax:
 ```
 
 The text `my-targets-label` is the label for the target.
-There are many ways that you can label a target, and the sections below describe the most common approaches.
+There are many ways that you can label a target, and the sections below describe the most common approaches.
```
rowanc1, line 155:
```suggestion
There are many ways that you can label a target, and the sections below describe the most common approaches.
```

rowanc1 (commented):
linting!

Issue #1222, opened 2024-05-22
Duplicated title in the HTML exports for markdown notebooks with title in a first cell with metadata


nthiery:
## Description

Assume a jupytext notebook, as in the example below, with title given
in the first markdown cell, and with metadata on that first cell. Then,
as of myst v1.2.3, the title appears twice in the HTML export:

```
---
jupytext:
  text_representation:
    extension: .md
    format_name: myst
    format_version: 0.13
kernelspec:
  display_name: Python 3 (ipykernel)
  language: python
  name: python3
---

+++ {"nbgrader": {"grade": false, "grade_id": "cell-2df857546d357cc2", "locked": true, "schema_version": 3, "solution": false, "task": false}}

# Introduction à la programmation avec Jupyter et Python

## Découvrir Jupyter

...
```

![image](https://github.com/executablebooks/mystmd/assets/545328/ab14c8ca-f5b5-4cfb-a07d-a8e1cd614094)

The duplication vanished when removing the first `+++ ...` line or moving the title in the frontmatter.

Motivation: the metadata here is used to configure the cell for use with nbgrader; in that case to make the cell readonly.

stevejpurves:
@nthiery I beleive also adding `title: null` to the frontmatter will prevent the heading from being "pulled up", although the duplication in the first case is stilll unexpected... @fwkoch?

nthiery:
Thanks for the quick feedback. I tried setting `title: null` in the frontmatter. The duplication is still there. Interestingly enough, the duplication now persists even if there is no cell metadata for the first cell as in:
```
---
jupytext:
  text_representation:
    extension: .md
    format_name: myst
    format_version: 0.13
kernelspec:
  display_name: Python 3 (ipykernel)
  language: python
  name: python3
title: null
---
# Introduction à la programmation avec Jupyter et Python
```

Sounds like just a little logic glitch somewhere.

parmentelat:
I am experiencing this too

parmentelat:
Trying this over again, it seems like the issue appears only when the **title cell is not the first** one in the notebook

parmentelat:
so maybe I jumped to conclusions..

my observation earlier was based on this repo
https://github.com/ue12-p24/intro/tree/myst    
(beware this is not the default branch, as main is still using jb1)
which renders into this
https://ue12-p24.github.io/intro/

all my notebooks used to follow the same scheme: cell#1: licence, cell#2: style, cell#3: title
[...]

PR #1383, opened 2024-07-14
🧱 Do not duplicate h1 title nested in block

fwkoch:
This should address the duplicate heading bug discussed in #1222

### DIFF
--- a/.changeset/little-shirts-decide.md
+++ b/.changeset/little-shirts-decide.md
@@ -1,0 +1,5 @@
+---
+'myst-transforms': patch
+---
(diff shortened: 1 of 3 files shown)


fwkoch:
This now prevents title/h1 duplication in the case of leading block breaks (as described in #1222) and leading comments.

Long term, this should probably also be handled on the frontend. In the frontmatter, we have a flag `content_includes_title`, which, when true, means the theme should not put the title in the frontmatter block, as it will show up somewhere else in the content where the author intended.

PR #1387, opened 2024-07-14
🚨 Do not warn on uppercase image extensions

fwkoch:
This should not error; uppercase image extensions are fine:

image


rowanc1:
@mepa, I think this solves one of the issues you ran into!

Issue #1170, opened 2024-04-30
Figure conversion darkens figure


roaldarbol:
## Description
Currently, if embedding (large) figures such as PDFs, the conversion step darkens them. It happens for both the web rendering and PDF generation with Typst.

## Proposed solution
Find the right settings for `imagemagick` which handles the conversion.

@rowanc1 previously pointed me to the code which is here:
https://github.com/executablebooks/mystmd/blob/77d4f58eb08e720ffacc5838f78e15777b00e33d/packages/myst-cli/src/utils/imagemagick.ts#L95
```
convert -density 600 -colorspace RGB ${image} -trim ${png}
```

stefanv:
Note that the `convert` command is deprecation in ImageMagick v7 for `magick`.

The current flags all appear to work "as-is".

I do not observe the darkening; I wonder if this is related to the colorspace. Do you have an example image you can share?

roaldarbol:
Yeah, I've written quite a bit about it on the Discord. Can add here too: :-)

If you're using `convert`, you're not getting any of the bug fixes, updates etc. that has been introduced in ImageMagick 7, but are stuck with the legacy IM6 edition of the command. Here's the differences between v6 and v7 magick: https://www.imagemagick.org/script/porting.php

And examples: With `convert -density 600 -colorspace RGB fig.pdf fig-2.webp`:
![](https://media.discordapp.net/attachments/1247486291130585088/1247961544810303549/fig-2.webp?ex=668ec142&is=668d6fc2&hm=8b204aef1855b12cdfe4087287d9d42646efb60d3456f2dc9a8b31928ac0d435&=&format=webp&width=1516&height=1138)

And with `magick -density 600 -colorspace RGB fig.pdf fig.webp`:
![](https://media.discordapp.net/attachments/1247486291130585088/1247961667724378112/fig.webp?ex=668ec15f&is=668d6fdf&hm=16c82a5de1489200c91f56c6386f1dd4c4fcc85d3f6e1f752525e6f8d0177d22&=&format=webp&width=1516&height=1138)

rowanc1:
#1388 should fix this!

https://github.com/user-attachments/assets/981501ff-4ba3-4e47-82ae-b6f8aca3b541

Issue #290, opened 2023-03-07
SVG images cause PDF export to fail on windows


riziles:
### What version of `mystjs` are you using?

0.1.17

### How did you install myst?

npm

### What operating system are you using?

Windows

### Which area is this feature request for?

Export: LaTeX or PDF

### Describe the Bug

Running `myst build --pdf` fails on if MyST document contains SVG images. Following error message is displayed:

```
Could not convert from SVG to PNG: 
_build\temp\mystmqG4xb\images\aaecccbb623db43cfbe1b76430ef9e1b.svg 
- Error: Command failed: convert -density 600  -colorspace RGB 
- _build\temp\mystmqG4xb\images\aaecccbb623db43cfbe1b76430ef9e1b.svg 
- _build\temp\mystmqG4xb\images\aaecccbb623db43cfbe1b76430ef9e1b.png
Invalid Parameter - 600
```

### Expected Behavior

An exported PDF document.

### To Reproduce

sample-doc.md:

````markdown
---
exports:
  - format: pdf
---

# SVG Export Test

```{figure} https://myst-parser.readthedocs.io/en/latest/_static/logo-wide.svg
---
name: svg-export-test
---
SVG MyST Logo
```

This is a test
````

**Comment by rowanc1** [comment]
Thanks @riziles for the report! We will take a look and get back to you soon!

**Comment by riziles** [comment]
FYI, I signed up for a paid Curvenote account, and I'm able to export docs with SVGs beautifully from the Curvenote website. However if I use the Curvenote CLI and try to do it locally, I receive the same error message, so I'm guessing it must be either a Windows issue or something specifically wrong with my local setup.

**Comment by rowanc1** [comment]
@fwkoch has been redoing all of the image exports in #317 based on thinking through some of this.

Can you run, to see if it is a version mismatch?
```
convert --version
```

riziles:
Error message is `Invalid drive specification.` .  FYI, my workaround of exporting the SVGs to PDF and just using those still works fine AND it keeps the images from getting rasterized.

riziles:
> convert --version

@rowanc1 , I am a little puzzled by this request. As far as I can tell, `convert` is a Windows system command to convert file systems on your hard drive. What does it have to do with rendering SVGs?

fwkoch:
[...]

PR #1388, opened 2024-07-14
🪄 Use 'magick' command instead of 'convert' for imagemagick

fwkoch:
This should address https://github.com/jupyter-book/mystmd/issues/1170 and partially https://github.com/jupyter-book/mystmd/issues/290


rowanc1:
This works well and actually fixes the coloring issue as well!!

https://github.com/user-attachments/assets/981501ff-4ba3-4e47-82ae-b6f8aca3b541

PR #1390, opened 2024-07-14
✍️ Keep all contributors/affiliations until final page resolution

fwkoch:
This addresses https://github.com/jupyter-book/mystmd/issues/1362

Previously, when extending from multiple config files, the contributors and affiliations are lost if they are not used on any of the intermediate steps.

With this PR, all contributors and affiliations are persisted right until the final page resolution, when unused affiliations/contributors are removed.

Issue #1312, opened 2024-06-13
Explicitly choose between config merging and clobbering configuration


agoose77:
At present, it seems from conversations with @rowanc1 and @fwkoch that we're selecting an extension policy for each particular key in `myst.yml`.

I wonder if it would be nice to have a user-facing control mechanism for this, e.g.
```yaml
site:
   options:
      extend: merge
```

I don't think this should be done at the config level, but rather per config-section. Thoughts?

agoose77:
Hmm, I don't love that either - how do you merge multiple configs? This needs more thought.

choldgraf:
Two quick thoughts for consideration, I don't feel strongly about either but in case they inspire:

### Could this be done more granularly at the parent config level? 

I think it would look less messy if the config file _doing the extending_ is the place where extends behavior is controlled, rather than in the file being pulled in.

E.g.:

In `myst.yml`:

```yaml
project:
  extends:
    otherconfig.yml: 
      - key1: replace 
      - key2: extend
    
```

### Could we overload symbol syntax to do this rather than using YAML tag features?

e.g.:

```yaml
project:
  extends:
    otherconfig.yml: 
      - key1++ # would mean "in addition to what's here" 
      - key2 # would mean "replace what's hear"
```

agoose77:
We certainly could. I think there are some considerations:

1. Defining the overload behaviour in the parent separates semantics. For some fields, I suspect it's important for readability and re-use that the fact we're extending is encoded in the same file. It also means we have two sets of state to keep in sync.
2. I assume you mean to introduce custom syntax? We could do this with a pre-processor, but it would be a lot of work. Tags are trivial and standard YAML features. Am I missing something here?

rowanc1:
I really like the idea of making the config easier to extend. My preference would be to do this through a plugin/code rather than yaml config. That is giving people the various objects (`[extends1, extends2, extends3]`, etc. along with `utils`) in a plugin. For example, then they could do a web-request to pull in abbreviations and transform and then merge selectively etc. or have their own YAML format, and provide some `extends` utils in the function (all the frontmatter ones).

I think this is probably easier to develop, document, learn and maintain than YAML?
e.g. this means that we could do the citation cff issue (#2401) with a plugin, which seems like a good sign of unlocking multiple features!

choldgraf:
Could you describe what this would look like from a user's perspective? I don't think I understand what "doing this with a plugin" means relative to the syntax proposals in this issue.

[...]

PR #1394, opened 2024-07-19
🐈 Concatenate list fields when extending project frontmatter

fwkoch:
# Background

We added the ability to compose yaml files when constructing your `myst.yml` file: https://github.com/jupyter-book/mystmd/pull/1215. This allows shared config, reducing duplication (and errors) across articles and just making authors' lives easier.

However, the original behavior was for fields to replace previous values. So if you define `keywords` in your `myst.yml`, any `keywords` defined in configuration files that the `myst.yml` `extends` are lost. Since then, we realized this meant `options` were not composed, so an option in your `myst.yml` could replace an inherited, totally unrelated option. This was fixed here: https://github.com/jupyter-book/mystmd/pull/1309. There were also some cases where contributors could be lost, addressed here: https://github.com/jupyter-book/mystmd/pull/1390.

These led to a further issue where we discussed allowing users to prescribe inheritance pattern for specific fields: https://github.com/jupyter-book/mystmd/issues/1312.

# PR Description

Recently, another issue came up (whoops, there's not actually an `issue` 😬): Inheriting one `export` from a shared config while also defining another `export` in your `myst.yml`. The new export would just clobber the shared export. This PR addresses this by _extending_ list fields when composing project configurations, rather than _replacing_.

Affected fields include:
- tags
- reviewers
- editors
- keywords
- bibliography
- requirements
- resources
- funding
- exports
- downloads

`exports` and `downloads` are deduplicated by `id` so you may replace an entry from an inherited config by redefining an entry with the same id. All the other fields just combine inherited and new values.

**Note:** When `page` frontmatter inherits `project` frontmatter, the old behavior remains; lists are not extended.

# Discussion

Overall, I think this behavior makes more sense and improves on the current behavior. For example:
- a shared config may define a few `keywords`. Then, if you have additional `keywords`, you do not need to rewrite the entire list
- a shared config may come bundled with `bibliography`/`requirements`/`resources` files. If you are providing additional files in these categories, you do not need to re-list the others.
- etc!

This also encourages authors to ensure shared config files actually contain content that "should" be shared (and on the flip side, discourages authors from inheriting just some other random `myst.yml` and accidentally pulling in unwanted data).

However, there is still an argument that this behavior is too prescriptive, e.g. if I inherit a config with any `keywords`, I _cannot_ get rid of those keywords. In that specific case, we are saying, "No, if you don't want those keywords, do not inherit that config!" The other option is giving the author control over inheritance; this is the same discussion we are already having in #1312.

### DIFF
--- a/packages/myst-frontmatter/src/exports/validators.ts
+++ b/packages/myst-frontmatter/src/exports/validators.ts
@@ -79,14 +79,26 @@ export const MULTI_ARTICLE_EXPORT_FORMATS = [
 
 export function validateExportsList(input: any, opts: ValidationOptions): Export[] | undefined {
   if (input === undefined) return undefined;
-  const output = validateList(
-    input,
-    { coerce: true, ...incrementOptions('exports', opts) },
-    (exp, ind) => {
-      return validateExport(exp, incrementOptions(`exports.${ind}`, opts));
-    },
-  );
+  const exportsOptions = { coerce: true, ...incrementOptions('exports', opts) };
+  const output = validateList(input, exportsOptions, (exp, ind) => {
+    return validateExport(exp, incrementOptions(`exports.${ind}`, opts));
+  });
   if (!output || output.length === 0) return undefined;
+  const duplicates = new Set<string>();
+  output.forEach((exp, ind) => {
+    if (
+      exp.id &&
+      output
+        .slice(ind + 1)
+        .map(({ id }) => id)
+        .includes(exp.id)
+    ) {
+      duplicates.add(exp.id);
+    }
+  });
+  if (duplicates.size) {
+    validationError(`duplicate export ids: ${[...duplicates].join(', ')}`, exportsOptions);
+  }
   return output;
 }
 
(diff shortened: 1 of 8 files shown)


fwkoch:
In writing up this PR, I realized I should also update the docs... 😅📚

packages/myst-frontmatter/src/exports/validators.ts
```diff
@@ -79,14 +79,26 @@ export const MULTI_ARTICLE_EXPORT_FORMATS = [
...
+      output
+        .slice(ind + 1)
+        .map(({ id }) => id)
+        .includes(exp.id)
+    ) {
+      duplicates.add(exp.id);
+    }
+  });
+  if (duplicates.size) {
+    validationError(`duplicate export ids: ${[...duplicates].join(', ')}`, exportsOptions);
```
fwkoch, line 100: This PR also adds validation on export / download ids to ensure they are unique.

rowanc1:
@fwkoch this just needs docs updates before merging?

fwkoch:
Ah, yes, if y'all are happy with it. I can update docs on this and #1396 this afternoon so we can get them in.

rowanc1 (approved):

Issue #1246, opened 2024-05-28
Improvements to support `parts` for exports


fwkoch:
Individual files allow `parts` to be defined in their frontmatter. These are respected for static exports of those files. However, parts cannot be defined at the project/export level. We should extend the frontmatter `parts` to be allowed there:

```yaml
...
project:
  exports:
    - format: pdf
      abstract: My Custom Export Abstract
```

Additionally, it would be nice to allow these to be paths to files rather than inline text.

rowanc1:
Very excited about this, especially around enabling parts to be defined by files in general.

fwkoch:
We should make sure this works for site templates, including live reload if the part is a file path.

fwkoch:
Ok - just hashing through things a bit to make sure I understand (1) the current state of things and (2) how to proceed...

# How do `parts` currently work?

- `parts` may be defined in four ways:
  1. `project frontmatter parts`
  2. `page frontmatter parts`
  3. directly in a page, under a `tagged block` part
  4. directly in a page, under a section with the same name as the part (these are `implicit parts`)
- During processing, this is simplified:
  1. `project frontmatter parts` are simply ignored currently
  2. `page frontmatter parts` become `tagged blocks` and are prepended to the page content. These blocks are marked as `visibility: remove` - the transform to do this is basically the first transform during mdast processing: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/process/mdast.ts#L170
  3. existing `tagged blocks` are unchanged
  4. existing `implicit part` sections are unchanged
- For static exports:
  - the export `template.yml` defines `parts`
  - these are then extracted from the page using `extractParts` - this pulls out `tagged blocks` (which now include `page frontmatter parts`) as well as `implicit parts`.
  - extracted parts are then available to the template
  - `page frontmatter parts` that were not extracted do not show up, due to their visibility; `tagged blocks` and `implicit parts` defined in the page that were not extracted _do_ show up, though.
  - for multi-page exports, parts are extracted across all pages, and if there are duplicates, they error (so, for example, `abstract` must be defined on one-page-and-only-one-page of the export)
  - an individual `export` cannot set it's own export-specific `parts`
- For site build:
  - the site theme `template.yml` _could_ define `parts`, but it currently does not
[...]

PR #1396, opened 2024-07-19
💇‍♀️ Allow parts in project frontmatter

fwkoch:
MyST projects allow authors to define `parts` on each page, for example `abstract`, `acknowledgments`, etc. However, in many cases, these parts apply to the entire project, not a single page. For example, if my project is a scientific article with a myst page for each section, it doesn't really make any sense for the abstract to be defined on only one random page. See https://github.com/jupyter-book/mystmd/issues/1246

This PR lifts all the types / validation of `parts` from `page` frontmatter to `project` frontmatter, allowing authors to define parts directly in their `myst.yml`

Since this logic is inherited from project to page, the `page` behavior is unchanged. Also, the values are _not_ inherited from project to page, so if you define `parts.abstract` in your project, it will not be defined on each page.

This PR does not yet consume these new project parts, it just opens the door to define them. There will be significant future work determining how to use these parts; for example, in a pdf export, how do we reconcile multiple `abstracts` defined at different levels...?

### DIFF
--- a/.changeset/silly-rivers-obey.md
+++ b/.changeset/silly-rivers-obey.md
@@ -1,0 +1,5 @@
+---
+'myst-frontmatter': patch
+---
+
+Allow parts in project frontmatter
--- a/docs/document-parts.md
+++ b/docs/document-parts.md
@@ -1,9 +1,9 @@
 ---
 title: Document Parts
-description: Parts allow you to specify special parts of your document, like abstract, keypoints acknowledgements.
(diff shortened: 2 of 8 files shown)


fwkoch:
Again, needs some docs... 🚨📚

stevejpurves:
@fwkoch is it am easy next step to read these parts from file?

fwkoch:
Absolutely - that's the second half of #1246. I think worth a separate PR, but it will be helpful immediately once we start consuming this change.

rowanc1:
Waiting on docs updates before merging!

PR #1398, opened 2024-07-20
🐱 Add support for pre-parsed mdast-type files (`.myst.json`)

agoose77:
This feature would be useful for enabling a nice workflow for pulling in mdast built from the Sphinx extension https://github.com/executablebooks/sphinx-ext-mystmd.

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -108,6 +108,11 @@ export function loadTexFile(
   return { kind: SourceFileKind.Article, mdast: tex.ast as GenericParent, frontmatter };
 }
 
+export function loadMySTJSON(content: string) {
+  const { mdast, frontmatter } = JSON.parse(content);
+  return { kind: SourceFileKind.Article, mdast, frontmatter };
+}
+
 /**
  * Attempt to load a file into the current session. Unsupported files with
  * issue a warning
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -185,6 +185,8 @@ export async function loadFile(
         // This MUST be the final case before `default`, as
         // we rely on falling through to the `default` case if
         // a non-MyST JSON file is encountered here
+        //
+        // falls through
       }
       default:
         addWarningForFile(session, file, 'Unrecognized extension', 'error', {
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -109,8 +109,8 @@ export function loadTexFile(
 }
 
 export function loadMySTJSON(content: string) {
-  const { mdast, frontmatter } = JSON.parse(content);
-  return { kind: SourceFileKind.Article, mdast, frontmatter };
+  const { mdast, frontmatter, kind } = JSON.parse(content);
+  return { kind: kind ?? SourceFileKind.Article, mdast, frontmatter };
 }
 
 /**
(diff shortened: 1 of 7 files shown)


rowanc1 (commented):
There are probably a lot of json files out there that are not mdast, and I know a few projects in myst that have these as data files and this would change the TOC and likely error out?

How do we make this more specific? Or is it already?

packages/myst-cli/src/project/fromPath.ts
```diff
@@ -127,8 +127,11 @@ function indexFileFromPages(pages: (LocalProjectFolder | LocalProjectPage)[], pa
   if (!indexFile) indexFile = matcher('.md');
   if (!indexFile) indexFile = matcher('.tex');
   if (!indexFile) indexFile = matcher('.ipynb');
+  if (!indexFile) indexFile = matcher('.json');
   if (!indexFile) [indexFile] = files.filter((file) => extname(file) === '.md');
   if (!indexFile) [indexFile] = files.filter((file) => extname(file) === '.tex');
+  if (!indexFile) [indexFile] = files.filter((file) => extname(file) === '.ipynb');
```
rowanc1, line 133: Was this a bug?

agoose77: I think so!

agoose77:
> There are probably a lot of json files out there that are not mdast, and I know a few projects in myst that have these as data files and this would change the TOC and likely error out?
> 
> How do we make this more specific? Or is it already?

Yes, good point. I wrote this on a recent flight and this is something I didn't worry about at the time. We probably should switch on e.g. myst.json

rowanc1:
Yeah, if this is `.myst.json` as the extension, I am much less worried. :)

agoose77:
@rowanc1 are you happy to merge this?

packages/myst-cli/src/project/fromPath.ts
fwkoch: (FWIW, this was the last extension type, so it would just get picked up in the final `[indexFile] = files;`. Definitely better as updated, since it's more explicit, but I don't think there is any worry that we inadvertently changed someones project with this change.)

packages/myst-cli/src/process/file.ts
```diff
@@ -108,6 +108,11 @@ export function loadTexFile(
   return { kind: SourceFileKind.Article, mdast: tex.ast as GenericParent, frontmatter };
 }
 
+export function loadMySTJSON(content: string) {
+  const { mdast, frontmatter } = JSON.parse(content);
+  return { kind: SourceFileKind.Article, mdast, frontmatter };
```
fwkoch, line 113: Should we also allow `kind` to be pulled from the content? (but still default back to `Article`)

rowanc1 (approved):
Looks good to me, @fwkoch has a few more questions, but when those are addressed, feel free to merge it in!

packages/myst-cli/src/process/file.ts
agoose77: Ooh, I like that idea. Feel free to tweak, or I'll merge.

fwkoch, line 113: I think just this?

```suggestion
  const { mdast, frontmatter, kind } = JSON.parse(content);
  return { kind: kind ?? SourceFileKind.Article, mdast, frontmatter };
```

fwkoch: Tweaked!

agoose77: MERGE AT WILL ✈️✈️

Issue #1323, opened 2024-06-15
Table of content, patterns and nested directories


nthiery:
Provide a way to control the hierarchy in the table of contents when using patterns, in particular to enable compatibility with JupyterBook.

## The problem

Let's consider some deep hierarchy of documents
```
- A1/
   - A11/
       - A111.md
       - A112.md
- A2
    - A21/
        - ...
``` 

**Use case 1:** you want to build automatically the whole table of content, with the hierarchy of directories being reflected in a hierarchy in the toc. In that case, you expect:
```
toc:
   - pattern: */*/*.md
```
to result in a nested table of content such as:
```
> A1
    > A11
          A111.md
          A112.md
> A2
        ...
```
That's what MyST does.

**Use case 2:** you want to control the nesting of your table of content. E.g. because you want to choose your own titles for the directories. Or maybe have a nesting that is different from the directory hierarchy. So you expect the following configuration:
```
toc:
   - title: The introduction
     content:
      - title: bar
        content:
          - pattern: A1/A11/*.md
   - title: User guide
     ...
``` 
to produce:
```
> The Introduction
  > Bar
        A111.md
        A112.md
> User Guide
            ...
```

However, with the current MyST and unlike JupyterBook, the produced TOC is doubly nested:
```
> The Introduction
  > Bar
     > A1
        > A11
              A111.md
              A112.md
> User Guide
            ...
```

## Proposal

I am not sure whether this should be qualified as a bug: automatic building the hierarchy in the toc can be useful as in Use Case 1. However we would need to support Use Case 2 as well, in particular for migrating from the current JupyterBook to MyST. Somehow, we would want the hierarchy to be built relatively to the current toc section, but I feel like matching a subdirectory with a toc folder is not necessarily 1 to 1, especially when having sections defined by just a title, not a file. Maybe we could have something like:
```
...
   - pattern: */*/*.md
```
would produces a flat table of content (as in JB), while:
```
...
   - pattern:
        - glob: foo/*/*.md
        - root: foo/ 
```
would produce a nested table of content w.r.t. the given root ???

## Additional notes

Thank you so much for the new `toc` syntax in MyST. It's one more step in the migration, and it so much systematic than the `toc` syntax in the current JB.

@choldgraf : you may want to mention this issue in the tracking issue for the JB->MyST migration.

agoose77:
[...]

PR #1399, opened 2024-07-20
📂 Drop folders in `pattern` expansion

agoose77:
This PR addresses #1323 by:

1. Simplifying patterns to only generate `file` entries
2. Drop support for `pattern` entries with `children` -- in the schema, `pattern` entries can only be leaves.

### DIFF
--- a/packages/myst-cli/src/project/fromTOC.ts
+++ b/packages/myst-cli/src/project/fromTOC.ts
@@ -1,14 +1,15 @@
 import fs from 'node:fs';
-import { join, parse, resolve, sep } from 'node:path';
+import { join, parse, resolve, sep, dirname, basename } from 'node:path';
 import { RuleId } from 'myst-common';
 import type { ISession } from '../session/types.js';
 import type { JupyterBookChapter } from '../utils/toc.js';
 import { readSphinxTOC, tocFile } from '../utils/toc.js';
-import { VALID_FILE_EXTENSIONS, isValidFile, resolveExtension } from '../utils/resolveExtension.js';
-import { createTitle, fileInfo } from '../utils/fileInfo.js';
+import { VALID_FILE_EXTENSIONS, resolveExtension } from '../utils/resolveExtension.js';
+import { fileInfo } from '../utils/fileInfo.js';
 import { addWarningForFile } from '../utils/addWarningForFile.js';
 import { nextLevel } from '../utils/nextLevel.js';
 import { selectors } from '../store/index.js';
+
 import type {
   PageLevels,
   LocalProjectFolder,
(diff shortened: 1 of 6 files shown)


packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -235,13 +104,15 @@ export function patternsToFileEntries(
         }
         ignore.push(...listExplicitFiles(newEntries, path));
         return newEntries;
+      } else {
+        const { children } = entry as ParentEntry;
```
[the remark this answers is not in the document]
agoose77: My view is that this was originally a bug - the spec doesn't allow for patterns to have children. But if you think it's already in the wild, I agree!

packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -220,8 +81,16 @@ export function patternsToFileEntries(
     .map((entry) => {
       if (isPattern(entry)) {
         const { pattern } = entry as PatternEntry;
-        const dirStructure = patternToDirectoryStructure(pattern, path);
-        const newEntries = directoryStructureToFileEntries(dirStructure, path, ignore);
+        const matches = globSync(pattern, { cwd: path })
+          .filter((item) => !ignore || !ignore.includes(item))
+          // Sort ignoring fodler
+          .sort((left, right) => sortByNumber(basename(left), basename(right)));
```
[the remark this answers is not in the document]
agoose77: No, you're right. I think we do want hierarchical sorting, and to prefer index.

packages/myst-cli/src/project/fromTOC.ts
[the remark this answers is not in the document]
fwkoch: Yeah - I think not allowing children on patterns is the right choice. We just need to make sure there is some sort of validation error if someone tries. Maybe this should actually be all the way up in the `myst-toc` validation? i.e. remove `'children'` here: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-toc/src/toc.ts#L132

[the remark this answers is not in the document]
agoose77: Oh, yes. That's a bug.

packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -31,156 +32,44 @@ import { globSync } from 'glob';
...
- */
-export function objFromPathParts(parts: string[], obj?: Record<string, any>) {
-  if (!obj) obj = {};
-  const [part, ...remainingParts] = parts;
-  if (!part) return obj;
-  if (!obj[part]) obj[part] = {} as Record<string, any>;
-  obj[part] = objFromPathParts(remainingParts, obj[part]);
-  return obj;
-}
+export function comparePaths(a: string, b: string): number {
```
agoose77, line 41: @fwkoch this function now performs the directory-sensitive string comparison:

1. Paths with non-matching directories are nat-sorted
2. Paths with matching directories are nat-sorted if neither file is an index, or both files are index files
3. Paths with non-matching directories are sorted in preference of the index file if one file is an index

Does this match your expectations?

packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -236,13 +133,15 @@ export function patternsToFileEntries(
...
         ignore.push(...listExplicitFiles(newEntries, path));
         return newEntries;
+      } else {
+        const { children } = entry as ParentEntry;
+        if (children) {
+          const newChildren = patternsToFileEntries(session, children, path, ignore, file);
+          return { ...entry, children: newChildren };
+        } else {
+          return entry;
+        }
```
agoose77, line 143: Make `children` a mutually exclusive property against `pattern`

packages/myst-toc/src/toc.ts
```diff
@@ -129,7 +129,7 @@ export function validatePatternEntry(
     entry,
     {
       required: ['pattern'],
-      optional: [...COMMON_ENTRY_KEYS, 'children'],
+      optional: [...COMMON_ENTRY_KEYS],
```
agoose77, line 132: Fix this bug!

packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -31,156 +32,44 @@ import { globSync } from 'glob';
...
+    // Prefer index files
+    else if (aIsIndex) {
+      return -1;
+    } else {
+      return +1;
+    }
+  }
+  // Otherwise, compare naively as strings
+  else {
+    return a.localeCompare(b, undefined, {
```
fwkoch, line 68: Another difference from implicit toc behavior: Here, folders and files get mixed together alphabetically; in the implicit toc, in a given directory, folders are always after files.

I feel like this one is totally fine. Worse case scenario, folders can always be renamed to order them differently.

packages/myst-cli/src/project/fromTOC.ts
fwkoch: I believe this comment has been addressed under `myst-toc` in this PR!

fwkoch (approved):
I like this much better. There's still a couple subtle differences between pattern expansion and implicit toc construction, but I don't think they matter? Just wanted to mention them so we are clear that the differences are intentional.

packages/myst-cli/src/project/fromTOC.ts
```diff
@@ -31,156 +32,44 @@ import { globSync } from 'glob';
...
-  let indexFile: string | undefined;
-  DEFAULT_INDEX_WITH_EXT.forEach((index) => {
-    if (indexFile) return;
-    indexFile = files.find((file) => file.toLowerCase() === index);
-  });
-  if (indexFile) {
-    files = [indexFile, ...files.filter((file) => file !== indexFile)];
+    // Are both a and b index files, or neither?
+    if (aIsIndex === bIsIndex) {
+      return aBaseName.localeCompare(bBaseName, undefined, {
```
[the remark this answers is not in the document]
agoose77: > This is an extreme edge case, but if there are two index files, index.md and index.ipynb, the .ipynb file will come first. It also means all INDEX files will be first. This is different from the implicit toc behavior where (1) index preference prioritizes .md over other formats and (2) once an index file is found, the other index files are just treated as normal files...

Yes -- although I hadn't thought about this regression, I think it's worth keeping for simplicity; then we can keep our sorting logic to be local rather than stateful?

packages/myst-cli/src/project/fromTOC.ts
agoose77: We can make this a choice. I'm perfectly happy with either solution. I perhaps slightly prefer the current logic, as most file-system browsers e.g. `ls` use this kind of ordering.

PR #1402, opened 2024-07-22
feat: allow user options to apply to mystParser

sglyon:
when using myst-parser as a unified plugin, any options supplied by the user were not actually being used. This fixes that problem

### DIFF
--- a/packages/myst-parser/src/myst.ts
+++ b/packages/myst-parser/src/myst.ts
@@ -116,7 +116,7 @@ export function mystParse(content: string, opts?: Options) {
  */
 export const mystParser: Plugin<[Options?], string, GenericParent> =
   function mystParser(opts?: Options) {
-    this.Parser = (content: string, _file: VFile) => {
+    this.Parser = (content: string) => {
       return mystParse(content, opts);
     };
   };
--- a/packages/myst-parser/src/myst.ts
+++ b/packages/myst-parser/src/myst.ts
@@ -117,7 +117,7 @@ export function mystParse(content: string, opts?: Options) {
 export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser(
   opts?: Options,
 ) {
-    this.Parser = (content: string) => {
-      return mystParse(content, opts);
-    };
+  this.Parser = (content: string) => {
+    return mystParse(content, opts);
   };
+};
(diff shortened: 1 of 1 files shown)


packages/myst-parser/src/myst.ts
```diff
@@ -114,8 +114,9 @@ export function mystParse(content: string, opts?: Options) {
 /**
  * MyST Parser as a Unified Plugin
  */
-export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser() {
-  this.Parser = (content: string, opts?: Options) => {
-    return mystParse(content, opts);
+export const mystParser: Plugin<[Options?], string, GenericParent> =
+  function mystParser(opts?: Options) {
+    this.Parser = (content: string, _file: VFile) => {
```
fwkoch, line 119: I think we can just leave off the unused `_file` argument. Feels a little wrong that we are passing `vfile` in on `opts` rather than grabbing it from here, but straightening that out is out of scope for this nice little PR.

```suggestion
    this.Parser = (content: string) => {
```

fwkoch (approved):
Thanks for catching and fixing this @sglyon!

packages/myst-parser/src/myst.ts
```diff
@@ -114,8 +114,9 @@ export function mystParse(content: string, opts?: Options) {
 /**
  * MyST Parser as a Unified Plugin
  */
-export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser() {
-  this.Parser = (content: string, opts?: Options) => {
-    return mystParse(content, opts);
+export const mystParser: Plugin<[Options?], string, GenericParent> =
+  function mystParser(opts?: Options) {
```
fwkoch, line 118:
```suggestion
export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser(
  opts?: Options,
) {
```

packages/myst-parser/src/myst.ts
```diff
@@ -114,8 +114,10 @@ export function mystParse(content: string, opts?: Options) {
...
-export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser() {
-  this.Parser = (content: string, opts?: Options) => {
-    return mystParse(content, opts);
+export const mystParser: Plugin<[Options?], string, GenericParent> = function mystParser(
+  opts?: Options,
+) {
+    this.Parser = (content: string) => {
+      return mystParse(content, opts);
+    };
   };
```
fwkoch, line 123:
```suggestion
  this.Parser = (content: string) => {
    return mystParse(content, opts);
  };
};
```

PR #1404, opened 2024-07-23
#️⃣ Fix HTML tags in inline contexts

### LINKED ISSUE #1400: Raw html tags like kbd transformed into paragraphs in certain cases
In certain cases, the raw html tag `<kbd>F</kbd>` gets transformed to `<p><code>F</code></p>`, which forces paragraph breaks. This affects other tags like `<code>`,  `<i>` or `<b>`, ...

Myst version: 1.2.9; also version currently deployed on mystmd.org

## How to reproduce

Inputing the following in the sandbox of mystmd.org:
```
-  Cliquez dans la cellule ci-dessous, puis tapez <kbd>Maj</kbd>+<kbd>Entrée</kbd>
```
Produces:
image

It would be expected to have everything in a single line as in:
image

agoose77:
Previously we were converting HAST to MDAST as full-documents, which was introducing paragraphs. 

`hast-util-from-util` always creates a `root` node, and `hast-util-to-mdast` introduces paragraphs for phrasing content inside a `root` node.

HTML nodes are `StaticPhrasingContent`, so we should already be e.g. inside a paragraph if we need one and thus can remove this wrapping? @rowanc1 you have touched this logic most recently.

Fixes #1400

### DIFF
--- a/.changeset/modern-wombats-boil.md
+++ b/.changeset/modern-wombats-boil.md
@@ -1,5 +1,5 @@
 ---
-"myst-transforms": patch
+'myst-transforms': patch
 ---
 
 Fix conversion of hast fragments to mdast fragments
--- a/.changeset/modern-wombats-boil.md
+++ b/.changeset/modern-wombats-boil.md
@@ -1,0 +1,5 @@
+---
+"myst-transforms": patch
+---
+
+Fix conversion of hast fragments to mdast fragments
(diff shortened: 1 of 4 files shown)


rowanc1 (commented):
Can you add a test for the `kbd` html element.

```yaml
- title: image
    before:
      type: root
      children:
        - type: html
          value: <kbd>Shift</kbd>
    after:
      type: root
      children:
        - type: keyboard
          children:
            - type: text
              value: Shift
```

Something like that.

agoose77:
Certainly!

rowanc1:
Testing this now.

rowanc1:
Looks good!

image

PR #1405, opened 2024-07-23
:wrench: Tighten up JSON input

agoose77:
This PR adds some validation for JSON-file input, and applies our frontmatter transforms.

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -21,6 +21,9 @@ import { loadBibTeXCitationRenderers } from './citations.js';
 import { parseMyst } from './myst.js';
 import { processNotebookFull } from './notebook.js';
 import { selectors } from '../store/index.js';
+import { defined, incrementOptions, validateObjectKeys, validateEnum } from 'simple-validators';
+import type { ValidationOptions } from 'simple-validators';
+import { fileError, fileWarn } from 'myst-common';
 
 type LoadFileOptions = { preFrontmatter?: Record<string, any>; keepTitleNode?: boolean };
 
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/process/file.ts
```diff
@@ -108,9 +111,84 @@ export function loadTexFile(
...
+  input: any,
+  opts: ValidationOptions,
+): { mdast: GenericParent; kind: SourceFileKind; frontmatter: PageFrontmatter } {
+  const value = validateObjectKeys(
+    input,
+    { required: ['mdast'], optional: ['kind', 'frontmatter'] },
+    opts,
+  );
+  // Although this is required, we'll set it to an empty article by default
+  const mdast: GenericParent = value?.mdast ?? { type: 'root', children: [] };
```
agoose77, line 141: Is this a reasonable thing to do? Declare this `mdast` property `required`, but add a default? I did this so that we don't need to throw an error in this function.

agoose77:
@rowanc1 this follows from our meeting, and fixes the frontmatter.

packages/myst-cli/src/process/file.ts
```diff
@@ -108,9 +110,84 @@ export function loadTexFile(
...
+export function mystJSONValidationOpts(
+  vfile: VFile,
+  opts?: { property?: string; ruleId?: RuleId },
+): ValidationOptions {
+  return {
+    property: opts?.property ?? 'file',
+    file: vfile.path,
+    messages: {},
+    errorLogFn: (message: string) => {
+      fileError(vfile, message, { ruleId: opts?.ruleId ?? RuleId.mystJsonValid });
```
agoose77, line 122: The name of the rule ID follows our existing convention.

agoose77:
@fwkoch friendly ping! :)

packages/myst-cli/src/process/file.ts
```diff
@@ -108,9 +110,84 @@ export function loadTexFile(
...
+    mdast,
+    vfile,
+    opts?.preFrontmatter,
+    opts?.keepTitleNode,
+  );
+  const frontmatter = fillProjectFrontmatter(
+    astFrontmatter,
+    jsonFrontmatter,
+    frontmatterValidationOpts(vfile),
+  );
```
fwkoch, line 187: I'm not totally sure how this should work, but to me, it feels like `jsonFrontmatter` is intended to be `PageFrontmatter`, rather than `ProjectFrontmatter`, and it should take priority, rather than be used as a fallback.

If that interpretation is correct, then I would use `jsonFrontmatter` as `preFrontmatter`, which would then potentially impact the `getPageFrontmatter` function... Something like:
```suggestion
  const { frontmatter, identifiers } = getPageFrontmatter(
    session,
    mdast,
    vfile,
    { ...astFrontmatter, ...opts?.preFrontmatter },
    opts?.keepTitleNode,
  );
```

fwkoch (commented):
Sorry to be slow on this! A few comments, hopefully they make sense. In particular, I think just throwing an error during validation would be fine (and probably better).

packages/myst-cli/src/process/file.ts
```diff
@@ -108,9 +110,84 @@ export function loadTexFile(
...
+      enum: SourceFileKind,
+    });
+  }
+  let frontmatter: undefined | PageFrontmatter;
+  if (defined(input.frontmatter)) {
+    frontmatter = validatePageFrontmatter(input.frontmatter, incrementOptions('frontmatter', opts));
+  }
+
+  return {
+    mdast,
```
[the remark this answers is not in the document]
agoose77: Right, I'd like that too. Perhaps we can open that goal down the road - maybe we only build upon the published msyt spec rather than breaking it in main?

packages/myst-cli/src/process/file.ts
agoose77: Ah, this was a good spot. I indeed wanted `jsonFrontmatter` to take precedence. I pulled in the `fill` function because I wanted to avoid a spread operator (thinking it was wrong to clobber things), but it looks like that's what we do internally anyway in `getPageFrontmatter`. So, I've made the changes!

agoose77:
Fab! Thanks @fwkoch, will merge this in!

PR #1410, opened 2024-07-24
🎆 Fall back to 'convert' command for imagemagick

fwkoch:
In #1388 we switched our ImageMagick command from `convert` (used by ImageMagick >= 6) to `magick` (used by ImageMagick 7). However, some linux distributions including, most notably, Ubuntu 22.04 - `ubuntu-latest` on github actions, still use ImageMagick 6.

This PR updates the invocation of ImageMagic commands to still use `convert` if that command is available and `magick` is not (`magick` still remains the preferred default if both are available).

### DIFF
--- a/packages/myst-cli/src/utils/imagemagick.ts
+++ b/packages/myst-cli/src/utils/imagemagick.ts
@@ -7,8 +7,23 @@ import { RuleId } from 'myst-common';
 import type { ISession } from '../session/types.js';
 import { addWarningForFile } from './addWarningForFile.js';
 
-export function isImageMagickAvailable() {
+function magickCommandAvailable() {
   return which.sync('magick', { nothrow: true });
+}
+
+function convertCommandAvailable() {
+  return which.sync('convert', { nothrow: true });
+}
+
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/utils/imagemagick.ts
```diff
@@ -187,14 +202,14 @@ export async function convertImageToWebp(
...
-    `magick -density 600 -colorspace RGB ${image} -trim ${png}`,
+    `${imageMagickCommand()} -density 600 -colorspace RGB ${image} -trim ${png}`,
     debugLogger,
   );
   const convertPdfWebP = makeExecutable(`cwebp -q ${quality} "${png}" -o "${webp}"`, debugLogger);
 
   try {
     if (path.extname(image) === '.pdf') {
-      if (!isImageMagickAvailable() && !isWebpAvailable()) {
+      if (!isImageMagickAvailable() || !isWebpAvailable()) {
```
fwkoch, line 212: This was also a bug - both of these are required for pdf webp conversion; previously this passed if only one was available.

PR #1411, opened 2024-07-25
🔭 Trim spaces in bibtex URLs and DOIs

rowanc1:


fwkoch (approved):
This is nice - I'm wondering if we could be even more liberal with trimming other fields? For example, [this bibtex](http://conference.scipy.org.s3-website-us-east-1.amazonaws.com/proceedings/scipy2018/bib/anton_malakhov.bib) is rendered with extra spaces (besides the DOI space which this PR should address):

image

PR #1416, opened 2024-07-29
:dart: Support `--execute` for building non-sites

### LINKED ISSUE #1415: Execution of code cells with when building tex / pdf
Dear mystmd team,

1. Thanks for developing mystmd - this is exactly what I was looking for since long time :-)
2. If have a few executable code cells in which are execute as expected when I build with `myst build --execute`. However when I build the `pdf` or `tex` version with `myst build --execute --pdf` or `myst build --execute --tex`. The code cells are not being executed. Is that expected?

agoose77:
Fixes #1415 

@fwkoch I'm not clear on what the purpose of `session.clone` is, so I'm not sure whether sharing a Jupyter session between all clones is reasonable. I *think* we want the Jupyter session manager to be a singleton so that we don't end up spinning up multiple local servers, but your input is useful here if I've got the wrong end of the stick.

### DIFF
--- a/.changeset/real-clouds-jam.md
+++ b/.changeset/real-clouds-jam.md
@@ -1,5 +1,5 @@
 ---
-"myst-cli": patch
+'myst-cli': patch
 ---
 
 Fix execution for non-site builds
--- a/.changeset/real-clouds-jam.md
+++ b/.changeset/real-clouds-jam.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Fix execution for non-site builds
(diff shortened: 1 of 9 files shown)


fwkoch (approved):
Oh nooo, I made basically this identical change at scipy, but it got lost in the shuffle 😅 - sorry! Thanks for getting it over the line. The diffs are almost exactly the same, which builds some confidence... https://github.com/jupyter-book/mystmd/commit/f96a8628ce78d3c66f80c239ac7f5292de4600e0

Also - regarding `Session` clones: This is required since sessions have a single store with a single mdast entry for each file. Each export runs its own transformations which mutate the mdast in place, so when running exports in parallel, they must each have their own session or these mutations will conflict. As far as I know, your Jupyter session can (and should) be shared across these, no problem.

agoose77:
Haha, that's sprints for you I suppose! Thanks for taking a look at this, I feel reassured that we settled on the same solutions. Merging :)

Issue #190, opened 2023-02-07
Index role / directive (similar to sphinx)


rowanc1:
### Which area is this feature request for?

Website User Interface, Command Line Tools

### Describe the feature you'd like to request

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup

````
### Inner Product and Norm

```{index} single: Vectors; Inner Product
```

```{index} single: Vectors; Norm
```
````

![image](https://user-images.githubusercontent.com/913249/193480195-dfc16c35-78a2-4d48-a70c-e46d824b39da.png)

**Comment by fwkoch** [comment]
We intend to support as much of the [Sphinx syntax](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup) as possible including:

- single/pair/triple
- see/seealso
- directive argument shorthand
- name/label
- index role

### Examples:

Index entries as directive body
````
```{index}
single: Vectors; Inner Product
single: Vectors; Norm
```
````

Index entry as options (n.b. no duplicates are allowed if `:option:` syntax is used)
````
```{index}
:single: Vectors; Inner Product
```
````

Index entry as directive argument
````
```{index} single: Vectors; Inner Product
```
````

Index entries as directive argument with shorthand, comma-separated `single` entries
````
```{index} Vectors; Inner Product, Vectors; Norm
```
````

Other index types `pair`/`triple`/`see`/`seealso`
````
```{index}
triple: Vectors; Inner Product; Norm
see: Dot Product; Inner Product
```
````

Index role
```
This is a {index}`paragraph` that contains several {index}`index entries <pair: index; entry>`.
```

rowanc1:
I have put together a PR to fix that up - thank you for the code snippet!

![image](https://github.com/user-attachments/assets/005c7800-3133-457e-bc67-fde872e9d5c8)

fwkoch:
#1418 addressed this a long time ago 🎉

Issue #1351, opened 2024-06-21
Allow site build to complete quickly then re-render as slower steps complete


fwkoch:
This issue is motivated by recent problems with slow DOI loading. In those cases, after running `myst start` everything just seemed to freeze for a long time before the built site was available. There are a few other slow steps in the build process, e.g. image processing. It would improve the user's experience if the build finished faster, leaving placeholder content for the slower steps, then, on completion of those steps, re-rendered.

This would take a little refactoring of the `processMdast`/`postProcessMdast` steps - e.g. transforms might need to check if prerequisites have been met, rather than assuming all the previous steps are complete; slow transforms will need to be async with a callback to re-render. We would also need to be careful about designing the placeholder content so the user understands the build has not yet finished. However, I think this would make the local authoring experience smoother, and I have a feeling it would also improve consistency of the transform code, forcing us to rely on fewer assumptions about the current state of the mdast when a transform runs.

agoose77:
I think this ties into more explicit dependency management, which we touched on somewhere else. This would be useful for plugins to be able to inject themselves into arbitrary points in the transform pipeline.

PR #1418, opened 2024-07-31
📖 Add Index support to MyST

fwkoch:
This addresses https://github.com/jupyter-book/mystmd/issues/190

It replicates the Sphinx functionality described here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup

There is an `index` role, `index` directive, and `genindex` directive to locate the index. (This last feature is slightly different from Sphinx where genindex is a "magic" toc entry; as a directive, genindex follows existing MyST patterns better.)

For the web build, we use a `definitionList` to make an index that looks something like:

image

The `index` functionality was also hooked into latex export using the `imakeidx` package:

image

This PR _does not_ implement support for generating index items (other than from glossary items), as this needs further discussion.


packages/myst-cli/src/process/site.ts
```diff
@@ -253,6 +261,19 @@ export function selectPageReferenceStates(
     })
     .filter((state): state is ReferenceState => !!state);
   if (!opts?.suppressWarnings) warnOnDuplicateIdentifiers(session, pageReferenceStates);
+  pageReferenceStates.forEach((state) => {
+    const { mdast } = cache.$getMdast(state.filePath)?.post ?? {};
+    if (!mdast) return;
+    const vfile = new VFile();
+    vfile.path = state.filePath;
+    buildIndexTransform(
```
fwkoch, line 269: This transform must occur after `processMdast` but before the parallel `postProcessMdast` step... For now, that is this overloaded `selectPageReferenceStates` function.

I think the complexity of all these different steps will be simplified once we tackle https://github.com/jupyter-book/mystmd/issues/1351

packages/myst-transforms/src/basic.ts
```diff
@@ -19,14 +19,14 @@ export function basicTransformations(tree: GenericParent, file: VFile, opts?: Re
   // lifting roles and directives must happen before the mystTarget transformation
   liftMystDirectivesAndRolesTransform(tree);
   // Some specifics about the ordering are noted below
+  // Target transformation must happen after lifting the directives, and before the heading labels
+  mystTargetsTransform(tree, file);
```
fwkoch, line 23: This rearranging is not totally necessary. However, in `getFrontmatter` we must run `mystTargetsTransform` and _all_ it's prerequisites. By moving this up earlier, we explicitly show that the only dependency is `liftMystDirectivesAndRolesTransform`.

packages/myst-transforms/src/frontmatter.ts
```diff
@@ -34,7 +35,10 @@ export function getFrontmatter(
   tree: GenericParent,
   opts: Options = { propagateTargets: true },
 ): { tree: GenericParent; frontmatter: Record<string, any>; identifiers: string[] } {
-  if (opts.propagateTargets) mystTargetsTransform(tree);
+  if (opts.propagateTargets) {
+    liftMystDirectivesAndRolesTransform(tree);
```
fwkoch, line 39: (This is where all the prerequisites are called - trying to keep it minimal)

packages/myst-transforms/src/glossary.ts
```diff
@@ -26,6 +26,7 @@ export function glossaryTransform<T extends Node | GenericParent>(mdast: T, file
           child.label = label;
           child.identifier = `term-${identifier}`;
           child.html_id = `term-${html_id}`;
+          child.indexEntries = [{ entry: toText(child) }];
```
fwkoch, line 29: All glossary terms get a corresponding index entry.

packages/myst-transforms/src/math.ts
```diff
@@ -43,16 +43,28 @@ function labelMathNodes(file: VFile, node: Math | InlineMath) {
...
-    if (node.enumerated === false) {
-      fileWarn(file, `Labelling an unnumbered math node with "\\label{${label}}"`, {
-        node,
-        source: TRANSFORM_NAME,
-        ruleId: RuleId.mathLabelLifted,
-      });
+    if (node.label) {
+      fileWarn(
+        file,
+        `Math node is already labeled "${node.label}" - ignoring inline "\\label{${label}}"`,
```
fwkoch, line 49: Given the edge case of a math node with both (1) `\label` in the content and (2) `mystTarget` label: This extra warning ensures behavior is unchanged now that `mystTargetsTransform` happens before these math transforms.

packages/myst-common/src/indices.ts
```diff
@@ -0,0 +1,139 @@
+import type { VFile } from 'vfile';
+import { fileError } from './utils.js';
+import type { GenericNode } from './types.js';
+
+export type IndexEntry = {
+  entry: string;
+  subEntry?: string;
+  emphasis?: boolean;
+  see?: boolean;
```
fwkoch, line 9: Make see/seeAlso enum or kind or something, so it's exclusive in the AST

packages/myst-directives/src/indices.ts
```diff
@@ -0,0 +1,83 @@
...
+        label: data.options?.label as string | undefined,
+        indexEntries: entries,
+      },
+    ];
+    return output;
+  },
+};
+
+export const genIndexDirective: DirectiveSpec = {
+  name: 'genindex',
```
fwkoch, line 61: Maybe this should be `show-index` (or something) and keep `genindex` as an alias, i.e. we are not generating the index at this time necessarily...

packages/myst-common/src/utils.ts
```diff
@@ -75,6 +78,31 @@ export function createHtmlId(identifier?: string): string | undefined {
...
+  }
+  if (sourceNode.identifier) {
+    destNode.identifier = sourceNode.identifier;
+    delete sourceNode.identifier;
+  }
+  if (sourceNode.html_id) {
+    destNode.html_id = sourceNode.html_id;
+    delete sourceNode.html_id;
+  }
+  if (sourceNode.indexEntries) {
```
agoose77, line 99: Just adding the comment here to anchor in review:

We should probably add `indexEntries` to our `myst-spec-ext` as another sibling to `identifier` and `label`.

packages/myst-transforms/src/indices.ts
```diff
@@ -0,0 +1,295 @@
...
+        nodes: IndexTargetInfo[];
+        see: IndexSeeInfo[];
+        seeAlso: IndexSeeInfo[];
+        subEntries: Record<string, IndexTargetInfo[]>;
+      }
+    >
+  > = {};
+  indexTargets.forEach(({ node }) => {
+    node.indexEntries?.forEach(({ entry, subEntry, emphasis, see, seeAlso }) => {
+      const letter = ALPHABET_UPPER.includes(entry[0]?.toUpperCase())
```
agoose77, line 61: This means we don't support grouping by number (only letters). Is that intentional?

agoose77 (approved):
We went through this review on Discord and then async.

One question that surfaces is on syntax; we've inherited the Sphinx syntax here, but is there a better MyST-focused variant that we can design? In general, I feel like avoiding DSLs is useful in MyST where possible. That might been using YAML-like constructs for e.g. `triple`.

An interesting feature of this work is that `index` is the first inline-AST node that can be linked to (?). Right now, we can't define custom targets for inline elements, only via
```markdown
(foo)=
...
```

We spoke about the need for a new transform stage being a symptom of the need to refactor our transform dependency management.

choldgraf (commented):
I've got a PR to add docs to this PR here:

- https://github.com/jupyter-book/mystmd/pull/1424

A few thoughts that came up as I was doing it:

- the semi-colon syntax is a bit weird. Prefacing entries with `single:` `double:` and `triple:` felt clunky just to have the semicolon `;` be overloaded in meaning. Could we not use some other symbol to separate those two and use the prefix just for Sphinx support?
- The `genindex` entries are only updated if the page with `{genindex}` is updated, not if only the entries are updated.

agoose77:
> So generally speaking, I'm in favor of @agoose77 and @fwkoch 's thinking that we can improve the UX here and it's OK if we don't 100% pull from Sphinx syntax.

It's good that we're all on a similar train of thought!

> I've got a PR to add docs to this PR here (I tried to push directly but we don't have permissions set up for that yet)
> 
>     * [Add documentation to index support #1424](https://github.com/jupyter-book/mystmd/pull/1424)

Fab! I know @fwkoch was going to touch on docs shortly, so hopefully this makes the job easier ;)

> A few thoughts that came up as I was doing it:
> 
>     * the semi-colon syntax is a bit weird. Prefacing entries with `single:` `double:` and `triple:` felt clunky just to have the semicolon `;` be overloaded in meaning. Could we not use some other symbol to separate those two and use the prefix just for Sphinx support?


:100: Franklin and I both agree here. The wider conversation we started with in review was the difference between Sphinx's "every role has a DSL" and MyST's "prefer commonmark vs custom roles/directives". The most recent example of this philosophy difference is in the link syntax for refs/xrefs vs `{ref}` roles. Another example is glossaries; we support slightly less than Sphinx at the moment, but as a result the content of a glossary directive is just MyST Markdown.

I have a suggestion; let's split this PR into the parsing and backend logic, such that we can immediately merge the non-parsing work, and have a bit more time to think about the syntax? @fwkoch what do you think? (I'll ping on Discord too).

>     * The `genindex` entries are only updated if the page with `{genindex}` is updated, not if only the entries are updated.

We don't have a great handle on xref-dependencies right now IIRC. 

>     * It would be really helpful if indices could be defined in the MyST frontmatter of a page. That feels cleaner than using a directive. E.g.:
>       ```yaml
>       ---
>       title: Some title
>       index:
>         - "entry parent; entry child"
>         - "pair: entry two; entry three"
>       ---
>       ```
I think these are orthogonal features, but I agree -- page entries would be nice. Another PR?

choldgraf:
You should revert whatever you like from that PR, I was just trying to get something up that people could start to iterate on. Take it where you'd like!

Though I'm a bit confused as to how people will manually define index entries if there is no role or directive.

rowanc1:
One of our requirements is that this works with both JupyterBook v1 and MyST, which means non backwards compatible syntax changes are not possible.

I agree whole heatedly that the syntax can be improved, but I would prefer we do that in a different PR, leave the directives in, and get something out that improves the compatibility and functionality now.

It should be easy to support both syntaxes in the future, as the current one is well constrained. Also agree that we should use less magic syntax in the future.

agoose77:
@rowanc1 do you have anything against splitting this PR? Is there an external deadline in play?

rowanc1:
Splitting the PRs is fine if you prefer. Our main constraint is the legacy syntax support, which we can't take the direction of the glossary (non backwards compatible).

For our deadlines, this week would be ideal to release and support an author who has a book editor breathing down their neck. :)

I would prefer that new syntax discussions (which I support!) are not coupled to supporting JBv1 books.

agoose77:
> One of our requirements is that this works with both JupyterBook v1 and MyST, which means non backwards compatible syntax changes are not possible.

This can be an upgrade step if the author can perform a one-way upgrade. I'm guessing from the context that this is not the case?

rowanc1:
Yes - in our current use case we require that both JupyterBook and MyST are running at the same time on the same content. MyST is still missing features that are preventing an upgrade including search, cross-page numbering, launch buttons, and a few other UI things.

choldgraf:
In that case, I'd recommend that we re-introduce the language about `{index}` role and directive that Angus removed, then add an admonition that is big and says "this is legacy support from Jupyter Book that we may remove soon", and then merge the PR ASAP. Future PRs can update and/or deprecate the syntax.

packages/myst-transforms/src/indices.ts
fwkoch: This is intentional, only to match legacy jupyter-book behavior. Entries that start with numbers appear under `Symbols`:
image

Could be a nice area to improve with user-facing directive options? But for now, I think we just keep this behavior.

packages/myst-transforms/src/frontmatter.ts
fwkoch: Yes - I'd agree with that. There were probably edge cases of, for example, a `(target)=` inside an admonition that would not resolve correctly. Now, `index` directives make this no longer an edge case.

fwkoch:
Thanks for all the comments (and the pinch hitting on the docs!) - I've addressed the comments from @agoose77 and made some updates to the docs so they reflect the implementation.

For now, we have leaned entirely into the legacy sphinx syntax. I've even introduced a warning if someone accidentally slips into directive option syntax (e.g. `:single: ...` instead of `single: ...`). This allows us more freedom if/when we redesign these roles/directives; the only legacy support is explicitly the old sphinx stuff - _nothing_ new has been introduced.

I really like the frontmatter-level index definitions, but also agree that is a separate PR.

The only think I'm not super happy with is `genindex` not updating when entries are added on other pages. We do track dependencies for `embed`/`include` content and rebuild dependent pages for those; `genindex` would be similar... However, the `genindex` update is a little more conditional, i.e. it is potentially dependent on _every_ other page, but only updates if those pages have an index entry. My inclination for this is to open an issue so we are not blocked from landing this PR.

choldgraf:
In my opinion this is a clear improvement on what's there, it unblocks a user that needs this functionality, we know we want the functionality we just need to improve the UX, but none of that is a critical blocker. I think we should open issues for anything we want to follow up on, and merge this in.

agoose77:
> pinch hitting

I just learned a new idiom lol

@fwkoch please merge this once you're happy! I agree that we can iterate.

choldgraf:
American baseball metaphors FTW! 😅

nthiery:
Excellent. Thank you for the support of indexes!

At this stage, is it possible in a custom role to generate an index role as part of its output? From the comment above, I assume not?

rowanc1:
There is some documentation on the index role here:
https://mystmd.org/guide/glossaries-and-terms#index-entry-role

It is possible to to create a different directive/role that would create an index by exporting an `indexEntries` list on any node ([type](https://github.com/executablebooks/mystmd/blob/main/packages/myst-spec-ext/src/types.ts#L26)).

If you have a plugin, you would add `node.indexEntries = [{ entry: 'My Index Entry' }];` to the node directly (ideally it also has a label/identifier pair). For example, the [term in glossaries does this](https://github.com/executablebooks/mystmd/blob/main/packages/myst-transforms/src/glossary.ts#L29)). Is that what you are asking?

nthiery:
Thanks @rowanc1! These are exactly the pointers I needed. Granted, in retrospect I could have guessed them :-)

PR #1422, opened 2024-08-06
add  jupyter book/ quarto labels for lighthouse runs in captions

brownsarahm:
they were not labeled in anything rendered on the page, only the references and image file names


agoose77 (approved):

agoose77:
Thanks @brownsarahm, that's a useful fix!

rowanc1:
I noticed that a while ago too, but didn't fix it! Thank you for your help!

PR #1423, opened 2024-08-07
🐍 Add parser support for additional markup features

fwkoch:
This PR includes some work from SciPy sprints to support a wider variety of markdown syntaxes:

- roles from shortcodes, e.g. `{{< role value >}}`
- labels at the end of headings/images/etc, e.g. `{#my-label}`
- classes on spans, e.g. `[content]{.python}`
- colon fences can be used to create divs with classes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13617,6 +13617,36 @@
         "url": "https://opencollective.com/unified"
       }
     },
+    "node_modules/unist-util-find-before": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/unist-util-find-before/-/unist-util-find-before-4.0.0.tgz",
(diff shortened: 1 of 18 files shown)


choldgraf:
Being able to add classes to spans and divs is really nice, especially if we can get custom CSS rules supported:

e.g. either:

- https://github.com/jupyter-book/myst-theme/issues/321
- Or an issue that lets people define custom CSS in page metadata

agoose77:
@rowanc1 we spoke about this at SciPy and I recall talking about which side of the markup the `{}` should go for spans. I think we also thought about whether roles should change too.

Can you remember any of the conclusions? It was frenetic time...

rowanc1:
I don't think this is ever intended to merge as is, just highlighting a few of the experiments we did at scipy!

PR #1424, opened 2024-08-07
Add documentation to index support

choldgraf:
This adds documentation about indices to the index support PR. I tried to push to your branch directly but it was rejected so this is a PR for the same thing.

I wrote the docs because I wanted to see how the index support felt in actual usage. Hopefully this is helpful as well!


choldgraf:
I'm gonna merge this in so that we can see what the previews look like

PR #1426, opened 2024-08-07
:date: Fix date representation (again)

### LINKED ISSUE #1357: date rendering is a little off
setting a date in the frontmatter results in a date being displayed in all the pages; except that it does not match 

1.2.5

## Description

if I set
```
date: 2024 June
```
I get rendered under the title and authors:
```
May 31, 2024
```

and if I set
```
date: 2024 June 25
```
I get
```
June 24, 2024
```

## Proposed solution

either use the date string as is, or make sure any performed date parsing/conversion does not alter the resulting rendering

## Additional notes

just in case the timezone could matter, my tests were run in Europe in CEST

agoose77:
Presently, our datetime handling is not quite right. We sometimes (due to `Date`) parse dates in the local timezone, but we render them in templates as UTC.

i.e. for BST=UTC+1
```node
> // A valid UTC date
> new Date(Date.UTC(2024, 6, 1))
2024-07-01T00:00:00.000Z
> // The same date without UTC
> new Date(2024, 6, 1)
2024-06-30T23:00:00.000Z
```
where the `Z` suffix indicates UTC. At the moment, we then ask for `getUTCDate()`, which returns `30` because the code parses the date as the `00:00:00` timestamp in BST, which is `23:00:00` the previous day in UTC.

Instead, we should assume that a user is entering a valid "local"/"wall time" date. This PR fixes mystmd's validation to explicitly ignore timezones and directly take the day/month information from the `Date` object.

Should fix #1357

See also https://tc39.es/proposal-temporal/docs/ambiguity.html

### DIFF
--- a/packages/myst-frontmatter/src/project/project.yml
+++ b/packages/myst-frontmatter/src/project/project.yml
@@ -27,7 +27,7 @@ cases:
       affiliations:
         - id: univa
           name: University A
-      date: 14 Dec 2021
+      date: '14 Dec 2021'
       doi: 10.1000/abcd/efg012
       arxiv: https://arxiv.org/example
       open_access: true
@@ -76,7 +76,7 @@ cases:
       affiliations:
         - id: univa
           name: University A
-      date: 14 Dec 2021
+      date: '2021-12-14'
       doi: 10.1000/abcd/efg012
       arxiv: https://arxiv.org/example
       open_access: true
--- a/.changeset/yellow-hotels-visit.md
+++ b/.changeset/yellow-hotels-visit.md
@@ -1,6 +1,6 @@
 ---
-"simple-validators": patch
-"myst-templates": patch
+'simple-validators': patch
+'myst-templates': patch
 ---
 
 Fix datetime handling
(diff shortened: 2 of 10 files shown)


packages/myst-frontmatter/src/project/project.yml
```diff
@@ -76,7 +76,7 @@ cases:
       affiliations:
         - id: univa
           name: University A
-      date: 14 Dec 2021
+      date: '2021-12-14'
```
agoose77, line 79: Escape YAML interpretation

agoose77:
Ugh, yikes, more problems.

`new Date(YYYY, MM, DD)` parses as a local date, whereas `new Date("YYYY-MM-DD")` parses as UTC. We really need a better way to handle this.

agoose77:
I *think* we can use `moment.js`'s `parseZone` function. This assumes _ambiguous_ input is UTC, and non-ambiguous input keeps the timezone. In other words, there's no timezone shift happening, which means we can use the raw-parsed components.

agoose77:
The history of this PR will be confusing. I'll open a new one!

PR #1427, opened 2024-08-07
␡ Enable strikethrough as markdownit extension

sglyon:
Ends up parsing out to a `delete` node and uses the rendering logic for the delete directive

### DIFF
--- a/packages/myst-parser/src/myst.ts
+++ b/packages/myst-parser/src/myst.ts
@@ -41,6 +41,7 @@ export const defaultOptions: Omit<AllOptions, 'vfile'> = {
     tasklist: true,
     tables: true,
     blocks: true,
+    strikethrough: false,
   },
   mdast: {},
   directives: defaultDirectives,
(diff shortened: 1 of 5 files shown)


sglyon:
Hi @rowanc1 is this ready to merge?

rowanc1:
Added a test - thanks for the nudge.

Will get this in when it goes green!

packages/myst-parser/src/myst.ts
```diff
@@ -75,6 +76,7 @@ export function createTokenizer(opts?: Options) {
   if (markdownit.linkify) {
     tokenizer.linkify.tlds(tlds.filter((tld) => !EXCLUDE_TLDS.includes(tld)));
   }
+  if (extensions.strikethrough) tokenizer.enable('strikethrough');
```
rowanc1, line 79: Note: This is not enabled by default and I am not sure if there is a way to turn this on from the CLI quite yet. We can add those features in future!

rowanc1:
Thanks @sglyon for your help!

PR #1428, opened 2024-08-08
📄 Small typst improvements

fwkoch:
This also depends on an update to `tex-to-typst`: https://github.com/curvenote/tex-to-typst/pull/8

### DIFF
--- a/packages/myst-to-typst/src/math.ts
+++ b/packages/myst-to-typst/src/math.ts
@@ -1,12 +1,15 @@
+import { normalizeLabel } from 'myst-common';
 import { texToTypst } from 'tex-to-typst';
 import type { Handler, ITypstSerializer, MathPlugins } from './types.js';
 
 function addMacrosToState(value: string, state: ITypstSerializer) {
   if (!state.options.math) return;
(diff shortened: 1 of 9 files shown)


packages/myst-to-typst/src/math.ts
```diff
@@ -55,16 +58,19 @@ export function resolveRecursiveCommands(plugins: MathPlugins): MathPlugins {
 }
 
 const math: Handler = (node, state) => {
-  const value = texToTypst(node.value);
+  const { value, macros } = texToTypst(node.value);
+  macros?.forEach((macro) => state.useMacro(macro));
+  const { identifier: label } = normalizeLabel(node.label) ?? {};
```
fwkoch, line 63: We should update `normalizeLabel` to not allow spaces at all - these break typst and seem funny for other contexts.

rowanc1: I think we can turn this into a bug and get this one out?

fwkoch: https://github.com/jupyter-book/mystmd/issues/1429

Issue #1361, opened 2024-06-27
Add support for icons


agoose77:
This issue tracks our missing support for icons used in the Executable Books and Sphinx ecosystems, defined in https://github.com/executablebooks/sphinx-design.

agoose77:
@rowanc1 @choldgraf I have started tackling this today, and I have some thoughts on things we could improve relative to SD.

- `sphinx-design` uses long role names e.g. `{material-twotone}`
- `sphinx-design` uses different specs depending on the icon family to define classes and size information, e.g.
   - `` {octicon}`report;1em;sd-text-info` `` (`name`;`size`;`class`)
   - `` {fab}`github;sd-text-info` `` (`name`;`class`)
- `sphinx-design` has a different role _type_ per icon, e.g. `` {fab}`github` ``, `` {octicon}`report` ``

I think we should write a new `icon` role that `myst init` upgrades the legacy role types to. This new role _wouldn't_ support size information (because that's styling), and we would use the following syntax:
```markdown
{icon}`fab:github`
{icon}`oct:report`
{icon}`mtt:settings`
```

If we want to add classes, I'm not sure whether we make an initial release _without_ them and then leverage a more generic inline-attribute syntax e.g. #1423

Otherwise, we could for now inline it in the role body e.g.
```markdown
{icon}`fab:github .large`
```

I'm obviously against that.

I also wonder whether we should formally acknowledge that really we have three icon packs with various `kinds`, e.g.

```markdown
{icon}`fa:b:github`
```
or
```markdown
{icon}`fa:r#github`
```

So far, I'm in favour of just having each combination of variant and family to be a new "icon pack" e.g. `far`, `fab`, `oct`, `mtt`, etc.

Does anyone have any thoughts here?

rowanc1:
Quite like `` {icon}`fab:github` ``, and bringing in classes/ids and other inline attributes in a standardized way later. Another alternative closer to the cite roles is `` {icon:fab}`github` ``, which also seems ~ok (this could be a standardized way of encoding `kind`? e.g. `prf:example`, `cite:p`).

I think we should transparently alias to previous functionality using the alias in the role if that works?

agoose77:
@rowanc1 I was thinking about `{icon:fab}` but completely forgot about it when writing here. I think that's exactly how we should do it.

agoose77:
@rowanc1 are you not in favour of just introducing this syntax exclusively and upgrading using the upgrade tool?

[...]

PR #1432, opened 2024-08-09
🖼 Add `icon` node

agoose77:
This PR adds support for a subset of the icon syntax from sphinx-design. Although this new role supports the legacy role names such as `{fab}`, the preferred aliases are:

```
'icon:oct',
'icon:mrg',
'icon:mol',
'icon:mrd',
'icon:msp',
'icon:mtt',
'icon:fab',
'icon:far',
'icon:fas'
```

### DIFF
--- a/packages/myst-cli/src/process/myst.ts
+++ b/packages/myst-cli/src/process/myst.ts
@@ -2,6 +2,7 @@ import { mystParse } from 'myst-parser';
 import { cardDirective } from 'myst-ext-card';
 import { gridDirectives } from 'myst-ext-grid';
 import { proofDirective } from 'myst-ext-proof';
+import { iconRole } from 'myst-ext-icon';
 import { exerciseDirectives } from 'myst-ext-exercise';
 import { reactiveDirective, reactiveRole } from 'myst-ext-reactive';
 import { tabDirectives } from 'myst-ext-tabs';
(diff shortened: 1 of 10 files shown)


packages/myst-ext-icon/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import type { RoleSpec, RoleData, GenericNode } from 'myst-common';
+
+export const iconRole: RoleSpec = {
+  name: 'fas',
+  alias: ['fab', 'far'],
```
rowanc1, line 5: Would have to update the kind logic below, but I think the role makes more sense as an icon role with aliases?

```suggestion
  name: 'icon',
  alias: ['fab', 'fas', 'far'],
```

agoose77: @rowanc1 see #1361 for where I think this PR is going (I'm working locally and have changed tack slightly).

agoose77:
@rowanc1 should we look to merging this ahead of getting theme support? Do you concur that it's worth targeting the same icon packs as sphinx-design?

packages/myst-cli/src/process/myst.ts
```diff
@@ -49,7 +50,7 @@ export function parseMyst(
     extensions: {
       frontmatter: !opts?.ignoreFrontmatter,
     },
-    roles: [reactiveRole, ...(session.plugins?.roles ?? [])],
+    roles: [reactiveRole, iconRole, ...(session.plugins?.roles ?? [])],
```
rowanc1, line 53: Happy with a merge except this, which we should coordinate with a theme release as that may change behaviour of the package.

rowanc1:
Let's talk about the icon packs we support when we build out the theme (probably lower priority than some of the other things we have on at the moment).

PR #1433, opened 2024-08-09
Update and refactor quickstart tutorials

choldgraf:
This is a light refactor and update of the Quickstart tutorial. It has the following goals:

- Make each tutorial scoped a bit more tightly. I've split the "Work with MyST documents" tutorial into two: one focused on enriched documents with MyST, and another about exporting. We can then flesh these out more independently.
- Reduce some of the clutter and redundant information, with point-backs rather than embeds in several spots. I found it complex to navigate all the different callouts etc, and thought this made it flow more easily.
- Tutorials should only have the most important and actionable information, and call out to other pages for more context etc. This tries to follow [the Diataxis guidelines](https://diataxis.fr/) for tutorials.
- Clean up some of the instructions for installing MyST, particularly because nodejs is a new workflow for many users.

The core of the content hasn't changed much, this is mostly cleaning things up, doing some refactoring, and trying to streamline a bit.


docs/installing.md
```diff
@@ -1,24 +1,70 @@
...
 description: MyST Markdown is available through Node and npm, install the package with `npm install mystmd`.
 ---
 
 +++
 
-The MyST Markdown Command Line Interface (CLI) is available through [NodeJS](./installing-prerequisites.md) and the node package manager, `npm`. Node is used by Jupyter as well as many other Python packages so you may already have it installed on your _PATH_ and the following command may just work 🤞.
+## Install MyST
 
-🛠️ [Install NodeJS](./installing-prerequisites.md), either version 18, 20, or 22
+The MyST Markdown Command Line Interface (CLI) is available through [NodeJS](./install-node.md) and the node package manager, `npm`. Node is used by Jupyter as well as many other Python packages so you may already have it installed on your _PATH_ and the following command may just work 🤞.
```
agoose77, line 11:
```suggestion
The MyST Markdown Command Line Interface (CLI) is available through [NodeJS](./install-node.md) and the node package manager, `npm`. Node is used by Jupyter (as well as many other Python packages) so you may already have it installed on your _PATH_, and the following command may just work 🤞.
```

agoose77:
I think a link broke on the index page that points to websites. I'll get to it once I've finished working on a bugfix PR.

agoose77:
I've fixed the broken quickstart link on the index page by _embedding_ the cards from `quickstart.md`.

docs/quickstart.md
```diff
@@ -1,182 +1,364 @@
...
-(installing-myst-tabs)=
-::::{tab-set}
-:::{tab-item} PyPI
+   ```shell
+   $ conda --version
+   conda 24.7.1
+   ```
 
-🛠 Install `node` (<https://nodejs.org>), see [Installing NodeJS](./installing-prerequisites.md):
+### Install Node and NPM
```
rowanc1, line 52: Is there a reason that these were removed?

There is now no longer a place where we can see how to install via pip or npm.

choldgraf: My reasoning was that we want quickstarts and tutorials to be opinionated, ready and executed linearly, and generally not provide many ways of doing something. Since we have [this page with much more information about installing MyST](https://mystmd.org/guide/installing) I figured we should define the one, simplest way to install it from the quickstart, and reference that page for more information.

PR #1435, opened 2024-08-09
:eye: Remove outputs and code cells with `visibility==='remove'` in exports

### LINKED ISSUE #1434: `remove-` tags do not affect non-site exports
Presently, the following cell is still visible in our TeX and Typst exports:
````markdown
```{code-cell}
:tags: remove-cell

# This entire cell should not appear
print(1 + 2)
```
````

### PR DESCRIPTION
**Pull request by agoose77**
This PR does the following:
- Make `typst` and `tex` exports aware of code-cell `visibility==='remove'`
- Remove outputs with `visibility==='remove'` during mdast finalisation

This last step has the highest probability of side-effects (in my opinion), but I think it's the "right" call. I'd prefer to handle visibility at the transform leaves (i.e. _during_ exports), but it's fine to do so earlier as long as we're mindful of it.

This PR fixes #1434

### DIFF
--- a/packages/myst-cli/src/transforms/outputs.ts
+++ b/packages/myst-cli/src/transforms/outputs.ts
@@ -38,12 +38,14 @@ export async function transformOutputsToCache(
   if (!outputs.length) return;
   const cache = castSession(session);
   await Promise.all(
-    outputs.map(async (output) => {
-      output.data = await minifyCellOutput(output.data as IOutput[], cache.$outputs, {
-        computeHash,
-        maxCharacters: opts?.minifyMaxCharacters,
-      });
-    }),
+    outputs
+      .filter((output) => output.visibility !== 'remove')
+      .map(async (output) => {
+        output.data = await minifyCellOutput(output.data as IOutput[], cache.$outputs, {
+          computeHash,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/myst-cli/src/transforms/outputs.ts [hunk 1]
```diff
@@ -38,12 +38,14 @@ export async function transformOutputsToCache(
...
   const cache = castSession(session);
   await Promise.all(
-    outputs.map(async (output) => {
-      output.data = await minifyCellOutput(output.data as IOutput[], cache.$outputs, {
-        computeHash,
-        maxCharacters: opts?.minifyMaxCharacters,
-      });
-    }),
+    outputs
+      .filter((output) => output.visibility !== 'remove')
```
rowanc1, line 42: This will break thebe. Can we do it in the render instead?

Similar to what we already do for the blocks:
https://github.com/executablebooks/mystmd/blob/main/packages/myst-to-typst/src/index.ts#L108

I think that just needs to be done for outputs/code.

rowanc1: Oh wait. This is outputs!

PR #1442, opened 2024-08-09
🍣 Add raw directives/roles for inserting tex/typst-specific content

fwkoch:
This PR expands the `raw` node/role/directive behavior to allow tex- and typst-specific raw content that is only included in the respective export.

Currently, there is some existing `raw` functionality:

````
```{raw}
This inserts *raw* `unparsed` text in all contexts.
```
````

````
```{raw} latex
This allows you to use \textit{latex} in your markdown.
It will be \texttt{parsed} to MyST AST and included in all contexts.
```
````

(I feel like the latter is a little funny, since it says "raw" but it is parsing to AST... but that's ok for now.)

This PR adds `raw:latex` and `raw:typst` directives and roles that allow inserting typst and tex into your document that _only_ show up in their respective exports. So, for example, you can insert typst-specific formatting that is not required elsewhere:

````
```{raw:typst}
#set page(columns: 2, margin: (x: 1.5cm, y: 2cm),);
```
````

The `{raw} latex` directive content is also now added directly into latex exports.

### DIFF
--- a/packages/myst-roles/src/raw.ts
+++ b/packages/myst-roles/src/raw.ts
@@ -31,7 +31,7 @@ export const rawTypstRole: RoleSpec = {
     doc: 'Raw typst content',
   },
   run(data): Raw[] {
-    const lang = 'tex';
+    const lang = 'typst';
     const typst = (data.body as string) ?? '';
     return [
       {
--- a/packages/myst-roles/src/raw.ts
+++ b/packages/myst-roles/src/raw.ts
@@ -1,0 +1,44 @@
+import type { RoleSpec } from 'myst-common';
+import type { Raw } from 'myst-spec-ext';
+
+export const rawLatexRole: RoleSpec = {
+  name: 'raw:latex',
+  alias: ['raw:tex'],
+  doc: 'Allows you to include tex in your document that will only be included in tex exports',
+  body: {
+    type: String,
+    doc: 'Raw tex content',
+  },
+  run(data): Raw[] {
+    const lang = 'tex';
+    const tex = (data.body as string) ?? '';
+    return [
+      {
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: packages/myst-roles/src/raw.ts [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+export const rawTypstRole: RoleSpec = {
+  name: 'raw:typst',
+  alias: ['raw:typ'],
+  doc: 'Allows you to include typst in your document that will only be included in typst exports',
+  body: {
+    type: String,
+    doc: 'Raw typst content',
+  },
+  run(data): Raw[] {
+    const lang = 'tex';
```
rowanc1, line 34:
```suggestion
    const lang = 'typst';
```

rowanc1 (approved):
One change!

rowanc1:
Nice - this does a good job at allowing us to pass through tex and typst directly to the templates. Super useful for some of the typst templates that we are doing in SciPy and other articles!

packages/mystmd/tests/raw/outputs/index.json
```diff
@@ -0,0 +1,113 @@
...
+              {
+                "type": "text",
+                "value": "This has raw ",
+                "position": {
+                  "start": { "line": 3, "column": 1 },
+                  "end": { "line": 3, "column": 1 }
+                }
+              },
+              { "type": "raw", "lang": "tex", "tex": "LaTeX" },
+              { "type": "raw", "lang": "tex", "typst": "Typst" },
```
rowanc1, line 38:
```suggestion
              { "type": "raw", "lang": "typst", "typst": "Typst" },
```

PR #1446, opened 2024-08-10
📖 Improve website navigation and TOC docs

choldgraf:
This tries to make our Table of Contents and Website Navigation documentation a little bit clearer and easier to find. Here are major changes:

- Move `Table of Contents` into `Authoring` instead of `Websites`. This is because most of the content there is not unique to websites, since you structure multi-page documents with the Table of Contents for any kind of export.
- Adds a `Websites` section on `Site Navigation` to make it clear where to find menu item configuration. It also adds a brief description of the layout of the page.
	- _Note: this will only apply to the default themes, if we create dedicated documentation for those themes, we can move this section there._


docs/website-navigation.md
```diff
@@ -0,0 +1,102 @@
...
+(site-navigation)=
+
+### Site navigation
+
+In addition to [your MyST document's Table of Contents](./table-of-contents.md), you may specify a top-level navigation for your MyST site.
+These links are displayed across all pages of your site, and are useful for quickly jumping to sections of your site, or for external links.
+
+Specify top-level navigation with the `site.nav` option, and provide a collection of navigation links similar to [how the Table of Contents is structured](./table-of-contents.md). For example:
+
+```{code-block} yaml
```
lwasser, line 39: @choldgraf i was able to find this section but what i had a hard time with is understanding how and when to use url vs file. 

in the example below you have url associated with https links to mystmd.org 

i suggest expanding this example to also include some urls associated with files in the mystmd docs such as a ipynb file and a markdown file.

then please add just a few sentences that clarify what @fwkoch  clarified for me [here](https://github.com/pyOpenSci/peer-review-metrics/pull/27/files#r1712550491) 

which essentially is 

1. if you use file: document.ipynb then it will allow you to provide a downloadable file
2. if you use url: document it will create a link to document.xxx 

it's confusing to me to think about 
1. i have a jupyter notebook that will eventually be a html file. what is the path to that file?
2. intuitivaly i want to use file: because that suggests to me that i'm telling mystmd to find the `document.ipynb` file that will have some html representation in the build
3. because the example shows .org urls i just assumed that url was for an external url not an in project url. 

i hope that helps!! it's working now except i want to get rid of the sidebar toctree now that i have a top bar one working!

choldgraf: Thanks for this feedback! I gave it another shot at the commit below.
Do you think this commit clarifies it?

https://github.com/jupyter-book/mystmd/pull/1446/commits/b7dd89a14bc33f33271f67337376b0818e6370cc

docs/deployment-github-pages.md
```diff
@@ -26,24 +26,20 @@ Navigate to your repository settings, click on Pages and enable GitHub pages. Wh
...
 
-:::{warning} Custom Domains
-GitHub allows you to host your static content on a custom domain, doing so _may_ require you to change the `BASE_URL` environment variable in the action. If you have unstyled content, try changing the `BASE_URL` to a blank string: `BASE_URL=''` (note the **single quotes**!); this serves the build assets from the root of your domain, rather than the default, which is the name or your repository.
-:::
-
-:::{tip} `BASE_URL` environment variable
-:class: dropdown
-The build and site assets are in the `/build` folder, which would point outside of the current repository to a repository called 'build', which probably doesn't exist!
+:::{tip} You will probably need to set a `BASE_URL`
+GitHub Pages likely requires you to set a `BASE_URL` because the URL for GitHub Pages defaults to a sub-folder of `username.github.io` (e.g. `username.github.io/myrepository/`).
```
lwasser, line 30:
```suggestion
:::{tip} `BASE_URL` Configuration for GitHub Pages

If you are deploying your mystmd website from a specific repository within an organization that has a domain url setup then you likely need to define a `BASE_URL` that includes the repository name.

For example, if your GitHub organization was connected to mystmd.org and you had a repository  called myrepository where your. mystmd website lives, you'd want the base url to be:

`username.github.io/myrepository/`

GitHub Pages likely requires you to set a `BASE_URL` because the URL for GitHub Pages defaults to a sub-folder of `username.github.io` (e.g. `username.github.io/myrepository/`).
```

docs/deployment-github-pages.md
```diff
@@ -26,24 +26,20 @@ Navigate to your repository settings, click on Pages and enable GitHub pages. Wh
...
+The easieset thing to do is use the GitHub Action variable for the repository name:
```
lwasser, line 31:
```suggestion
The BASEURL variable is only needed for deploying your site to GitHub pages. If you are using GitHub actions to deploy your site, you can add a GitHub Action variable that defins the repository name and creates your BASE_URL like this:
```

docs/deployment-github-pages.md
```diff
@@ -26,24 +26,20 @@ Navigate to your repository settings, click on Pages and enable GitHub pages. Wh
...
 
-To fix this, we can change the base url that the site is mounted to, which can be done through the `BASE_URL` environment variable:
-
-```bash
-export BASE_URL="/repository_name"
+```yaml
+BASE_URL: /${{ github.event.repository.name }}
```
lwasser, line 34: Your GitHub action would then look something like this:

```suggestion
name: MyST GitHub Pages Deploy
on:
  schedule:
    - cron: '0 0 * * 0' # Run every Sunday @ midnight (UTC)
  pull_request:
  push:
    branches:
      - main
env:
  # `BASE_URL` determines the website is served from, including CSS & JS assets
  # You may need to change this to `BASE_URL: ''`
  BASE_URL: /${{ github.event.repository.name }}```

Comment (lwasser) [reply docs/deployment-github-pages.md:34]: i'm not sure about the comments here in my action. but i'm just thinking it would be good to be super explicit regarding where that base_url is set in the action.

File: docs/deployment-github-pages.md [hunk 5]
```diff
@@ -26,24 +26,20 @@ Navigate to your repository settings, click on Pages and enable GitHub pages. Wh
...
 ```
 
-The base URL is _absolute_ and should not end with a trailing slash. This can be done automatically in a GitHub Action by looking to the `github.event.repository.name` variable.
+If you're using a custom domain (e.g. `mydomain.org`) then you may need to manually set `BASE_URL` to an empty string, like `export BASE_URL=''`.
```
lwasser, line 37: Just a note that i did not need to do this (i think)

see [here](https://github.com/pyOpenSci/peer-review-metrics/blob/main/.github/workflows/deploy.yml#L12)

now i may not understand how this works yet but i didn't have to tell it my domain name, i did have to define the subfolder / repo name

docs/deployment.md
```diff
@@ -82,7 +82,31 @@ site:
...
 
-### Deploying to a Static Web Server
+(deploy:base-url)=
+### Custom domains and the base URL
+
+When deploying static HTML to the web, you may have to specify a Base URL so that links resolve properly.
+This happens when you serve your MyST site from a *subfolder* of a parent URL.
+For example:
+
+- `mysite.org/`: **Does not** require a `BASE_URL` to be set, because the root site is where your MyST site lives.
```
lwasser, line 92: This is super helpful, chris. i may be confused about the text above as it seems to contradict what is written VERY CLEARLY here.

docs/table-of-contents.md
```diff
@@ -4,7 +4,7 @@ description: The Table of Contents is the left-hand navigation for your site, it
 thumbnail: thumbnails/table-of-contents.png
 ---
```
lwasser, line 6: @choldgraf the other thing i'm trying to figure out that is confusing related to TOC / NAV
if you go to this page - notice there is no toctree on some pages and sidebar toctrees on others. 
https://mystmd.org/overview/ecosystem 

in one of the repos that i saw, there was a _toc.yml file that in the docs said was deprecated.

can you add info on (maybe i missed this) how you control what's on the left hand side bar for a folder / section ? if it's in the docs i just had a hard time finding it. if this is out of scope for this pr that is fine too - i know this is a nerd snipe situation which is not my goal! 🙃

docs/table-of-contents.md
```diff
@@ -13,6 +13,11 @@ To automatically add a `toc` section to your `myst.yml` file using filenames to
 myst init --write-toc
```
lwasser, line 13: This seems to just add all of the base level files. what if we want to define a structure for toctrees or no toctree in each section of the site?

docs/table-of-contents.md
choldgraf: The mystmd.org docs are actually not a great example of this because I think the content of those repositories is actually hosted in different places. It is confusing I agree. In my opinion it would be better for us to `embed` that content and try to accomplish the myst docs with a single repository without stitching together a bunch of repos in a hacky way.

I opened up this issue which I think would help your situation:

- https://github.com/jupyter-book/mystmd/issues/1450

docs/table-of-contents.md
choldgraf: The idea is that this just creates the initial structure for you, and you can then modify it however you want. re: `no toctree` could you describe better what the end result looks like that you're shooting for?

agoose77:
If @choldgraf gives this the go ahead, I'm then going to merge; I think we can iterate from here, and I'm worried about things becoming stagnant as we have an ever-growing list of PRs! Thanks @lwasser for your reviews here too :)

docs/myst.yml
```diff
@@ -79,6 +79,7 @@ project:
         - file: quickstart-jupyter-lab-myst.md
     - title: Authoring
       children:
+        - file: table-of-contents.md
```
rowanc1, line 82: I think this should be moved slightly further down the list. Maybe at the bottom?

rowanc1 (approved):
One minor thing - we can also rearrange these later.

choldgraf:
+1 to @rowanc1's suggestion as well as to your changes @agoose77 . I'm happy w the PR as-is but if you wanna move the TOC section lower before merging that works for me

PR #1452, opened 2024-08-12
:mermaid: Add conversion for static exports of mermaid diagrams

agoose77:
@roaldarbol this is a WIP, but I think you could try it out and see what else needs doing. I'm off this week, so I'll have little time to iterate.

You'll need `mmdc` on your `PATH`, as well as `inkscape`. I suspect things like figure centering might need additional work.


roaldarbol:
Hi @agoose77! Awesome, that's great to hear! There's no rush on my side - I'm just trying to assure that I won't be caught in a pickle close to submission. Enjoy your week off! 🥳

agoose77:
This was a PoC. I'm closing it for now.

PR #1454, opened 2024-08-12
⏩ Simplify quickstart guide to use `mamba`

agoose77:
Instead of going into detail about provisioning `node`, let's just go all-in on conda-forge; there's already a detailed article on installing manually

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -55,7 +55,7 @@ If you're familiar with either Python or JavaScript-based workflows, see [](inst
 :::
 ### Install the `mystmd` package
 
-Once `conda` has been installed, we can install the `mystmd` package from `conda-forge`:
+Once `mamba` has been installed, we can install the `mystmd` package (which `mamba` pulls from the `conda-forge` channel by default):
 ```shell
 mamba install mystmd
 ```
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -57,7 +57,7 @@ If you're familiar with either Python or JavaScript-based workflows, see [](inst
 
 Once `conda` has been installed, we can install the `mystmd` package from `conda-forge`:
 ```shell
-conda install -c conda-forge mystmd
+mamba install mystmd
 ```
 
 Once installed you should be able to print the version of MyST like so:
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -30,7 +30,7 @@ The current tutorial will help you get up and running from scratch.
 `mystmd` is a {abbr}`CLI (Command Line Interface)` that provides modern tooling for technical writing, reproducible science, and creating scientific & technical websites.
 These instructions help you install the CLI.
 
-### Install `conda` and `conda-forge`
+### Install `mamba`
 
 The easiest way to install MyST is with [the `mamba`/`conda` package manager](https://mamba.readthedocs.io/en/latest/index.html).
 `mamba` is a multi-language manager that is useful for users across many data science languages like Python, R, Julia, and JavaScript.
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -49,6 +49,10 @@ The easiest way to install `mamba` is with [the `miniforge` distribution](https:
    conda 24.7.1   
    ```
 
+:::{seealso} There are many ways to install MyST
+This tutorial focuses on the simplest possible way to install MyST from scratch.
+If you're familiar with either Python or JavaScript-based workflows, see [](installing.md) for more information on how to install MyST.
+:::
 ### Install the `mystmd` package
 
 Once `conda` has been installed, we can install the `mystmd` package from `conda-forge`:
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -71,7 +71,7 @@ v1.3.1
 
 MyST needs both NodeJS and NPM to build sites locally. By installing `mystmd` from conda-forge, we do not need to install NodeJS manually. 
 
-See [](./install-node.md) for more information about installing a JavaScript environment.
+See [](./install-node.md) for more information about installing a JavaScript environment if you opt not to use `conda-forge`.
 :::
 
 For more information about installing MyST, see [installing MyST](./installing.md).
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -32,7 +32,7 @@ These instructions help you install the CLI.
 
 ### Install `mamba`
 
-The easiest way to install MyST is with [the `mamba`/`conda` package manager](https://mamba.readthedocs.io/en/latest/index.html).
+The easiest way to install MyST is with [the `mamba` package manager](https://mamba.readthedocs.io/en/latest/index.html).
 `mamba` is a multi-language manager that is useful for users across many data science languages like Python, R, Julia, and JavaScript.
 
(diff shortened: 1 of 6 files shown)


agoose77:
@jnywong would you possibly be happy to take a look at this?

jnywong:
Happy to! 👍

docs/quickstart.md
```diff
@@ -32,47 +32,28 @@ These instructions help you install the CLI.
...
-:::
-
-### Install the `mystmd` Python package
-
-The MySTMD Python package is a wrapper around the MyST JavaScript library, and makes it easier to upgrade and use MyST with Python workflows.
-
-🛠 Install Python and the `mystmd` package:
+### Install the `mystmd` package
 
+Once `conda` has been installed, we can install the `mystmd` package from `conda-forge`:
```
jnywong, line 54:
```suggestion
Once `mamba` has been installed, we can install the `mystmd` package which `mamba` pulls from `conda-forge` channel by default:
```

rowanc1:
It looks like the tabs for different options have been removed in #1433. Can we bring them back?

https://github.com/jupyter-book/mystmd/pull/1433/files#r1713825789

There is currently no longer any documentation on how to install from pip and npm. The current structure also buries the lead significantly - if you know what you are doing `pip install mystmd` works out of the box, and then instructs you on how to install node. If you know npm, it is again one line. Tabs are able to highlight those workflows for people who are not using conda.

jnywong (approved):
LGTM! A few small suggestions added.

jnywong:
This PR came off the back of feedback from a tutorial we just gave. But I like your suggestion @rowanc1 of using the tabbed component for discoverability of other methods.

agoose77:
@rowanc1 a signal that we're picking up increasingly is that you use Python and most likely know about / use conda. In reality, if you are having to use new tools, just using conda is the fastest way to get people started. 

Being on PyPI is useful in-and-of itself as a binary story for our compiled code. But there's a lot less to go wrong in "Just use conda". 

I think we can better surface the "advanced installation" information that's already partly linked, though. I'll iterate on Jenny's points and think that through next week.

choldgraf:
Oops I didn't see this PR get opened, sorry! I actually did some of the same content work over in this PR as well:

- #1457 

happy to remove that and focus on this one though.

@rowanc1 to your question about installing, my reasoning was that we want tutorials to be as simple as humanly possible, and meant to be followed literally from top to bottom. That's based on [the principles of tutorials from the Diataxis documentation framework](https://diataxis.fr/tutorials/#key-principles) that I've found helpful to keep things simple and focused. We don't have to literally follow that if you think there's a better way of presenting the information.

My feeling is that we want all the options, considerations, etc to be documented [on the install page](https://mystmd.org/guide/installing). That way the tutorial stays as simple and actionable as possible.

However, as @rowanc1 notes we lost a lot of cross-references to that page. I think we should liberally cross-link that page from the `Get Started` tutorial.

docs/quickstart.md
```diff
@@ -32,47 +32,28 @@ These instructions help you install the CLI.
...
 2. Download the release for your platform and follow the instructions.
 3. To check that you've installed properly, you should be able to execute the following command to print the version of conda:
 
    ```shell
-   $ conda --version
-   conda 24.7.1
+   $ mamba --version
+   mamba 1.5.8
+   conda 24.7.1   
    ```
```
choldgraf, line 51:
````suggestion
   mamba 1.5.8
   conda 24.7.1   
   ```

:::{seealso} There are many ways to install MyST
This tutorial focuses on the simplest possible way to install MyST from scratch.
If you're familiar with either Python or JavaScript-based workflows, see [](installing.md) for more information on how to install MyST.
:::
````

File: docs/quickstart.md [hunk 3]
```diff
@@ -32,47 +32,28 @@ These instructions help you install the CLI.
 
 ### Install `conda` and `conda-forge`
 
-The easiest way to install MyST is with [the `conda` package manager](https://conda.io/projects/conda/en/latest/).
-`conda` is a multi-language manager that is useful for users across many data science languages like Python, R, Julia, and JavaScript.
+The easiest way to install MyST is with [the `mamba`/`conda` package manager](https://mamba.readthedocs.io/en/latest/index.html).
```
choldgraf, line 35:
```suggestion
The easiest way to install MyST is with [the `mamba`/`conda` package manager](https://mamba.readthedocs.io/en/latest/index.html).
```

I'd recommend just using mamba, or conda, but not both. We should assume that a user has no background in any of the installation methods we're talking about. Introducing multiple tools or concepts will just increase cognitive load. I'd suggest we just pick one way that is reasonable and explain that here, and then defer to `installing.md` for all the options, explanation, etc.

choldgraf: Also just noting the section calls it `conda` but the instructions call it `mamba` . This will confuse users who do not understand the relationship between the two projects (this is 99% of users).

docs/quickstart.md
```diff
@@ -32,47 +32,28 @@ These instructions help you install the CLI.
...
-The MySTMD Python package is a wrapper around the MyST JavaScript library, and makes it easier to upgrade and use MyST with Python workflows.
-
-🛠 Install Python and the `mystmd` package:
+### Install the `mystmd` package
 
+Once `conda` has been installed, we can install the `mystmd` package from `conda-forge`:
 ```shell
-conda install -c conda-forge pip
-pip install mystmd
+conda install -c conda-forge mystmd
```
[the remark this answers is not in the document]
choldgraf: Just noting that the miniforge distribution comes with `conda-forge` as the default (and only) channel, so you don't need to do `-c conda-forge` with `conda` either

choldgraf (commented):
A few quick thoughts on how we can streamline this a bit, but in general I think it's a helpful addition. Anything that removes steps needed to install, so we can focus on steps to learn the tools, is a good idea.

I also just want to re-iterate that my comments here are from the following assumptions, so we should make sure that others agree they are reasonable assumptions:

- We want these tutorials to be as to-the-point and action-oriented as possible.
- They are meant to be followed exactly, and linearly, top-to-bottom.
- We should not use them to provide many options, explanations, etc. That's what `installing.md` is for. This tutorial is meant to users momentum and quick wins.
- It should be designed for the most common user, which is probably somebody with a little bit of either Python or R experience, and that is it.

docs/quickstart.md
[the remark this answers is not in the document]
agoose77: Thanks both -- rushed this during the tutorial and clearly missed some bits.

docs/quickstart.md
```diff
@@ -30,49 +30,34 @@ The current tutorial will help you get up and running from scratch.
 `mystmd` is a {abbr}`CLI (Command Line Interface)` that provides modern tooling for technical writing, reproducible science, and creating scientific & technical websites.
 These instructions help you install the CLI.
 
-### Install `conda` and `conda-forge`
+### Install `mamba`
 
-The easiest way to install MyST is with [the `conda` package manager](https://conda.io/projects/conda/en/latest/).
-`conda` is a multi-language manager that is useful for users across many data science languages like Python, R, Julia, and JavaScript.
+The easiest way to install MyST is with [the `mamba`/`conda` package manager](https://mamba.readthedocs.io/en/latest/index.html).
```
agoose77, line 35:
```suggestion
The easiest way to install MyST is with [the `mamba` package manager](https://mamba.readthedocs.io/en/latest/index.html).
```

rowanc1 (approved):
🚀

PR #1461, opened 2024-08-13
🎓 Add DOIs for venue and biblio

fwkoch:
Journals and Issues may have DOIs, in addition to articles. This PR adds `doi` to `venue` and `biblio`, for journals and issues respectively. It also adds `short_title` to `venue`, as often journals have an abbreviation.

I also tweaked `fillProjectFrontmatter` so `biblio` fields fill (rather than replacing the entire `biblio` object). This allows you to define `volume`/`issue`/`doi` in a shared config and `first_page`/`last_page` on an article.

### DIFF
--- a/.changeset/curly-ads-repeat.md
+++ b/.changeset/curly-ads-repeat.md
@@ -1,0 +1,5 @@
+---
+'myst-templates': patch
+---
+
+Eliminate frontmatter validation warnings for template doc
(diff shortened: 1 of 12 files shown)


rowanc1 (approved):
Thanks @fwkoch! This looks good - tracking additional scholarly metadata about volumes/issues/journals.

PR #1470, opened 2024-08-20
:date: Fix date representation

### LINKED ISSUE #1357: date rendering is a little off

agoose77:
Our existing handling of dates parsing and rendering is vulnerable to some quirks (bugs) in different timezones. Internally we assume that we represent dates in UTC. Any parsing logic that breaks this assumption, such as non-ISO 8601 strings (which parse in the local timezone), leads to formatting errors down the road.

1. ISO 8601 date string (in UTC+1):
```node
> process.env.TZ = "Europe/London"
> new Date("2024-06-06")
2024-06-06T00:00:00.000Z
```
2. Non-ISO 8601 date string (in UTC+1):
```node
> process.env.TZ = "Europe/London"
> new Date("6 Jun 2024")
2024-06-05T23:00:00.000Z
```
The YAML spec supports ISO 8601 date strings, which are parsed (without using the local timezone) into `Date` objects. But strings of the form `6 Jun 2024` are handled by `new Date` directly, which causes trouble!

To further complicate things, it it not possible to identify the providence of a `Date` object, i.e.
```node
> new Date("2024-06-06T00:00:00+03:00")
2024-06-05T21:00:00.000Z
```
looks identical to 
```node
> new Date("2024-06-05T21:00:00.000Z")
2024-06-05T21:00:00.000Z
```

This PR proposes:
1. Represent dates in validated frontmatter as ISO 8601 date-strings, i.e. `YYYY-MM-DD`.
2. Never parse a date string in the _local_ timezone
3. Change from non UTC timezones to UTC using conversion

These changes ensure that we _never_ use the local timezone in date parsing, and we _always_ have valid (UTC) ISO 8601 date strings.

Fixes #1357


packages/myst-frontmatter/src/page/page.yml
```diff
@@ -40,7 +40,7 @@ cases:
         a: b
       subtitle: sub
       short_title: short
-      date: 14 Dec 2021
+      date: '14 Dec 2021'
```
agoose77, line 43: These quotes don't change how the contents are parsed, but it makes it clear we are ingesting strings not YAML timestamps.

agoose77:
This PR introduces a new problem; `moment.js` complains about non RFC 2822/ISO 8601 formatted strings. 

@rowanc1 all of this work leads me to think we should be stricter here and _not accept_ under-specified dates. 

My original approach here was to drop time information altogether, so that we can reasonably talk about "plain dates" (e.g. https://tc39.es/proposal-temporal/docs/plaindate.html). By adopting ISO 8601 _dates_, we can communicate this concept.

However, there's still ambiguity about what users are inputting. I _think_ the solution might be to choose a small set of valid dates, i.e. ISO 8601 and a `[ day-of-week "," ] date` scheme taken from RFC 2822 (DAY MONTH YEAR).

agoose77:
Following a meeting, we discussed next steps:
- Ask for only RFC 2322 or ISO 8601 
- Support out-of-schema formats by assuming local parsing and reverse the offset
- Warn if the user passes invalid date (or includes timestamp)

agoose77:
@rowanc1 one of the quirks of these changes is that we can't detect timestamps in YAML frontmatter, because the YAML loader implicitly casts these to `Date` objects. This is not required by the YAML spec; it's an application-specific thing, and Python's `PyYaml` library does not do this.

We can build our own schema to turn this off pretty easily: https://github.com/nodeca/js-yaml/blob/0d3ca7a27b03a6c974790a30a89e456007d62976/lib/schema/default.js#L13

What do you think?

packages/simple-validators/src/validators.ts
```diff
@@ -207,22 +206,122 @@ export function validateEnum<T>(
...
+}
+
 /**
  * Validate date string or object
  *
- * Uses javascript Date() constructor; any input to the constructor that returns
- * a valid date is valid input. This includes ISO 8601 formatted strings and
- * IETF timestamps are valid.
+ * Parses strings as ISO 8601 dates, or a variant of RFC 2822 dates, falling back to the Date constructor otherwise.
+ * Parses date objects as UTC or local dates according to the given options.
```
agoose77, line 232:
```suggestion
 * Parses Date objects as UTC or local dates according to the given options.
```

Capitalise `Date`.

packages/myst-frontmatter/src/project/validators.ts
```diff
@@ -34,7 +34,7 @@ export function validateProjectAndPageFrontmatterKeys(
 ) {
   const output: ProjectAndPageFrontmatter = validateSiteFrontmatterKeys(value, opts);
   if (defined(value.date)) {
-    output.date = validateDate(value.date, incrementOptions('date', opts));
+    output.date = validateDate(value.date, { ...incrementOptions('date', opts), dateIsUTC: true });
```
agoose77, line 37: This encodes the assumption that `Date` objects passed in to the validator should be treated as plain dates in UTC (i.e. take components in UTC).

.changeset/yellow-hotels-visit.md
```diff
@@ -0,0 +1,6 @@
+---
+'simple-validators': patch
+'myst-templates': patch
```
rowanc1, line 3: I think that these are out of date? Can you double check?

packages/simple-validators/src/validators.spec.ts
```diff
@@ -246,14 +246,14 @@ describe('validateDate', () => {
...
+    '2021/12/14',
+    '2021-12-14',
   ])('valid date: %p', async (date: any) => {
-    expect(validateDate(date, opts)).toEqual(date);
+    expect(validateDate(date, opts)).toEqual('2021-12-14');
   });
   it('date object is valid', () => {
-    const date = new Date();
-    expect(validateDate(date, opts)).toEqual(date.toISOString());
+    const date = new Date('2024-11-03');
```
rowanc1, line 255: Do we have a test for a time?

packages/simple-validators/src/validators.ts
```diff
@@ -207,22 +206,122 @@ export function validateEnum<T>(
...
+        validationWarning(
+          `invalid date "${input}" - specialised RFC 2822 dates must not include a time component ("${tail}"), which has been ignored`,
+          opts,
+        );
+      }
+
+      const numericYear = parseInt(year);
+
+      // Convert Jan to 1 etc.
+      const numericMonth = MONTH_TO_NUMBER.get(month)!;
```
rowanc1, line 276: Is there a reason we didn't just do `new Date(str)` if it didn't pass the initial checks?

I think that if people are doing their own thing, and not following the standard `YYYY-MM-DD` then we should just use the default javascript parser rather than writing our own?

Regardless, if we are adding our own simple date parsing, I would love to see some test coverage of these.

packages/simple-validators/src/validators.ts
```diff
@@ -207,22 +206,122 @@ export function validateEnum<T>(
...
+      localDate.getDate(),
+    );
+    validationWarning(
+      `invalid date "${input}" - must be ISO 8601 or (specialised) RFC 2822 format calendar date. Attempting to interpret date as "${result}"`,
+      opts,
+    );
+    return result;
+  }
+  // Handle null-like as "current time"
+  else if (input == null) {
```
rowanc1, line 309: Can we add some docs - and update the frontmatter section with the new recommendation? I think this should just be removing quite a bit?

packages/simple-validators/src/validators.ts
```diff
@@ -207,22 +206,122 @@ export function validateEnum<T>(
   return input;
 }
 
+// This pattern implements the date pattern from ISO8601, with a trailing capture group for timestamps
+const ISO8601_DATE_PATTERN = /^(\d\d\d\d)(?:-(\d\d))?(?:-(\d\d))?(T.*)?$/;
```
rowanc1, line 210: This looks like it could also parse just a YEAR or YYYY-MM? Is that true - what happens then?

`(new Date('2023')).getUTCFullYear()`

Does this behave correctly? 🧪!

packages/simple-validators/src/utils.ts
```diff
@@ -1,28 +0,0 @@
...
-  if (typeof object === 'string') {
-    return new Date(object);
-  }
-  if (object?.toDate !== undefined) {
-    return object.toDate();
-  }
-  throw new Error(`Could not parse date: ${object}`);
-}
-
-export function formatDate(date: Date | { toDate: () => Date }): string {
```
rowanc1, line 17: I can't find anywhere externally that we are using this, so should be fine to remove.

rowanc1 (changes requested):
If we are starting to implement more date parsing functionality (month names, etc) it would be great to see some test coverage for that.

Is there a reason we can't use native `new Date(str)` for that?

The docs should also get a short update with the new recommendation!

.changeset/yellow-hotels-visit.md
agoose77: > Is there a reason we can't use native `new Date(str)` for that?

Yes -- the RFC 2822 handling is not "standard" -- we're building a new rule from the ABNF grammar for RFC2822. That is, it's defined by the grammar in that RFC but we're not using their `date-time` rule. So, we need to roll our own.

packages/simple-validators/src/validators.ts
agoose77: `YYYY` is a valid ISO8601 date, but actually I don't think we want to support it now that I come back to this PR. There's a profile of ISO8601, RFC3339, that I think is better describing what we _do_ want: 
`YYYY-MM-DD`. I'll make this stricter.

packages/simple-validators/src/validators.ts
agoose77: Yes -- the RFC 2822 handling is not "standard" -- we're building a new rule from the ABNF grammar for RFC2822. That is, it's defined by the grammar in that RFC but we're not using their `date-time` rule. So, we need to roll our own.

We also want to identify rogue time-like components.

package-lock.json
```diff
@@ -15998,7 +15997,7 @@
     "packages/simple-validators": {
       "version": "1.0.6",
       "license": "MIT",
-      "devDependencies": {
+      "dependencies": {
```
rowanc1, line 16000: Can we change these back!

rowanc1:
Ok - I think just a few tests that show the various parts of the parsing logic (including some invalid dates), and this is good to go!

package-lock.json
agoose77: Ah, forgot to touch the lockfile.

agoose77:
@rowanc1 looks like we didn't handle `null` dates in the frontend (see https://github.com/jupyter-book/myst-theme/pull/451). This PR now passes-through null (like it used to) to signal "today", rather than encoding the time-of-build into frontmatter. Do you think that's wrong?

rowanc1:
I don't remember the null/today date being a feature previously, and left a comment on the `null`:

> What is the use case for this? Can we send through "today" instead? `null`-ish to me means _no_ date.
>
> I can (maybe) see that a preprint/blog gets a `today` string, that is recognized by the CLI and then serialized to the frontend as the date it was built on. That is pretty dangerous though - it means all your blog posts are always written "today" whenever you rerender them (which is probably not what people want).
> 
> I don't know if I understand the use case where we want the frontend to show today's date whenever you refresh .. ?

I would opt to remove that and constrain to valid dates.

agoose77:
Yes, I think the semantics of 'today' should be to encode the authorship date.

I propose that no `date` value in frontmatter means that we generate the current date and store it. If the user sets `date` to `null`, then the frontend doesn't display a date.

Do you think that sits well with how people are already using myst?

agoose77:
I reread your last - rebuilds changing metadata is a serious thing to avoid, so we should make it hard to do that.

As such, I think this PR is correct - and we just need to continue hiding null values in the frontend.

rowanc1:
I do **not** think we should auto-magic the date in.

I am coming at this from (1) documentation - the default is not to have a date in the title, if you want it you should put it in explicitly; (2) scientific articles - the date is meaningful and should be put in intentionally, having the date be auto-magic-ed in means that we don't have that information in the YAML, it is not committed to git, and difficult to ensure at a check level that it has actually been input.

rowanc1:
> rebuilds changing metadata is a serious thing to avoid

Yes - that is very succinct way to put it. :)

packages/simple-validators/src/validators.ts
```diff
@@ -207,22 +206,132 @@ export function validateEnum<T>(
...
+
 /**
  * Validate date string or object
  *
- * Uses javascript Date() constructor; any input to the constructor that returns
- * a valid date is valid input. This includes ISO 8601 formatted strings and
- * IETF timestamps are valid.
+ * Parses strings as ISO 8601 dates, or a variant of RFC 2822 dates, falling back to the Date constructor otherwise.
+ * Parses Date objects as UTC or local dates according to the given options.
+ * Parses `null` dates as "now".
```
rowanc1, line 238: I think we are coming to the conclusion that this should not be allowed (at least from the frontmatter?)

agoose77:
@rowanc1 this is now done imo.

rowanc1:
Minor updates to cover a few more cases (e.g. `2023-02-29` is not valid!)

PR #1473, opened 2024-08-20
🐌 Fix slug generation for `.myst.json` files

agoose77:
Presently we use `path.parse` to strip away file extensions before slug generation. This PR replaces that logic with a custom `parse` function that understands the MyST extension list.

I'm not 100% sure whether talking about `.myst.json` as an _extension_ rather than a "suffix" is a poor choice of language, but we're walking down this path for now.

### DIFF
--- a/packages/myst-cli/src/process/file.ts
+++ b/packages/myst-cli/src/process/file.ts
@@ -16,6 +16,7 @@ import { castSession } from '../session/cache.js';
 import { warnings, watch } from '../store/reducers.js';
 import type { PreRendererData, RendererData } from '../transforms/types.js';
 import { logMessagesFromVFile } from '../utils/logging.js';
+import { parseFilePath } from '../utils/resolveExtension.js';
 import { addWarningForFile } from '../utils/addWarningForFile.js';
 import { loadBibTeXCitationRenderers } from './citations.js';
 import { parseMyst } from './myst.js';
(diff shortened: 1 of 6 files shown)


rowanc1:
Can you add a test case on the function!

This is something that we are doing in parseFile as well right - and I think there was custom logic for the extensions there that we had a fall through case or something - not sure if it is related?

agoose77:
@rowanc1 ah, brilliant -- well remembered. I've updated that method.

packages/myst-cli/src/process/file.ts
```diff
@@ -258,16 +259,9 @@ export async function loadFile(
...
-      case '.json': {
-        if (file.endsWith('.myst.json')) {
-          loadResult = loadMySTJSON(session, content, file);
-          break;
-        }
-        // This MUST be the final case before `default`, as
-        // we rely on falling through to the `default` case if
-        // a non-MyST JSON file is encountered here
-        //
-        // falls through
```
rowanc1, line 270: I like when we get rid of comments like these. Nice work. :)

packages/myst-cli/src/utils/resolveExtension.spec.ts
```diff
@@ -76,3 +76,13 @@ describe('isValidFile', () => {
...
+
+describe('parseFilePath', () => {
+  it.each([
+    ['/tmp/foo/bar.md', { dir: '/tmp/foo', name: 'bar', ext: '.md' }],
+    ['/tmp/foo/bar/bat.txt', { dir: '/tmp/foo/bar', name: 'bat', ext: '.txt' }],
+    ['/tmp/baz.myst.json', { dir: '/tmp', name: 'baz', ext: '.myst.json' }],
+  ])('%s parses properly', (path, result) => {
+    expect(parseFilePath(path)).toEqual(result);
+  });
+});
```
rowanc1, line 88: Thank you!

PR #1477, opened 2024-08-21
🔢 Add enumeration options to common directives

rowanc1:
This adds `enumerated` / `enumerator` (aliased as `numbered` / `number`) to all common directives (code, figures, etc.). This allows for things like turning off enumeration for a specific figure / equation as well as taking control of numbering completely.

In a recent example, we had to create a figure with a specific number `S5` and then carry on numbering as usual.

This also introduces some places to centralize directive option logic - which we could extend to classes in the future.


packages/myst-transforms/src/enumerate.ts
```diff
@@ -385,6 +385,12 @@ export class ReferenceState implements IReferenceStateResolver {
    * If node is subcontainer/subequation, a sub-count is incremented
    */
   incrementCount(node: TargetNodes, kind: TargetKind | string): string {
+    if (node.enumerator) {
+      // If the enumerator is explicitly defined, return early
+      // This is the case if the figure, for example, has an enumerator set (e.g. `2a`)
+      // The other numbering will not be effected, and may be wrong
```
agoose77, line 391:
```suggestion
      // The other numbering will not be affected, and may be wrong
```

docs/cross-references.md
```diff
@@ -371,3 +371,5 @@ Finally, under the `numbering` object, you may specify `enumerator`. For now, th
 numbering:
   enumerator: A1.%s
 ```
+
+If you want to control the numbering for a specific figure, you can use the {myst:directive}`figure.enumerator` option. This will give the figure a specific enumerator and not increment the counting for other figures. This is helpful if you want to explicitly count figure `2a` and then carry on counting figures as normal; alternatively you can take control of numbering entirely.
```
agoose77, line 375:
```suggestion
If you want to control the numbering for a specific figure, you can use the {myst:directive}`figure.enumerator` option. This will give the figure a specific enumerator, and will not increment the counting for other figures. This is helpful if you want to explicitly count figure `2a` and then carry on counting figures as normal; alternatively you can take control of numbering entirely.
```

@rowanc1 can you elaborate on 
> alternatively you can take control of numbering entirely.

Is this saying 
> you can just set `enumerator` on every enumerable?

agoose77 (commented):
I like the refactoring of options for directives. We should extend this to `class` in the future.

docs/cross-references.md
rowanc1: Yes!

rowanc1:
Yes - I purposely left `class` out as there is some extra work to do on the UI side to actually make that possible. But I think the pattern is good (and easy) to follow!

packages/myst-directives/src/blockquote.ts
```diff
@@ -1,17 +1,16 @@
...
   alias: ['epigraph', 'pull-quote'],
   doc: 'Block quotes are used to indicate that the enclosed content forms an extended quotation. They may be followed by an inscription or attribution formed of a paragraph beginning with `--`, `---`, or an em-dash.',
   options: {
-    label: {
-      type: String,
-      alias: ['name'],
-    },
+    ...labelDirectiveOption('blockquote'),
+
+    // TODO: Add enumeration in future
```
fwkoch, line 13: Good that we are not adding `enumerator` on places where we do not number stuff yet.

packages/myst-directives/src/mermaid.ts
```diff
@@ -1,17 +1,21 @@
 import type { DirectiveSpec, DirectiveData, GenericNode } from 'myst-common';
+import { addCommonDirectiveOptions, commonDirectiveOptions } from './utils.js';
 
 export const mermaidDirective: DirectiveSpec = {
   name: 'mermaid',
+  options: {
+    ...commonDirectiveOptions('mermaid'),
```
fwkoch, line 7: Should `mermaid` have enumerator? Are these currently treated as figures...?

fwkoch (approved):
This looks good - relies on the fact that we already respect `enumerated` flag. I like the shared option refactor.

packages/myst-directives/src/mermaid.ts
rowanc1: Changed to just label on mermaids!

packages/myst-directives/src/blockquote.ts
rowanc1: Looking forward to being able to number everything ... !

PR #1478, opened 2024-08-21
🆎 Explicitly ignore longer abbreviations

rowanc1:
This ignores longer abbreviations:

```yaml
HR: Heart Rate
RHR: null
```

This would make: HR an abbreviation, but leave RHR alone. Previously there was no way to let the `HR` in `RHR` not be abbreviated with `R<HR>`.

### DIFF
--- a/packages/myst-transforms/src/abbreviations.ts
+++ b/packages/myst-transforms/src/abbreviations.ts
@@ -31,8 +31,7 @@ function replaceText(mdast: GenericParent, opts: Options) {
         abbr,
         (value: any, { stack }: RegExpMatchObject) => {
           if (!title) {
-            // Explicitly do not match this abbreviation
-            // This does need to mark and split the text, which is undone below
+            // Change the type of this match to a transient node to guard from further replacements
             return u('__skippedAbbreviation__', value);
           }
           if (stack.slice(-1)[0].type !== 'text') return false;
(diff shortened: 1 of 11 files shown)


packages/myst-transforms/src/abbreviations.ts
```diff
@@ -30,14 +30,22 @@ function replaceText(mdast: GenericParent, opts: Options) {
       .map(([abbr, title]) => [
         abbr,
         (value: any, { stack }: RegExpMatchObject) => {
+          if (!title) {
+            // Explicitly do not match this abbreviation
+            // This does need to mark and split the text, which is undone below
```
agoose77, line 35:
```suggestion
            // Replace this match with a transient node to guard from further replacements
```

packages/myst-transforms/src/abbreviations.ts
```diff
@@ -30,14 +30,22 @@ function replaceText(mdast: GenericParent, opts: Options) {
...
           if (parent) return false;
-          return u('abbreviation', { title }, [u('text', value)]) as any;
+          return u('abbreviation', { title }, [u('text', value)]);
         },
       ]),
   );
   findAndReplace(mdast as any, replacements);
+  selectAll('__skippedAbbreviation__', mdast).forEach((n) => {
+    n.type = 'text';
+  });
```
agoose77, line 48:
```suggestion
  // Restore the original `text` type of the transient replacements performed above
  selectAll('__skippedAbbreviation__', mdast).forEach((n) => {
    n.type = 'text';
  });
```

agoose77 (dismissed):
Approve but suggest these comments!

PR #1486, opened 2024-08-22
🗃 Start admonitions in the open position

### LINKED ISSUE #1319: allow dropdown admonitions to start open
currently an admonition declared with `:class: dropdown` always starts closed
there are cases where it makes sense to have both the ability to open/close *and* to start open

## Proposal

add an `:open: ` flag to the `admonition` directive; its default would then be `false` if the `dropdown` class is mentioned, and `false` otherwise

rowanc1:
Fixes #1319

Requires: https://github.com/jupyter-book/myst-theme/pull/452

### DIFF
--- a/docs/admonitions.md
+++ b/docs/admonitions.md
@@ -167,7 +167,7 @@ This is the body.
 To turn an admonition into a dropdown, add the `dropdown` {myst:directive}`admonition.class` to them.
 Dropdown admonitions use the `<details>` HTML element (meaning they also will work without Javascript!),
 and they can be helpful when including text that shouldn't immediately visible to your readers.
-To have the admonition be a dropdown, and also start open, add the {myst:directive}`admonition.open` option.
+To have a dropdown-style admonition start open, add the {myst:directive}`admonition.open` option.
 
 ```{myst}
 :::{note} Click Me! 👈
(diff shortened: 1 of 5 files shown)


docs/admonitions.md
```diff
@@ -167,6 +167,7 @@ This is the body.
 To turn an admonition into a dropdown, add the `dropdown` {myst:directive}`admonition.class` to them.
 Dropdown admonitions use the `<details>` HTML element (meaning they also will work without Javascript!),
 and they can be helpful when including text that shouldn't immediately visible to your readers.
+To have the admonition be a dropdown, and also start open, add the {myst:directive}`admonition.open` option.
```
agoose77, line 170:
```suggestion
To have a dropdown-style admonition start open, add the {myst:directive}`admonition.open` option.
```

agoose77: The `open` option doesn't add the `class` for dropdown, right?

agoose77: Perhaps we should add the `dropdown` class if `open` is defined?

agoose77: @rowanc1 should we warn the user if they don't define `class: dropdown` then?

rowanc1: Yeah - I will add that. Right now it just doesn't do anything silently.

rowanc1:
- [ ] Add error if there is a clash between dropdown and open options

fwkoch (commented):
This is good - not changing the default but adding a simple `open` option.

Rather than introducing validation errors if `dropdown` and `open` disagree, could we just have `open: true/false` imply `class: dropdown`?

rowanc1:
We took @fwkoch's suggestion of coercing and working either way!

Issue #1443, opened 2024-08-09
Bundle NodeJS with `mystmd` so users don't have to install it on their own


choldgraf:
Our documentation asks users to install NodeJS and NPM on their own in addition to installing MyST. Most users aren't familiar with these workflows, and installing nodejs from scratch is an extra barrier to trying out MyST. We get a bit closer by telling users to install `conda`, but this has some weird stuff going on with Anaconda supposedly suing organizations that aren't paying them for using their default packaging channels.

Could we bundle the right NodeJS version with the `mystmd` package? This would reduce some install complexity for users if they didn't have to think about NodeJS at all.

### Implementation ideas

I know that [the `sphinx-theme-builder` package](https://github.com/pradyunsg/sphinx-theme-builder) does this so perhaps we can learn from that.

It uses [the `nodeenv` package](https://github.com/ekalinin/nodeenv) to download Node into an isolated environment in the package. [Here's where a lot of the nodejs logic is inside Furo](https://github.com/pradyunsg/sphinx-theme-builder/blob/45d3f2f53f190864f91f8aa98f52ad8bab178f4e/src/sphinx_theme_builder/_internal/nodejs.py#L76)

agoose77:
@choldgraf yes, and this is on my roadmap for 1.0. I need to revisit it because IIRC I don't think nodeenv uses per-platform wheels, so I want to check how robust it is

rowanc1:
This is only for the pypi/conda package right? Could easily see something that checks for node and a version and installs it if it isn't there?

It is nice that the npm version installs in less than a second on CI.

choldgraf:
@rowanc1 yes this is just for the `pip` install of MyST. If I recall, `nodeenv` creates a `.nodeenv` directory and saves a node environment there, only for use by the package.

We could ask @pradyunsg for some pointers if he'd be willing to give them, I think he's wrestled with nodeenv to get this working for the sphinx-theme-builder

pradyunsg:
> if he'd be willing to give them

Always happy to share knowledge. ^.^

> Could easily see something that checks for node and a version and installs it if it isn't there?

[...]

PR #1487, opened 2024-08-22
:package: Use `nodeenv` when Node.js can't be found by Python package

agoose77:
Fixes https://github.com/jupyter-book/mystmd/issues/1443 by using `nodeenv` to provision a Node.js environment for installs of our Python package.

Here's an example running in Podman (the final crash appears container related)
https://github.com/user-attachments/assets/9ca90520-f1fa-4d37-a6da-a64edc5f0a83


packages/mystmd-py/.gitignore
```diff
@@ -3,3 +3,5 @@ dist
 src/mystmd_py/myst.cjs
 **/__pycache__/
 _package.json
+
+/_build/
```
rowanc1, line 7: Probably you running myst in this folder?!

packages/mystmd-py/src/mystmd_py/main.py
```diff
@@ -4,41 +4,87 @@
...
+
+def find_any_node(binary_path):
+    node_path = find_installed_node()
+    if node_path is not None:
+        return pathlib.Path(node_path).absolute(), binary_path
+
+    nodeenv_path = find_nodeenv_path()
+    if not nodeenv_path.exists():
+        print(f"🔍 Couldn't find installed `node`.\n\n⚙️ Installing Node.js in {nodeenv_path}")
+        create_nodeenv(nodeenv_path)
```
rowanc1, line 45: We should give the user control here - "Would you like to install node now using nodeenv?" y/n

packages/mystmd-py/src/mystmd_py/main.py
```diff
@@ -4,41 +4,87 @@
...
     if not (major_version in {18, 20, 22} or major_version > 22):
         raise SystemExit(
             f"MyST requires node 18, 20, or 22+; you are running node {major_version}.\n\n"
             "Please update to the latest LTS release, using your preferred package manager\n"
             "or following instructions here: https://nodejs.org/en/download"
         )
+
+    # Start MyST CLI
+    PATH_TO_BIN_JS = (pathlib.Path(__file__).parent / "myst.cjs").resolve()
     os.execve(
```
rowanc1, line 84: Somewhat unrelated: There is a bug on windows (maybe others) where this is execution does not get keyboard commands (so the y/n input stuff doesn't work at all). Do you have any idea how to fix that @agoose77?

packages/mystmd-py/.gitignore
agoose77: Lol the one time I just use git add update!

packages/mystmd-py/src/mystmd_py/main.py
```diff
@@ -4,41 +4,99 @@
...
+    return platformdirs.user_data_path(
+        appname="myst", appauthor=False, version=NODEENV_VERSION
     )
-    PATH_TO_BIN_JS = (pathlib.Path(__file__).parent / "myst.cjs").resolve()
 
-    if not NODE_LOCATION:
+
+def ask_to_install_node(path):
+    if env_value := os.environ.get(INSTALL_NODEENV_KEY, "").lower():
+        return env_value in {"yes", "true", "1"}
```
agoose77, line 26:
```suggestion
        return env_value in {"yes", "true", "1", "y"}
```

agoose77:
@rowanc1 I've added an `input` call (that can be bypassed by defining the `MYSTMD_ALLOW_NODENV` env var, where the result depends upon the truthiness (`{"yes", "true", "1", "y"}`) of the lowercased value.

packages/mystmd-py/src/mystmd_py/main.py
agoose77: I can't immediately see why the exec isn't working properly. will need to play around on Windows later.

rowanc1:
Can you rebase this on the docs PR that just got merged. My understanding is that this will simplify the pip install instructions?

packages/mystmd-py/.gitignore
rowanc1, line 7:
```suggestion
```

packages/mystmd-py/pyproject.toml
```diff
@@ -20,6 +20,10 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Topic :: Scientific/Engineering",
 ]
+dependencies = [
+  "platformdirs~=4.2.2",
+  "nodeenv~=1.9.1"
```
rowanc1, line 25: How does this play with conda? Do we have other node deps over there that do the install for us?

packages/mystmd-py/.gitignore
agoose77: Lol yep, still todo.

packages/mystmd-py/pyproject.toml
agoose77: Oh right, that's a good point. I didn't think about how we would repackage that. Let me have a think.

packages/mystmd-py/pyproject.toml
```diff
@@ -34,6 +34,14 @@ myst = "mystmd_py.main:main"
...
 path = "_package.json"		
 
+# Ensure that we only bring in nodeenv for PyPI
+[tool.hatch.build.hooks.custom]
+path = "build_plugin.py"
+pypi_dependencies = [
+  "platformdirs~=4.2.2",
+  "nodeenv~=1.9.1"
+]
+
```
agoose77, line 44: @rowanc1 this is a build plugin for hatch that injects the PyPI dependencies at build time. The conda-forge build pulls in the sdist (which is mainly just a tar of the package contents), so it does not have this injected content. When the conda-forge build happens, we'll need to update the environment to set MYST_IS_CONDA_PACKAGE, which will disable the injection.

agoose77:
I have lifted the custom build hook into https://github.com/jupyter-book/hatch-deps-selector/, which means that others can also use it.

Simply put, this plugin makes it possible for us to add additional predefined dependencies to our package metadata at build-time by setting an environment variable.

agoose77:
This PR now should hold until https://github.com/conda-forge/staged-recipes/pull/27417 is merged, so that we can immediately update our feedstock recipe.

agoose77:
Adding a pre-built wheel to test on Windows...
[mystmd-1.3.6-py3-none-any.whl.zip](https://github.com/user-attachments/files/16847273/mystmd-1.3.6-py3-none-any.whl.zip)

agoose77:
@rowanc1 this has been tested on Windows and Linux, and works as expected. We can review during today's meeting if helpful.

PR #1490, opened 2024-08-23
🤖 Make `name` & `display_name` required properties for `KernelSpec`

agoose77:
The `nbformat` spec requires that `name` and `display_name` are set in a kernelspec object. Presently, we don't require either of these fields.

This PR makes `name` and `display_name` "soft" required; default values will be chosen if they're missing, but warnings will also be issued. 

In future, we can eventually just make them required.

**This PR also changes our execution hash key strategy**. Now, it includes the kernelspec name! It's highly unlikely to be a common bug, but users could e.g. switch from one Python kernel to another, and the cache would remain unchanged. This would definitely be a problem if different kernels had different environments.

**As a result of this change, existing caches will be invalidated**. But, I *think* that's OK given how young execution is (I don't think we advertise it as stable yet).


packages/myst-frontmatter/src/kernelspec/validators.ts
```diff
@@ -6,29 +6,52 @@ import {
...
+
+  // name is a required member of KernelSpec, but hasn't previously been required
+  let name: string;
   if (defined(value.name)) {
-    output.name = validateString(value.name, incrementOptions('name', opts));
-  }
-  if (defined(value.language)) {
-    output.language = validateString(value.language, incrementOptions('language', opts));
+    const validatedName = validateString(value.name, incrementOptions('name', opts));
+    if (validatedName === undefined) return undefined;
```
rowanc1, line 26: Is this correct? If it doesn't validate we bail and get no kernel?

rowanc1 (commented):
Looks good, haven't tested. Let me know if you want me to do that, but the changes look straight forward.

packages/myst-execute/src/execute.ts
```diff
@@ -281,6 +284,15 @@ export type Options = {
  */
 export async function kernelExecutionTransform(tree: GenericParent, vfile: VFile, opts: Options) {
   const log = opts.log ?? console;
+
+  // We need the kernelspec to proceed
+  if (opts.frontmatter.kernelspec === undefined) {
```
rowanc1, line 289: This should never be triggered in the current CLI workflow, right? Because we always have a valid object coming out of frontmatter validation?

packages/myst-frontmatter/src/kernelspec/validators.ts
rowanc1: I think that makes sense.

packages/myst-execute/src/execute.ts
agoose77: If you don't define frontmatter.kernelspec, you don't get a kernelspec object: https://github.com/jupyter-book/mystmd/blob/4a5d3b789cd9c85f253593367eda2bf681b10c53/packages/myst-frontmatter/src/page/types.ts#L18

packages/myst-frontmatter/src/kernelspec/validators.ts
agoose77: Yes, if the user gives a `name`, it must be valid. Otherwise, we handle the case that it's omitted for legacy reasons (up until now, it wasn't required for execution).

agoose77:
@rowanc1 are you happy to merge this in now?

PR #1508, opened 2024-09-02
🌍 Add `LanguageInfo` to frontmatter

agoose77:
Recently, I authored #1490 to ensure that we don't break `jupytext` by having a frontmatter schema that is too loose. This PR is slightly different.

I've been noticing the "undefined language" warning during testing of execution. It turns out that [we look for a `language` key in `kernelspec`](https://github.com/jupyter-book/mystmd/blob/a6b59d6b88ecdce5b3231d18bebf336ebd871a74/packages/myst-cli/src/process/mdast.ts#L235) as a default for this field. When hand-authoring notebooks, I do not define such a field. I looked at the nbformat spec, and they don't define this value; it probably comes from JupyterLab. So, I think we should look to read an additional _well-specified_ field, and probably better to use one that _is_ defined in the schema rather than push harder on `kernelspec.language`.[^why]

> [!NOTE]
> My short reasoning is:
> 1. `language` of code-cells is related to the kernel
> 2. `nbformat` defines a proper place for this to live
> 3. Let's follow `nbformat` for ease & simpler down-stream conversion of md:myst documents to ipynb etc.

Thus, this PR adds support for validating `language_info` from nbformat:
https://github.com/jupyter/nbformat/blob/ba2c6f53afbeb18f6fad0c9b9f67def34e757cb4/nbformat/v4/nbformat.v4.5.schema.json#L33

> [!TIP]
> `language_info` also captures the language _version_ e.g. for Python: `3.12.0`
> For now, I didn't validate that, but it might be useful when reading `ipynb` files.

More broadly than this PR, we do need to be mindful of _why_ we are adopting external frontmatter schemas, with consideration of _who_ is ingesting our frontmatter. In my view, there are two cases:
1. Where we are the only consumers of MyST frontmatter. There, we should (in my view) opt for permissive-but-validatable definitions, and adopting external schemas is merely a good idea from the perspective of building on the shoulders of giants, or using external tools (like `jupyter server`). 
2. Where we are _not_ the only clients, e.g. `jupytext`. There, we need to ensure that they can read our metadata. 

RE (2): Jupytext _right now_ doesn't apply all the transforms to frontmatter that we do. Thus, we can end up in situations where `jupytext` either fails to validate (or actually, produces _invalid_ notebooks) e.g. the `authors` field set to a list of strings.

Finally, it's worth thinking about the distinction between hand-authored metadata, and programmatically generated metadata. `nbformat` does not distinguish between the two, but it's highly likely that `authors` is hand-authored whilst `language_info` is not (as it comes from the kernel).

[^why]: whilst `kernelspec.language` is fine, others could write something there that we don't want, precisely because it is not specified.

### DIFF
--- a/.changeset/fair-colts-design.md
+++ b/.changeset/fair-colts-design.md
@@ -1,6 +1,6 @@
 ---
-"myst-frontmatter": patch
-"myst-cli": patch
+'myst-frontmatter': patch
+'myst-cli': patch
 ---
 
 Add support for `language_info` frontmatter
--- a/.changeset/fair-colts-design.md
+++ b/.changeset/fair-colts-design.md
@@ -1,0 +1,6 @@
+---
+"myst-frontmatter": patch
+"myst-cli": patch
+---
+
+Add support for `language_info` frontmatter
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -232,7 +232,9 @@ export async function transformMdast(
   transformFilterOutputStreams(mdast, vfile, frontmatter.settings);
   transformCitations(session, file, mdast, fileCitationRenderer, references);
   await unified()
-    .use(codePlugin, { lang: frontmatter?.kernelspec?.language })
+    .use(codePlugin, {
+      lang: frontmatter?.language_info?.name ?? frontmatter?.kernelspec?.language,
+    })
     .use(footnotesPlugin) // Needs to happen near the end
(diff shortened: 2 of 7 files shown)


rowanc1:
Can you add a test that captures the logic?

rowanc1:
After the call today - this seems like an upstream change?! @agoose77 please confirm. :)

agoose77:
After today's call, we decided to find a more user-friendly syntax here, and require that jupytext validate / normalise our frontmatter.

PR #1509, opened 2024-09-02
🤖 Add `flake.nix` for Nix users

agoose77:
This is a dev PR that adds `flake.nix` for Nix users. This makes it as easy to start running/developing MyST as `nix develop`. 

> [!NOTE]
> Developers will need to modify their NodeJS prefix in order for `myst` to be found on `PATH`

I will take responsibility for keeping this up to date, and it's only intended as a convenience for developers.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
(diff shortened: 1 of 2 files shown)


rowanc1:
Sounds good. Happy for you to bring this in when you are ready.

Issue #1425, opened 2024-08-07
Keyboard input hangs on windows


rowanc1:
There is a full discussion of the bug here:
https://discord.com/channels/1083088970059096114/1265372315282051134

When wrapped in the python client, the input does not make it through to the node process. That means that people are hitting `yes` etc. or typing a name of a file, and that input does not get connected properly to the underlying kernel, and they cannot complete actions such as `myst init` and `myst init --gh-pages`.

This is only a problem when using myst through the `mystmd-py` python wrapper.

Duplicates #1176

PR #1510, opened 2024-09-02
👷 Check `platform.system` to fix `myst` on Windows

agoose77:
We use `os.exec` to make the MyST binary an invisible wrapper after startup. However, on Windows this doesn't work as intended. So, this PR sets up a subprocess on Windows instead.

Fixes https://github.com/jupyter-book/mystmd/issues/1425


agoose77:
Yep! It works :)

PR #1516, opened 2024-09-06
📖 Update to installation docs

rowanc1:
This puts the actual install code front and center, and de-emphasizes the knowledge of `nodejs` now that this is installed for you on both pip and conda-forge.

Also removed a section on separating the site config which is no longer relevant.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -5,33 +5,16 @@ short_title: Install MyST
 description: MyST Markdown is available through Node and npm, install the package with `npm install mystmd`.
 ---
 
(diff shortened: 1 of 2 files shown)


agoose77:
@rowanc1 did you remove the emoji in the PR that merged in nodeenv?

rowanc1:
I did fix the alignment in code (two spaces for some emojis) - can revert to just fixing that. Although the font isn't showing a gear on my end, meaning the font isn't doing a good job there.

Happy to revert that change.

PR #1517, opened 2024-09-06
🐞 Directive option flag is always a boolean

### LINKED ISSUE #1504: Directive option flag is always a boolean
When the option is a string, for example, it can be interpreted and passed to renderers as a boolean:

```
options: {
  path: { type: String } 
}
```

when the directive has the option:

```
:path:
```

That will be passed as `true` to the rendering code.

It should be passed as an empty string in this case I think: `""`.

rowanc1:
Fixes #1504

### DIFF
--- a/packages/markdown-it-myst/src/directives.ts
+++ b/packages/markdown-it-myst/src/directives.ts
@@ -64,7 +64,7 @@ function runDirectives(state: StateCore): boolean {
         directiveOpen.map = map;
         directiveOpen.meta = {
           arg,
-          options: simplifyDirectiveOptions(options),
+          options: getDirectiveOptions(options),
           // Tightness is computed for all directives (are they separated by a newline before/after)
           tight: computeBlockTightness(state.src, token.map),
         };
@@ -109,7 +109,7 @@ function parseDirectiveContent(
 ): {
   body: string[];
   bodyOffset: number;
-  options?: [string, string][];
+  options?: [string, string | true][];
 } {
   let bodyOffset = 1;
   let yamlBlock: string[] | null = null;
@@ -145,7 +145,7 @@ function parseDirectiveContent(
       );
     }
   } else if (content.length && COLON_OPTION_REGEX.exec(content[0])) {
-    const options: [string, string][] = [];
+    const options: [string, string | true][] = [];
     let foundDivider = false;
     for (const line of content) {
       if (!foundDivider && !COLON_OPTION_REGEX.exec(line)) {
@@ -158,7 +158,7 @@ function parseDirectiveContent(
       } else {
         const match = COLON_OPTION_REGEX.exec(line);
         const { option, value } = match?.groups ?? {};
-        if (option) options.push([option, value || 'true']);
+        if (option) options.push([option, value || true]);
         bodyOffset++;
       }
     }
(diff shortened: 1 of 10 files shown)


packages/markdown-it-myst/src/directives.ts
```diff
@@ -171,44 +171,44 @@ function directiveArgToTokens(arg: string, lineNumber: number, state: StateCore)
...
     if (simplified[key] !== undefined) {
       return;
-    } else if (!isNaN(Number(val))) {
-      simplified[key] = Number(val);
-    } else if (typeof val === 'string' && val.toLowerCase() === 'true') {
-      simplified[key] = true;
-    } else if (typeof val === 'string' && val.toLowerCase() === 'false') {
-      simplified[key] = false;
-    } else {
-      simplified[key] = val;
```
rowanc1, line 187: I am moving this logic to `myst-parser` and this is just passing either a `true` or the actual content.

packages/markdown-it-myst/src/directives.ts
```diff
@@ -158,7 +158,7 @@ function parseDirectiveContent(
       } else {
         const match = COLON_OPTION_REGEX.exec(line);
         const { option, value } = match?.groups ?? {};
-        if (option) options.push([option, value || 'true']);
+        if (option) options.push([option, value || true]);
```
rowanc1, line 161: This is the major change. Most other things are chasing types.

packages/markdown-it-myst/src/directives.ts
```diff
@@ -171,44 +171,44 @@ function directiveArgToTokens(arg: string, lineNumber: number, state: StateCore)
...
+          )
+        : [
+            new state.Token('directive_option_open', '', 1),
+            new state.Token('directive_option_close', '', -1),
+          ];
     if (optTokens.length) {
       optTokens[0].info = key;
-      optTokens[0].content = value;
+      optTokens[0].content = typeof value === 'string' ? value : '';
+      optTokens[0].meta = { value };
```
rowanc1, line 211: I am now putting this in `meta`, which doesn't change the value - content will always change to a string.

packages/myst-parser/tests/myst.spec.ts
```diff
@@ -91,6 +91,21 @@ function fixMystDirectives(tree: GenericParent) {
...
+      } else if (!isNaN(Number(val))) {
+        options[key] = Number(val);
+      } else if (typeof val === 'string' && val.toLowerCase() === 'true') {
+        options[key] = true;
+      } else if (typeof val === 'string' && val.toLowerCase() === 'false') {
+        options[key] = false;
+      } else {
+        options[key] = val;
+      }
+    });
```
rowanc1, line 108: This should eventually be a spec change. In that the directive options are not coerced in the parser - they are just strings or a `flag` at that point. They becoming meaningful when the directive logic is applied.

rowanc1 (commented):
This moves the logic from `markdown-it-myst` to `myst-parser` for the directive options. I think this is a good idea, because that is when we know what the options actually are - e.g. a number, string, boolean, inline, etc.

This should change to a spec change in the future.

packages/markdown-it-myst/src/directives.ts
fwkoch: Ah, so this stuff isn't really actually needed at all - it is redundant to the coercion in `myst-parser` (which you've fixed up for the `true/'true'` change).

packages/myst-parser/tests/myst.spec.ts
fwkoch: Just to make sure I'm clear - you only added this here to avoid updating the test data, right? The "spec change" is only at the `mystDirective` level, which feels pretty inconsequential?

packages/markdown-it-myst/src/directives.ts
fwkoch: That makes sense - and we never use `content` either now. We just want `meta.value`

fwkoch (approved):
This looks great and makes sense. No longer coercing values twice and I think doing a better job of keeping track of the user's intent for `true`/`''`/etc.

My only question is if we should just update the test data rather than adding more postprocessing stuff...?

packages/myst-parser/tests/myst.spec.ts
rowanc1: Yes - I do think we should followup with a spec PR.

rowanc1:
I agree that we should open a PR to fix the spec. I will do that today. I would like this one to move ahead though! 🚀

PR #1518, opened 2024-09-06
🧟‍♀️ Revive link-block directive to create linkBlock node

fwkoch:
The `link-block` directive used to exist, see this commit: https://github.com/jupyter-book/mystmd/commit/920c087ccf092bbf57c65833e139e34ee913e63d but somewhere in the monorepo refactor I believe it was lost.

There is still lingering support for `linkBlock` nodes in `mystmd`, e.g. https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/process/mdast.ts#L83, and in `myst-theme`, https://github.com/jupyter-book/myst-theme/blob/main/packages/myst-to-react/src/links/index.tsx#L108

This PR revives the `link-block` directive to create `linkBlock` nodes. The `url` is required as directive arg, and `title` option and description in the body are optional.

image

### DIFF
--- a/.changeset/nasty-cougars-smile.md
+++ b/.changeset/nasty-cougars-smile.md
@@ -1,0 +1,6 @@
+---
+'myst-directives': patch
+'myst-transforms': patch
+---
+
+Revive link-block directive to create linkBlock node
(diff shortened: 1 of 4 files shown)


fwkoch:
After chatting a bit with @rowanc1  - `linkBlock` was intentionally removed in favor of better `cards`. The remaining references to `linkBlock` in the code are deprecated and should not be re-expanded!

PR #1524, opened 2024-09-10
👩‍💻 Allow index entries on many more node types for myst-to-tex

fwkoch:
Previously, index entries only showed up in `pdf` exports if they were defined above headings or figures. For example, this would create an index entry:

````
```{index} my topic
```
# My Topic

Let me talk about my topic.
````

But this would _not_ create an index entry:

````
# My Topic

```{index} my topic
```
Let me talk about my topic.
````

This PR adds PDF index entry creation on many more node types, meaning users can insert their index entries exactly where they want without worrying they will be lost.

### DIFF
--- a/.changeset/unlucky-timers-clap.md
+++ b/.changeset/unlucky-timers-clap.md
@@ -1,0 +1,5 @@
+---
+'myst-to-tex': patch
+---
+
+Allow index entries on many more nodes for myst-to-tex
--- a/packages/myst-to-tex/src/index.ts
+++ b/packages/myst-to-tex/src/index.ts
@@ -118,6 +118,7 @@ const handlers: Record<string, Handler> = {
     state.text(node.value);
   },
   paragraph(node, state) {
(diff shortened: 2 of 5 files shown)


dellaert:
Will this be in mystmd as of 1.3.9 then? Or is it already in 1.3.8?

rowanc1:
1.3.9, not yet released. Hopefully by tomorrow!

dellaert:
> 1.3.9, not yet released. Hopefully by tomorrow!

Hey y'all, any chance of rolling this into an update? I wrote a script to create index entries into our notebooks automatically, but it'd only work well with this change :-)

rowanc1:
Thanks for the nudge. @fwkoch can you release in an hour?

fwkoch:
@dellaert This is released - sorry for the delay!

dellaert:
Awesome, it works !!!! Thanks !

PR #1530, opened 2024-09-12
:wof Add search

agoose77:


packages/myst-cli/src/process/search.ts
```diff
@@ -0,0 +1,152 @@
+import type { GenericNode } from 'myst-common';
+import type { DocumentHierarchy } from 'myst-spec-ext';
+import type { Heading } from 'myst-spec';
+import { toText } from 'myst-common';
+export const SKIP = () => {};
+
+type Visit = (content: GenericNode, path: string) => typeof SKIP | void;
+type Depart = (content: GenericNode, path: string) => void;
+
+function visitNode(content: GenericNode, visit: Visit, depart: Depart | undefined, path: string) {
```
agoose77, line 10: This stuff is all hand-rolled visitor logic. Maybe we can refactor it now to use the unist utilities. At one point it did more work.

packages/myst-cli/src/process/search.ts
```diff
@@ -0,0 +1,152 @@
...
+  depth: number;
+  html_id?: string;
+};
+
+export type Section = {
+  heading?: HeadingInfo;
+  parts: string[];
+};
+
+export function toSectionedParts(content: GenericNode) {
```
agoose77, line 53: This naively sections a document upon encountering a heading. Seems to work well so far.

packages/myst-cli/src/process/search.ts
rowanc1: Yeah - this looks identical to the visit code. I think it is worth factoring it out and using the existing tools.

packages/myst-cli/src/process/site.ts
```diff
@@ -149,6 +152,56 @@ export async function writeMystXRefJson(session: ISession, states: ReferenceStat
   writeFileToFolder(filename, JSON.stringify(mystXRefs));
 }
 
+export async function writeMystSearchJson(session: ISession, pages: LocalProjectPage[]) {
```
rowanc1, line 155: Can we have a high level test on this?

packages/myst-cli/src/process/site.ts
```diff
@@ -557,10 +610,12 @@ export async function processSite(session: ISession, opts?: ProcessSiteOptions):
   if (opts?.writeFiles ?? true) {
     await writeSiteManifest(session, opts);
     const states: ReferenceState[] = [];
+    const allPages: LocalProjectPage[] = [];
     await Promise.all(
       siteConfig.projects.map(async (project) => {
         if (!project.path) return;
         const { pages } = await loadProject(session, project.path);
+        allPages.push(...pages);
```
rowanc1, line 618: This will be a non-deterministic order. Which we may want to avoid.

rowanc1 (commented):
A few comments! I think at least a few tests on the integration or the toText stuff would be good.

Might be a weird out-of-order bug in the async list generator. I think we might just want to avoid those, they are the worst to track down later...!

packages/myst-cli/src/process/site.ts
agoose77: I assume the ordering of `pages` is not guaranteed? OK, I'll impose an arbitrary sorting criterion.

rowanc1: I meant something more like this:

```
const allPages: LocalProjectPage[] = (await Promise.all(
  ...
  ...
  return pages
)).flatten()
```

More the pattern that I am against, as I don't think it will make a difference now that I am looking closer.

packages/myst-cli/src/build/site/start.ts
```diff
@@ -51,6 +51,7 @@ export async function startContentServer(session: ISession, opts?: ServerOptions
   app.use('/config.json', express.static(join(session.sitePath(), 'config.json')));
   app.use('/objects.inv', express.static(join(session.sitePath(), 'objects.inv')));
   app.use('/myst.xref.json', express.static(join(session.sitePath(), 'myst.xref.json')));
+  app.use('/myst.search.json', express.static(join(session.sitePath(), 'myst.search.json')));
```
rowanc1, line 54: We will also need to export this for the static sites.

agoose77:
Just working on improving the AST to text transform.

rowanc1:
We should support 1 apparently. Ideally the CLI should not crash if search index cannot be made!

image

Issue #1450, opened 2024-08-12
Per-page customization of website theme options


choldgraf:
Currently, theme options are set at the **project level**, meaning that you define it once in `myst.yml` and it applies to the entire site.

However, there are some cases where users want to apply options at a **page level**. It would be useful if we had the same `project -> page` inheritance for site options.

### Example: Splash pages vs. documentation

Consider a site that has three top-level sections they wish to put in the navbar, one of which is `docs/`. Their Table of Contents defines the structure of `docs/`, and they **only want it to show up when users are in `docs/*`**.

To accomplish this, we could use **page-level theme options** to disable the left and right sidebar only for the non-docs pages.

Issue #1308, opened 2024-06-13
Flatten `myst.yml`


rowanc1:
Right now we have a few high level keys in the `myst.yml` including project/site/version/extends. It is confusing to users where to put the keys (mostly always under project). Additionally, when writing a shared configuration, for example, abbreviations, authors, etc. that requires you to write it under the project key.

Instead, I suggest that we lean into the single project, and flatten all keys that are in the project. Moving from:

```yaml
version: 1
project:
  open_access: true
  license:
    content: Apache-2.0
    code: CC-BY-4.0
  references:
    foundations: https://projectpythia-mystmd.github.io/pythia-foundations
site:
  template: book-theme
```

to

```yaml
version: 1
open_access: true
license:
  content: Apache-2.0
  code: CC-BY-4.0
references:
  foundations: https://projectpythia-mystmd.github.io/pythia-foundations
site:
  template: book-theme
```

There are still potentially some site options to configure, but they will be obviously related to the theme (nav links, footer, etc.).

@jmunroe

fwkoch:
Some additional thoughts around simplifying/flattening the `myst.yml`:

- Many of the `project` fields are also allowed on `site`, but it's ambiguous why they would be defined in one place or the other. Almost exclusively, they are only needed on `project`. I think we should move all "SiteFrontmatter" (https://github.com/executablebooks/mystmd/blob/main/packages/myst-frontmatter/src/site/types.ts#L58) to "ProjectFrontmatter" and remove them from the `site` config
- With the above change, `site` will have way fewer fields; this means we can remove the `site.options` key and just define those options directly on the `site`.

rowanc1:
Thanks @fwkoch -- I am a big fan of these changes and think it would make the onboarding simpler, and the code a lot simpler to reason about. This also is closer to mimic-ing how the exports options work, which is helpful.

Issue #1103, opened 2024-04-15
Eliminate `site` key in `myst.yml`?


fwkoch:
Currently the `myst.yml` file has a `site` key and a `project` key - there is lots of confusion around the difference between these, how inheritance may or may not happen, where new keys go, etc. (as one example, see the discussion on this PR: https://github.com/executablebooks/myst-theme/pull/325)

If you look in the docs about frontmatter, `site` isn't even mentioned: https://github.com/executablebooks/mystmd/blob/main/docs/frontmatter.md In the code, there there isn't even a `validateSiteFrontmatter(...)` function; "site frontmatter" isn't really ever used. The closest we get is when building the "site manifest" we validate the frontmatter against the _site theme template_; this looks much more like `export` validation than frontmatter validation.

Why does the `site` section of the `myst.yml` exist at all? I think the only reason for this comes from the idea that a site may have multiple MyST projects: https://github.com/executablebooks/mystmd/blob/main/packages/myst-config/src/site/types.ts#L72 - that means `site` is a separate entity higher in the hierarchy than project.

In practice, however, `project:site` is always `1:1`, especially as we improve "intersphinx/inter-MyST" - there's no reason to combine multiple projects into one site; just link two sites.

---

Because of all of this, I think it would be useful to eliminate the `site` from the project config. Configuring `site` would then work the same way as configuring another `export` - specify the format, a template, and any options required by the template.

```
---
exports:
  - format: site
    template: book-theme
    logo: my-logo.png
    ...
---
```

There would no longer be the same ambiguity around inheritance; project frontmatter _is_ site frontmatter, and anything defined on the site `export` would override `project` frontmatter.

choldgraf:
It sounds like my mental model of `site:` isn't what it was originally intended for. I had assumed it was short for `website:` and was only for configuration that was related to HTML themes.

The way you describe it here sounds reasonable to me, at least as a starting point. It feels like this would be a natural progression to start with:

1. Begin with the assumption of `1 myst project per myst.yml file`. Use the minimal number of keys / config options to support this (e.g., prefer one `project:` key over 2 keys)
2. See what people do with this limitation. See if key stakeholders or many users hit an issue of "I want to have configuration that spans multiple myst projects)
3. If so, then do some UX research to understand their needs more deeply.
4. Use this to decide if / how to expand the scope of a MyST build to incorporate multiple MyST sites.

[...]

PR #1536, opened 2024-09-14
💽 Add site options to each page

rowanc1:
This enables per-page options for the site.

Picked up in the theme in https://github.com/jupyter-book/myst-theme/pull/472

### DIFF
--- a/.changeset/sour-apes-swim.md
+++ b/.changeset/sour-apes-swim.md
@@ -1,0 +1,6 @@
+---
+"myst-frontmatter": patch
+"myst-cli": patch
+---
+
+Enable site options on each page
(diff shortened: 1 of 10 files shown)


fwkoch:
Chatting some with @rowanc1 - we will keep the new `site` key, as a place to override site options at a page level. This is somewhat similar to how we treat other `exports`, where in the export definition, you can define other options and override fields. (There is some discussion of sites vs. exports here: https://github.com/jupyter-book/mystmd/issues/1103) Also, once we tackle removing the `project` key in `myst.yml` (https://github.com/jupyter-book/mystmd/issues/1308), page frontmatter with `site` alongside other keys will start looking very similar to `myst.yml` content with `site` alongside other keys.

As for `options` - these can remain, as a place to define any arbitrary key/value pair. The site/export templates should have access to these `options`, but they are less strictly validated than options defined directly on `site`.

Also note, I think this PR addresses: https://github.com/jupyter-book/mystmd/issues/1450

Issue #1539, opened 2024-09-17
Bib items should parse `urldate` in addition to `year`


rowanc1:
There are some citations that look like this:

```
@misc{echoview_software_pty_ltd_echoview_nodate,
	title = {Echoview - {Sound} {Knowledge}},
	url = {https://www.echoview.com/},
	urldate = {2020-10-04},
	author = {{Echoview Software Pty Ltd}},
}
```

They do not have a `year` but do have a `url` date. We should try to parse the date from that field if the `year` is not available. Currently these are showing as `n.d.` (no date) in the inline citations and in the reference list.

See:
- https://github.com/scipy-conference/scipy_proceedings/pull/914

PR #1541, opened 2024-09-20
🎓 Replace n.d. with access date for urls in citations

fwkoch:
This addresses #1539 - `citation-js-utils` now looks at `accessed` for date in addition to `issued`.

Currently, this change eliminates the "Accessed..." text from the reference:

image

### DIFF
--- a/packages/citation-js-utils/src/index.ts
+++ b/packages/citation-js-utils/src/index.ts
@@ -283,7 +283,6 @@ export function getCitationRenderers(data: CSL[]): CitationRenderer {
       });
       // Trim the DOIs and URLs (these are encoded) on load
       if (c.URL) c.URL = c.URL.replace(/^(%20)*/, '').replace(/(%20)*$/, '');
-      if (!c.issued && c.accessed) c.issued = c.accessed;
       return [
         c.id,
         {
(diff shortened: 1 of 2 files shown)


packages/citation-js-utils/src/index.ts
```diff
@@ -281,6 +283,7 @@ export function getCitationRenderers(data: CSL[]): CitationRenderer {
       });
       // Trim the DOIs and URLs (these are encoded) on load
       if (c.URL) c.URL = c.URL.replace(/^(%20)*/, '').replace(/(%20)*$/, '');
+      if (!c.issued && c.accessed) c.issued = c.accessed;
```
fwkoch, line 286: If we remove this line, then the _inline_ citation still has a year, but the reference still has "n.d." with "Accessed..." - looks like:

image

Possibly this is better...? Not sure.

rowanc1: I think this is the one we want. I have asked @mepa from SciPy to chime in!

cbcunc: This is better, and honest, and much like what you'd see on Wikipedia, which deals with lots of url refs.

cbcunc (approved):

mepa:
Thanks for making this update, @rowanc1 and @fwkoch!

PR #1561, opened 2024-10-03
🧹 Clean up `nodeenv` and add CI tests

agoose77:
In some recent refactoring work, the `nodeenv` integration added some duplicate code. This PR removes that, and ensures that we're testing its usage.

I couldn't find an elegant way to create an environment without Node.js without using Docker. There's also Nix, but I know that it is not something people are familiar with. Although I am confident it would be simple to understand, if it goes wrong, things get hard.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -1,0 +1,96 @@
+name: Python CI
+
+on:
+  push:
+    #branches: [main, ci-*]
+    #paths:
+    #    - 'packages/mystmd-py/**'
+  pull_request:
+    #branches: [main]
+    #paths:
+    #  - 'packages/mystmd-py/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,96 @@
...
+on:
+  push:
+    #branches: [main, ci-*]
+    #paths:
+    #    - 'packages/mystmd-py/**'
+  pull_request:
+    #branches: [main]
+    #paths:
+    #  - 'packages/mystmd-py/**'
+  workflow_dispatch:
```
agoose77, line 12: These will be un-commented once the tests pass.

rowanc1:
Feel free to take this over the line whenever you feel like it @agoose77!

PR #1563, opened 2024-10-03
🪪 Allow custom licenses outside SPDX

fwkoch:
Currently `license` values in MyST frontmatter are strict. They must be from the [SPDX license list](https://spdx.org/licenses/), referenced by ID, with no extra, custom information. This is probably how licenses _should_ be used - just pick a good, existing one and stick with that, don't try to embellish it or make up your own. However, in reality, this is not the case. Publishers, authors, etc often have their own license notes, urls for their license policies, or entirely new licenses.

This PR updates license validation in MyST to enable more flexibility when defining license.

First, you may define a license as an SPDX ID, as before. The behavior for this is unchanged: it will be automatically expanded to contain all the relevant license metadata, including name, url, OSI status, etc:

```yaml
license: CC-BY-4.0
```

And, as before, you may still separate `content` and `code` licenses - again, no changes to this.

```yaml
license:
  content: CC-BY-4.0
  code: MIT
```

However, you may now add your own `note` - in this case, other metadata will still be populated as before, alongside your note:

```yaml
license:
  id: CC-BY-4.0
  note: When attributing this content, please indicate the Source was MyST Documentation.
```

You may also override metadata fields - in this case, the new `url` will be used, but other metadata will be populated as before:

```yaml
license:
  id: CC-BY-4.0
  url: https://example.com/licenses/how-our-journal-uses-cc-by
```

Finally, you may define your own totally new license:

```yaml
license:
  id: im-not-a-lawyer
  name: I'm Not A Lawyer License
  url: https://example.com/licenses/these-wont-stand-up-in-court
```

Note, in cases all where you (1) do not use an SPDX license or (2) override default SPDX fields, MyST will raise a warning.

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -636,7 +636,7 @@ This field can be set to a string value directly or to a License object.
 
 Available fields in the License object are `content` and `code` allowing licenses to be set separately for these two forms of content, as often different subsets of licenses are applicable to each. If you only wish to apply a single license to your page or project use the string form rather than an object.
 
-String values for licenses should be a valid “Identifier” string from the [SPDX License List](https://spdx.org/licenses/). Identifiers for well-known licenses are easily recognizable (e.g. `MIT` or `BSD`) and MyST will attempt to infer the specific identifier if an ambiguous license is specified (e.g. `GPL` will be interpreted as `GPL-3.0+` and a warning raised letting you know of this interpretation). Some common licenses are:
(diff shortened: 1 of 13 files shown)


docs/frontmatter.md
```diff
@@ -658,7 +658,25 @@ String values for licenses should be a valid “Identifier” string from the [S
...
 ```
 
-By using the correct SPDX Identifier, your website will automatically use the appropriate icon for the license and link to the license definition.
+By using the correct SPDX Identifier, your website will automatically use the appropriate icon for the license and link to the license definition.  The simplest and most common example is something like:
+
+```yaml
+license: CC-BY-4.0
+```
+
+### Nonstandard licenses
```
rowanc1, line 667: Something that we are seeing a lot of in the publishing world. Exciting to be able to support these in MyST.

rowanc1:
Went through this with Franklin! Looking good, and opens up a lot more possibilities to support real-world use cases for capturing license information.

PR #1576, opened 2024-10-09
🛑 Stop the process when the CLI task is complete

rowanc1:
Fixes hanging node process waiting on unresolved promises.

### DIFF
--- a/packages/myst-cli-utils/src/utils.ts
+++ b/packages/myst-cli-utils/src/utils.ts
@@ -19,6 +19,9 @@ export function clirun<S extends ISession>(
     program: Command;
     getSession: (logger: Logger, opts?: SessionOpts) => S;
   },
+  runOptions?: {
+    forceExit?: boolean | ((...args: any[]) => boolean);
+  },
 ) {
   return async (...args: any[]) => {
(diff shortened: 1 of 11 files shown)


packages/myst-cli-utils/src/utils.ts
```diff
@@ -19,6 +19,9 @@ export function clirun<S extends ISession>(
     program: Command;
     getSession: (logger: Logger, opts?: SessionOpts) => S;
   },
+  runOptions?: {
+    forceExit?: boolean | ((...args: any[]) => boolean);
```
rowanc1, line 23: Maybe we should flip this to have this be the default and then processes that need to hang (e.g. start/watch) can add work arounds.

agoose77:
@rowanc1 do we know which processes are hanging? For example, execution is supposed to ensure that everything is cleaned up.

rowanc1:
I have spent at least four hours trying to track in down to no avail over the last few weeks. Happy for someone else to take a look, but it is a pretty hard thing to find imo....

There were anywhere between 10-100 open promises at the end of the build function. for a Typst build. And it wasn't overly consistent.

I think this hits the user's perceived problem. But for sure there are some things we need to clean up internally.

PR #1586, opened 2024-10-16
🏗 Support parts as external files and in the site config

fwkoch:
This PR addresses part of this issue: https://github.com/jupyter-book/mystmd/issues/1246 - specifically,

> Allow frontmatter part to be a file
>
> Modify behaviour of in-line frontmatter parts be processed as a "pseudo-file" rather than injected into the page tree
>
> Allow site to define parts
>
> Update static export templating logic and extractPart for new part definition locations

The parts of that issue that are not yet addressed in this PR are

> Allow exports to define parts
>
> Update site templating logic to respect parts in template.yml

~Additionally, project-level parts are still _valid_, but are not processed in the same way as page or site parts (i.e. they are still left as text in-line in the frontmatter)~ This PR has been updated so project-level parts are processed to mdast/frontmatter objects, as suggested in the review.

### DIFF
--- a/packages/myst-cli/src/build/cff.ts
+++ b/packages/myst-cli/src/build/cff.ts
@@ -14,6 +14,7 @@ import { KNOWN_IMAGE_EXTENSIONS } from '../utils/resolveExtension.js';
 import type { ExportWithOutput, ExportFnOptions } from './types.js';
 import { cleanOutput } from './utils/cleanOutput.js';
 import { getFileContent } from './utils/getFileContent.js';
+import { resolveFrontmatterParts } from '../utils/resolveFrontmatterParts.js';
 import { parseMyst } from '../process/myst.js';
 
 function exportOptionsToCFF(exportOptions: ExportWithOutput): CFF {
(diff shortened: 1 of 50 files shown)


packages/myst-cli/src/build/cff.ts
```diff
@@ -40,19 +41,27 @@ export async function runCffExport(
   let abstract: string | undefined;
   if (projectPath && selectors.selectLocalConfigFile(state, projectPath) === sourceFile) {
     frontmatter = selectors.selectLocalProjectConfig(state, projectPath);
+    // TODO: This uses project-config-level parts which are not yet supported in the same way
+    // as page- and site-level parts. Once those are supported, this needs updating.
     const rawAbstract = frontmatter?.parts?.abstract?.join('\n\n');
```
fwkoch, line 46: This is one place where we already use `project` parts - it has to be handled in the old way still

fwkoch:
Before merging, we should also process project parts the same way as site parts, even though they are unused everywhere (except the CFF export...)

packages/myst-cli/src/build/cff.ts
fwkoch: Project part processing has been updated, along with the logic here.

PR #1597, opened 2024-10-22
🐛 Typst imports can be conflict prone

rowanc1:


fwkoch (approved):
This is good to avoid conflicts if your export is defined in `myst.yml`. I guess the next step, to prevent _all_ conflicts, would be something like keeping the template files in a separate folder from the content maybe...?

rowanc1:
I think the actual solution is to rename the exports based on not conflicting with the known template.yml files + (myst-exports.typ in this case).

Issue #670, opened 2023-10-13
Support folder hierarchies and nested files in URLs without flattening them


nthiery:
When building a web site, nested file names are flattened and replaced
with slugs:

https://mystmd.org/guide/table-of-contents#nesting-of-files-in-urls

While I can see the value for the end-user (simple urls), there are use cases
where this is detrimental, as the nested file name actually carries semantic.

For example, in my [course](https://Nicolas.thiery.name/Enseignement/Info111/),
the name of a document, says `/Solutions/Semaine3/foo.html` tells that
this is a solution for notebook foo of week 3. While `Devoirs/Semaine3/foo.html`
is the original assignment. This also helps the user match the pages on the
web with the notebooks that they have dowloaded separately.

## Proposal

Provide a configuration option `flatten_filename=true/false`, with `true` as default
value, to enable or disable flattening of file names.

Provide a configuration option 'transform_filename=true/false`, with `true` as default
value, to enable or disable [file name transformation](https://mystmd.org/guide/table-of-contents#filename-transformations).

When both options are set to false, one recovers JupyterBook's behavior.

nthiery:
I believe this is my last blocker for migrating my course notes to myst :-)

Next in line: a course starting in January:

https://nicolas.thiery.name/Enseignement/IntroScienceDonnees/

Thanks in advance!

choldgraf:
Could I signal-boost this one? Is it a particularly complicated one to implement? I ask because I think this is one of the only remaining "hard blockers" for any pre-existing users of Jupyter Book or Sphinx, given that currently, switching to MyST will break everybody's URLs.

agoose77:
Talking to @choldgraf, another point to consider is the ability to preserve existing links. If we implement a jb1 slug scheme, then we could easily ensure that no links break.

However this is not necessarily a requirement.

rowanc1:
That is links in the table of contents?

I have talked to @fwkoch and we are going to do some planning on this this week, and ideally get it over the line next week.

agoose77:
I mean that should a user have some uri to `foo/bar.md`, we should maybe have a mode for preserving that (which goes beyond supporting folders to ensuring that the entire url for a page is identical).

Maybe it should be a configurable setting.

choldgraf:
As a few driving examples:

- Jupyter Book's default HTML builder uses `.html` files, here's what they look like: `https://jupyterbook.org/en/stable/publish/gh-pages.html`
[...]

PR #1601, opened 2024-10-23
🗂 Add site option to include folder structure in url paths

fwkoch:
This PR adds an option to support nested folder structure in URLs. See https://github.com/jupyter-book/mystmd/issues/670

This uses an option at the `site` level of your `myst.yml` - `site.options.url_folders: true`. Given this option, slugs are computed to contain dot-delimited folder structure (as opposed to the previous behavior where slugs were computed entirely based on filename). Consuming these new slugs with folders requires an updated myst theme, see https://github.com/jupyter-book/myst-theme/pull/489 (without the updated theme, the dots in the slugs will just show up in your MyST site; with the updated theme, your site URLs will have slashes instead of dots)


packages/myst-cli/src/project/load.ts
```diff
@@ -57,9 +57,13 @@ export async function loadProjectFromDisk(
   let newProject: Omit<LocalProject, 'bibliography'> | undefined;
   let { index, writeTOC } = opts || {};
   let legacyToc = false;
+  const siteConfig = selectors.selectLocalSiteConfig(state, path);
+  const urlFolders = !!siteConfig?.options?.url_folders;
```
fwkoch, line 61: I'm happy to take suggestions about a better name for this option...

agoose77: @fwkoch do we have a policy about `snake_case` vs `pascalCase` for frontmatter/config item names?

stevejpurves:
@fwkoch seems like there will be no impact on the `location` field in the output mdast and in `source` on notebook outputs?

fwkoch:
@stevejpurves - yeah, `location` and `source` should be unchanged - these are still the relative file paths.

packages/myst-cli/src/process/mdast.ts
```diff
@@ -262,6 +262,7 @@ export async function transformMdast(
     url = `/${useSlug ? pageSlug : ''}`;
     dataUrl = `/${pageSlug}.json`;
   }
+  url = url?.replace('.', '/');
```
rowanc1, line 265: do we need the `.index` and replaceAll?

packages/myst-cli/src/process/site.ts
```diff
@@ -172,14 +172,14 @@ export async function writeMystSearchJson(session: ISession, pages: LocalProject
...
-      if (!mdast || !frontmatter) {
+      if (!mdast || !frontmatter || !slug) {
         return [];
       }
       const title = frontmatter.title ?? '';
 
       // Group by section (simple running accumulator)
       const sections = toSectionedParts(mdast);
-      const pageURL = slug && DEFAULT_INDEX_FILENAMES.includes(slug) ? '/' : `/${slug}`;
+      const pageURL = DEFAULT_INDEX_FILENAMES.includes(slug) ? '/' : `/${slug.replace('.', '/')}`;
```
rowanc1, line 182: replaceAll

packages/myst-cli/src/project/load.ts
rowanc1: lets do `site.folders` and get around it!

packages/myst-cli/src/process/mdast.ts
rowanc1:
```suggestion
  url = url?.replace(/\.index$/, '').replaceAll('.', '/');
```

packages/myst-cli/src/process/site.ts
rowanc1:
```suggestion
      const pageURL = DEFAULT_INDEX_FILENAMES.includes(slug) ? '/' : `/${slug.replace(/\.index$/, '').replaceAll('.', '/')}`;
```

rowanc1 (approved):

PR #1603, opened 2024-10-25
🧹 Upgrade unified/unist stack

agoose77:
This PR starts the process of upgrading our internals to use `unified@^11`, `@types/mdast@^4`, and `@types/unist@^3`. We will need to make this change _at some point_, and given the nature of unifiedjs, this is a broad-reaching change.

It massages our existing code to stop complaining about the `GenericParent <-> Node` differences that arise in `@types/mdast@^4`, and pulls in the latest version of the various unist utilities. This is mostly done by additions of type asserts (`as XXX`) that will need to be removed in future.

Related to this work is jupyter-book/myst-spec#67, whereby the `myst-spec` types have been modified to use the extension mechanism supported by `@types/mdast@>=4`. After _this_ PR in jupyter-book/mystmd#1603 is merged, the integration of jupyter-book/myst-spec#67 should remove the need for `GenericParent` altogether.

To be clear, this PR is independent of jupyter-book/myst-spec#67

> [!NOTE]  
> We can't completely upgrade the stack: the unified-latex ecosystem is stuck on an older spec.
> However, this is an internal tool, so we just need to ensure that it generates valid mdast@^4, and ignore any type errors. I think that it _does_ from glancing at it.


rowanc1:
🚀

packages/myst-cli/src/transforms/crossReferences.ts
```diff
@@ -140,6 +140,7 @@ export async function transformMystXRefs(
       } else {
         const data = await fetchMystXRefData(session, node as CrossReference, vfile);
         if (!data) return;
+        console.debug({ identifier: node.identifier });
```
rowanc1, line 143: remove!

packages/myst-parser/tests/myst.spec.ts
```diff
@@ -163,6 +162,7 @@ describe('Testing mdast --> html conversions', () => {
   test.each(htmlCases)('%s', (name, { html, mdast }) => {
     const modified = replaceCommentNodes(mdast);
     if (html) {
+      console.log({ name, skip: SKIP_TESTS.some((p) => p.test(name.trim())) });
```
rowanc1, line 165: Probably remove this?

packages/myst-cli/src/transforms/crossReferences.ts
agoose77:
```suggestion
```

packages/myst-parser/tests/myst.spec.ts
agoose77:
```suggestion
```

PR #1622, opened 2024-11-04
🔧 Typst card support, multipage pagebreaks, tex image argument fix

fwkoch:


packages/myst-to-typst/src/index.ts
```diff
@@ -454,6 +456,23 @@ const handlers: Record<string, Handler> = {
...
+  card(node, state) {
+    if (node.url) {
+      node.children?.push({ type: 'paragraph', children: [{ type: 'text', value: node.url }] });
+    }
+    state.renderChildren(node);
+    state.ensureNewLine();
+    state.write('\n');
+  },
+  cardTitle(node, state) {
+    headingHandler({ ...node, depth: 3 }, state);
```
rowanc1, line 468: I think this should just be bold here. Otherwise all of these will end up in  table of contents, etc.

fwkoch: Ah yeah, that `depth 3` was also totally arbitrary. Thanks

packages/myst-to-typst/src/index.ts
```diff
@@ -454,6 +454,27 @@ const handlers: Record<string, Handler> = {
...
+  card(node, state) {
+    if (node.url) {
+      node.children?.push({ type: 'paragraph', children: [{ type: 'text', value: node.url }] });
+    }
+    state.renderChildren(node);
+    state.ensureNewLine();
+    state.write('\n');
+  },
+  cardTitle(node, state) {
+    state.write('*');
```
fwkoch, line 466: These are now just bold, rather than headings. That means they will not show up in the toc, but they may now be dangling at the bottom of a page in the final PDF. Not really a big deal. Probably need to give cards more thought if they become widespread in typst exports.

PR #1628, opened 2024-11-07
Improve web documentation and document styles

choldgraf:
This PR has two focuses:

1. Adds documentation about how to customize style for the web, based on a recent conversation I had with @rowanc1 
2. Moves our web docs to a new `web/` folder and enables folder-based URLs for the MyST guide. I didn't do this for all docs because I first wanted to see how this looked to folks before suggesting we do it with other sections in a follow-up PR.


docs/website-style.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Use `div` and `span` elements
+
+You can attach classes directly to [`div` and `span` elements](#div-and-span).
+This works with `style` properties as well.
+
+For example:
+
+<div class="col-gutter-right" style="font-weight: bold;">Here's my div</div>
+Here's some <span class="col-gutter-right" style="font-weight:bold;">Span</span> content
```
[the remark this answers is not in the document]
rowanc1: Note that inline elements can't be in a column though, can update this to a slightly different example!

docs/website-style.md
```diff
@@ -0,0 +1,94 @@
...
+```{note}
+:class: col-gutter-right
+I'm on the right!
+```
+````
+
+```{note}
+:class: col-gutter-right
+I'm on the right!
+```
```
agoose77, line 76:
```suggestion
You can attach classes directly to [`div` and `span` elements](#div-and-span).

{myst:directive}`div` and {myst:role}`span` are analogous to their HTML counterparts. The HTML elements can also be given `style` options, e.g.

<div class="col-gutter-right" style="font-weight: bold;">Here's my div</div>

Here's some <span class="col-gutter-right" style="font-weight:bold;">Span</span> content

## Add CSS classes to directives

:::{note} Not all directives support the `:class:` option
If you wish to attach classes to a directive that doesn't seem to support it, please [open an issue](https://github.com/jupyter-book/mystmd/issues)
:::

Many directives and content blocks have a `:class:` option that can be used to add arbitrary CSS classes.
For example, below we add a CSS class to an admonition directive to snap it to the right:

````md
```{note}
:class: col-gutter-right
I'm on the right!
```
````

```{note}
:class: col-gutter-right
I'm on the right!
```
```

**Comment by choldgraf** [comment]
@agoose77 I see a bunch of updates in here - can you merge this whenever you like? Alternatively I will merge this PR the next time I think about it if I don't hear otherwise.

File: docs/website-style.md [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+
++++
+
+Normal-styled
+
+## Use `div` and `span` elements
+
+You can attach classes directly to [`div` and `span` elements](#div-and-span).
+
+{myst:directive}`div` and {myst:role}`span` are analogous to their HTML counterparts. The HTML elements can also be given `style` options, e.g.
```
agoose77, line 50:
```suggestion
{myst:directive}`div` and {myst:role}`span` are analogous to their HTML counterparts. Unlike the directive and roles, the HTML elements can also be given `style` options, e.g.
```

agoose77 (approved):

PR #1634, opened 2024-11-13
🪪 Improvements to spdx license validation

fwkoch:
This just makes two minor improvements to license validation - I'll point them out in the comments.

### DIFF
--- a/packages/myst-frontmatter/src/licenses/licenses.yml
+++ b/packages/myst-frontmatter/src/licenses/licenses.yml
@@ -280,6 +280,53 @@ cases:
         content:
           url: https://example.com
     warnings: 1
+  - title: url string coerces to license from spdx list
+    raw:
+      license: https://creativecommons.org/licenses/by/4.0/
+    normalized:
+      license:
+        content:
+          id: CC-BY-4.0
+          name: Creative Commons Attribution 4.0 International
+          url: https://creativecommons.org/licenses/by/4.0/
+          CC: true
+          free: true
+  - title: url coerces to license from spdx list
+    raw:
+      license:
+        url: https://creativecommons.org/licenses/by/4.0/
+    normalized:
+      license:
+        content:
+          id: CC-BY-4.0
+          name: Creative Commons Attribution 4.0 International
+          url: https://creativecommons.org/licenses/by/4.0/
+          CC: true
+          free: true
+  - title: url coerces to license from spdx list with http
+    raw:
+      license:
+        url: http://creativecommons.org/licenses/by/4.0/
+    normalized:
+      license:
+        content:
+          id: CC-BY-4.0
+          name: Creative Commons Attribution 4.0 International
+          url: https://creativecommons.org/licenses/by/4.0/
+          CC: true
+          free: true
+  - title: url coerces to license from spdx list without trailing slash
+    raw:
+      license:
+        url: https://creativecommons.org/licenses/by/4.0
+    normalized:
+      license:
+        content:
+          id: CC-BY-4.0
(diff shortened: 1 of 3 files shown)


packages/myst-frontmatter/src/licenses/licenses.yml
```diff
@@ -280,6 +280,53 @@ cases:
         content:
           url: https://example.com
     warnings: 1
+  - title: url string coerces to license from spdx list
+    raw:
+      license: https://creativecommons.org/licenses/by/4.0/
```
fwkoch, line 285: Previously, this test case passed - i.e. setting `license` to a known SPDX url would validate to the correct SPDX license object.

packages/myst-frontmatter/src/licenses/licenses.yml
```diff
@@ -280,6 +280,53 @@ cases:
...
+        content:
+          id: CC-BY-4.0
+          name: Creative Commons Attribution 4.0 International
+          url: https://creativecommons.org/licenses/by/4.0/
+          CC: true
+          free: true
+  - title: url coerces to license from spdx list
+    raw:
+      license:
+        url: https://creativecommons.org/licenses/by/4.0/
```
fwkoch, line 297: However, this would fail - i.e. if the url was explicitly provided as `license.url`, it would _not_ validate to the SPDX license.

Now, this coerces correctly to SPDX.

packages/myst-frontmatter/src/licenses/licenses.yml
```diff
@@ -300,11 +347,21 @@ cases:
...
       license: my custom license
     normalized:
       license:
         content:
           name: my custom license
     warnings: 1
+  - title: Long string with "mit" buried in it is not coerced to MIT license
+    raw:
+      license: |
+        This is an open access article. Unrestricted non-commercial use is permitted provided the original work is properly cited
```
fwkoch, line 361: The utility MyST uses to match license IDs with SPDX is _way_ too generous. It would pick up on the `mit` in the middle of "permitted" here, and coerce this to the MIT license. 😱

packages/myst-frontmatter/src/licenses/validators.ts
```diff
@@ -103,7 +103,8 @@ export function validateLicense(input: any, opts: ValidationOptions): License |
   if (typeof input === 'string') {
     const value = validateString(input, opts);
     if (value === undefined) return undefined;
-    const valueSpdx = correctLicense(value);
+    // Do not try to coerce long values into SPDX licenses
+    const valueSpdx = value.length < 50 ? correctLicense(value) : undefined;
```
fwkoch, line 107: Now, we still use the same utility, but we only use it if the provided ID is a reasonable ID length...

Maybe this still isn't the best solution - It would be nice to just have a smarter tool that doesn't just pick up on any substring. But for now, this is an easy improvement.

rowanc1: Can we cut it down further to 10 characters? Not sure what the longest id string is, but that probably is the longest we would want to coerce?

fwkoch: Yeah, that makes sense. The longest spdx id is 36 characters I think, but that includes a lot of long weird ones.

I might cut it down to 15 - Looking at `CC-BY-NC-ND-4.0` which is a normal, CC license (and 15 chars).

fwkoch:
```suggestion
    const valueSpdx = value.length < 15 ? correctLicense(value) : undefined;
```

PR #1636, opened 2024-11-13
🏷️ Do not report available updates for whitelabelled applications

agoose77:
Although we may re-think the white-labelling approach in future, for now it's useful.

### DIFF
--- a/packages/myst-cli/src/session/session.ts
+++ b/packages/myst-cli/src/session/session.ts
@@ -103,6 +103,7 @@ export class Session implements ISession {
       })
       .catch(() => null);
   }
+
   showUpgradeNotice() {
     if (
       this._shownUpgrade ||


packages/myst-cli/src/session/session.ts
```diff
@@ -102,9 +103,17 @@ export class Session implements ISession {
       })
       .catch(() => null);
   }
-
+  /**
+18:export function isWhiteLabelled(): boolean {
```
rowanc1, line 107: Can remove this?

rowanc1 (changes requested):

packages/myst-cli/src/session/session.ts
agoose77: Yes, please.

packages/myst-cli/src/session/session.ts
```diff
@@ -102,9 +103,17 @@ export class Session implements ISession {
...
+ */
```
rowanc1, line 108:
```suggestion

```

rowanc1 (approved):

agoose77:
:rocket: :rocket: :rocket:

PR #1650, opened 2024-11-18
🎁 Use `commonDirectiveOptions` in all directives

agoose77:
This PR tackles #1657. We will need an upstream `myst-theme` PR to ensure that these are actually set in the DOM.

I omitted the documentation directives `{myst:role}` et al. I didn't think they needed it.

### DIFF
--- a/docs/exercises.md
+++ b/docs/exercises.md
@@ -7,16 +7,6 @@ thumbnail: ./thumbnails/exercise.png
 
 There are two directives available to add exercises and solutions to your documents: (1) an `exercise` directive; and (2) a `solution` directive. The exercises are enumerated by default and can take in an optional title argument as well as be "gated" around Jupyter Notebook cells.
 
-:::{note} Same as Sphinx Exercise 🎉
-:class: dropdown
-
-The implementation and documentation for exercises and solutions is based on [Sphinx Exercise](https://ebp-sphinx-exercise.readthedocs.io), the syntax can be used interchangeably. We have reused the examples in that extension here to show off the various parts of the MyST extension.
-
-Changes to the original extension include being able to click on the exercise label (e.g. "Exercise 1"), and having a link to that exercise anchor. We have also updated the styles from both Sphinx and Jupyter Book to be more distinct from admonitions.
-
-You can also reference exercises with any cross-reference syntax (including the {myst:role}`ref` and {myst:role}`numref` roles). We recommend the markdown link syntax.
-:::
-
 ## Exercise Directive
 
 ::::{tab-set}
--- a/docs/exercises.md
+++ b/docs/exercises.md
@@ -104,7 +104,7 @@ _Source:_ [QuantEcon](https://python-programming.quantecon.org/functions.html#Ex
 
 The following options for solution directives are supported:
 
-:::{myst:directive} exercise
+:::{myst:directive} solution
 :::
 
 ## Referencing Exercises & Solutions
--- a/docs/exercises.md
+++ b/docs/exercises.md
@@ -19,7 +19,8 @@ You can also reference exercises with any cross-reference syntax (including the 
 
 ## Exercise Directive
 
-**Example**
+::::{tab-set}
+:::{tab-item} Example
 
 ```{exercise}
 :label: my-exercise
@@ -34,7 +35,8 @@ In particular, write a function `factorial` such that `factorial(n)` returns $n!
 for any positive integer $n$.
 ```
 
-**MyST Syntax**
+:::
+:::{tab-item} MyST Syntax
 
 ````markdown
 ```{exercise}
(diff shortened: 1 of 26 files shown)


agoose77:
@fwkoch can you help clarify the changeset semver policy? I'm not clear for mystmd when we can make a minor bump.

agoose77:
Requires https://github.com/jupyter-book/myst-theme/pull/500 to see some of the improvements.

fwkoch:
I put my thoughts about the semver policy here: https://github.com/jupyter-book/mystmd/pull/1658 - nothing official in that yet, just my opinions!

fwkoch (approved):
This is really nice for adding `class` to all node types, it adds that feature while simplifying the code.

My only hesitation is that it also adds `enumerated`/`enumerator` to all the node types. To me, this feels slightly less general than `class`, which is just a transparent string that themes may or may not support...

Enumeration gets picked up by the `enumerate` transforms and impacts the mdast processing. Testing a little bit, it seems ok... good actually... I can now enumerate many things that I could not before, and it _seems_ to just work. However, it makes the scope of this PR bigger than it might seem at first - it doesn't just add some simple properties to directives, it enriches (hopefully without breaking) fairly significant myst functionality.

Anyway, after going through all this, I'm still approving - I don't actually think there is a problem! Mostly just had to wrap my head around the implications of `enumerated`.

packages/myst-ext-exercise/src/exercise.ts
```diff
@@ -36,19 +32,21 @@ export const exerciseDirective: DirectiveSpec = {
...
     const exercise: GenericNode = {
       type: 'exercise',
-      label,
-      identifier,
-      class: data.options?.class as string,
       hidden: data.options?.hidden as boolean,
       enumerated: !nonumber,
       children: children as any[],
     };
+    addCommonDirectiveOptions(data, exercise);
```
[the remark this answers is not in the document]
fwkoch: Ah, I guess this is a small change requested, if you agree @agoose77

agoose77:
Yes @fwkoch, and helpful of you to explicitly surface that change --I made the same judgment that it feels like we should be able to identify and reference most (all) directives, but that isn't a small conceptual change by any means.

packages/myst-ext-exercise/src/exercise.ts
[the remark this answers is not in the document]
fwkoch: Maybe this is just a documentation thing. `nonumber: true` is really just equivalent to `enumerated: false`, and I think it's fair that the former is just legacy support and the latter takes precedence?

Ugh - looking at the docs, there are pages where these directive properties are manually documented. e.g. https://mystmd.org/guide/exercises#exercise-directive - we should update those to use the `myst:directive` directive (though we would need to update that plugin to also import the `myst-ext-*` directives...).

packages/myst-ext-exercise/src/exercise.ts
```diff
@@ -8,12 +9,7 @@ export const exerciseDirective: DirectiveSpec = {
...
   options: {
-    label: {
-      type: String,
-    },
-    class: {
-      type: String,
-    },
+    ...commonDirectiveOptions('exercise'),
     nonumber: {
       type: Boolean,
```
fwkoch, line 14:
```suggestion
      type: Boolean,
      doc: 'Legacy flag to disable numbering of exercises; equivalent to `enumerated: false`',
```

packages/myst-ext-proof/src/proof.ts
```diff
@@ -24,13 +24,7 @@ export const proofDirective: DirectiveSpec = {
...
-    label: {
-      type: String,
-      alias: ['name'],
-    },
-    class: {
-      type: String,
-    },
+    ...commonDirectiveOptions('proof'),
     nonumber: {
       type: Boolean,
```
fwkoch, line 29:
```suggestion
      type: Boolean,
      doc: 'Legacy flag to disable numbering of proofs; equivalent to `enumerated: false`',
```

fwkoch (commented):
Just thinking a little more about some documentation loose ends here...

agoose77:
@fwkoch thanks for your review! Really good catch on the `enumerated` - `nonumber` overlap! I've updated the directives to prefer `nonumber` to `enumerated`, but respect the latter if given.

I've also reworked the docs to use `{myst}` or tabs for pairing syntax and rendered examples. **I removed reference to `sphinx-exercise`**, because at this point we are unifying things in a breaking manner, and we're trying to distance ourselves from the notion that MyST is 100% compatible with Sphinx.

docs/exercises.md
```diff
@@ -113,21 +97,15 @@ factorial(4)
...
+::::
 
-  Value of the solution’s class attribute which can be used to add custom CSS or JavaScript.
+_Source:_ [QuantEcon](https://python-programming.quantecon.org/functions.html#Exercise-1)
 
-- `hidden` : flag (empty)
+The following options for solution directives are supported:
 
-  Removes the directive from the final output.
+:::{myst:directive} exercise
```
fwkoch, line 107:
```suggestion
:::{myst:directive} solution
```

fwkoch:
The last thing I did was add `commonDirectiveOptions` to `solution` directive. I think that makes sense - you can now have, for example, `Solution 1` and `Solution 2` which may both be solutions to `Exercise 1`.

PR #1658, opened 2024-11-19
👩‍💻 Proposed notes on versioning for contributors

fwkoch:
This PR is an attempt to slightly formalize my answer to this: https://github.com/jupyter-book/mystmd/pull/1650#issuecomment-2485557941

I'm totally open to feedback here - this is just the general rules I personally follow when bumping versions (and I always try to avoid any changes that don't just fall under `patch`...).

These rules are different from "official" semantic versioning which is basically - major: any breaking change, minor: any feature, patch: bugfix. But I'd argue in favor of maintaining the distinction between user-facing breaking change vs. developer-facing breaking change. (Possibly this suggests we should have different versioning on `mystmd` - "user-facing" - and `myst-cli` - "dev-facing" - where breaking changes can imply different levels of impact...)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,6 +105,12 @@ Running in live-changes mode: depending on the package you are working in we hav
 
 `mystmd` uses [changesets](https://github.com/changesets/changesets) to document changes to this monorepo, call `npm run changeset` and follow the prompts. Later, `npm run version` will be called and then `npm run publish`.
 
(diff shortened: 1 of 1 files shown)


choldgraf:
For what it's worth, I'd be keen on trying out Jacob Tomlinson's effver approach, which I've found to more naturally map onto the ways that devs think about versioning their code. It is basically semver, but with the guiding question.being "how much effort would it likely take your target user to upgrade?".

That said I'm not sure how it works in practice so I don't feel super strongly about this.

https://jacobtomlinson.dev/effver/

agoose77:
@fwkoch I like this PR. If you are happy that it represents the current state of thinking on your part and Rowan's, then let's merge?

PR #1660, opened 2024-11-20
🖼️ Use Lorem Picsum instead of Unsplash for demo images

agoose77:
This fixes up some errors in our build console.


fwkoch (approved):
Nice! Replaces the last bits of broken unsplash stuff with an equivalent service.

PR #1661, opened 2024-11-20
⬆️ Lift and parse Markdown/text outputs from Jupyter outputs

agoose77:
A highly-requested feature on our roadmap for Jupyter Book 2.0.0b0 is #1026. This issue requests support for parsing Markdown in Jupyter code-cell outputs, so that they can participate in our build-time transforms. 

Presently, Markdown is left in its raw-text form until render time. At the time of render, we cannot faithfully do much with this Markdown, unless we parse it as CommonMark (stateless), because MyST Markdown can alter the state of the build (e.g. by incrementing reference counts).

Fixing this in MyST-native manner involves some refactoring of our AST. Presently, the AST structure built by parsing a Jupyter Notebook `.ipynb` file is:
```yaml
- type: block
  children:
    - type: code
      ...
    - type: output
      data: 
        - output_type: display_data
          ...
        - output_type: display_data
          ...
```

In order for MyST to see these outputs, we need to rework this so that each `output.data[]` item is an AST node, e.g.:
```yaml
- type: block
  children:
    - type: code
      ...
    - type: outputs
      children:
        - type: output
          data: 
            output_type: display_data
            ...
        - type: output
          data: 
            output_type: display_data
            ...
```

By doing this, we can then add `children` to each `output`, and embed the result of e.g Parsing Markdown. An important point to note is that there can be multiple MIME keys. We need, at build time, to _choose_ which MIME type to "process" and embed in the AST. There are some consequences to this (see footnote).[^parse]


agoose77:
This is a sub-optimal PR vs https://github.com/jupyter-book/mystmd/pull/1671. Let's just do https://github.com/jupyter-book/mystmd/pull/1671

PR #1665, opened 2024-11-21
🔧 More typst export improvements

fwkoch:
This PR just has a handful of typst improvements from working on formatting a PDF export:

Newly supported small typst features include:
- ordered lists that do not start at 1
- table cell background color

Areas where typst now matches other myst exports better:
- cross-references maintain their children (e.g. `[see here](#fig1)` will render as "see here" in typst instead of the default "Figure 1")
- table legends now appear below tables

Areas where it is easier to customize typst:
- default "breakable" value for figures may be set in your source content
- "columnStyle" can be defined on a per-table basis, allowing injecting specific column widths (and possibly other table styles)

### DIFF
--- a/packages/myst-to-typst/tests/cross-references.yml
+++ b/packages/myst-to-typst/tests/cross-references.yml
@@ -21,7 +21,7 @@ cases:
     typst: |-
       = My Heading <section-one>
 
-      Please see @section-one for more information!
+      Please see #link(<section-one>)[] for more information!
   - title: references followed by text
     mdast:
       type: root
@@ -35,4 +35,4 @@ cases:
             - type: text
               value: 'a for the first part of the figure!'
     typst: |-
-      See #[@fig1]a for the first part of the figure!
+      See #link(<fig1>)[]a for the first part of the figure!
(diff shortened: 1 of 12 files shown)


packages/myst-to-typst/tests/cross-references.yml
```diff
@@ -35,4 +35,4 @@ cases:
             - type: text
               value: 'a for the first part of the figure!'
     typst: |-
-      See #[@fig1]a for the first part of the figure!
+      See #link(<fig1>)[]a for the first part of the figure!
```
rowanc1, line 38: I think that this actually is the case where we want this to continue happening with an `@id` now.

Right now this is empty link text.

rowanc1 (changes requested):
Want to keep the current functionality if there is empty link text.

rowanc1 (approved):

packages/myst-to-typst/tests/cross-references.yml
fwkoch: Nice, yeah, I was thinking about this - wasn't sure if we wanted MyST default text or typst default text. Thanks for fixing it up!

PR #1671, opened 2024-11-22
🪓 Split `Output` into `Outputs` → `Output[]`

### LINKED ISSUE #1674: Add support for representing per-output sub-ASTs in MyST AST
MyST AST does not support per-output AST nodes, which is required for parsing full (MyST) Markdown. We need to find a way to support introducing a singular AST tree per output.

agoose77:
This PR closes #1674 by introducing two new versions (`1*` and `2`) of the MyST AST: 

<details><summary>

Version `2`

</summary>

```typescript
type CodeBlock = {
  type: "block";
  kind: "notebook-code",
  children: [
    Code,
    Outputs
  ]
}

type Outputs = {
  type: "outputs";
  children: Output[];
  visibility?: ...;
  identifier?: ...;
  html_id?: ...;
  label?: ...;
}

type Output = {
  type: "output";
  children: GenericNode[];
  jupyter_data: IOutput;
  identifier?: ...;
  html_id?: ...;
  label?: ...;
}
```

</details> 

<details><summary>

Version `1*`

</summary>

```typescript
type CodeBlock = {
  type: "block";
  kind: "notebook-code",
  children: [
    Code,
    Output
  ]
}
type Output = {
  type: "output";
  children: GenericNode[];
  data: IOutput[];
  _future_ast: Outputs
}
```

</details> 

This PR does the following things with these AST types:

1. Recognise version `1*` and `2` in input data (e.g. `.myst.json` or XRefs) and upgrade it to version 2
2. Process MyST AST in version `2`
3. Export MyST AST to disk as version `1*` (to avoid breaking old consumers)

The idea here is to start publishing a modified version of the current schema that is forward-compatible (version `1*`) ASAP. Then, at some future moment, we will trigger an AST upgrade and start publishing version `2` in a _future_ build of MyST.

Versions of `mystmd` that include #1671 will recognise this AST, and have no problems in handling it. 

> [!WARNING]
> Upgrading from `1` to `2` is a lossy process — there is no obvious way to figure out which `output.children` belong to which `output.data` iff. the `output.data.length > 1`. I think that's a rare case. If we really care about it, we can look to store the original version 1 AST to support round-tripping.

This PR does not yet solve #1734 entirely -- for that, we need to pull in the `myst-compat` package more dynamically, so that it can be pulled in for old `mystmd` binaries.

### To Do
- [x] Validate behavior of `reduceOutputs` -- do we want this removing `Outputs` nodes?
- [ ] Validate change to `blockMetadataTransform` that leaves empty `Outputs` nodes
- [ ] Apply block data labelling _after_ execution (follow up)
- [ ] Investigate consequences of parsing Markdown & images etc. _before_ making a release


packages/myst-cli/src/process/mdast.ts
```diff
@@ -380,6 +385,7 @@ export async function finalizeMdast(
 ) {
   const vfile = new VFile(); // Collect errors on this file
   vfile.path = file;
+
```
agoose77, line 388:
```suggestion
```

packages/myst-cli/src/process/notebook.ts
```diff
@@ -165,17 +165,20 @@ export async function processNotebookFull(
...
-        // Embed outputs in an output block
-        const output: { type: 'output'; id: string; data: IOutput[] } = {
-          type: 'output',
-          id: nanoid(),
-          data: [],
+        const outputsChildren = (cell.outputs as IOutput[]).map((output) => {
+          // Embed outputs in an output block
+          const result: { type: 'output'; id: string; jupyter_data: IOutput } = {
+            type: 'output',
+            id: nanoid(),
```
agoose77, line 172: Not certain of the purpose of `id`, but I suspect we actually want to lift this up.

rowanc1:
From meeting:

- Split out the compat repo.
- Make it simple, e.g. footnote number --> enumerated (currently)
- merge that!
- Open an outputs pr on top (only focus on breaking changes, no `_futureAst`)
- Make that at the same time for a theme PR. We release those at the same time.
- Them do the markdown from outputs PR.

PR #1699, opened 2024-12-12
☣️ Refactor `transformMdast` and `postProcessMdast` to a single pipeline

agoose77:
This spike investigates fusing our single and multi-document pipelines. The motivation for this follows from previous conversations at SciPy 2024 about making more responsive MyST `start` builds, and capturing transform dependencies.

We originally discussed building MyST transforms as a directional graph. This PR is much less ambitious; it's not clear that a directional graph is necessarily desirable; MyST would become much harder to reason about if we can't treat it as a linear series of transforms.

There were other reasons not do go as far as a graph, but I can't remember them at this very moment.

1. Support fine-grained `before` and `after` constraints on plugins
2. Reason about MyST "pipeline" in one place
3. Prepare for customisable pipelines e.g. fast and slow passes

We did not have to fuse the pre and post phases; that was a choice to put logic in one place. The rationale is that we can treat the need perform referencing across pages as a "sync" point for builds within a single-document pipeline.

### DIFF
--- a/packages/myst-cli/src/build/utils/getFileContent.ts
+++ b/packages/myst-cli/src/build/utils/getFileContent.ts
@@ -80,31 +80,53 @@ export async function getFileContent(
   // Keep 'files' indices consistent in 'allFiles' as index is used for other fields.
   const allFiles = [...files, ...projectFiles, ...projectParts];
 
-  const { dispatch, promises: filePromises } = makeSyncPoint(allFiles);
+  const { dispatchReferencing, promises: referencingPromises } = makeSyncPoint(allFiles);
+  const { dispatchIndexing, promises: indexingPromises } = makeSyncPoint(allFiles);
 
   // TODO: maybe move transformMdast into a multi-file function
   const referenceStateContext: {
     referenceStates: ReturnType<typeof selectPageReferenceStates>;
   } = { referenceStates: [] };
-  Promise.all(filePromises).then(() => {
-    const pageReferenceStates = selectPageReferenceStates(
-      session,
-      allFiles.map((file) => {
-        return { file };
-      }),
-    );
+  const referencingPages = allFiles.map((file) => {
+    return { file };
+  });
+  Promise.all(referencingPromises).then(() => {
+    const pageReferenceStates = selectPageReferenceStates(session, referencingPages);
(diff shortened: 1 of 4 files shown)


agoose77:
I'm breaking for Christmas, but I will take notes here of what is left to do!

- [ ] Simplify `set-shared-state` & `reference-resolution` -- this obfuscates parsing of shared states and then generation of `MultiPageReferenceResolver`. This can be done in a single blocking promise, and should be written somewhere for index generation to use.
- [ ] Move body of `indexingPromise.then()` to `transformMdast`, and introduce a blocking sync point after? 
- [ ] Maybe add implementation of barrier `makeBarrier(name: string, impl: () => void)` to allow barrier body (e.g. cross-reference generation) to be defined in `transformMdast`? 
- [ ] Change `transformMdast` to _build_ a pipeline, and have the caller invoke it.
- [ ] Change `finalizeMdast` to _build_ a pipeline, and have the caller invoke it.

rowanc1:
This needs to be updated when we merge #1727, the gates plugin needs to happen before the blocks.

agoose77:
I took another look at this again, thinking it might provide a pathway for making our internals easier to reason about. I've been unsure, and now have a firmer idea.

Things I like:
- Easy to reason about transforms in one place
- Elegance of sync points to represent "perform referencing", etc.
- Improvement in plugin UX to plug in to more parts of the build
- By naming individual transforms, we can trace overheads of each step. 

Things I don't like:
- Hard to elegantly update only referencing e.g. after rebuilding documents A and B, update A, B, and C post-referencing. 
- Awkward state management through use of the builder pattern (`builder.build().run()`) which is immediately realised

I like thinking about sync points as _phases_ that we can revert back to. It might be a useful way to generalise the need to re-reference, re-index, etc. The builder pattern itself was intended as a way to let us build a graph including plugins. A smaller jump would be to drop the builder and instead run the transform in a single function with regular event dispatchers. This is much more like Sphinx. We can't then use constraints to figure out when to run plugins, because we're discovering phases as we run the function rather than at build time. But that might be fine -- most people want to say "run this plugin after XXX".

One of the original goals was progressive enhancement. Maybe by thinking about that some more we can figure out whether there's a better design.

Issue #1157, opened 2024-04-24
using `figure` placeholders to get hard copy placeholders, can break the web experience for altair


stevejpurves:
Current the [figure directive](https://mystmd.org/guide/directives#directive-figure) has a `:placeholder:` option that I have been using in the following ways:

1. Given an figure based on an `ipywidgets` notebook output, I can provide an image to use in a myst website on immediate load, where no output is otherwise available, and this will also be used in hard copy pdf export
2. Given [an altair based figure](https://www.stevejpurves.com/computational-finance/mpt-ef-customer-report), I can use the placeholder to provide an image that is used when generating a pdf from the document

However, in case 2 the interactive Altair outputs are no longer rendered by myst-theme, and the placeholder is used instead, meaning I need to connect to a kernel (if possible/enabled) and execute the figure/notebook before I can interact with the plot. this defeats the object of the altair plot.

## Proposal

We add a `:hard-copy:` option to the figure such that we can satisfy case 2. without having to accept the `:placeholder:` behaviour, which is desirable in case 1. `:hard-copy:` should be preferred over `:placeholder:` in pdf/doc builds if both are set, and `:placeholder:` would be used in absence of explicit `:hard-copy:`

## Additional notes

I have considered that the theme could use logic to somehow prioritise altair/vega's mimetype over the static image or placeholder. While that is possible, it leads in to very specific prioritisation and other/future jupyter outputs that warrant the same changes would require theme updates to support. A new option communicates the author's intent better too.

agoose77:
@stevejpurves @fwkoch I had some more thoughts about this. 

Is the issue really just that in web builds, we want placeholders for widgets that require a kernel, but for non kernel-backed widgets, we want the best output?

Do you know of any examples where it's not that simple?

If not, I wonder whether we need this additional placeholder at all, or whether it's really only widgets haht are special.

stevejpurves:
The intention of placeholders so that the author can take control of what is first displayed in any figure either on first load or statically. 

Right now in myst-theme that only makes sense for kernel backed widgets, as they are the only ones that can change. Other themes that support different things might use placeholders differently (e.g. while dynamic javascript is loaded, or before a non jupyter based interactive output is activated), but the point remains the same in that it gives the author the mechanism to decide what is "best output" to be displayed - and the point of this issue is that the `placeholder` mechanism (being web build fcoussed) does not give a way to provide a high quality hard copy version of the figure for print (so I think your comment is prob. a bit off topic here).

[...]

PR #1701, opened 2024-12-18
🪄 Updates for multi-article typst export with interactive figures

fwkoch:
I was working on building this PDF: https://github.com/msa-em/nanocartography/pull/7 - It is a multi-article, typst export with many interactive figures and complex equations.

This triggered some updates to `tex-to-typst`, mainly matrices, brackets, arrows. See https://github.com/continuous-foundation/tex-to-typst/compare/f4ca4533aea3af2fc33e0e7d59a31fa0e010b3f5...v0.0.12

Additionally, it encouraged some significant updates here in MyST:

- `static` images to use _only_ in hard-copies, different than `placeholders` which modify the web experience - this addresses an existing issue: https://github.com/jupyter-book/mystmd/issues/1157
- usage of project-level parts in exports defined along-side them (i.e. parts and exports defined in `myst.yml#project`)


packages/myst-directives/src/figure.ts
```diff
@@ -56,6 +56,10 @@ export const figureDirective: DirectiveSpec = {
       type: String,
       doc: 'A placeholder image when using a notebook cell as the figure contents. This will be shown in place of the Jupyter output until an execution environment is attached. It will also be used in static outputs, such as a PDF output.',
     },
+    static: {
```
fwkoch, line 59: I suspect we do not want to call this `static`. @stevejpurves suggested `hard-copy` in #1157 . We could also do something like `placeholder:pdf`, but really, this applies to all "paper" exports: docx, tex, typst, and pdf. `placeholder:paper`? Not sure how confusing `placeholder` vs `placeholder:pdf` might be.

packages/myst-to-typst/src/container.ts
```diff
@@ -104,6 +104,11 @@ export const containerHandler: Handler = (node, state) => {
     return;
   }
 
+  if (node.enumerator?.endsWith('.1')) {
```
fwkoch, line 107: This could be improved. It places strict assumptions on the `enumerator` in `numbering`. e.g. It works great for:
```
numbering:
  enumerator: 3.%s
```

But doesn't work at all for:
```
numbering:
  enumerator: (3-%s)
```

fwkoch: Possibly we need better translation between myst and typst enumerators.

packages/myst-to-typst/src/utils.ts
```diff
@@ -49,8 +49,8 @@ const textOnlyReplacements: Record<string, string> = {
   '©': '#emoji.copyright ',
   '®': '#emoji.reg ',
   '™': '#emoji.tm ',
-  '<': '\\< ',
-  '>': '\\> ',
+  '<': '\\<',
```
fwkoch, line 52: Making this change did not feel good - I'm not sure why we were previously adding a space after these characters. For the examples I was looking at, the addition of this space looked very wrong.

@rowanc1 perhaps you have some recollection of this...?

packages/myst-directives/src/figure.ts
agoose77: Can we block this PR until we resolve https://github.com/jupyter-book/mystmd/issues/1157#issuecomment-2580015766?

rowanc1: Lets break this into ~6 prs and get the ones in that are simple and make this one smaller.

fwkoch:
This PR is superseded by the 6 PRs linked above. Most are small and simple. We can continue discussion about `static` placeholder in #1744

PR #1721, opened 2025-01-06
⌨️ Do not export code block options when undefined

rowanc1:


rowanc1:
This was changing simple code blocks into `{code}` directives. This fixes that!

PR #1725, opened 2025-01-06
🖼️ Read iframe options from html

rowanc1:


packages/myst-transforms/src/html.ts
```diff
@@ -124,6 +124,13 @@ const defaultHtmlToMdastOptions: Record<keyof HtmlTransformOptions, any> = {
       if (node.properties.alt) attrs.alt = node.properties.alt;
       return h(node, 'image', attrs);
     },
+    iframe(h: H, node: any) {
+      const attrs = addClassAndIdentifier(node);
+      attrs.url = String(node.properties.src || '');
+      if (node.properties.src) attrs.src = node.properties.src;
```
rowanc1, line 130: Hm, I think this is a mistake. Should only be one!

PR #1726, opened 2025-01-06
⌨️ Code block options can have caption

rowanc1:
We already allow this for options defined with the `#| caption:` syntax, this expands to being allowed in the code-block options as well.

### DIFF
--- a/.changeset/sixty-swans-pull.md
+++ b/.changeset/sixty-swans-pull.md
@@ -1,0 +1,6 @@
+---
+"myst-directives": patch
+"myst-transforms": patch
+---
+
+Add caption to code-block options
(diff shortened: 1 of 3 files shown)


choldgraf (commented):
This looks great! Though I am also confused as isn't this already possible? Here's a screenshot that seems to show a code block with a caption:

https://mystmd.org/guide/code#code-blocks

CleanShot 2025-01-06 at 08 51 52@2x

rowanc1:
This is code-cell! Not code.

rowanc1:
Yes, docs autoupdate for all directive options.

I have added additional docs about the caption.

The mystmd docs should **also** be stand-alone from Jupyter Book (i.e. a user coming to mystmd.org can get up and running with the myst cli with the relevant tutorials, without having to know about jupyter book), so there still needs to be **some** guides and tutorials in the mystmd docs, just more of a focus on reference material over time. This will likely become more clear as JBv2 gets released fully and has more docs focused on building books and how-tos etc.

PR #1727, opened 2025-01-06
🌁 Join gates plugin before blocks

rowanc1:


fwkoch (approved):
This change makes sense - keeping the gate joining closer to author's intent.

Issue #1704, opened 2024-12-19
Dollar math equations incorrectly do not continue the same paragraph after if first $$ has a newline after it


dellaert:
Still some issues with display equations

myst v.1.3.9

## Description

 It seems

```
Paragraph
$$Ax=b$$
remainder of paragraph
```
does work, but

```
Paragraph
$$
Ax=b
$$
remainder of paragraph
```

does not render properly without a newline :-/

FreekPols:
With line before and after $$?

dellaert:
I updated the description. The idea was that omitting the line before and after would prevent starting a new paragraph in latex. This (now, after @rowanc1 PR) works fine if I use

```
Paragraph
\begin{equation}
Ax=b
\end{equation}
remainder of paragraph
```

or 

```
Paragraph
$$Ax=b$$
remainder of paragraph
```

bot not if I use

```
Paragraph
$$
Ax=b
$$
remainder of paragraph
```

choldgraf:
Hmm I can't replicate this on the sandbox. It seems to behave the same in each case here:

https://mystmd.org/sandbox?tab=demo&myst=UABhAHIAYQBnAHIAYQBwAGgACgBcAGIAZQBnAGkAbgB7AGUAcQB1AGEAdABpAG8AbgB9AAoAQQB4AD0AYgAKAFwAZQBuAGQAewBlAHEAdQBhAHQAaQBvAG4AfQAKAHIAZQBtAGEAaQBuAGQAZQByACAAbwBmACAAcABhAHIAYQBnAHIAYQBwAGgACgAuAAoACgBQAGEAcgBhAGcAcgBhAHAAaAAKACQAJABBAHgAPQBiACQAJAAKAHIAZQBtAGEAaQBuAGQAZQByACAAbwBmACAAcABhAHIAYQBnAHIAYQBwAGgACgAKAEIAdQB0ACAAbgBvAHQACgAKAFAAYQByAGEAZwByAGEAcABoAAoAJAAkAAoAQQB4AD0AYgAKACQAJAAKAHIAZQBtAGEAaQBuAGQAZQByACAAbwBmACAAcABhAHIAYQBnAHIAYQBwAGgA

dellaert:
Thanks for checking @choldgraf ! I just ran the same experiment locally, and can confirm the same behavior for latex. It seems the extra newline after \begin{equation} in the third case messes up rendering the latex to pdf:

[...]

PR #1728, opened 2025-01-06
🧮 Trim all math to ensure no new lines in latex output

rowanc1:
See #1704

### DIFF
--- a/packages/mystmd/tests/basic-tex-math/input.md
+++ b/packages/mystmd/tests/basic-tex-math/input.md
@@ -4,6 +4,12 @@ Paragraph
 $$Ax=b$$
 Paragraph
 
+Paragraph around dollar math
+$$
+Ax=b
+$$
+Paragraph
+
 Paragraph
 \begin{equation}
 Ax=b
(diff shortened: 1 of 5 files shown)


packages/mystmd/tests/basic-tex-math/input.md
```diff
@@ -4,6 +4,12 @@ Paragraph
 $$Ax=b$$
 Paragraph
 
+Paragraph around dollar math
+$$
+Ax=b
+$$
+Paragraph
+
```
rowanc1, line 12: @dellaert can you confirm that this is the output that you would expect (see the outcome right below).

packages/mystmd/tests/outputs/basic-tex-math.tex
```diff
@@ -4,6 +4,12 @@
 \end{equation}
 Paragraph
 
+Paragraph around dollar math
+\begin{equation}
+Ax=b
+\end{equation}
+Paragraph
+
```
rowanc1, line 12: This is the result in latex (no more new line **inside** the equation).

packages/mystmd/tests/basic-tex-math/input.md
dellaert: Yep!

dellaert (approved):
Thanks !!!

PR #1734, opened 2025-01-08
⏬ Add mechanism for downgrading MyST AST

### LINKED ISSUE #1724: Define a policy for XRef version incompatibility
## Problem
A well-defined problem in distributing software is the need to handle forward compatibility, whereby "future" input is encountered by legacy application versions. Conventionally this is hard, meaning that people try to avoid making breaking changes.

Meanwhile, we _know_ that as a young ecosystem we likely have breaking changes coming down the wire (such as changing the output format). Right now, our best strategy is to anticipate these changes in the application  before writing updated outputs.

### Old `mystmd`, new XRef
Consider site A wishes to pull in an XRef from site B that was built by a newer version of `mystmd` than the binary that builds A. At present, if B includes a breaking AST change, the build will fail. 

Users can choose to mitigate this by upgrading their binary of `mystmd`. However, this only protects the builds of site A whilst site B is not re-deployed; down the road, if the version of mystmd that builds site A is pinned, the build will fail once again.

### Old `myst-theme`, new XRef
In deployed sites, we use `fetch` to load the latest version of the xref. This allows for hover-previews to reflect the latest state of the linked resource. By pulling in new foreign AST, we might pull in a breaking change. This means that in the likely scenario that site A is rarely deployed, it can still be broken by updates to site B.

## Solutions
Right now, there are no guardrails against legacy MyST applications pulling in incompatible XRef data. We have two options:

### Reject (or error) xrefs that are "too new"
Here, we would effectively need to "safely break" xrefs that pull in new data. For theme deployments, we could show a warning message in the hover previews, whilst for `mystmd` builds we could show a console warning.

This is a very strict approach -- **anything new is considered breaking.**

### Declare the known-degradable xref versions in published JSON
It might be the case that version 2 consumers can actually safely read versions 3, 4, and 5. Publishers of these newer versions will know that, and could store that information in the output e.g.
```yaml
version: 5
supported_consumers:
  - 2
  - 3
  - 4
  - 5
```

This would allow us to accurately version changes to our schema, but ultimately cannot handle the case that `supported_consumers` doesn't include the current consumer version.

### Attempt to downgrade xrefs that are "too new"
We can publish a small bundle of code to a public repository that downgrades new XRefs to older versions. Then, both `mystmd` and `myst-theme` can be instructed to pull ESM from this repository that is able to convert new XRefs into something the existing application can handle.

This raises concerns about security that we would need to robustly address.

agoose77:
As outlined in #1724, as the MyST user-base grows, we are increasingly likely to find ourselves in a situation whereby we feel reluctant to make breaking changes to our AST due to the impact upon the wider ecosystem of published knowledge. 

We already know of one breaking change (e.g. #1671), and it is not unlikely that there will be more in coming releases.

This PR introduces a mechanism for legacy `mystmd` applications to handle newer AST through a downgrade process. We would also need to use the same approach for `myst-theme`. Ideally, these will use the same downgrade bundles (with different build targets if needs be).

Closes #1724

### DIFF
--- a/packages/myst-cli/src/transforms/crossReferences.ts
+++ b/packages/myst-cli/src/transforms/crossReferences.ts
@@ -32,10 +32,6 @@ export type MystData = {
   references?: References;
 };
 
-async function maybeDowngradeMystData(rawData: MystData): Promise<MystData> {
-  return rawData;
-}
-
 async function fetchMystData(
   session: ISession,
   dataUrl: string | undefined,
@@ -52,8 +48,7 @@ async function fetchMystData(
     try {
       const resp = await session.fetch(dataUrl);
       if (resp.ok) {
-        const rawData = (await resp.json()) as MystData;
-        const data = await maybeDowngradeMystData(rawData);
+        const data = (await resp.json()) as MystData;
         writeToCache(session, filename, JSON.stringify(data));
         return data;
       }
@@ -84,7 +79,18 @@ export async function fetchMystXRefData(session: ISession, node: CrossReference,
   if (node.remoteBaseUrl && node.dataUrl) {
     dataUrl = `${node.remoteBaseUrl}${node.dataUrl}`;
   }
-  return fetchMystData(session, dataUrl, node.urlSource, vfile);
+  const rawData = await fetchMystData(session, dataUrl, node.urlSource, vfile);
+  let data: MystData | undefined;
+  if (node.remoteBaseUrl) {
+    // TODO
+  } else {
+    fileWarn(
+      vfile,
+      `Unable to determine XRef AST version for external MyST reference: ${node.urlSource}`,
+    );
+    data = rawData;
+  }
+  return data;
 }
 
 export function nodesFromMystXRefData(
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/transforms/crossReferences.ts
```diff
@@ -74,12 +79,93 @@ export async function fetchMystLinkData(session: ISession, node: Link, vfile: VF
...
+  data: any,
+): Promise<any> {
+  const downgradeCachePath = `myst-downgrade-${fromVersion}-${toVersion}.mjs`;
+  if (
+    !checkCache(session, downgradeCachePath, {
+      maxAge: 7, // days
+    })
+  ) {
+    try {
+      const response = await fetch(`http://localhost:9000/${downgradeCachePath}`);
```
rowanc1, line 102: What are your thoughts on just treating this as a normal npm package that we install in the `_build` directory and check for updates (in the same way we do for mystmd now)?

Then there would be a version check against npm for that package, an npm install (already doing that for the themes).

That would mean we don't have to host these, we have versions, publishing, deprecations, etc. AND other tools can use this in a normal way - it is just a package.

PR #1738, opened 2025-01-09
🎭 Support more characters in typst export

fwkoch:
(This is a small part of the changes originally in #1701)


packages/myst-to-typst/src/utils.ts
```diff
@@ -49,8 +49,8 @@ const textOnlyReplacements: Record<string, string> = {
   '©': '#emoji.copyright ',
   '®': '#emoji.reg ',
   '™': '#emoji.tm ',
-  '<': '\\< ',
-  '>': '\\> ',
+  '<': '\\<',
+  '>': '\\>',
```
fwkoch, line 53: Making this change did not feel good - I'm not sure why we were previously adding a space after these characters. However, for the examples I was looking at, the addition of this space looked very wrong, and simply removing it here fixed the output.

@rowanc1 perhaps you have some recollection of this...?

rowanc1: Lets go with an oversight/bug for now.

rowanc1 (approved):
Looks good.

PR #1739, opened 2025-01-09
🎆 Prevent html outputs that translate to empty images

fwkoch:
If `output` has `html` media type that includes `image` node with no `url`, that output is entirely ignored. This approach is a little heavy-handed, but likely if we encounter an image with no url, the html is too complex to convert to markdown at all.


rowanc1 (approved):
Ran into this in the QuantEcon books, thanks @fwkoch.

PR #1741, opened 2025-01-09
🖐 Support restarting counter for each typst article in multi-article export

fwkoch:
For the most part, typst handles its own numbering. This PR forces typst to reset numbering where myst resets numbering. However, the way this PR determines where numbering resets is by looking at the `enumerator` - if this value is customized with `template` to not look like `#.#.#`, this counter reset will not happen...

### DIFF
--- a/packages/myst-to-typst/src/container.ts
+++ b/packages/myst-to-typst/src/container.ts
@@ -104,6 +104,10 @@ export const containerHandler: Handler = (node, state) => {
     return;
   }
 
+  // This resets the typst counter to match MyST numbering.
+  // However, it is dependent on the resolved enumerator value. This will work given
+  // default enumerators, but if the user sets numbering 'template' it will not work.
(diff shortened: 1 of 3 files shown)


packages/myst-to-typst/src/container.ts
```diff
@@ -104,6 +104,11 @@ export const containerHandler: Handler = (node, state) => {
     return;
   }
 
+  if (node.enumerator?.endsWith('.1')) {
```
fwkoch, line 107: There should be a better way to do this... But it's not simple, since `myst-to-typst` is taking a fully-processed tree, it's lost some of the intermediate knowledge of _how_ things are counted, and just has the final results.

Doing something like stashing the entire `numbering` object on each enumerated node could work, but that feels yucky. Maybe there is a happy medium?

But for now, I think the existing shortcut in this PR is ok. It handles default enumerators.

rowanc1: Can you add a comment to this line with that?

PR #1742, opened 2025-01-09
🔧 Fix project cross-references in typst

fwkoch:
This prevents cross-references to other pages in typst from being written as (invalid) links. They now remain as (correct) cross-references.


packages/myst-to-typst/src/index.ts
```diff
@@ -363,10 +363,10 @@ const handlers: Record<string, Handler> = {
   legend: captionHandler,
   captionNumber: () => undefined,
   crossReference(node: CrossReference, state, parent) {
-    if (node.remote) {
```
fwkoch, line 366: xrefs to other pages are still considered `remote` - we only want to do this if there is a `remoteBaseUrl`, i.e. they are xrefs to a totally separate project.

PR #1744, opened 2025-01-09
🎆 Add static figure placeholder for images

fwkoch:
This is the subset of the changes from https://github.com/jupyter-book/mystmd/pull/1701 which focuses on the `static` image field, an alternative to `placeholder` that will only be used in static exports, e.g. PDF. There is a little more discussion in that previous PR.

### DIFF
--- a/.changeset/kind-lies-kick.md
+++ b/.changeset/kind-lies-kick.md
@@ -1,0 +1,7 @@
+---
+'myst-directives': patch
+'myst-transforms': patch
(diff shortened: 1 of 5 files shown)


rowanc1:
@fwkoch and I were talking about this in the context of another feature: `only` and I think that is both more flexible and hits this use case.

Let's explore that (which is necessary for supporting some of the books we are working on), and see if that replaces the need here.

PR #1746, opened 2025-01-09
🚮 Remove broken and unused reactive roles

rowanc1:
The reactive roles were an early prototype we were doing at curvenote. These shouldn't be included in the myst-cli as they don't work and have no renderer.

This removes them from the docs and default directive/roles.


packages/myst-cli/package.json
```diff
@@ -89,8 +89,8 @@
     "myst-to-md": "^1.0.14",
     "myst-to-tex": "^1.0.40",
     "myst-to-typst": "^0.0.28",
-    "myst-transforms": "^1.3.28",
     "myst-toc": "^0.1.2",
+    "myst-transforms": "^1.3.28",
     "nanoid": "^4.0.0",
```
rowanc1, line 94: No change here, just ordering.

fwkoch (approved):
Just cleaning up legacy, unused stuff!

PR #1750, opened 2025-01-09
🙈 Notebook Cell Tags Documentation

### LINKED ISSUE #1108: Control code cell visibility and behavior with tags and metadata
Currently, all code cells that are parsed by MyST will include their input and outputs together.

Outputs can be inserted into the page via the `{embed}` directive, which allows you to remove the input, output, or include both.

However it's not possible to modify the behavior of the code cell itself, which makes it clunky if you want the code cell to exist on the same page where you want to embed its output.

It would be helpful if there were a way to control the visibility of the cell input and cell output from within a code cell.

### How MyST-NB handles this

MyST NB accomplishes this by [using Cell Metadata tags](https://myst-nb.readthedocs.io/en/latest/configuration.html#cell-tags), with tags like the following:

- `remove-cell`
- `remove-input`
- `remove-output`
- `hide-input` (hides behind a toggle)
- `hide-output`
- `skip-execution`
- `raises-exception` (to prevent errors from stopping the build process)

rowanc1:
See also:

- jupyter-book/myst-theme#517
- Fixes #1108

### DIFF
--- a/packages/myst-execute/src/execute.ts
+++ b/packages/myst-execute/src/execute.ts
@@ -5,7 +5,7 @@ import type { Kernel, KernelMessage, Session, SessionManager } from '@jupyterlab
 import type { Block, Code, InlineExpression, Output } from 'myst-spec-ext';
 import type { IOutput } from '@jupyterlab/nbformat';
 import type { GenericNode, GenericParent, IExpressionResult, IExpressionError } from 'myst-common';
-import { NotebookCell, fileError } from 'myst-common';
+import { NotebookCell, NotebookCellTags, fileError } from 'myst-common';
 import type { VFile } from 'vfile';
 import path from 'node:path';
 import assert from 'node:assert';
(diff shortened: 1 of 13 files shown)


packages/myst-execute/src/execute.ts
```diff
@@ -175,15 +175,15 @@ function isCellBlock(node: GenericNode): node is CodeBlock {
...
+  return !!node.data?.tags?.includes?.(NotebookCellTags.raisesException);
 }
 /**
  * Return true if the given code block should not be executed
  *
  * @param node block to test
  */
 function codeBlockSkipsExecution(node: CodeBlock) {
-  return !!node.data?.tags?.includes?.('skip-execution');
+  return !!node.data?.tags?.includes?.(NotebookCellTags.skipExecution);
```
agoose77, line 186: Much better!

rowanc1: Small little improvements. I was in there looking to implement this and was surprised it was already done!

PR #1751, opened 2025-01-10
🚀 Always launch a new Jupyter Server

### LINKED ISSUE #1716: Executing using active kernels anywhere on the computer is confusing and hard to debug
When executive a MySTMD document with `--execute`, MyST will search for any active kernel and use it rather than spinning up a new kernel with the local environment. This happens more-or-less silently.

It is really confusing when you're trying to run Jupyter Book in an isolated environment. For example, say you want to build a book with `nox` using an environment that's installed locally. But, you've got a Jupyter kernel running _anywhere else on your machine_. When the book executes, it won't spin up a kernel for the isolated environment, it'll just re-use the existing kernel and whatever things are installed there.

_Note: I think we've discussed this before but I couldn't find an issue, so please close and re-direct if there's a better place_.

rowanc1:
Fixes #1716

### DIFF
--- a/.changeset/strange-baboons-build.md
+++ b/.changeset/strange-baboons-build.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Do not findExistingJupyterServer when using execute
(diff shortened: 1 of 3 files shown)


choldgraf:
I feel like I've been given like 45 minutes at least of my future life back

Issue #1760, opened 2025-01-11
Add a helpful message box to the CLI when there is a new theme version available


choldgraf:
I regularly get burned because the documentation for mystmd relates to a newer theme than whatever exists in the `_build` folder. It would be helpful if there were a way to really clearly tell users "you're using an outdated theme, here's how to upgrade".

The message that mystmd itself sends is quite helpful and noticeable, would it be possible to check for new theme versions and surface a similar message?

![Image](https://github.com/user-attachments/assets/f40c7d21-a605-4ab1-8448-ab3e78806bab)

rowanc1:
@fwkoch wrote a bit about this here:

- https://github.com/jupyter-book/mystmd/issues/854

PR #1761, opened 2025-01-11
🦋 Change from twitter link to bsky link

rowanc1:
@choldgraf can you give a thumbs up?

image

Reminded from #1760


choldgraf (approved):
good call 👍

Issue #1654, opened 2024-11-19
Add a button for call to action and noticeable links


choldgraf:
Sometimes it's useful for authors to put a button on a page to draw more attention to it. For example, a "Get started" button on a documentation landing page, or a "Donate" button to direct users to a donation page.

Currently the closest thing you can get is manually defining an HTML element like:

````html
<a href="/start">
  <div style="border-radius: 1em; background-color: #e07330; width: 8em; height: 3em; margin: .5em auto; color: white;">
  Get started
  </div>
</a>
````

However this has a few shortcomings:

- It is verbose and depends on enough CSS knowledge to hand-write the rules.
- style works with `<div>` but not with `<a>` so you can't style the text inside the button.

It would be much easier if we either:

1. Supported native button syntax, like:

   ```
   :::{button} Button text
   :style: some ways to style the button, center it, etc
   :::
   ```
2. Fixed #1617 so that you could style the `<div>` as you wished with classes from tailwind (and make this work for `<a>` as well)

choldgraf:
We discussed this in the Jupyter Book team meeting today. There are two PRs that implement this:

- https://github.com/jupyter-book/mystmd/pull/1773
- https://github.com/jupyter-book/myst-theme/pull/521

The main complexity that we'd run into was expressing buttons as a `{role}` but without the ability to define classes etc.

I believe that we decided to just move forward with an MVP approach, to add functionality to support `xref` links via the button link, and to leave things like styling etc to future improvements.

PR #1773, opened 2025-01-13
🖲 Add `button` role

jnywong:
Ref: https://github.com/2i2c-org/infrastructure/issues/5346

Co-authored-by: Angus Hollands <angus@oieltd.com>

> This PR is a spike for a `button` role. Sphinx-design [has button directives](https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html), but these feel under-constrained for `mystmd`. Furthermore, buttons feel like "special links", and links are inline elements. As a consequence, @jnywong and I decided to implement a button as a special `kind` of link. We also added a `class` attribute to the `Link` node. This will enable primary-secondary styling of buttons in future. This is probably the way that we'll go for all node types — add `class` and maybe `label` etc. to every node.
> With a recent discussion for adding inline-attributes to roles (``{button .secondary}`My button <https://>.` ``), this might become a nice UX. This may also be relevant if we support inline attribute annotation of Markdown syntax, e.g. `{kind=button}[a button](https://...)`
> To support sphinx-design, we might consider "upgrading" button directives into block-divs containing ``{button}`...` `` in future.
> -- @agoose77

### References

- Related PR: https://github.com/jupyter-book/myst-theme/pull/521
- closes https://github.com/jupyter-book/mystmd/issues/1654

### TODO
- [ ] Update myst spec!


packages/myst-directives/src/button.ts
```diff
@@ -0,0 +1,37 @@
...
+        required: false,
+  },
+  run(data: DirectiveData): GenericNode[] {
+    const children: GenericNode[] = [];
+    if (data.body) {
+      children.push(...data.body as GenericNode[]);
+    }
+    const node: Link = {
+      type: 'link',
+      kind: 'button',
```
agoose77, line 28: Buttons feel inherently link-like, i.e. they are not _new_ node types, but more of a styling augmentation.

packages/myst-ext-button/CHANGELOG.md
```diff
@@ -0,0 +1,125 @@
+# myst-ext-card
```
rowanc1, line 1: Can delete this file, it will get recreated in the next release!

packages/myst-ext-button/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+        required: true,
+  },
+  run(data: RoleData): GenericNode[] {
+    const body = data.body as string;
+    const match = REF_PATTERN.exec(body);
+    const [, modified, rawLabel] = match ?? [];
+    const url = rawLabel ?? body;
+    const node: Link = {
+      type: 'link',
+      kind: 'button',
```
rowanc1, line 21: Nice - this means there is a good fallback.

rowanc1 (commented):
Thanks for the PR (taking a quick look while in draft!). I am curious if there is going to be conceptually more to `myst-ext-button` package in the future? In which case, maybe expanding the package name to cover it?

jnywong:
Hey @rowanc1 ! Yeah @agoose77 and I were prototyping today and made a lot of pivots!

> I am curious if there is going to be conceptually more to myst-ext-button package in the future? In which case, maybe expanding the package name to cover it?

We very quickly found there's a myriad ways authors may want to style links, and initially borrowed some inspiration from the existing `card` component. For this PR, we're going to be focused on a button package specifically.

rowanc1:
Makes sense, forgot about the card package as an analogy.

choldgraf (commented):
Can you provide short user-facing documentation for how this would be used? That would make it easier to understand the implications of role vs. directive etc.

The most common use-case I can imagine for a button is in situations like this:

````md
Some paragraph with a call to action.

[Click me button]

And some more paragraph text.
````

It sounds like this implementation would be accomplished with the following, is that right?

````md
Some paragraph with a call to action.

{button}`Some call to action text`

And some more paragraph text.
````

In the context of landing pages you might see one or more buttons next to each other, e.g.


````md
# Title

![A wide hero-style image]()

[Click me button] [Another button]
````

In this case, is it correct that accomplishing these would be accomplished with:


````md
# Title

![A wide hero-style image]()

{button}`Button text 1` {button}`Button text 2`
````

agoose77:
I can engage on a few of these right now (though I will circle back for a lengthier reply).

Your examples are how we envisaged using a button — that's the benefit of a role vs directive.[^rvd]. 

> E.g. do you imagine users will be doing things like writing some text and {button}`Click Me` putting an inline button like that?

Yes, in fact that was one of the things I wanted to do for the AGU demo gallery, but couldn't get the styling right with the constraints at the time.

> Finally, just noting that one of the useful features of sphinx-design was that buttons could link to more than just URLs. 

Yes, this is a useful feature. 

We borrowed the `label <reference>` Sphinx syntax (used in `download`, for example) to allow separating the label from the URL. Therefore, the following would be the proposed syntax for linking to documents and/or identifiers:
````markdown
{button}`Click me <#identifier>`
{button}`#identifier`
{button}`Click me <./document.md>`
{button}`./document.md`
````

To implement this, we also need to extend `crossReference` with a `kind`, just like `link`.

An uncertainty that I have is whether the existence of two nodes that need "button-like" features implies the need to have
````json
{
  "type": "button",
  "children": [
    {
      "type": "link",
      "url": "..."
    }
  ]
}
````

instead of 
````json
{
  "type": "link",
  "kind": "button",
  "url": "..."
}
  

````

However, I think in the long run I *might* prefer using annotations e.g.
````markdown
{kind=button}[Click me](#identifier)
{kind=button}[](#identifier)
{kind=button}[Click me](./document.md)
{kind=button}[](./document.md)
````

[^rvd]: For completeness, there's nothing in mystmd that precludes us from having directives produce inline node types. But we tend to consider directives as "block like" rather than "inline like".

choldgraf (approved):
This and https://github.com/jupyter-book/myst-theme/pull/521 look really nice to me. It sounds like we might want to change the syntax of how buttons are defined (e.g. by attaching them via classes), but that this implementation wouldn't close any of those doors. I'm +1 on merging this in so that we can see how this feature gets used and get a feel for it ourselves.

My only request is that we add a short snippet of docs, perhaps just a short section to https://mystmd.org/guide/dropdowns-cards-and-tabs ? I don't think we strictly need to block on that though because the role will be auto-documented if I understand properly.

rowanc1 (approved):
Looks good, feel free to merge whenever @agoose77 and @jnywong.

agoose77:
I'm rushing off at the end of the day to get to a car dealership in time! I will circle back to  this to get it merged on Monday (or sooner if possible), but I am 100% confortable with anyone else merging it if they give it a quick once over!

agoose77:
@rowanc1 quick sense check -- are we intentionally not documenting the extension directives (like cards, tabs) in the main reference (https://deploy-preview-1773--mystmd.netlify.app/directives) in favour of in-line in the extension pages?

rowanc1:
No, I think that is an oversight.

rowanc1:
If they are documented in both places, likely because we are thinking of extensions as separable. Either way though, as long as that exists somewhere.

agoose77:
I've touched-up the docs here, and it looks ready to merge!

The formatting on the embedded directive definitions is not ideal (the header depth is non-parametrised at the moment), but we can easily fix that down the road.

agoose77:
Thank you @jnywong for a FAB first (?) `mystmd` core feature!! :sparkles: 

@rowanc1 & @choldgraf  hank you for the extra help and review :robot:

jnywong:
Woohoo! And thank you @agoose77 for getting this over the line 🏁

mathieuboudreau:
Just FYI, the website examples for the buttons are broken: https://mystmd.org/guide/dropdowns-cards-and-tabs#buttons

Screenshot 2025-02-12 at 10 20 59 AM

jnywong:
Oh interesting! Thanks for letting us know @mathieuboudreau 👍

jnywong:
Looks like we just need to wait for a theme release for this to sort itself out :)

rowanc1:
This will need an explicit tick of the dependencies here as mystmd.org is on a custom theme:

https://github.com/jupyter-book/mystmd.org

jnywong:
Ohhh okay! I can create an issue there

jnywong:
https://github.com/jupyter-book/mystmd.org/issues/22

rowanc1:
Fixed.

image

Issue #1336, opened 2024-06-19
Tracking: Turing Way


rowanc1:
We had a wonderful conversation this morning with the TuringWay team on how best to support the TuringWay team in their adoption of JupyterBook v2, which will be using `mystmd`. There were a lot of exciting possibilities that we talked about in terms of creating an ecosystem of content that the TuringWay could expose (figure, equations, glossaries), and that be included, linked to and cross-referenced by other books, tutorials, and communities in the wider open-science ecosystem. There are also some current deficiencies in MyST MD at the moment, one of the core ones is around internationalization. We got a really good demo from @BatoolMM on how that process is working, and some of the sticking points on how to bring content together to have a unified experience between languages. The localization also must include right-to-left languages, which will require a number of style updates and improvements across the components.

# Resources for translations:

- https://turingway-arabic.netlify.app/welcome.html
- https://github.com/TWTranslation/the-turing-way/tree/main/book/translation

# Example of Turing Way content deployed in MyST

- https://rowanc1-turingway.curve.space/  (this is deployed using Curvenote)

# Some cool things

![image](https://github.com/executablebooks/mystmd/assets/913249/8a5fb0a3-98dc-49a6-b8cc-3d6d692b18b6)

![turing-way-hover](https://github.com/executablebooks/mystmd/assets/913249/6590d090-1eed-48ff-98e4-aea70edfc80e)

# Updates that need to happen

- Glossary #618
- Improve the upgrade process #1223 (in progress)
- Internationalization and localization (this is a big task that we need to break down, what is necessary is a lot clearer to me notw after @BatooIMM's demo) #166
- Host multiple languages in a single theme
- PR: fix admonitions with `Note` --> `note` (that can be upstreamed right away)
- Way to turn off `@username` references on a page (e.g. changelogs) or have those fallback to a configured list of IDs and improve, e.g. our GitHub person cards.
- Discuss and demo different ways to build out the infrastructure for the book (and tradeoffs between the different approaches)
- Idea (curious on your take): #1305
- Improved header / footer information

There are probably a number of other things that will come out as we work towards these goals, I encourage members of the TuringWay team to jump in here and let us know how we can support you. Please try the preview above and get a sense of where we are at. One of the cool wins is that the performance of navigating between pages should be improved quite a bit. There are also some abbreviations that will work and increase accessibility etc. for various pages.

[...]

PR #1793, opened 2025-01-20
🔄 Fetch with Retry for HTML build

### LINKED ISSUE #1775: Export to html often fails when trying to serve many pages
I can build the site, but export to html often fails and in a stochastic way.

v1.3.20

See attachments for log files.

[log5.log](https://github.com/user-attachments/files/18398049/log5.log)
[log6.log](https://github.com/user-attachments/files/18398047/log6.log)
[log7.log](https://github.com/user-attachments/files/18398048/log7.log)

## Description

As my site has grown, the conversion to html has started to fail often. I get ECONNRESET or EPIPE errors after the app begins making requests to the web server to pull html pages. If I run `myst build --execute --html` 20 times, it might succeed once or twice.

## Proposed solution

My best guess is that the server is becoming overwhelmed with what amounts to a DOS attack. I tried setting `NODE_OPTIONS="--max-old-space-size=4096"`, but it did not seem to help.

Maybe the exporter can slow down page requests a little, or buffer them?

rowanc1:
Limits outgoing connections to five.

Fixes #1775
See also #1336

### DIFF
--- a/packages/myst-cli/src/build/html/index.ts
+++ b/packages/myst-cli/src/build/html/index.ts
@@ -9,6 +9,10 @@ import type { StartOptions } from '../site/start.js';
 import { startServer } from '../site/start.js';
 import { getSiteTemplate } from '../site/template.js';
 import { slugToUrl } from 'myst-common';
+import pLimit from 'p-limit';
+import { fetchWithRetry } from '../../utils/fetchWithRetry.js';
+
+const limitConnections = pLimit(5);
 
 export async function currentSiteRoutes(
   session: ISession,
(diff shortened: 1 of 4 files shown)


packages/myst-cli/src/build/html/index.ts
```diff
@@ -141,25 +145,27 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
...
-          fileWriteStream.on('finish', resolve);
-        });
-      } else {
-        const content = await resp.text();
-        writeFileToFolder(path.join(htmlDir, route.path), content);
-      }
-    }),
+    routes.map(async (route) =>
+      limitConnections(async () => {
+        const resp = await fetchWithRetry(session, route.url);
```
rowanc1, line 150: Main changes here, limiting connections to 5 and having a retry with exponential backoff.

PR #1798, opened 2025-01-22
✏️ Parse markdown outputs into AST

### LINKED ISSUE #1026: Parse markdown code-cell outputs as part of the document AST

agoose77:
This PR closes #1026, and implicitly tests #1671.

Depends upon #1671 

:warning: WIP

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15751,6 +15751,7 @@
         "chalk": "^5.2.0",
         "myst-cli-utils": "^2.0.11",
         "myst-common": "^1.7.6",
+        "nbtx": "^0.2.3",
         "node-fetch": "^3.3.0",
         "unist-util-select": "^4.0.3",
         "vfile": "^5.3.7",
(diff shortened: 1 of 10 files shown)


agoose77:
Closed as needs re-work on separate PR

PR #1799, opened 2025-01-22
🐛 Fix options key present with parts during JB 1 upgrade

### LINKED ISSUE #1796: `options` in Jupyter Book TOC fails upgrade
See Discord for context: https://discord.com/channels/1083088970059096114/1126886232060870678/1330929529652252704

bryanwweber:
- **fix: allow options key in toc with parts**

Closes #1796

### DIFF
--- a/packages/myst-cli/src/init/jupyter-book/toc.ts
+++ b/packages/myst-cli/src/init/jupyter-book/toc.ts
@@ -10,7 +10,7 @@ const TOCTreeOptions = z
     caption: z.string(),
     hidden: z.boolean(),
     maxdepth: z.number(),
-    numberted: z.boolean(),
+    numbered: z.boolean(),
     reversed: z.boolean(),
     titlesonly: z.boolean(),
   })
--- a/packages/myst-cli/src/init/jupyter-book/toc.ts
+++ b/packages/myst-cli/src/init/jupyter-book/toc.ts
@@ -149,6 +149,7 @@ const BookShorthandInnerSubtree: z.ZodType<BookShorthandInnerSubtreeType> = z.ob
 
 const BookHasOuterSubtrees = z.object({
   parts: BookOuterSubtree.array(),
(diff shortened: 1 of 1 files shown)


packages/myst-cli/src/init/jupyter-book/toc.ts
```diff
@@ -10,7 +10,7 @@ const TOCTreeOptions = z
     caption: z.string(),
     hidden: z.boolean(),
     maxdepth: z.number(),
-    numberted: z.boolean(),
+    numbered: z.boolean(),
```
rowanc1, line 13: 🚀

agoose77: Brilliant catch!

agoose77:
We definitely should loosen the strictness of TOC handling during upgrade. However, I think the proper solution is to do this in the validation step: https://github.com/jupyter-book/mystmd/blob/fad8f6730b6461cb70cafb1b2960f4a0dd13c6a7/packages/myst-cli/src/init/jupyter-book/toc.ts#L172-L176

I will circle around to this shortly.

bryanwweber:
> We definitely should loosen the strictness of TOC handling during upgrade. However, I think the proper solution is to do this in the validation step: https://github.com/jupyter-book/mystmd/blob/fad8f6730b6461cb70cafb1b2960f4a0dd13c6a7/packages/myst-cli/src/init/jupyter-book/toc.ts#L172-L176
> 
> I will circle around to this shortly.

Thanks! I'd written to Rowan on discord, I tried to find a way to get Zod to issue a warning, but it doesn't really support that: https://discord.com/channels/1083088970059096114/1177359316886503464/1331659677485891664

packages/myst-cli/src/init/jupyter-book/toc.ts
rowanc1: Merged here:
- #1839

PR #1802, opened 2025-01-23
⬆️ Add simple upgrade/downgrade package for MyST AST

agoose77:
This PR adds a simple package that sets an API for upgrading and downgrading an AST. Over time, we can use this to define imperative motions between AST schema versions.

For now, this PR implements a new "version 2" of the MyST schema which drops the use of `number` in favour of `enumerator` in footnotes.

## Next Steps
1. Hook this package into AST ingest routes in `mystmd`
2. Separately update this package to pull in newer-than-mystmd releases


rowanc1 (commented):
I think the upgrade downgrade should be on the output of what is written to json in the site.

```
{
   version: 1,
   sha: "",
   mdast: {},
   citations: {order:} ... etc,
}
```

That way we can execute and move around.

Currently the upgrade/downgrade is on the mdast only (likely the bulk of the changes, but citations were a poor choice to do outside of the tree as an example...!).

packages/myst-compat/src/upgrade/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Parent } from 'mdast';
+
+import { upgrade as upgrade1To2 } from './version_1_2.js';
+
+export function upgrade(from: string, to: string, ast: Parent): void {
+  if (from === '1' && to === '2') {
```
fwkoch, line 6: This means we need to be careful and explicit about bumping `version` in the myst metadata.

An alternative could be having tests against the content to determine if the upgrade is required (e.g. "if footnoteReference numbering is defined") - that could allow more flexibility to introduce compatibility updates that missed a version bump on the myst side.

I think the explicit version numbering is good, just musing on alternatives.

fwkoch (commented):
This is looking good to me - appreciate how it is simple and explicit.

packages/myst-compat/src/upgrade/index.ts
rowanc1: Maybe a slightly different thought - do we need/want a `minor_version` field here? Where we can have a bit more intentionality around what constitutes a major change (outputs) vs something that is pretty minor and can be dealt with with fallbacks in the theme, etc.

agoose77: @fwkoch right now we need to do some work to ensure that we're talking about the same "version" (being the schema version), certainly.

I have flipped between thinking about using a minor version vs major version. These kind of changes are hard to categorise. For example, whilst adding a new field to existing nodes should be considered non-breaking, adding new node types should be breaking (because downstream clients won't know what to do with them). But this case is one where the theme can probably just ignore unrecognised nodes and look at their children instead. 

The angle I've been working of late is to prefer handling compatibility explicitly by downgrading/upgrading, rather than implicitly encoding compatibility in the version number semantics.

Do either of you see a case where this might be a problem?

agoose77:
> I think the upgrade downgrade should be on the output of what is written to json in the site.
> 
> ```
> {
>    version: 1,
>    sha: "",
>    mdast: {},
>    citations: {order:} ... etc,
> }
> ```
> 
> That way we can execute and move around.
> 
> Currently the upgrade/downgrade is on the mdast only (likely the bulk of the changes, but citations were a poor choice to do outside of the tree as an example...!).

I've been toying with the same idea too. Doing that would require that "version" here refers to both the AST version and the version of the page data. We probably _can_ tie the two together.

packages/myst-compat/src/upgrade/index.ts
agoose77: @fwkoch yes, this reminds me why I originally scoped the package to the AST; we don't export a promise about our page types.

Here's my view of where we go.

1. At some point in future, we move to https://github.com/jupyter-book/myst-spec/pull/67 
2. We add types for `$.json` files to `myst-spec`
3. `myst-spec-ext` becomes a shim that exports from `myst-spec`
4. We define a single version number for `myst-spec` that changes whenever we make an AST or auxiliary change.
5. We _define_ a breaking change to be one that removes fields or changes types (in this case, "change" excludes "new fields").

This means that the MyST "spec" version is not _just_ the MyST AST. I think that is probably a good thing, because the AST and the auxiliary state are intimately connected. This is especially pertinent now that MyST can consume `.myst.json` files generated by other tools.

Consequently, we have the `myst.xref.json` and `myst.search.json` files which are _separately_ versioned.

@rowanc1 @fwkoch do you have any feelings about this approach?

rowanc1: We can not strongly type the interface. The file type will change over time, so that upgrade `File` type will also change (e.g. citations is in there now, and won't be in the future). The types **are** known after the version is known, and then we can cast to the right type, which can be `File_v1` in this package, or something.

For the options I would prefer them in a dict, so that we can add to it over time `{ throwError: false, inPlace: true }` etc. The upgrade options, should ideally be _optional_, and then the function should just upgrade to the latest.

```
upgrade(file);
upgrade(file, { from: 1, to: 15 });
```

The upgrade process should stamp the version into the file if it isn't there (where we are right now, which is `1`).

The mdast version and the file version should be the same, and I agree that `myst.xref.json` and `myst.search.json` have different version numbers.

agoose77: I think we can strongly type the interface; we already do that here for the AST. The idea is not to exhaustively type the AST, though; only the types that you actually need to work with during the upgrade step.

> The upgrade options, should ideally be optional, and then the function should just upgrade to the latest.

Yes, I think if we add the proper version field, we can do-away with the `from` version arg. We'll still want `to` if the version consuming the package is older than the latest, though.

packages/myst-compat/src/downgrade/index.ts
```diff
@@ -0,0 +1,17 @@
+import type { IFile } from '../types/index.js';
```
agoose77, line 1: The idea here is that we accept unvalidated inputs, and enforce strict types inside the upgrade/downgrade routines

packages/myst-compat/src/types/index.ts
```diff
@@ -0,0 +1,4 @@
+export interface IFile {
+  version: any;
+  mdast: any;
+}
```
agoose77, line 4: The generic "untyped input" interface.

packages/myst-compat/src/types/v1.ts
```diff
@@ -0,0 +1,22 @@
+// TODO: use mdast once we rely on mdast types
+export interface IFile {
+  version: '1';
+  mdast: any;
+}
```
agoose77, line 5: A "more typed" per-version interface.

I don't imagine that we'll actually type `mdast` — this would require bundling the entire MyST JSON spec with each iteration, which is probably overkill. Ultimately, we're still mostly going to need to just tell the type system that what we did was allowed. Let's not drown in types along the way.

packages/myst-compat/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import { downgrade } from './downgrade/index.js';
+import { upgrade } from './upgrade/index.js';
+import type { IFile } from './types/index.js';
+
+export function makeCompatible(src: IFile, to: string) {
```
rowanc1, line 5: Change this to `opts: {to: string }`?
```suggestion
export function migrate(src: IFile, to: string) {
```

packages/myst-compat/src/upgrade/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { IFile } from '../types/index.js';
+import { upgrade as upgrade1To2 } from './version_1_2.js';
+
+export function upgrade(src: IFile, to: string): void {
+  const { version: from } = src;
```
rowanc1, line 5:
```suggestion
  const { version: from = '1' } = src;
```

packages/myst-compat/src/types/v1.ts
rowanc1: Agree that we don't need to do strong typing here at all.

packages/myst-compat/src/index.ts
rowanc1: In meeting the version is required. Can add options as third arg.

rowanc1:
I will take a pass through this and get it merged @agoose77.

packages/myst-migrate/src/migrations.ts
```diff
@@ -0,0 +1,4 @@
+import type { Migration } from './types.js';
+import * as v1 from './v1_footnotes.js';
+
+export const MIGRATIONS: Migration[] = [v1];
```
rowanc1, line 4: This types each module for an upgrade/downgrade/description, and we then just put it in an array.

rowanc1:
I have merged a few of the files together, keeping the upgrade/downgrade/types all in one file. I have also added a description and ensured that this will work with chains of migrations (0-->5, etc.). In discussion with @agoose77 I have changed the spec to a number.

I also put the tests in their own folder and changed to `myst-migrate` as the name.

PR #1804, opened 2025-01-23
🔧 Add edit url to file frontmatter

fwkoch:
This is to enable an `Edit on Github` button in the theme. The `edit` url will be computed `github` is provided in the frontmatter and `git` is available (unless `edit` is explicitly set to `null`)

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -175,8 +175,8 @@ The following table lists the available frontmatter fields, a brief description 
 * - `github`
   - a valid GitHub URL or `owner/reponame`
   - page can override project
-* - `edit`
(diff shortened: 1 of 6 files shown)


docs/frontmatter.md
```diff
@@ -175,6 +175,9 @@ The following table lists the available frontmatter fields, a brief description
 * - `github`
   - a valid GitHub URL or `owner/reponame`
   - page can override project
+* - `edit`
```
rowanc1, line 178: `edit_url`?

rowanc1:
This looks good! Theme change can happen next, then we can further document.

PR #1812, opened 2025-01-24
💬 Preserve quotes in inline expressions upon request

agoose77:
Quotes are semantically important things, so we should provide a way for users to preserve them if necessary. This PR ensures that we only strip quotes if the user does not specify (through various channels) that they should not be stripped.

I have no strong feelings on whether this field should be `strip-quotes` or `keep-quotes`.

Example notebook:
````markdown
---
kernelspec:
  name: python3
  display_name: python3
---


```{code-cell} python3
from IPython.utils.capture import capture_output

with capture_output() as c1:
    display("hi", metadata={"strip-quotes": False})

with capture_output() as c2:
    display("hi")


class CustomString:
    def __init__(self, text: str, keep_quotes: bool):
        self._text = text
        self._keep_quotes = keep_quotes
      
    def _repr_mimebundle_(self, include=None, exclude=None):
        return (
          {"text/plain": self._text},
          {"strip-quotes": not self._keep_quotes}
    )
c3 = CustomString("'with quotes'", True)
c4 = CustomString("'without quotes'", False)
```

With quotes, we have {eval}`c1.outputs[0]`. Without quotes, we have {eval}`c2.outputs[0]`

With custom objects, we have {eval}`c3`. Without quotes, we have {eval}`c4`

````

### DIFF
--- a/.changeset/shy-spoons-exist.md
+++ b/.changeset/shy-spoons-exist.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Add support for `strip-quotes` metadata
--- a/packages/myst-cli/src/transforms/inlineExpression.spec.ts
+++ b/packages/myst-cli/src/transforms/inlineExpression.spec.ts
@@ -41,7 +41,7 @@ describe('transformRenderInlineExpressions', () => {
     } as InlineExpression;
     const tree = { type: 'root', children: [expr] };
     transformRenderInlineExpressions(tree, vfile);
-    // Children are added and quotes are removed
+    // Children are added and quotes are preserved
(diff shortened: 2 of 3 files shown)


rowanc1:
Thoughts on just `"{eval}`text`"`, or `{eval}`"'text'"`? Are there quotes that can be part of a plain text repr that are not a string?

agoose77:
@rowanc1 users can definitely do this at the markup stage. However, it's not perfect - we're making assumptions that might not always hold. For example, a widget that holds an enum of values - `1.0` and `'1.0'` would render identically. 

Quote stripping is really about fixing up the default repr, so applying it without an escape hatch to objects that define an explicit repr feels a bit too strong.

This pr just gives people an opt out.

PR #1814, opened 2025-01-24
⚠️ Add warnings for missing alt-text and auto-generated alt-text

JimMadge:
This PR adds two warnings with rules allowing them to be disabled,

- Warning for images missing alt-text (`image-has-alt-text`)
- Warning for images where alt-text was auto-generated, by checking for the `altTextIsAutoGenerated` tag (`image-alt-text-generated`)

Example output,

```command
> npx myst build -d
...
⚠️  index.md missing alt text for ./cat1.jpg
To suppress this message, add rule: "image-has-alt-text" to "error_rules" in your project config
⚠️  index.md alt text for ./cat2.jpg was auto-generated
   You can remove this warning by writing your own alt text
To suppress this message, add rule: "image-alt-text-generated" to "error_rules" in your project config
...
```

I have a example project to test this, but haven't added a test here.


packages/myst-transforms/src/images.ts
```diff
@@ -29,6 +32,28 @@ export function imageAltTextTransform(tree: GenericParent) {
...
+        node: image,
+        source: TRANSFORM_SOURCE,
+      });
+    }
+    if (image.data?.altTextIsAutoGenerated) {
+      fileWarn(file, `alt text for ${image.url} was auto-generated`, {
+        ruleId: RuleId.imageAltTextGenerated,
+        node: image,
+        source: TRANSFORM_SOURCE,
+        note: 'You can remove this warning by writing your own altxtext',
```
rowanc1, line 50:
```suggestion
        note: 'You can remove this warning by writing your own alt text',
```

rowanc1:
I am not sure how these will behave for notebooks, etc. Which has limited/no way to add alt text to for outputs at the moment, and so there would be many warnings you can't get rid of.

Thoughts on this being opt-in vs opt out?

agoose77:
@rowanc1 that's an interesting point. Jim and I were thinking about hand authored markup. What about stopping at output nodes, and not recursing into their children? Then, as long as this happens before embeds, we're OK?

Incidentally, can't we use block labels (`#| foo`) to add alt? That might be useful (feature request I think).

choldgraf:
I think a nice short term solution would be to not add this warning for output nodes, and to document that a good strategy is to use embed logic to get nicer things like captions, alt text, etc for notebook outputs

JimMadge:
I don't have strong feelings about opt-in or opt-out. On one hand, alt text is important for accessibility and it would be good to promote using it. On the other, it could make output very noisy and push people to ignore the warning.

Thanks for the improvements 🚀.

rowanc1:
There is the capability to stop warnings for a certain image, if we pass the key through. (That is via `addWarningForFile` not `fileWarn` though (these are logged in `logMessagesFromVFile`), so we ned a way to pass the `key` through to that function on the `VFileMessage`.) Happy to jump in and make that change.

This would enable turning on/off on a per-image basis, for example, if you are using a decorative image that does not have alt text, but want to warn/error on the rest.

My preference is to change this to opt-in. I am more swayed by the `it could make output very noisy and push people to ignore the warning` part of the argument. We are also currently not enforcing any other best practices in warnings, only content errors and syntax-related requirements. If it moves to opt-in, we should add a section in the docs.

JimMadge:
That all sounds good to me. I would also be happy to work on those things if, perhaps, @agoose77 can spare some time to mentor me on it.

agoose77:
Thanks for the helpful wider context, @rowanc1!

It looks like we have two separate enhancements: 
1. Support per-file selection of this rule
2. Support disabling certain rules by default

Both sound useful. @rowanc1, RE 
> Happy to jump in and make that change

I'm not 100% following the suggestion, but I understand that we have session-free and session-aware logging utilities. If you're happy to make that change _at some point_, that would be nice. I can put some cycles in this week to think about specifying a default set of error rules that merges with the user configuration.

rowanc1:
I will jump in and make it possible to add a `key` this afternoon!

JimMadge:
Just checking in on this PR. We are waiting for features so that,

- We can sensibly disable errors/warnings for certain types of image (like outputs in notebooks)
- We set rules as opt-in to avoid noisy warning output when users don't want to enforce alt-text

Is that right? Anything I can do to help with those?

rowanc1:
I am going to try and take a pass on this over the next week to get this over the line.

JimMadge:
@rowanc1 That's great :pray:. I expect I'm not much use with the MyST internals but do say if there is anything I can do to help.

JimMadge:
@agoose77 what was the key you recommended adding to the object in `fileWarn`?

agoose77:
I *think* you just want to set `key` to something that uniquely identifies the image, e.g. the URL. It doesn't make as much sense as other uses -- in the long run we might also want `error_rules` to be able to select against `label` etc, but for now it would enable per-URL selection of this rule.

JimMadge:
I suppose `url` may not be unique though 🤔. It would be _better_ to embed an image if you've already used it (I think). But I could see a project adding the same image twice by path or url.

agoose77:
Yeah, that didn't come through in my message - it's not guaranteed to be injective. But, it's good enough for a first pass where you want to turn off the warning for a particular image, and perhaps catch more than one. 

We definitely need to make surgical targeting of this easier!

JimMadge:
Hmmm, the build fails as `key` is not a valid property for `MessageInfo`.

```ts
export type MessageInfo = {
  node?: Node | Position;
  note?: string;
  source?: string;
  url?: string;
  fatal?: boolean;
  ruleId?: RuleId | string;
};
```

maybe `source` is what we want, that is used by `file.message`.

agoose77:
Ah, @JimMadge I think I remember what the issue is here — the limited logging API does not accept this `key`. Hmm... one solution is to move this to `myst-cli` transforms, and use `addWarningForFile`. Perhaps start there, whilst we figure out a better API.

JimMadge:
I'm happy to be directed. `myst-cli` sounds like it might be a more appropriate place, what we want is a warning message when running a build?

agoose77:
@JimMadge my intuition is that it's going to be easier to just get this in via `myst-cli`, e.g. https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/transforms/raw.ts.

It will be mostly the same code, but importantly you can use `addWarningForFile` instead of `fileWarn` from `Session`.

Issue #7, opened 2020-08-02
Roles with data / parameters


rowanc1:
I am wondering what the best way to add roles that have multiple attributes that need to be set, but no content. For example, creating a simple example for reactive document like (variables, displays and sliders):

![image](https://user-images.githubusercontent.com/913249/89132974-4ccf5480-d4d5-11ea-8ed4-444e1f77f31d.png)

````md
# Hello World

```{variable}
:name: x
:value: 2
```

The value of x is {display}`value:x format:.1f`.

{range}`value:x min:0 max:10`
````

I am looking to suggest a format for the internals of a role, which I don't think are suggested at all by MyST.

Something like: `attribute:value attribute2:value2`?

I think the only overlap in functionality would be the `glue` role, but I don't think it has any arguments at the moment?

choldgraf:
That's a great question - I feel like this one we should make sure to get everybody's input on, and then agree on a standard for all of the EBP roles. @chrisjsewell @AakashGfude @mmcky what do you think?

AakashGfude:
To add to this issue. Roles with multiple attributes that also have **content** , is also a necessary feature to have.  As indicated by @choldgraf in this PR https://github.com/executablebooks/jupyter-book/pull/837

I thnk in rst, it is achieved by declaring it beforehand to create specialized roles? :-  https://docutils.sourceforge.io/docs/ref/rst/roles.html#id17

chrisjsewell:
Linking https://github.com/executablebooks/MyST-Parser/issues/69

choldgraf:
oi I can't believe it's been 6 months since those conversations, I totally forgot about that issue.

Looking through the ideas there, what about something like:

* `` {role}`content <option option key=val key="val w spaces">` ``

It piggy-backs on the current rST convention of `< >` to define a link. More generally, do we like the idea of having some kind of syntax to explicitly separate the "content" parts of the role from the "options" part of the role? (above, it is `< >` but could be something else like ` ||` or something)

fperez:
FWIW, before finding this issue I was wrestling with this problem with ChatGPT, and it told me this would work:

![Image](https://github.com/user-attachments/assets/0518fb0c-ab58-42e0-9f8b-32c2ec611810)

That doesn't actually work, but worth mentioning for reference :) I don't know which system does have that behavior - quarto, pandoc?

PR #1822, opened 2025-01-27
🧮 Add inline options to roles and directives

rowanc1:
Work in progress to allow for parsing of inline attributes.

- [x] Improve the parsing of attributes without quotes
- [x] Apply this to directives as well (dry up code for option parsing)

- ref MEP: https://github.com/jupyter-book/myst-enhancement-proposals/pull/28
- closes https://github.com/jupyter-book/mystmd/issues/7
- see: https://github.com/jupyter-book/myst-spec/issues/68


rowanc1:
I have followed up with the changes requested from the last MyST meeting where we went through this PR in person. We are now no longer defaulting to span with roles (keeps open the possibility of other ways to work with span-like attributes in the future) and there is improved documentation as well as error messages on `.class.without.spaces` (which @choldgraf requested).

Going to get this merged, released and docs pushed.

choldgraf:
Woot - thanks to everybody for discussing this one and getting it over the line!

Issue #1809, opened 2025-01-23
Papercuts & Feedback from Discord


rowanc1:
This is a staging ground for launching other issues.

- Would be nice if SVGs were used as (AFAIU) <object> rather than <img>, so that e.g. links within can be clicked
- Hover-previews of links to other documents could be more useful, e.g. contents of first section/heading that's nonempty (I ended up directly linking to that subsection instead)
- if I make a ::::{grid} with two :::{card}s in it, it renders more nicely at some zoom levels if I explicitly specify ::::{grid} 2, even though there's two cards/columns in either case
- might be nice as an option to be able to have footnotes not present in the main document text at all, only as hover tooltips
- there's an outline_maxdepth site option (book theme), but anything above 3 has no (additional) effect
- when there are a lot of items in the RHS outline, the last few are cut off (not visible even when scrolled all the way down)
- there's a horizontal scrollbar on the page for no obvious reason (afaict, scrolls everything except the LHS ToC sidebar by a few tens of px left/right)
- the :linenos option for {code} affects indentation of the contained code -- i.e., lines starting from line 10+ are indented 1 column further than lines 1-9 due to the width of the line numbers themselves
- the :emphasize-lines: option only seems to have an effect if :linenos: is also enabled (but IMHO it seems independently useful)
- some way to influence the widths of table columns would be nice -- we have a bunch of "key-value" style tables with the name of a property in the first column and a description in the second; and the width of the first column (which always has the same contents) varies wildly per-table based on whether there are any long lines in the second column (in which case the first column takes only as much space as it needs; otherwise, ~half the table width)
- the :align: option in a :::{table} seems to have no effect (initially I thought it was being implicitly overridden by the per-column alignments inferred from markdown header syntax... but testing it out at https://mystmd.org/guide/tables, it doesn't seem to do anything for list-table or csv-table either)
- strong emphasis on literal text (**`asdf`**) has no effect (fwiw it seems it does in here :)

From: [glaebhoerl on discord](https://discord.com/channels/1083088970059096114/1126886232060870678/1330705364441759804)

glaebhoerl:
TYVM! Would've gotten around to it eventually...

(Of the ones listed above, the 3-4 of them beginning from "when there are a lot of items in the RHS outline, the last few are cut off" seem the more significant ones to me.)

I ended up collecting a few more in the meantime:

### Perceptible annoyance
[...]

PR #1825, opened 2025-01-28
🌎 Remove TLD of `.java` and `.zip` from linkify

rowanc1:
See #1809

cc @glaebhoerl

### DIFF
--- a/.changeset/large-clouds-mix.md
+++ b/.changeset/large-clouds-mix.md
@@ -1,0 +1,5 @@
+---
+"myst-parser": patch
+---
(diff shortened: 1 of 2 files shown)


glaebhoerl:
Thanks! :pray: 

Had I known this would be addressed so quickly, and that the fix would be TLD-specific, I would've been more exhaustive: of the ones I've personally run into, this also affects `.properties` and `.id` (the latter being something we'd chosen as an extension for our own custom thing, rather than something prevalent in the wild).

Issue #1800, opened 2025-01-22
List page headers / table of contents in-text


choldgraf:
When you have a page with multiple headers (especially nested ones), it is useful to be able to list each header with a link throughout the page. This is different from the sidebar contents list, because it flows within the text and often only points to a subset of headers.

This is particularly useful when you have a large page and specific sub-sections can get lost within the page-wide contents list, and so you want to define a short list of a section's headers in order to orient the user.

For example, given a document like this:

````md
# Types of things

```{tableofcontents} Things in this section
```
## Type 1

Foo

## Type 2

Bar

## Type 3

Baz
````

It would output something like:

```
Types of things

**Headers in this section**
- [Type 1](link)
- [Type 2](link)
- [Type 3](link)

**Type 1**

Foo

**Type 2**

Bar

**Type 3**

Baz
```

### Implementations

Sphinx has two ways of doing this:

- **[toctree](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree)** is how you define the structure of a book, and also results in a list of toctree items in the page content (wherever the toctree was placed).
- **[contents](https://docutils.sourceforge.io/docs/ref/rst/directives.html#table-of-contents)** is a docutils directive that lists the content headers for the current section.

_Converting this into an issue since it was a draft issue on our board_

PR #1826, opened 2025-01-28
✨ New table of contents directive

fwkoch:
This PR addresses https://github.com/jupyter-book/mystmd/issues/1800

It introduces a new `toc` directive that, for site builds, is transformed into a list-based table of contents. You may specify `context` to switch between project, page, and section contents. The default context is project.

- `project` - for this context, entries in the table of contents are all project pages and folders. You may also use `toctree` as an alias for the directive. (This provides partial functionality similar to the sphinx toctree directive, although that directive allows you to _define_ child pages and their order. The MyST toc directive is only for display and entirely dependent on the toc structure defined in the `myst.yml` file. See sphinx toctree here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree)

  The following are all equivalent

  ````
  ```{toc}
  ```

  ```{tableofcontents}
  :context: project
  ```
  
  ```{toctree}
  ```
  ````

- `page` - for this context, entries in the table of contents are all the headings on a page.

  ````
  ```{toc}
  :context: page
  ```
  ````

- `section` - for this context, entries in the table of contents are only the headings of the current section you are in (sections are delimited by headings of the same depth). You may also use `contents` as an alias for toc directive with section context, to provide functionality similar to the docutils contents directive: https://docutils.sourceforge.io/0.4/docs/ref/rst/directives.html#table-of-contents)

  The following are all equivalent

  ````
  ```{toc}
  :context: section
  ```
  
  ```{contents}
  ```
  ````

You may also provide a heading for your table of contents:

````
```{toc} My Contents
```
````

Currently `toc` directives are ignored by PDF and other non-site build targets.

- closes https://github.com/jupyter-book/jupyter-book/issues/2312
- closes https://github.com/jupyter-book/mystmd/issues/1800

### DIFF
--- a/.changeset/brave-dragons-pump.md
+++ b/.changeset/brave-dragons-pump.md
@@ -1,0 +1,7 @@
+---
+'myst-directives': patch
+'myst-transforms': patch
+'myst-cli': patch
+---
+
+New TOC directive
--- a/docs/directives.md
+++ b/docs/directives.md
@@ -62,3 +62,6 @@ description: A full list of the directives included in MyST Markdown by default.
 
 :::{myst:directive} table
 :::
+
+:::{myst:directive} toc
+:::
--- a/packages/myst-cli/src/build/site/manifest.ts
+++ b/packages/myst-cli/src/build/site/manifest.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import { SPEC_VERSION } from '../../spec-version.js';
 import { hashAndCopyStaticFile } from 'myst-cli-utils';
 import { RuleId, TemplateOptionType } from 'myst-common';
 import type { SiteAction, SiteExport, SiteManifest } from 'myst-config';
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

**Comment by fperez** [comment]
Thanks @fwkoch! Quick question (and sorry if I missed something, but I don't see docs in the PR), how does it work for PDF output? For example, for standalone documents I'm currently using raw typst and sticking to typst templates, via

```{raw:typst}
#outline(indent:auto, depth:2)
```

This works fine, but obviously would be best to have toc management be done at the mystmd level so it's consistent across output types.

Thoughts on how to gracefully handle this? (and I totally understand if it's too much to deal with for now and that's a future enhancement).

fwkoch:
Hey @fperez - for this PR, PDFs are out of scope; the new directive is just ignored. Your approach of raw typst is still best.

I agree consistency across output types would be nice, but for that to be smooth, I think we need more control over the different export types (e.g. https://github.com/jupyter-book/mystmd/issues/1833) so features can easily be "turned on and off."

PR #1835, opened 2025-02-04
🐛 Improve error message for missing javascript and python plugins

### LINKED ISSUE #1813: Improve error message for missing plugin `.mjs` file
When running through including a new transform to `mystmd` with @JimMadge, we observed that the error message for a non-existent plugin file isn't clearly referring to a plugin!

JimMadge:
Closes #1813

It looks like the problem was here

https://github.com/jupyter-book/mystmd/blob/d1f8768eb4dbada0a370765ba696fbb3c00ba66f/packages/myst-cli/src/plugins.ts#L87

`fs.statSync` raises an error if there is no file/directory.
This PR adds an option so that `fs.statSync` returns `undefined` if there is no file, then handles that possibility when calling `isFile()`.

I see that there is another instance of `fs.statSync(...).isFile()` in `plugins.ts` that might cause similar problems.

### DIFF
--- a/packages/myst-cli/src/plugins.ts
+++ b/packages/myst-cli/src/plugins.ts
@@ -84,7 +84,7 @@ export async function loadPlugins(session: ISession): Promise<ValidatedMystPlugi
           return { path, module: { plugin } };
         }
         case 'javascript': {
-          if (!fs.statSync(path, {throwIfNoEntry: false})?.isFile() || !path.endsWith('.mjs')) {
+          if (!fs.statSync(path, { throwIfNoEntry: false })?.isFile() || !path.endsWith('.mjs')) {
             addWarningForFile(
               session,
               path,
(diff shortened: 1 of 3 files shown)


rowanc1:
Thanks, this looks good. Will merge when tests go green.

Thanks @JimMadge.

packages/myst-cli/src/plugins.ts
```diff
@@ -84,7 +84,7 @@ export async function loadPlugins(session: ISession): Promise<ValidatedMystPlugi
           return { path, module: { plugin } };
         }
         case 'javascript': {
-          if (!fs.statSync(path).isFile || !path.endsWith('.mjs')) {
+          if (!fs.statSync(path, {throwIfNoEntry: false})?.isFile() || !path.endsWith('.mjs')) {
```
rowanc1, line 87:
```suggestion
          if (!fs.statSync(path, { throwIfNoEntry: false })?.isFile() || !path.endsWith('.mjs')) {
```

rowanc1:
Have expanded to cover the python ones as well. cc @agoose77.

PR #1837, opened 2025-02-04
🔗 Add citation for the project

rowanc1:


rowanc1:
This is setup through the Zenodo plugin, and will be updated on subsequent releases.

PR #1840, opened 2025-02-06
📖 Update to installing instructions to include conda

mmcky:
This PR updates the installation instructions to include `conda` as an option (in addition to `mamba`) for clarity as many end users of Jupyter Book have access to `conda`.


choldgraf (commented):
Thanks Matt - this feels fine to me, especially if it may confuse people to not see conda up there.

Issue #1498, opened 2024-08-28
Refine task of improving JupyterHub/Binder launch feature


agoose77:
The scene of execution and launching of interactive content in MyST is complex (because the scene is complex!). 

At the heart of this, I think there are three possible ways a user might want to provide launch and/or execution support for their content:

1. Build image from content & environment repo (repo2docker) using a BinderHub
   - MyBinder.org
   - Generic BinderHub
2. Build image from environment repo (repo2docker) using a BinderHub, and provision content from separate repo with `nbgitpuller`
3. Use pre-existing image and provision with `nbgitpuller` on a JupyterHub
4. Talk to pre-existing server without provisioning environment or content (manual)[^only-thebe]

The above strategies should work for both Jupyter and JupyterLite distributions, both for execution (Thebe) and launch links. 

I think our config probably needs a re-work to reflect this, and Thebe might need some more support to facilitate it.

[^only-thebe]: This only makes sense for Thebe, probably. But perhaps not.

Here's a sketch of how our config might change:

#### Existing configuration (simplified)
```yaml
project:
  jupyter: undefined(false) | boolean | object | 'lite' | 'binder'
    lite: boolean
    binder: undefined(false) | boolean | object
      url: string (url)
      provider: string (git | gitlab | github | or custom)
      repo: string (org-name/repo-name | url | string)
      ref: string (undefined | string)
    server:  undefined | object
      url: string (url)
      token: string (any)
```

#### Suggested configuration (simplified)
```yaml
project:
  jupyter:
    lite:
      # JupyterLite can build custom environment
      environment:
        url: string (url)
        provider: string (git | gitlab | github | or custom)
        repo: string (org-name/repo-name | url | string)
        ref: string (undefined | string)

      # Does it make sense to pull content at runtime? 
      # Binder does this because images are ephemeral
      # We have to build the env at build-time anyway
      # content:
      
    binder:
      # Binder can build images
      environment:
        url: string (url)
        provider: string (git | gitlab | github | or custom)
        repo: string (org-name/repo-name | url | string)
        ref: string (undefined | string)

      # nbgitpuller can provision image
      content:
        url: string (url)
        provider: string (git | gitlab | github | or custom)
        repo: string (org-name/repo-name | url | string)
        ref: string (undefined | string)
    hub:
      # Hubs can't build images
      # environment:

      # But can configure spawn
      options:
        ...

      # nbgitpuller can provision image
      content:
        url: string (url)
[...]

Issue #1373, opened 2024-07-05
Add support for execution strategies as an array of named subconfigurations


agoose77:
1. This would allow us to have multiple kinds of local execution with unique IDs.
2. Allow frontmatter to specify execution strategy e.g. per page.
3. Allow CLI to override strategy if not defined.

agoose77:
Also: move `execute` flag to config? Default on.

agoose77:
There are a number of use cases feeding this user-story:
1. Switching between local and remote execution when building locally
2. Executing particular pages in remote contexts to sparingly use remote compute
3. Executing particular pages in remote contexts where different datasets are available.

Even if we _don't_ allow per-page execution strategies, the machinery for defining it remains the same.

PR #1842, opened 2025-02-06
🚫 Add page frontmatter option to disable execution

### LINKED ISSUE #1619: Allow execution of specific notebooks to be skipped
Currently it's possible to tag specific **cells** for skipping execution (with `skip-execution`) but there's not a way to do this for **entire notebooks**. It would be useful if this were possible either in page metadata, or in a list of `exclude` glob patterns in `myst.yml`.

This is useful when you have:

1. A notebook that you know is broken, and so you don't want to clog up your build with broken execution
2. A notebook that takes a lot of resources to build, and you want to run it once and embed the outputs inside without re-executing

agoose77:
This PR closes #1619 by adding a new `skip_execution` field to the page frontmatter. I don't anticipate it being a stable addition — we'll probably rethink at the very least _where_ it lives when we tackle #1498 and #1373. But, let's just get something in for now.

### DIFF
--- a/.changeset/dirty-ducks-tie.md
+++ b/.changeset/dirty-ducks-tie.md
@@ -1,0 +1,6 @@
+---
+"myst-frontmatter": patch
+"myst-cli": patch
+---
+
+Add support for skipping execution of individual notebooks
--- a/docs/execute-notebooks.md
+++ b/docs/execute-notebooks.md
@@ -65,6 +65,24 @@ name = input("What is your name?")
 
 Additional [cell tags](#tbl:notebook-cell-tags) to hide, remove, or raise exceptions are also possible.
 
+## Skip entire notebooks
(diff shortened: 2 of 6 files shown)


rowanc1 (approved):
Looks awesome. Docs are great.

Haven't tested, but this looks like everything makes sense.

agoose77:
I haven't added tests here because they'd need to be E2E, and I want to get around to that once we overhaul how execution works. Are you OK merging without tests?

I've run a test locally with three files permuting `skip_execution` missing, defined `true`, defined `false`, and only the `skip_execution: true` notebook was skipped.

rowanc1:
🚀🚀

choldgraf:
yesss amazing, I can finally run execution on my blog now lol

PR #1864, opened 2025-02-17
🛫 Add documentation for Landing Pages

agoose77:
>[!NOTE]
> This PR depends upon jupyter-book/myst-theme#531

This PR modifies our block transform to pull `class` and `kind` from `.data`, and adds documentation for the landing-page feature.

To try this PR out, build `mystmd`, and run it locally in the docs with e.g. `env -C docs -- npx myst start --headless`, then run the book-theme from the `myst-theme` PR.

Screenshot 2025-02-17 at 12-56-00 Landing Pages - MyST Markdown Guide


mmcky:
genuinely excited for this feature. Thanks @agoose77

agoose77:
@rowanc1 note that if we use `kind`, it will mean that `ipynb` files that want to define landing pages will lose the `notebook-cell` block kind.

rowanc1:
@agoose77 you can have blocks in blocks ... so I think this should be fine?

agoose77:
Er @rowanc1 our schema states that we don't support that, which is why it's in my :heavy_plus_sign: column for using blocks rather than directives :sweat_smile: 

What's the incentive for nested blocks? Shouldn't we use div for that?

agoose77:
I feel as though `notebook-content` isn't that useful, so I'm OK with it being easy to override.

PR #1867, opened 2025-02-18
 📂 Set `site.options.folders: true` for JB upgrades

### LINKED ISSUE #1866: JB2 breaks links to JB1 projects unless `folders: true` is enabled
My understanding is that JB1 reflects folder structure in URLs by default (I need to check this).
JB2 does not reflect folder structure in URLs by default: https://mystmd.org/guide/website-templates#site-url-options

A problem occurs when upgrading from JB1 to JB2 in that if the defaults are untouched, links to existing resources which used JB1 will get broken.

## Proposal

Potential options:
1. Make `folders: true` the default in JB2 projects
2. Make `folders: true` the default in JB2 projects that are upgraded from JB1
3. Add a prompt within the JB1->JB2 upgrade process which asks if the user would like to enable this option.

agoose77:
Fixes #1866 

I made `_toc.yml` a requirement for upgrading too, as this is a required file in JB1. I'm not sure why I didn't do this to begin with.

### DIFF
--- a/packages/myst-cli/src/init/jupyter-book/upgrade.ts
+++ b/packages/myst-cli/src/init/jupyter-book/upgrade.ts
@@ -41,6 +41,14 @@ export async function upgradeJupyterBook(session: ISession, configFile: string) 
     (config as any).project.toc = upgradeTOC(session, tocData);
   }
 
+  config.site = config.site ?? {};
+  config.site.options = config.site.options ?? {};
+
+  // Ensure we enable folders
+  session.log.debug(
+    chalk.dim(`Setting ${chalk.blue('site.options.folders')} to ${chalk.blue('true')}`),
+  );
+  config.site.options.folders = true;
 
   // Upgrade legacy syntax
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/init/jupyter-book/upgrade.ts
```diff
@@ -30,15 +30,25 @@ export async function upgradeJupyterBook(session: ISession, configFile: string)
...
+  }
+
+  config.site = config.site ?? {};
+  config.site.options = config.site.options ?? {};
+
+  // Ensure we enable folders
+  session.log.debug(
+    chalk.dim(`Setting ${chalk.blue('site.options.folders')} to ${chalk.blue('true')}`),
+  );
+  config.site.options.folders = true;
```
rowanc1, line 51: 👍

rowanc1 (approved):
> I'm not sure why I didn't do this to begin with.

I don't think it _existed_ to begin with!

PR #1872, opened 2025-02-23
Handle unsupported options for include directive

### LINKED ISSUE #813: Allow non-literal `include` to respect start/end options
The `include` directive has two "modes" - `literal: true` where file content is directly pasted into the document, and `literal: false` where file content is parsed as MyST if possible.

The former allows you to specify starting/ending lines in various ways; the latter does not, you must parse the whole file. I do not think this restriction is necessary. Maybe you just want to include one section of a markdown document? This should be able to do a standalone parse of just that section.

JimMadge:
Closes #813

- [x] Add warnings for unused options in non-literal include
- [x] Add support for some/all start/end options for non-literal includes (requested in #813 and https://github.com/the-turing-way/the-turing-way/pull/4049

The filters get applied for non-literal includes. The code change is just removing a conditional block which made an early return when `literal === false`. This does mean includes are slightly more expensive now, and I think some of the processing of filter/options doesn't make sense when `literal === false`. It does keep the code quite clean and simple though.

I've tested this locally, but haven't added any tests.

````markdown
# index.md
start-after and end-before

```{include} ./data.txt
:filename: true
:start-after: BEGIN
:end-before: END
```

lines

```{include} ./data.txt
:lines: 2,3
```

literal with start-after and end-before

```{include} ./data.txt
:literal: true
:start-after: BEGIN
:end-before: END
```

Test "upgrading" to literal when `lang` is defined

```{include} ./data.txt
:lang: Python
:start-after: BEGIN
:end-before: END
```
````

```
# data.txt
BEGIN
Hello
World!

*emphasis*
**bold**
END

```

Screenshot 2025-02-26 at 11 50 06

### DIFF
--- a/packages/myst-directives/src/include.ts
+++ b/packages/myst-directives/src/include.ts
@@ -111,13 +111,15 @@ export const includeDirective: DirectiveSpec = {
         ];
       }
     }
+    // Are any filter properties set?
+    const usesFilter = Object.values(filter).some((value) => value !== undefined);
     const include: Include = {
       type: 'include',
       file,
       literal,
       lang,
       caption: data.options?.caption as any[],
-      filter: Object.keys(filter).length > 0 ? filter : undefined,
+      filter: usesFilter ? filter : undefined,
       ...opts,
     };
     addCommonDirectiveOptions(data, include);
--- a/packages/myst-directives/src/include.ts
+++ b/packages/myst-directives/src/include.ts
@@ -82,6 +82,7 @@ export const includeDirective: DirectiveSpec = {
           file.split(/\/|\\/).pop(),
         )
       : {};
+    const caption = literal ? (data.options?.caption as any[]) : undefined;
     const filter: Include['filter'] = {};
     ensureOnlyOneOf(data, vfile, ['start-at', 'start-line', 'start-after', 'lines']);
     ensureOnlyOneOf(data, vfile, ['end-at', 'end-line', 'end-before', 'lines']);
@@ -118,7 +119,7 @@ export const includeDirective: DirectiveSpec = {
       file,
       literal,
       lang,
-      caption: data.options?.caption as any[],
+      caption,
       filter: usesFilter ? filter : undefined,
       ...opts,
     };
--- a/packages/myst-directives/src/include.ts
+++ b/packages/myst-directives/src/include.ts
@@ -1,5 +1,5 @@
 import type { DirectiveData, DirectiveSpec, GenericNode } from 'myst-common';
-import { RuleId, fileWarn, normalizeLabel } from 'myst-common';
+import { RuleId, fileWarn } from 'myst-common';
 import { CODE_DIRECTIVE_OPTIONS, getCodeBlockOptions } from './code.js';
 import type { Include } from 'myst-spec-ext';
 import type { VFile } from 'vfile';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by agoose77** [comment]
The usual pattern in `mystmd` is that you have a particular transform(s) that looks for a particular node type. I found one at `packages/myst-transforms/src/include.ts` using `ripgrep`:
```bash
rg -t ts -i "include\S*(directive|transform)"
```

The non-`literal` block is here:
https://github.com/jupyter-book/mystmd/blob/0ca2c6792b3f83d58942788cf499d495a9e3e73d/packages/myst-transforms/src/include.ts#L124-L126

agoose77:
Taking a look at this, I think you're right that the only required change is to pass-through the filter options from the directive, i.e. you don't need to touch the transform. 

The following lines define the filter: 
https://github.com/jupyter-book/mystmd/blob/229947182f1b43d226e31605ff39c7072c6d490f/packages/myst-directives/src/include.ts#L97-L125

JimMadge:
> Taking a look at this, I think you're right that the only required change is to pass-through the filter options from the directive, i.e. you don't need to touch the transform.

:tada: That was going to be the first thing I tried hoping it worked 😆.

I can see that the code essentially says "if not literal ignore all options, otherwise apply options" so I was hoping I could simply change that :+1:.

packages/myst-directives/src/include.ts
```diff
@@ -68,25 +68,11 @@ export const includeDirective: DirectiveSpec = {
...
-      }
-      const include: Include = {
-        type: 'include',
-        file,
-      };
-      addCommonDirectiveOptions(data, include);
-      return [include];
-    }
     const lang = (data.options?.lang as string) ?? extToLanguage(file.split('.').pop());
     const opts = getCodeBlockOptions(
```
agoose77, line 77: I'd be tempted to gate `opts` behind a test for `literal`, e.g.
```typescript
const opts = literal ? getCodeBlockOptions(...) : {};
```

Literal includes are the only includes that could support code options.

JimMadge:
I'm getting the same test errors on main as on this branch, so I `think` it isn't related to my changes.

packages/myst-directives/src/include.ts
JimMadge: That looks good. I wanted to do a bit more of that but was struggling with the JS/TS syntax 😅.

JimMadge:
Perhaps `caption` is also makes no sense for a non-literal include?

https://github.com/jupyter-book/mystmd/blob/cf18f2027f8df741b499093a0974c5d88d1bbb8e/packages/myst-directives/src/include.ts#L119

Would this work?

```javascript
      caption: literal ? data.options?.caption as any[] : [],
```

agoose77:
@JimMadge I've fixed the tests — this PR broke them because the test fails if the new AST has attributes that the expected AST does not.

Regarding `caption`, I agree we should hide it. I think instead of setting to `[]`, let's pull out `caption` as a local variable and set it to `undefined` for `literal === true`.

JimMadge:
Ah, I think I must have forgotten to build after checking out main 🤦

JimMadge:
I think these changes would mean a patch release from my reading of https://github.com/jupyter-book/mystmd/blob/main/CONTRIBUTING.md

agoose77 (approved):

PR #1888, opened 2025-03-03
📖 Adds note on Windows builds to Contribution guide

nmielec:
The build process uses unix commands that probably won't work in cmd.exe. Switching to WSL, or a unix like shell such as Git Bash and configuring npm to use it solves the issue.


rowanc1:
Thank you!

PR #1900, opened 2025-03-06
📠 Add up/downgrade for `outputs` AST change

agoose77:
This PR is the first step towards changing our AST for reworking output handling.

### DIFF
--- a/packages/myst-migrate/src/v3_outputs.ts
+++ b/packages/myst-migrate/src/v3_outputs.ts
@@ -1,0 +1,93 @@
+import { selectAll } from 'unist-util-select';
+import { assert } from 'console';
+import type { IFile } from './types.js';
+
+export const description = `
+Output nodes previously could not represent AST trees for each output. 
+Now, the Outputs node has Output children with a 1:1 correspondence to IOutput bundles.
+`;
+
+type OutputV2 = {
+  type: 'output';
+  children?: any[];
+
+  data?: any[];
+  visibility?: any;
+
+  html_id?: string;
+  label?: string;
+  identifier?: string;
+
+  id?: string;
+};
+
+type OutputV3 = {
+  type: 'output';
+  children: any[];
+
+  jupyter_data?: any;
+
+  label?: string;
+  identifier?: string;
+  html_id?: string;
+};
+
+export function upgrade(file: IFile): IFile {
+  const { version, mdast } = file;
+  assert(version === 2, 'Version must be 2');
+  const nodes = selectAll('output', mdast) as OutputV2[];
+  nodes.forEach((node) => {
+    // We can only correlate output children with the IOutput objects if there's only one IOutput
+    // Additionally, there may be placeholders that need to be removed
+    const numOutputs = node.data?.length ?? 0;
+    const children = numOutputs === 1 ? node.children ?? [] : [];
+    const placeholders = children.filter((child) => !!child.placeholder);
+    const notPlaceholders = children.filter((child) => !child.placeholder);
+
+    const outputsChildren = (node.data ?? []).map((outputData) => {
+      const result: OutputV3 = {
+        type: 'output',
+        jupyter_data: outputData,
+        children: notPlaceholders, // FIXME: ignoring children here
+      };
+      notPlaceholders.length = 0;
+      return result;
+    });
+
+    // Restore placeholders at the end
+    outputsChildren.push(...placeholders);
+
+    // Convert Output into Outputs
+    if (node.data !== undefined) {
+      delete node.data;
+    }
+    (node as any).type = 'outputs';
+    (node as any).children = outputsChildren;
+  });
+  return file;
+}
+
+export function downgrade(file: IFile): IFile {
+  const { version, mdast } = file;
+  assert(version === 3, 'Version must be 3');
+  const nodes = selectAll('outputs', mdast) as OutputV2[];
+  nodes.forEach((node) => {
+    const outputsChildren = node.children as any[];
+    const data = outputsChildren
+      .filter((output) => output.type === 'output')
+      .map((output: any) => output.jupyter_data)
+      .filter((datum) => !!datum);
+
+    const notPlaceholders = outputsChildren.filter((child: any) => !child.placeholder);
+    const children = notPlaceholders.map((output) => (output as OutputV3).children ?? []).flat();
+    const placeholders = outputsChildren.filter((child: any) => !!child.placeholder);
+    children.push(...placeholders);
+
+    // Convert Outputs into Output
+    (node as any).data = data;
+
+    (node as any).type = 'output';
+    (node as any).children = children;
+  });
+  return file;
+}
--- a/packages/myst-migrate/src/migrations.ts
+++ b/packages/myst-migrate/src/migrations.ts
@@ -1,5 +1,6 @@
 import type { Migration } from './types.js';
 import * as v1 from './v1_footnotes.js';
 import * as v2 from './v2_blockClasses.js';
+import * as v3 from './v3_outputs.js';
 
-export const MIGRATIONS: Migration[] = [v1, v2];
+export const MIGRATIONS: Migration[] = [v1, v2, v3];
--- a/packages/myst-migrate/src/tests/version_2_3.spec.ts
+++ b/packages/myst-migrate/src/tests/version_2_3.spec.ts
@@ -1,0 +1,139 @@
+import { describe, expect, it } from 'vitest';
+import { migrate } from '../index';
+import type { Parent } from 'mdast';
+
+const SIMPLE_AST: Parent = {
+  type: 'root',
+  children: [
+    {
+      // @ts-expect-error: unknown type
+      type: 'block',
+      children: [
+        {
+          // @ts-expect-error: invalid child type
+          type: 'paragraph',
+          children: [
+            {
+              type: 'text',
+              value: 'This is an ',
+            },
+            {
+              type: 'emphasis',
+              children: [
+                {
+                  type: 'text',
+                  value: 'interesting',
+                },
+              ],
+            },
+            {
+              type: 'text',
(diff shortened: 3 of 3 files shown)


packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,93 @@
+import { selectAll } from 'unist-util-select';
+import { assert } from 'console';
+import type { IFile } from './types.js';
+
+export const description = `
+Output nodes previously could not represent AST trees for each output. 
+Now, the Outputs node has Output children with a 1:1 correspondence to IOutput bundles.
```
rowanc1, line 7:
```suggestion
Now, the Outputs node has Output children with a 1:1 correspondence to IOutput bundles.
The IDs do not change...
```

rowanc1:
Once the docs are good, can you copy them over to the `./docs/upgrade.ts` or something. Right now this is duplicated.

stevejpurves:
@agoose77 @choldgraf - In future I think changes to the `myst-migrate` package should be included in the PR that is making the AST changes. This is akin to including your DB migrate scripts in the same changeset that updates your DB schemas.

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
+import { selectAll } from 'unist-util-select';
+import { assert } from 'console';
```
stevejpurves: Why are we asserting rather than throwing?

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+    const numOutputs = node.data?.length ?? 0;
+    const children = numOutputs === 1 ? node.children ?? [] : [];
+    const placeholders = children.filter((child) => !!child.placeholder);
+    const notPlaceholders = children.filter((child) => !child.placeholder);
+
+    const outputsChildren = (node.data ?? []).map((outputData) => {
+      const result: OutputV3 = {
+        type: 'output',
+        jupyter_data: outputData,
+        children: notPlaceholders, // FIXME: ignoring children here
```
stevejpurves: @agoose77 what is the relevance of this FIXME? what need to happen here?

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+      };
+      notPlaceholders.length = 0;
+      return result;
+    });
+
+    // Restore placeholders at the end
+    outputsChildren.push(...placeholders);
```
stevejpurves, line 76: this seems fine, so curious as to the FIXME above?

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+  identifier?: string;
+  html_id?: string;
+};
+
+export function upgrade(file: IFile): IFile {
+  const { version, mdast } = file;
+  assert(version === VERSION, `Version must be ${VERSION}`);
+  const nodes = selectAll('output', mdast) as OutputV2[];
+  nodes.forEach((node) => {
+    // We can only correlate output children with the IOutput objects if there's only one IOutput
```
stevejpurves: @agoose77 can you expand on this and explain the impact on a round trip downgrade-> upgrade->downgrade, is this lossy?

What do we expect to be in the children array in V2?

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Additionally, there may be placeholders that need to be removed
+    const numOutputs = node.data?.length ?? 0;
+    const children = numOutputs === 1 ? node.children ?? [] : [];
+    const placeholders = children.filter((child) => !!child.placeholder);
```
stevejpurves, line 60: The test case so far does not cover this logic (placeholders, multiple outputs, children preset), and should be extended to cover them.

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    // Restore placeholders at the end
+    outputsChildren.push(...placeholders);
+
+    // Convert Output into Outputs
+    if (node.data !== undefined) {
+      delete node.data;
+    }
+    (node as any).type = 'outputs';
+    (node as any).children = outputsChildren;
```
stevejpurves, line 83: 👍

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+  });
+  return file;
+}
+
+export function downgrade(file: IFile): IFile {
+  const { version, mdast } = file;
+  assert(version === 3, 'Version must be 3');
+  const nodes = selectAll('outputs', mdast) as OutputV2[];
```
stevejpurves: should be a cast to V3? is this runs it means our types are too flaky/any?

packages/myst-migrate/src/v3_outputs.ts
```diff
@@ -0,0 +1,98 @@
...
+    const outputsChildren = node.children as any[];
+    const data = outputsChildren
+      .filter((output) => output.type === 'output')
+      .map((output: any) => output.jupyter_data)
+      .filter((datum) => !!datum);
+
+    const notPlaceholders = outputsChildren.filter((child: any) => !child.placeholder);
+    const children = notPlaceholders.map((output) => (output as OutputV3).children ?? []).flat();
+    const placeholders = outputsChildren.filter((child: any) => !!child.placeholder);
+    children.push(...placeholders);
```
stevejpurves, line 104: I'm confused by the placeholder treatment, if we had some test fixtures taken from actual V2 AST with these different flavours of outputs that would help. These could be easily harvested from https://github.com/jupyter-book/example-outputs

stevejpurves (changes requested):
* There are a few open questions mainly regarding comments in the code.
* We need to extend test coverage, the current fixture does not cover subtitles in the transformation logic. Test fixtures should document our understanding of the AST in the more complex situations like placeholders, multiple outputs and V2 outputs with children (I am not clear on where the latter arises)

packages/myst-migrate/src/v3_outputs.ts
agoose77: This routine is supposed to be called by the version upgrade dispatch. It's a fatal error to try and upgrade with the wrong version, and we can prefer asserting this rather than throwing explicitly because users _should_ never encounter this code path (i.e. we're optimising for writing nice upgrades here, not user-facing errors).

packages/myst-migrate/src/v3_outputs.ts
agoose77: In v2, if there is:
```
Output {
  data: [
    IOutput { ... }
  ],
  children: [
    ...
  ]
}
```
then it's easy to map to/from the v3 `outputs` structure: there is a single `Output` child (of `Outputs`), and it takes `data[0]` and `children` from the v2 `Output` node.

However, if in v2 we have:
```
Output {
  data: [
    IOutput { ... },
    IOutput { ... },
  ],
  children: [...]
}
```
then there is no obvious way say whether `Output.children[i]` belongs to `Output.data[i]` or `Output.data[j]` --- I don't think we guarantee that there will always be one-and-exactly-one child for each IOutput object.

For the most part this should be fairly rare — I think most of the time people have singular outputs.

I don't think we presently promise round-tripping.

agoose77:
OK, I've added a test that handles placeholders and multiple outputs.

@stevejpurves good spot with the assertion. At one point we were using Node's assert, which throws errors. Now it's using the console assert, which doesn't, so it was failing silently!

packages/myst-migrate/src/v3_outputs.ts
agoose77: The tests are tricky -- we really don't want to rely on any MyST types because they will change naturally between versions. But, I've improved this so that the bug is now raised (and fixed).

packages/myst-migrate/src/v3_outputs.ts
stevejpurves: plotly/bokeh, does not have singular outputs. There might be a way by cross referencing into the `jupyter_data` structure to figure stuff out, that could be made more convenient if V3 were to include additional information about where `children` came from. BUT for this downgrade, at this stage of adoption I don't think it is worth it.

But to be clear this is not round-tripping, this means that that content would break for V2 compatible themes trying to consume V3 content, forcing an upgrade (or a custom/extended myst downgrade to be created)

stevejpurves:
@agoose77 I just checked out the branch to change the `assert`s to `throw`s but see you've already done it (I miss read your comment 🤦). On that subject though, I think we should strive to keep this package pure JS and free of all Node, Web and DOM APIs 👍🏼 

Just about to remove the unused console imports. 🧹

packages/myst-migrate/src/v3_outputs.ts
agoose77: I don't think this is a problem:
1. There are no v2 themes that use `myst-migrate`, so adding `myst-migrate` support means just releasing a v3 theme
2. Downgrading only drops `output[i].children`, _not_ `output[i].jupyter_data`, so Bokeh would be fine — see the Bokeh example, there are no output children: 
   ```shell
   curl https://jupyter-book.github.io/example-outputs/bokeh.json | jq .mdast.children[0].children[1].children
   ```

agoose77:
> I think we should strive to keep this package pure JS and free of all Node, Web and DOM APIs 👍🏼

Yes, this is a hard requirement. I don't actually know how easy it is to setup the package so that e.g. adding a Node.js dependency fails the build?

> Just about to remove the unused console imports. 🧹

Thanks! I forgot that step.

agoose77:
OK, you've approved this PR _and_ it isn't in use anywhere. So, we can merge into main to reduce our merge burden!

Issue #1899, opened 2025-03-06
Custom CSS breaks on certain pages for HTML export


agoose77:
## Description
When serving custom CSS with static builds, it sometimes fails to render:

https://luukfroling.github.io/custom-css-test/contentpage vs https://luukfroling.github.io/custom-css-test/second

This is because hydration doesn't match, and the regenerated HTML does not include the base URL.

rowanc1:
Two PRs up should fix this. There was a mismatch between the JS and the HTML, that is now more controlled by JS and has the baseUrl baked in directly. This also means that we don't need to rewrite the html as we put it through and it is closer to the direct output from Remix.

choldgraf:
Just to make it easier to track, I think these are the two PRs @rowanc1 is talking about (correct me if I'm wrong!):

- https://github.com/jupyter-book/mystmd/pull/1901
- https://github.com/jupyter-book/myst-theme/pull/557

rowanc1:
This will be fixed on a redeploy!

https://rowanc1.github.io/custom-css-test/second
https://rowanc1.github.io/custom-css-test/contentpage

PR #1901, opened 2025-03-06
🐛 Ensure CSS has baseURL directly

rowanc1:
See:
- #1899
- jupyter-book/myst-theme#557

### DIFF
--- a/.changeset/small-days-arrive.md
+++ b/.changeset/small-days-arrive.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/small-days-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Ensure that javascript and html bunble match.
```
rowanc1, line 5:
```suggestion
Ensure that javascript and html bundle match.
```

PR #1914, opened 2025-03-12
⌨️ Add command line argument to `myst start` for selecting template

stefanv:
- This command line argument uses the same mechanism as the manifest loader
- The theme manifest currently has a files section that includes build products that are not present when working locally. This commit therefore drops the `files` key from the `template.yml`.
- When determining whether to obtain the manifest locally or download it, we currently check whether `template_dir/node_modules` exists; again, this is not the case when working locally, and merely wanting to point to a different manifest file, so this check was simplified to only check for the presence of the manifest file itself.

Since I do not know the project history, removing `files` and the check for `node_modules` may well violate some known constraint. Happy to be pointed in a better direction!

### DIFF
--- a/packages/myst-cli/src/build/site/template.ts
+++ b/packages/myst-cli/src/build/site/template.ts
@@ -11,7 +11,7 @@ import { castSession } from '../../session/cache.js';
 const DEFAULT_TEMPLATE = 'book-theme';
 const DEFAULT_INSTALL_COMMAND = 'npm install';
 
-export async function getSiteTemplate(session: ISession, opts?: { defaultTemplate?: string }) {
+export async function getSiteTemplate(session: ISession, opts?: { template?: string, defaultTemplate?: string }) {
   const cache = castSession(session);
   const state = cache.store.getState();
   if (cache.$siteTemplate) return cache.$siteTemplate;
@@ -19,7 +19,7 @@ export async function getSiteTemplate(session: ISession, opts?: { defaultTemplat
   const file = selectors.selectCurrentSiteFile(state) ?? session.configFiles[0];
   const mystTemplate = new MystTemplate(session, {
     kind: TemplateKind.site,
-    template: siteConfig?.template ?? opts?.defaultTemplate ?? DEFAULT_TEMPLATE,
+    template: opts?.template ?? siteConfig?.template ?? opts?.defaultTemplate ?? DEFAULT_TEMPLATE,
     buildDir: session.buildPath(),
     errorLogFn: (message: string) => {
       addWarningForFile(session, file, message, 'error', {
@@ -41,7 +41,7 @@ export async function installSiteTemplate(
   session: ISession,
   mystTemplate: MystTemplate,
 ): Promise<void> {
-  if (fs.existsSync(join(mystTemplate.templatePath, 'node_modules'))) return;
+  if (fs.existsSync(mystTemplate.templatePath)) return;
   const toc = tic();
   session.log.info('⤵️  Installing web libraries (can take up to 60 s)');
   await makeExecutable(
--- a/packages/myst-cli/src/build/site/template.ts
+++ b/packages/myst-cli/src/build/site/template.ts
@@ -31,6 +31,7 @@ export async function getSiteTemplate(session: ISession, opts?: { template?: str
         ruleId: RuleId.validSiteConfig,
       });
     },
+    validateFiles: false
   });
   await mystTemplate.ensureTemplateExistsOnPath();
   cache.$siteTemplate = mystTemplate;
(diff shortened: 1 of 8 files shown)


packages/myst-templates/src/template.ts
```diff
@@ -65,8 +65,15 @@ class MystTemplate {
...
+
+    interface TemplateYaml {
+      files: string[];
+      [key: string]: any;
+    };
     const content = fs.readFileSync(templateYmlPath).toString();
-    return yaml.load(content);
+    const yamlData : TemplateYaml = yaml.load(content) as TemplateYaml;
+    const { files, ...filteredYaml } = yamlData;
+    return filteredYaml;
```
rowanc1, line 76: The `files` are used primarily in the tex/typst templates to copy assets like styles, images, etc. over into a new build folder. So removing files here will break that.

I am assuming that the `package-lock.json` is complaining (not sure on your local setup)? We could pass a flag, like `ignoreTemplateFiles`, into the MystTemplate class and do this logic if true. e.g. if the `opts?.template` is passed into `getSiteTemplate` above.

rowanc1 (commented):
A few suggestions/changes.

Beyond that, to make this merge ready it would be awesome to see a changeset (either in the github comment or `npm run changeset` locally).

Not sure if we have a docs page really on this at the moment, or if your other PR is the best place to pick up on this option and document it there.

packages/myst-templates/src/template.ts
```diff
@@ -65,8 +65,15 @@ class MystTemplate {
...
       throw new Error(`The template yml at "${templateYmlPath}" does not exist`);
     }
+
+    interface TemplateYaml {
+      files: string[];
+      [key: string]: any;
+    };
     const content = fs.readFileSync(templateYmlPath).toString();
-    return yaml.load(content);
+    const yamlData : TemplateYaml = yaml.load(content) as TemplateYaml;
```
rowanc1, line 74: I think this is the lint failure:

```suggestion
    const yamlData: TemplateYaml = yaml.load(content) as TemplateYaml;
```

packages/myst-cli/src/build/site/template.ts
```diff
@@ -41,7 +41,7 @@ export async function installSiteTemplate(
   session: ISession,
   mystTemplate: MystTemplate,
 ): Promise<void> {
-  if (fs.existsSync(join(mystTemplate.templatePath, 'node_modules'))) return;
+  if (fs.existsSync(mystTemplate.templatePath)) return;
```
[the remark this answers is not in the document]
stefanv: There is node `node_modules` directory under `theme/book`, so this early return does not happen.

packages/myst-templates/src/template.ts
stefanv: Yes, no `page-lock.json`, no `build` dir, etc.

packages/myst-templates/src/template.ts
stefanv: Heh, so if you run `npm run format` and then `npm run lint`, you get 12 errors! So doesn't look like eslint and prettier are configured to work together correctly.

packages/myst-templates/src/template.ts
stefanv: Added a flag `validateFiles` in the appropriate places, which seems to do the trick.

packages/myst-cli/src/build/site/template.ts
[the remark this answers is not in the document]
stefanv: Addressed by skipping validation for `--template` specified templates.

packages/myst-cli/src/build/site/template.ts
```diff
@@ -31,6 +34,7 @@ export async function getSiteTemplate(session: ISession, opts?: { defaultTemplat
         ruleId: RuleId.validSiteConfig,
       });
     },
+    validateFiles: false,
```
rowanc1, line 37: Can we only turn this on if the template is specified?

```suggestion
    validateFiles: opts?.template ? false : true,
```

packages/myst-cli/src/build/site/template.ts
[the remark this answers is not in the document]
rowanc1: thank you!

packages/myst-templates/src/template.ts
rowanc1: I think just doing this logic when the template is specified makes sense, left a different comment on that!

packages/myst-templates/src/template.ts
rowanc1: CI is checking `npm run lint` and `npm run lint:format`. They should be working locally and do work together... ?!

I am using vscode with eslint/prettier default extensions, that formats on save.

stefanv: Would you be open to a git hook to achieve something similar?

rowanc1:
🚀🚀🚀

Thank you!

PR #1882, opened 2025-02-28
📓 Add `ipynb` as export format

rowanc1:
cc @mmcky 

## To Do
- [ ] Fix redundant `+++` markers in Markdown cells.

PR #1915, opened 2025-03-12
Complete ipynb export support

kp992:
Hi, 

I am trying to complete the feature in #1882. And building my first PR on top of it. I tried to add the `ipynb` writer by visiting the blocks and merging all the markdown blocks to a single cell until a code cell appears. Please review the PR.

I have a small question:
How do I test it using a CLI locally for a single file? I currently do `npm run test` from `myst-to-ipynb` directory. Is there a way I can test some `ast.yml` or `.md` file to `ipynb`?

I am sorry if this is a very newbie question and documented somewhere.

### DIFF
--- a/packages/myst-to-ipynb/src/index.ts
+++ b/packages/myst-to-ipynb/src/index.ts
@@ -12,42 +12,25 @@ function sourceToStringList(src: string): string[] {
   return lines;
 }
 
-function markdownString(file: VFile, md_cells: Block[]) {
-  const md = writeMd(file, { type: 'root', children: md_cells }).result as string;
-  return {
-    cell_type: 'markdown',
-    metadata: {},
-    source: sourceToStringList(md),
-  };
-}
-
 export function writeIpynb(file: VFile, node: Root, frontmatter?: PageFrontmatter) {
-  const cells = [];
-  const md_cells: Block[] = [];
-
-  for (const block of node.children as Block[]) {
+  const cells = (node.children as Block[]).map((block: Block) => {
     if (block.type === 'block' && block.kind === 'notebook-code') {
-      if (md_cells.length != 0) {
-        cells.push(markdownString(file, md_cells));
-        md_cells.length = 0;
-      }
       const code = select('code', block) as Code;
-      cells.push({
+      return {
         cell_type: 'code',
         execution_count: null,
         metadata: {},
         outputs: [],
         source: sourceToStringList(code.value),
-      });
-    } else {
-      md_cells.push(block);
+      };
     }
-  }
-
-  if (md_cells.length != 0) {
-    cells.push(markdownString(file, md_cells));
-    md_cells.length = 0;
-  }
+    const md = writeMd(file, { type: 'root', children: [block] }).result as string;
+    return {
+      cell_type: 'markdown',
+      metadata: {},
+      source: sourceToStringList(md),
+    };
+  });
 
   const ipynb = {
     cells,
@@ -59,6 +42,7 @@ export function writeIpynb(file: VFile, node: Root, frontmatter?: PageFrontmatte
     nbformat: 4,
     nbformat_minor: 2,
   };
+
   file.result = JSON.stringify(ipynb, null, 2);
   return file;
 }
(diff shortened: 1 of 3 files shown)


packages/myst-to-ipynb/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    metadata: {},
+    source: md,
+  };
+}
+
+export function writeIpynb(file: VFile, node: Root, frontmatter?: PageFrontmatter) {
+  const cells = [];
+  const md_cells: Block[] = [];
+
+  for (const block of node.children as Block[]) {
```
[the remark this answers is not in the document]
agoose77: Hi @kp992, thanks for your PR!

`block` nodes are _usually_ top level -- particularly in the case of notebooks.

The test-case in `myst-execute` is much less opinionated about what a valid notebook looks like. We support non-block top-level content at parse time, it just gets transformed under a top-level block by the end.

packages/myst-to-ipynb/src/index.ts
```diff
@@ -6,31 +6,43 @@ import type { PageFrontmatter } from 'myst-frontmatter';
...
-    return {
-      cell_type: 'markdown',
-      metadata: {},
-      source: sourceToStringList(md),
-    };
-  });
+  }
+
+  if (md_cells.length != 0) {
+    cells.push(markdownString(file, md_cells));
```
agoose77, line 42: @kp992 this is a "policy" decision to fuse Markdown cells. Is there a particular reason for doing this? With the base PR #1882, if you ingest an `ipynb` file and export an `ipynb`, then you'll end up with the same number of cells (perhaps the frontmatter might be moved, though). 

If you have a use-case for fusing Markdown cells, I'd suggest adding that as an option!

agoose77:
> @rowanc1 I have added/fixed the test cases from your PR. Should I work on capturing more metadata and output of the cells in the same PR or the follow up is okay?

I've changed the base of this PR so that it will be merged into the original PR. If the original PR changes, we'll need to rebase this one first. So, the goal should be to get this one over the line ASAP by keeping it small! The test changes are welcome — I haven't looked at them in any detail, but it would be wonderful to get the test suite passing. If you have time to address my query about the fusing of Markdown cells, we can then look at what needs to be done to close this out and open a follow-up PR if needs be!

packages/myst-to-ipynb/src/index.ts
agoose77: I still think these changes need reverting. Consider the following Markdown notebook
````markdown
---
kernelspec:
  name: python3
  display_name: Python 3
---

```{code-cell} python3
import sys
```

+++ { "kind": "notebook-content" }
Desc 1

+++ { "kind": "notebook-content" }

Desc 2

```{code-cell} python3
import os
```
````

It's the MD equivalent of having four cells: 
- code
- md
- md
- code

In this PR, that gets flattened to 
- code
- md
- code

(and the `myst-to-md` also writes the block separator `+++` inside the cell, which is not quite right).

Can you elaborate on why you made this particular change? Did the original PR fail to convert something properly?

File: packages/myst-to-ipynb/tests/run.spec.ts [hunk 3]
```diff
@@ -3,11 +3,11 @@ import fs from 'node:fs';
 import path from 'node:path';
 import yaml from 'js-yaml';
 import { unified } from 'unified';
-import mystToMd from '../src';
+import writeIpynb from '../src';
 
 type TestCase = {
   title: string;
-  markdown: string;
+  ipynb: string;
```
agoose77, line 10: We _probably_ don't care about the _exact_ text representation of the `ipynb` file on disk. Could you change the `ipynb` entry to be a `Record<string, any>?`

Then, you'd need to update the tests to be

````yaml
- title: link reference
  mdast:
   type: root
   children:
     - type: text
       value: markdown
  ipynb: 
    cells:
      - 
````
instead of 

````yaml
- title: link reference
  mdast:
   type: root
   children:
     - type: text
       value: markdown
  ipynb: |-
      {
        "cells": [
````

This would enable us to write the regression test in YAML (less verbose) _and_ to choose to omit parts of the output that we don't want to test (e.g. if there were timestamps)

agoose77 (changes requested):
Thanks for adding the tests @kp992! I made some comments about some of the changes that we would probably like to revert, etc.

PR #1919, opened 2025-03-14
✅ Add husky and lint-staged to help devs fix code before pushing

stefanv:
For folks without VSCode, this could help by providing a git pre-commit hook that is run via eslint and prettier to fix things up.

----

Running `npx husky` installs the git pre-commit hook.

From then on, whenever you make a commit, eslint and prettier will be invoked, re-formatting the commit automatically.

If you want to revert this behavior, remove `core.hooksPath` from `.git/config`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "version": "changeset version && npm install",
     "publish": "npm run clean && npm run build -- --force && changeset publish && git push --follow-tags",
     "publish:ci": "npm run build && changeset publish && git push --follow-tags",
-    "prepare": "husky"
+    "install-precommit-hooks": "husky"
   },
   "devDependencies": {
     "@changesets/cli": "^2.26.1",
--- a/package.json
+++ b/package.json
@@ -17,13 +17,16 @@
     "changeset": "changeset",
     "version": "changeset version && npm install",
     "publish": "npm run clean && npm run build -- --force && changeset publish && git push --follow-tags",
-    "publish:ci": "npm run build && changeset publish && git push --follow-tags"
+    "publish:ci": "npm run build && changeset publish && git push --follow-tags",
+    "prepare": "husky"
   },
   "devDependencies": {
     "@changesets/cli": "^2.26.1",
     "@types/node": "^20.2.5",
     "esbuild": "^0.17.19",
     "eslint-config-curvenote": "^0.0.4",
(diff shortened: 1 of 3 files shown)


agoose77:
I'm _not_ merging this because it's a workflow PR; @fwkoch are you happy to merge?

stefanv:
Also happy to iterate on feedback! Thanks for taking a look.

package.json
```diff
@@ -17,13 +17,16 @@
     "changeset": "changeset",
     "version": "changeset version && npm install",
     "publish": "npm run clean && npm run build -- --force && changeset publish && git push --follow-tags",
-    "publish:ci": "npm run build && changeset publish && git push --follow-tags"
+    "publish:ci": "npm run build && changeset publish && git push --follow-tags",
+    "prepare": "husky"
```
fwkoch, line 21: `prepare` is automatically run durning `npm install` - that means these precommit hooks will always run unless users explicitly opt out. This might be off-putting if someone is particular about their git workflow (or they already have linting from VSCode or the build is running in CI...). Can we just change this to a different command name, making it easy, but not automatic, to opt in? Maybe something like:

```suggestion
    "install-precommit-hooks": "husky"
```

fwkoch (changes requested):
This is great! CI failures for linting are annoying for developers who are not setup to run the linting themselves. This PR just makes it easier for them to do that.

My only request is to make husky "easy opt-in" rather than installed automatically and less easy to opt out of - let me know if you agree with my suggested change.

package.json
stefanv: Ah, yes, definitely did not mean to trigger it automatically!

choldgraf (commented):
This looks good to me, with one request: can you document this? I think [just a little section in the developer guide](https://mystmd.org/guide/developer) would be fine. Something like `### How to set up the linting pre-commit hook` and copy/paste the brief instructions that you gave. I think that would help make this more maintainable and discoverable for those who aren't familiar

stefanv:
> can you document this?

Good idea! Done.

choldgraf (approved):

choldgraf:
I believe that @stefanv implemented all of @fwkoch 's suggestions so I'm going to merge!

choldgraf (approved):

PR #1922, opened 2025-03-16
feat(tex): support page parameter for includegraphics with multi-page pdf

fk128:
Add support for selecting specific pages in multi-page PDFs

In LaTeX, when including a multi-page PDF as a graphic, it's possible to specify a page number:
```
 \includegraphics[width=1.0\textwidth,page=3]{figures/my_pic.pdf}
```

However, the page parameter is currently ignored, causing all the links to point only to the first page, even though ImageMagick extracts and converts all pages.

This PR adds support for the page parameter, ensuring that the correct page is selected when converting multi-page PDFs.

### DIFF
--- a/packages/myst-cli/src/utils/imagemagick.ts
+++ b/packages/myst-cli/src/utils/imagemagick.ts
@@ -155,7 +155,7 @@ export async function convert(
       options?.trim ? ' -trim' : ''
     } ${output}`;
 
-    session.log.info(`Executing: ${executable}`);
+    session.log.debug(`Executing: ${executable}`);
     const exec = makeExecutable(executable, createImagemagikLogger(session));
     try {
       await exec();
--- a/packages/myst-cli/src/utils/imagemagick.ts
+++ b/packages/myst-cli/src/utils/imagemagick.ts
@@ -138,12 +138,12 @@ export async function convert(
   session: ISession,
   input: string,
   writeFolder: string,
-  options?: { trim?: boolean },
+  options?: { trim?: boolean; page?: number },
 ) {
   if (!fs.existsSync(input)) return null;
   const { name, ext } = path.parse(input);
   if (ext !== inputExtension) return null;
-  const filename = `${name}${outputExtension}`;
+  const filename = `${name}${options?.page ? '-' + options.page : ''}${outputExtension}`;
   const output = path.join(writeFolder, filename);
   const inputFormatUpper = inputExtension.slice(1).toUpperCase();
   const outputFormatUpper = outputExtension.slice(1).toUpperCase();
@@ -151,10 +151,11 @@ export async function convert(
     session.log.debug(`Cached file found for converted ${inputFormatUpper}: ${input}`);
     return filename;
   } else {
-    const executable = `${imageMagickCommand()} -density 600 -colorspace RGB ${input}${
+    const executable = `${imageMagickCommand()} -density 600 -colorspace RGB ${input}${options?.page ? '[' + options.page + ']' : ''}${
       options?.trim ? ' -trim' : ''
     } ${output}`;
-    session.log.debug(`Executing: ${executable}`);
+
+    session.log.info(`Executing: ${executable}`);
     const exec = makeExecutable(executable, createImagemagikLogger(session));
     try {
       await exec();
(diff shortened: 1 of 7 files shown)


packages/myst-cli/src/utils/imagemagick.ts
```diff
@@ -138,23 +138,24 @@ export async function convert(
...
     session.log.debug(`Cached file found for converted ${inputFormatUpper}: ${input}`);
     return filename;
   } else {
-    const executable = `${imageMagickCommand()} -density 600 -colorspace RGB ${input}${
+    const executable = `${imageMagickCommand()} -density 600 -colorspace RGB ${input}${options?.page ? '[' + options.page + ']' : ''}${
       options?.trim ? ' -trim' : ''
     } ${output}`;
-    session.log.debug(`Executing: ${executable}`);
+
+    session.log.info(`Executing: ${executable}`);
```
rowanc1, line 158:
```suggestion
    session.log.debug(`Executing: ${executable}`);
```

packages/tex-to-myst/src/figures.ts
```diff
@@ -1,7 +1,8 @@
 import type { GenericNode } from 'myst-common';
 import { u } from 'unist-builder';
 import type { Handler, ITexParser } from './types.js';
-import { getArguments, texToText } from './utils.js';
+import { getArguments, extractParams, texToText } from './utils.js';
+import { group } from 'console';
```
rowanc1, line 5: I don't think this is used?

packages/tex-to-myst/src/figures.ts
```diff
@@ -45,18 +46,33 @@ const FIGURE_HANDLERS: Record<string, Handler> = {
...
+    if (params.width) {
+      if (typeof params.width === 'number') {
+        params.width = `${Math.round(params.width * 100)}%`;
+      } else {
+        delete params.width; // If width is a string, we don't know what it is, so we ignore it
+      }
+    }
+    if (params.page) {
+      if (typeof params.page === 'number') {
+        params.page = Number(params.page) - 1; // Convert to 0-based for imagemagick
```
rowanc1, line 70: Do we need to round or parse this or have something similar to `Number.isFinite(Number.parseFloat(params.page))`?

rowanc1 (commented):
Nice, looks great. Left a few minor things.

Could you also add `page` with a comment here:

https://github.com/executablebooks/mystmd/blob/main/packages/myst-spec-ext/src/types.ts#L117

e.g.

```
export type Image = SpecImage & {
  urlSource?: string;
  urlOptimized?: string;
  height?: string;
  placeholder?: boolean;
  /** Optional page number for PDF images, this ensure the correct page is extracted when converting to web and translated to LaTeX */
  page?: number;
};
```

agoose77:
@rowanc1 I'm wondering whether it makes sense to bring page awareness to our normal AST. Should we not treat this as a post-tex transform, and rewrite the image node with the proper figure?

packages/tex-to-myst/src/utils.ts
```diff
@@ -138,6 +138,23 @@ export function getArguments(
   );
 }
 
+export function extractParams(args: { content: string }[]): Record<string, string | number> {
```
fwkoch, line 141: This seems like it has potential for wider use across other macros/parameter types. 👍

fwkoch (commented):
I just looked at this PR now for the first time - looks great, and I think we should work on getting it landed.

First, just to summarize my understanding, this has two parts: (1) On parsing a tex file, this looks at image params and adds `page` to the `image` node (alongside `width`, which was previously supported). (2) On image conversion, if `page` is present, imagemagick extracts the correct page.

A couple points:
- @agoose77 - I think your concern was around adding `page` to the image node. It feels a little clutter-y since it only applies in a few specific cases. Thinking through the alternative a bit: If we do not store `page` on the node, we would need to do initial image processing at parse time, pulling out the specific page as a separate file. This separate file would still need final conversion to correct format. This means we have extra intermediate files stored on users' machines.
- I noticed if this page-specific PDF image is selected as the "implicit" thumbnail, it also causes problems: `thumbnail` is just a file, not an entire image node. That means `thumbnail` does not get the `page` value, and is stuck trying to process/convert the full PDF file. There are certainly ways around this where we could get these thumbnails working... but I also think we might not need to worry about this edge case. An "explicit" thumbnail can always be set if there are errors with the "implicit" thumbnail.

My inclination is to keep this as-is. It works nicely and it's relatively simple. The only downside is an extra `page` field on image nodes that will usually be undefined.

PR #1925, opened 2025-03-19
Cannot load plugins without proj config

stefanv:


agoose77:
@stefanv can you share the thought process behind why this PR is necessary? From a first glance, it means that `loadPlugins` may silently _not_ load plugins, which is not ideal.

agoose77:
Can you publish a reproducer? The myst docs use plugins, and the content server loads them just fine.

stefanv:
I added the scienceicons extension to the landing-pages repo. I can share the exact repo, and log, showing that the plugin doesn't load, when I'm back at a computer.

agoose77:
Hi @stefanv, any update on a reproducer here?

stefanv:
Looks like an interaction between using parts + plugin.

Try this content repo: https://github.com/stefanv/cryocloud-landing-demo/tree/no-plugin-reproducer

And then use this PR branch vs main branch of mystmd. You'll (hopefully) see that the plugins load with one, but not the other.

stefanv:
Superseded by #2093

PR #1930, opened 2025-03-25
Clarify BASE_URL instructions in GH action for custom domains.

fperez:
Extends the existing comment to help users who may change deployment options after their original setup, as they may encounter CSS issues.

I just spent an hour tracking this down - while the comment was there as a hint, I wasn't quite sure this was the problem. I just added a slightly more explicit note as my situation may happen again: a user who initially tests as site on `<username>.github.io`  and later decides to move to `theirwebsite.org`, sets everything up correctly for custom domains on GH and their DNS machinery, but gets a borked website with zero CSS.

I'm also happy to add more to the docs if desired, though I didn't quite stumble due to docs issues, it really was this action that threw me for a loop.

### DIFF
--- a/packages/myst-cli/src/init/gh-actions/index.ts
+++ b/packages/myst-cli/src/init/gh-actions/index.ts
@@ -26,6 +26,12 @@ on:
     branches: [${defaultBranch}]
 env:
   # \`BASE_URL\` determines the website is served from, including CSS & JS assets
+  #
+  # The code below was automatically set when you first ran \`myst init --gh-pages\`,
+  # but the correct values may change if you update where you deploy your site. 
+  # For repositories served from custom domains (e.g. "you.com"), you may need to set
+  # this value to "BASE_URL: ''".
+  #
   # You may need to change this to \`${
     isGithubIO ? 'BASE_URL: /${{ github.event.repository.name }}' : "BASE_URL: ''"
   }\`
--- a/packages/myst-cli/src/init/gh-actions/index.ts
+++ b/packages/myst-cli/src/init/gh-actions/index.ts
@@ -25,8 +25,6 @@ on:
     # Runs on pushes targeting the default branch
     branches: [${defaultBranch}]
 env:
-  # \`BASE_URL\` determines the website is served from, including CSS & JS assets
-  #
   # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
   # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
   # If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''".
--- a/packages/myst-cli/src/init/gh-actions/index.ts
+++ b/packages/myst-cli/src/init/gh-actions/index.ts
@@ -27,10 +27,9 @@ on:
 env:
   # \`BASE_URL\` determines the website is served from, including CSS & JS assets
   #
-  # The code below was automatically set when you first ran \`myst init --gh-pages\`,
-  # but the correct values may change if you update where you deploy your site. 
-  # For repositories served from custom domains (e.g. "you.com"), you may need to set
-  # this value to "BASE_URL: ''".
+  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
+  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
+  # If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''".
   #
   # You may need to change this to \`${
     isGithubIO ? 'BASE_URL: /${{ github.event.repository.name }}' : "BASE_URL: ''"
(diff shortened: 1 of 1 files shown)


fperez:
I now wonder if instead of that more verbose comment, instead a better solution would simply be to generate the file as:

```typescript
  # `BASE_URL` determines the website is served from, including CSS & JS assets
  BASE_URL: /${{ github.event.repository.name }} # For user.github.io/repo
  # BASE_URL: ''  # For custom domains
```

With only one option uncommented, so if `isGithubIO` is false, then it would be instead

```typescript
  # `BASE_URL` determines the website is served from, including CSS & JS assets
  # BASE_URL: /${{ github.event.repository.name }} # For user.github.io/repo
   BASE_URL: ''  # For custom domains
```

But leaving both lines for the user to uncomment one only.

If this approach is preferred, I'm happy to amend the PR to go that route instead.

stefanv:
The latter approach seems more helpful to the user. But, either way, the language doesn't parse right now:

```
  # `BASE_URL` determines the website is served from, including CSS & JS assets
  # You may need to change this to `BASE_URL: ''`
  BASE_URL: /${{ github.event.repository.name }}
```

Maybe we need something like:

```
    `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
    E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
    If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''".
```

stefanv:
Related question: do we currently allow URLs to resources to be relative, so that BASE_URL isn't necessary under most circumstances?

choldgraf:
I like Stefan's suggestion. My only twist would be to put this into the documentation, and then the Yaml could just be a link to it (eg "see xxx for more context"). But I don't think that should block this if others disagree, we should just merge it when @stefanv approves

fperez:
@choldgraf - I'm happy to add to the docs, but I hesitate to hardcode links to specific docs in auto-generated files... It's very easy for those files to be long-lived, and they can easily fall out of sync with the layout of the docs. I think information that helps someone get out of a jam should be actionable right there, if possible.

An _extra_ link to docs with more explanation is good, but I wouldn't do it _instead_ of having the file be self-explanatory.

agoose77:
Users are likely to need to modify this workflow, and at the same time it is not something that needs to worry about being terse or concise. I think we should merge this as-is!

choldgraf:
@agoose77 fair point! I'm happy to see this merged if you think the text is accurate!

packages/myst-cli/src/init/gh-actions/index.ts
```diff
@@ -26,6 +26,12 @@ on:
     branches: [${defaultBranch}]
 env:
   # \`BASE_URL\` determines the website is served from, including CSS & JS assets
+  #
+  # The code below was automatically set when you first ran \`myst init --gh-pages\`,
+  # but the correct values may change if you update where you deploy your site. 
+  # For repositories served from custom domains (e.g. "you.com"), you may need to set
+  # this value to "BASE_URL: ''".
```
choldgraf, line 33:
```suggestion
  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
  # If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''".
```

packages/myst-cli/src/init/gh-actions/index.ts
```diff
@@ -26,6 +26,11 @@ on:
     branches: [${defaultBranch}]
 env:
   # \`BASE_URL\` determines the website is served from, including CSS & JS assets
+  #
+  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
+  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
```
choldgraf, line 31:
```suggestion
  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
```

packages/myst-cli/src/init/gh-actions/index.ts
```diff
@@ -25,7 +25,10 @@ on:
...
     branches: [${defaultBranch}]
 env:
-  # \`BASE_URL\` determines the website is served from, including CSS & JS assets
+  # `BASE_URL` determines, relative to the root of the domain, the URL that your site is served from.
+  # E.g., if your site lives at `https://mydomain.org/myproject`, set `BASE_URL=/myproject`.
+  # If, instead, your site lives at the root of the domain, at `https://mydomain.org`, set `BASE_URL=''".
+  #
   # You may need to change this to \`${
     isGithubIO ? 'BASE_URL: /${{ github.event.repository.name }}' : "BASE_URL: ''"
   }\`
```
choldgraf, line 34:
```suggestion
```

choldgraf (approved):
Given that we had two 👍s for @stefanv 's suggestion, including from the PR author @fperez , I applied Stefan's suggestion to the PR and think it's ready to go. Could @stefanv confirm that the current state of the PR reflects what he was suggesting? If so I think we should merge

fperez:
Thanks a lot @choldgraf for these fixes! Happy to see it in with these changes.

stefanv:
Looks good to me, thanks Chris!

PR #1937, opened 2025-03-31
📥 Support embedding raw code with `remove-input` flag

### LINKED ISSUE #1932: Embed inline shorthand produces different results to explicit directive.
The [docs](https://mystmd.org/guide/embed) describe the inline and explicit syntaxes for `{embed}` as equivalent. However, I have a document with:

```
# Explicit Embed:
:::{embed} #mylabel
:::

# Inline Embed:
![](#mylabel)
```

and elsewhere in another file (included in the toc), I have:

````
:::{note} My info block
:label: mylabel
**A literal block:**

```
Line 1
Line 2
etc...
```
:::
````

The two forms of embedding are supposed to produce the same results, but I'm getting:

Image

### PR DESCRIPTION
**Pull request by agoose77**
Fixes #1932

### DIFF
--- a/packages/myst-cli/src/transforms/embed.ts
+++ b/packages/myst-cli/src/transforms/embed.ts
@@ -17,6 +17,7 @@ import {
   fileError,
   isTargetIdentifierNode,
   selectMdastNodes,
+  NotebookCell,
 } from 'myst-common';
 import type { Dependency, Embed, Container, CrossReference, Link } from 'myst-spec-ext';
 import { selectFile } from '../process/file.js';
@@ -39,9 +40,19 @@ function mutateEmbedNode(
     });
   }
   if (targetNode && node['remove-input']) {
-    targetNode = filter(targetNode, (n: GenericNode) => {
-      return n.type !== 'code' || n.data?.type === 'output';
-    });
+    targetNode = filter(
+      targetNode,
+      (
+        n: GenericNode,
+        index: number | null | undefined,
+        parent: GenericNode | undefined | null,
+      ) => {
+        return (
+          !(n.type === 'code' && parent?.type === 'block' && parent?.kind === NotebookCell.code) ||
+          n.data?.type === 'output'
+        );
+      },
+    );
   }
   selectAll('[identifier],[label],[html_id]', targetNode).forEach((idNode: GenericNode) => {
     // Non-target nodes may keep these properties

### CODE REVIEW

**Review by choldgraf — APPROVED** [review]
I added some brief documentation to clarify this for future reference.

This looks fine to merge to me, assuming the logic you're following to detect a notebook cell is correct. My one suggestion would be to add a test to make sure that this is behaving as expected, but I don't think it's strictly required if you are confident that this fixes the bug.

File: packages/myst-cli/src/transforms/embed.ts [hunk 1]
```diff
@@ -35,13 +36,26 @@ function mutateEmbedNode(
...
+        n: GenericNode,
+        index: number | null | undefined,
+        parent: GenericNode | undefined | null,
+      ) => {
+        // If we have a code cell with a NotebookCell parent, assume its an input cell and remove.
+        return !(
+          n.type === 'code' &&
+          parent?.type === 'block' &&
+          parent?.kind === NotebookCell.code
+        );
```
agoose77: The previous logic had a special case for code-nodes that were tagged as outputs. Now, we want to be even stricter --- only code nodes that are children of code-cells _and_ not tagged as outputs.

packages/myst-cli/src/transforms/embed.ts
```diff
@@ -35,13 +36,28 @@ function mutateEmbedNode(
...
+      (
+        n: GenericNode,
+        index: number | null | undefined,
+        parent: GenericNode | undefined | null,
+      ) => {
+        // Code nodes that are children of notebook cells and _not_ tagged as outputs
+        // should be removed
+        return !(
+          n.type === 'code' &&
+          n.data?.type !== 'output' &&
```
agoose77, line 55: @fwkoch I don't think I need this branch actually, because `reduceOutputs` is always called _after_ this transform runs? So there should never be such data in the file.

packages/myst-cli/src/transforms/embed.ts
```diff
@@ -35,13 +36,28 @@ function mutateEmbedNode(
   const { url, dataUrl, targetFile, sourceFile } = opts ?? {};
   if (targetNode && node['remove-output']) {
     targetNode = filter(targetNode, (n: GenericNode) => {
+      // After reduction, 'output' nodes may be replaced by their children which are then tagged as outputs
       return n.type !== 'output' && n.data?.type !== 'output';
```
agoose77, line 40: See lower comment about `reduceOutputs`

agoose77:
@fwkoch I am not 100% clear on the `data.input` tagging -- where do we set this, is it something we care about in the embed transform? Or does it happen later? Same for `data.output` tag!

When you get a moment, could you take this over the line and add in-line comments to explain for future reference?

docs/embed.md
```diff
@@ -60,7 +60,9 @@ For example, the following references the admonitions list in [](admonitions.md)
 
 ### The `![](#embed)` short-hand
 
-The embedding markdown shorthand lets you quickly embed content using the Markdown image syntax (see more about [images](./figures.md)).
+The embedding shorthand lets you embed content using the Markdown image syntax (see more about [images](./figures.md)).
+This **removes the input cell** if you are [embedding from a Jupyter notebook](./reuse-jupyter-outputs.md).
```
fwkoch, line 64: 👍

packages/myst-cli/src/transforms/embed.ts
fwkoch: I think you are correct that `reduceOutputs` always comes after this. However, I think we should just leave this check in place. At worst, it's just a no-op, but, at best, it would allow tagging nodes as an `output` in contexts other than the standard Jupyter Notebook processing (e.g. a custom transform plugin).

fwkoch (approved):
Ok, finally took the time to understand the purpose of this PR and what it is fixing. The scope is actually quite small, simple, and safe; it just makes `remove-input` a bit stricter so it's actually "remove input" rather than "remove code." Easy to reproduce and see the improvement.

@agoose77 - Regarding your question about `data.input`, I don't see this anywhere, either in this PR or in `reduceOutputs` or anything, so I don't think we need to worry about it? We _could_ add it in as a way for users to tag specific, non-code nodes as `input`? But that's a separate feature. Regarding `output` tags - I made a comment, but I think it doesn't hurt to just leave as-is.

Also, I noticed an edge case - if you try to embed a notebook code cell using `` that has not been executed (i.e. just input, no output) it will still be removed, leaving nothing. I think this is fine, though, especially with the sentence added to the docs explicitly stating notebook input will be removed.

Sorry this review took 3 months 😬

PR #1942, opened 2025-04-01
🪄 Improve error handling during cross reference resolution

agoose77:
I created this PR when working on output upgrading a few months ago. It's a simple change to make it easier to determine why xref processing fails.

### DIFF
--- a/packages/myst-cli/src/transforms/crossReferences.ts
+++ b/packages/myst-cli/src/transforms/crossReferences.ts
@@ -39,36 +39,39 @@ async function fetchMystData(
   urlSource: string | undefined,
   vfile: VFile,
 ) {
-  let note: string;
-  if (dataUrl) {
-    const filename = mystDataFilename(dataUrl);
-    const cacheData = loadFromCache(session, filename, { maxAge: XREF_MAX_AGE });
-    if (cacheData) {
-      return JSON.parse(cacheData) as MystData;
-    }
-    try {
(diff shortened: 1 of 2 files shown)


packages/myst-cli/src/transforms/crossReferences.ts
```diff
@@ -39,36 +39,39 @@ async function fetchMystData(
...
+  const filename = mystDataFilename(dataUrl);
+  const cacheData = loadFromCache(session, filename, { maxAge: XREF_MAX_AGE });
+  if (cacheData) {
+    return JSON.parse(cacheData) as MystData;
+  }
+  let data: MystData;
+  try {
+    const resp = await session.fetch(dataUrl);
+    if (!resp.ok) {
+      return onError('Could not fetch data from URL');
```
agoose77, line 65: Error (2) -- If the HTTP return code is not in the 200-299 range

packages/myst-cli/src/transforms/crossReferences.ts
```diff
@@ -39,36 +39,39 @@ async function fetchMystData(
...
     }
-    note = 'Could not load data from external project';
-  } else {
-    note = 'Data source URL unavailable';
+    data = (await resp.json()) as MystData;
+  } catch {
+    return onError('Could not load fetched data');
```
agoose77, line 69: Error (3) -- Unable to load the JSON body

PR #1949, opened 2025-04-04
🔙 Move execution transform earlier in mdast pipeline

agoose77:
Subsequent PRs that look to introduce Markdown output support from code-cells will want to make use of as much of the AST transform pipeline as possible. This PR moves kernel execution before much of the basic transforms, so that we can later do this. It should not break / change anything.

### DIFF
--- a/.changeset/weak-bananas-cover.md
+++ b/.changeset/weak-bananas-cover.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


rowanc1:
@fwkoch can you take a quick think on this if there will be any impacts?

agoose77:
@fwkoch thanks for pointing that out. I had not considered that particular avenue. We could always add more granular stages (although, yes, the PR in #1699 would solve that too!)

PR #1959, opened 2025-04-10
🔍 Do not treat wiki permalinks as wiki-links for now

### LINKED ISSUE #1954: Wiki link heuristic does not understand permalinks
As per [this comment](https://github.com/jupyter-book/mystmd/issues/1336#issuecomment-2593403732):
   > Our wiki heuristic doesn't understand permalinks to wiki content: https://en.wikipedia.org/w/index.php?title=Tidy_data&oldid=962241815 fails:

error:
> ⛔️ reproducible-research/rdm/rdm-spreadsheets.md:52 Wikipedia pagenames should not contain "/" in link: https://en.wikipedia.org/w/index.php?title=Tidy_data&oldid=962241815
   Only point to the final page name, do not include any other parts of the Wikipedia URL.

Links: 
- https://github.com/jupyter-book/mystmd/issues/1336
- https://github.com/the-turing-way/the-turing-way/pull/4057

agoose77:
To fix #1954, we need to understand permalinks. I don't have time _right now_ to add support for transforming and understanding permalinks, as this will likely require support in myst-theme too. This PR, instead, just silences any warnings by not touching permalinks and instead treating them as regular old links.

### DIFF
--- a/packages/myst-transforms/src/links/plugin.ts
+++ b/packages/myst-transforms/src/links/plugin.ts
@@ -20,7 +20,7 @@ export function formatLinkText(link: Link) {
   if (link.children?.length !== 1 || link.children[0].type !== 'text') return;
   const url = link.children[0].value;
   // Add an exception for wiki transforms links.
-  if (url.length < 20 || url.match(/\s/) || url.startsWith('wiki:')) return;
+  if (url.length < 20 || url.match(/\s/)) return;
   if (url.includes('​')) return;
   // Split the URL into an array to distinguish double slashes from single slashes
   const doubleSlash = url.split('//');
@@ -48,7 +48,10 @@ export function linksTransform(mdast: GenericParent, file: VFile, opts: Options)
     const result = transform?.transform(link, file);
     // The link transform may compare the text
     // Formatting adds no-width spaces to some URLs
-    formatLinkText(link);
+    // Don't format text if transform already does this
+    if (!(transform && transform.formatsText)) {
+      formatLinkText(link);
+    }
     if (!transform || result === undefined) return;
     if (result) {
       delete link.error;
--- a/packages/myst-transforms/src/links/types.ts
+++ b/packages/myst-transforms/src/links/types.ts
@@ -5,6 +5,7 @@ import type { Inventory } from 'intersphinx';
 
 export interface LinkTransformer {
   protocol?: string;
+  formatsText?: boolean;
   test: (uri?: string) => boolean;
   transform: (link: Link, file: VFile) => boolean;
 }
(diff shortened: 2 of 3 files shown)


packages/myst-transforms/src/links/plugin.ts
```diff
@@ -20,7 +20,7 @@ export function formatLinkText(link: Link) {
   if (link.children?.length !== 1 || link.children[0].type !== 'text') return;
   const url = link.children[0].value;
   // Add an exception for wiki transforms links.
-  if (url.length < 20 || url.match(/\s/) || url.startsWith('wiki:')) return;
```
rowanc1, line 23: I think this should stay.

packages/myst-transforms/src/links/types.ts
```diff
@@ -5,6 +5,7 @@ import type { Inventory } from 'intersphinx';
 
 export interface LinkTransformer {
   protocol?: string;
+  formatsText?: boolean;
```
rowanc1, line 8: This is not set to true anywhere?

packages/myst-transforms/src/links/plugin.ts
agoose77: The other changes in this PR that generalise link formatting should ensure that this function is only called when its required (i.e., not for wiki links)

packages/myst-transforms/src/links/types.ts
agoose77: It should be opt-out (explicit `false`). Looks like I missed it in `wiki.ts`. Let's fix that.

packages/myst-transforms/src/links/plugin.ts
```diff
@@ -48,7 +48,10 @@ export function linksTransform(mdast: GenericParent, file: VFile, opts: Options)
     const result = transform?.transform(link, file);
     // The link transform may compare the text
     // Formatting adds no-width spaces to some URLs
-    formatLinkText(link);
+    // Don't format text if transform already does this
+    if (!(transform && transform.formatsText)) {
```
agoose77, line 52:
```suggestion
    if (!transform?.formatsText)) {
```

agoose77:
@rowanc1 is this OK to ship?

rowanc1:
Will get to this by Monday or you can merge.

agoose77:
As per the above comment, I'm merging this. I think it's a safe merge -- it generalises our link formatting mechanism, and adds a special case for permalinks so that they are treated as regular hyperlinks for now.

rowanc1:
Thanks. :)

At our continuous science workshop this week.

PR #1961, opened 2025-04-10
✍️ Parse Markdown and LaTeX outputs into AST

### LINKED ISSUE #1827: mystmd does not render `text/html` MIME-type for inline `{eval}` role
## Description

mystmd doesn't render the `text/html` MIME-type if it is present when the `{eval}` role is used. I produced an explainer/reproducer repository over here: https://github.com/bryanwweber/mystmd-mime-ignored-repro This is a significant (IMO) difference from the `{glue:}` role developed for JB1 and I really miss being able to style content when inserting it inline.

## Proposed solution

@agoose77 suggested that the fix belonged in `myst-theme` on [Discord](https://discord.com/channels/1083088970059096114/1126886232060870678/1333567984530358324):

> I suspect the bug is that we always prefer AST children over the mime bundle (IE it's in myst theme). This is something we'll be fixing soon anyway with output parsing.

## Additional notes

@rowanc1 suggested I create this issue on Discord 😄

### LINKED ISSUE #2114: Parse code-cell outputs into MyST AST
## Proposal

Generated content like Markdown, LaTeX, and tables should be parsed from code-cell outputs into MyST AST. This will enable code-cells to generate content that participates in static export builds (like PDFs).

### Definition of done
- Generated Markdown, LaTeX, tables, etc, is represented in the MyST document as sub-AST.

agoose77:
> [!Important]
> This PR really touches on important questions of content negotiation and variation.
> I've written up my view in [a HackMD](https://hackmd.io/c7Gqw53DRWa9cJdR1iQojA). 
> This PR as-is doesn't fully represent that yet.

This PR closes #2114 and closes #1827 by parsing Markdown and LaTeX outputs into the MyST AST. This step takes place immediately after execution / cache loading, and is intended to allow code-cells to produce AST via a MIME rendering process.

The idea is that
```yaml
type: outputs
children:
  - type: output
    children: []
    jupyter_data:
      output_type: execute_result
      data:
        text/markdown: |
            This is _emphasized_ text. See @ref for more.
```
becomes
```yaml
type: outputs
children:
  - type: output
    children: 
      - type: paragraph
        children:
          - type: text
            value: 'This is '
          - type: emphasis
            children:
              - type: text
                value: emphasized
          - type: text
            value: ' text. See '
          - type: cite
            identifier: ref
            label: ref
            kind: narrative
            error: true
          - type: text
            value: ' for more.'
    jupyter_data:
      output_type: execute_result
      data:
        text/markdown: |
            This is _emphasized_ text. See @ref for more.
```
Following this transform, it is possible for Markdown (or AST) output by code-cells to participate in referencing and other MyST transformations.

So far, the MIME rendering step occurs after the following transforms (meaning they're skipped):
- `importMdastFromJson`
- `includeFilesTransform`
- `rawDirectiveTransform`
- `liftCodeMetadataToBlock`

We _might_ want to apply those again to the output children.

I also moved the cache transform later -- it's only used for writing out-of-band outputs.


stevejpurves:
@agoose77 I am right in thinking that this does nothing to provide for markdown or latex rending when using in browser compute correct (we may already have LaTeX)? and if we in future support MyST rendering in outputs then in a similar vein they would not be rendered by the native Jupyter outputs areas we use for in-browser compute.

What is your longer term view on getting that support? are you thinking that we will move away from using native jupyter services code for in-browser support? or you think this will just stay unsupported`?

agoose77:
@stevejpurves I think this is an area where a conversation will be easiest! I'm totally OK with there being some limitations between thebe and build-time experiences, and I think that there are a few different ways we can think about it.

stevejpurves:
@agoose77 this is part of the issue, significant differences between build-time renderered / static content and results of in-browser may not be OK at all for a wider group. 

I'm happy for a conversation if it can generate some output / assessment of the impact of the change that can be visibly discussed and decided on in a larger group. (i.e. significant enough to warrant an enhancement proposal)

agoose77:
@stevejpurves can you elaborate on what you mean by "a wider group"?

I'm going to save this for our meeting because it will be a lot quicker to do, but in brief:
>  significant differences between build-time renderered / static content and results of in-browser may not be OK at all for a wider group.

I think this is the wrong reference frame. The existing status-quo involves major differences (not working) between static and Thebe compute outputs for Markdown and LaTeX. What we're talking about is inverting that, so that build-time is more feature-rich than compute-time. I think that's a reasonable (and necessary) trade-off. 

the most important thing for us to do here is have clear user-stories and personas in mind for who is using MyST/Thebe, and how they are using them. In my mind, there are a lot of differences. Where they intersect, we can and should add ways to reduce the gap (such as opt-in ignoring recomputed cell outputs, etc.)

stevejpurves:
@agoose77 let's try and get some written down output from our meeting, clear users stories would be a good output from that!

agoose77:
Steve and I are preparing some user stories to frame this work in https://hackmd.io/H7ERZTycQlKy52JIlEWhDw

rowanc1:
@agoose77 is this still the best branch to move forward on parsing outputs? Currently has conflicts.

agoose77:
@stevejpurves I've rebased this on main, and it's looking much smaller :P

akhmerov:
Can this also help with https://github.com/jupyter-book/mystmd/issues/2504

agoose77:
Yep! This PR doesn't add the feature, but we will want to start consuming the outputs in a theme PR.

agoose77:
This is nearly there. The AST renderer for `application/vnd.mystmd.ast+json;version=1` expects full documents, rather than inline AST for inline expressions. We might change those rules later.

It's failing because the renderer now tries to lift HTML script tags, which become empty divs. This differ from our tests. This is possibly a policy decision, or we could shake our empty divs.

packages/myst-cli/src/process/notebook.ts
```diff
@@ -34,6 +32,28 @@ function blockParent(cell: ICell, children: GenericNode[]) {
...
+export interface IUserExpressionsMetadata {
+  [metadataSection]: IUserExpressionMetadata[];
+}
+
+export function findExpression(
+  expressions: IUserExpressionMetadata[] | undefined,
+  value: string,
+): IUserExpressionMetadata | undefined {
+  return expressions?.find((expr) => expr.expression === value);
+}
```
agoose77, line 55: I removed the file this came from, because it was too niche.

agoose77:
This is ready for review, though I'd like to add a test for LaTeX/Typst exports that include to `outputs`. I've been testing this with https://jupyter-book.github.io/example-outputs/

agoose77:
I'm vacillating on how we should think about post-execution processing transforms. This comment serves to help me think this problem through.

Immediately after executing/parsing a notebook, we need to handle the `jupyter_data` object for each `outputs.children[i]`. This step serves two purposes:
1. We want to parse impure content (like Markdown or LaTeX) that can define new reference targets (modifies the state of the build). This permits code cells to dynamically create new content.
2. We want to reduce the AST + MIME bundle duplication into a single AST tree for simplicity and consistency. This relates to https://hackmd.io/q0KUwIf_TLCgMyEORaK8kA, and means that e.g. plugins can check all images for a11y criteria without special-casing output cells.

## Parsing impure content :ballot_box_with_check: 
An important contextual point here is that outputs can have multiple MIME types for each output. Jupyter Book 1 uses this to specialise the doctree to the build target (e.g. web vs PDF). We want the same result — to support PDF images in LaTeX, vs webp/svg on web.

Regarding (1), multiple MIME types is probably not a problem. We can probably define Markdown and AST outputs from code-cells as "super MIME types" that are preferred for all export types (and we can define a strict precedence). If we _can't_ do this, then we might need to think about the concept of an **alternation** construct (and imposing constraints on the alternations, to handle one MIME type introducing a `label`, and the others omitting it). More on this later.

## Comprehensive parsing :thinking: 
(2) is more fiddly — it's really an expression of the idea that "why stop at just Markdown and/or LaTeX?". Regarding (2), we can't lift all MIME outputs into AST unless we support **alternations**. We need a way to say "this content only renders under X conditions". This feels very similar to the Future Work section in (2), which proposes a construct for doing this at the AST level, so content authors can scope subtrees of AST to e.g. Typst output only, e.g.
```
type: output
children:
  - type: only
    tags:
      - mime:application/x-latex
    children:
      - type: raw
        kind: latex
        value: |
           ...
  - type: only
    tags:
      - mime:text/vnd.typst
    children:
      - type: raw
        kind: typst
        value: |
           ...
```
The uAST contains the above, whilst the export-to-PDF AST drops the Typst or LaTeX. As such, the tag `mime:XXX` is a convention that MyST conforms to — the `only` node can (and should) support other generic entries that users can control and filter on (e.g. `audience:student` vs `audience:instructor`).

Maybe, though, this is hammer-nail thinking. We could also just treat `output` nodes as special -- require the renderers to filter the `output.children` by inspecting each `child.data` namespace for a MIME key that `mystmd` populates during rendering.

packages/myst-cli/src/process/mdast.ts
```diff
@@ -194,6 +194,11 @@ export async function transformMdast(
       log: session.log,
     });
   }
+  await transformLiftExecutionResults(mdast, vfile, {
+    parseMyst: (content: string) => parseMyst(session, content, file),
+  });
```
agoose77, line 210: Immediately try to render outputs

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -25,6 +36,122 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+  new LaTeXRenderer(),
+  new ImageRenderer(),
+  new HTMLRenderer(),
+  new TextRenderer(),
+];
+const OUTPUT_MIME_RENDERERS: MimeRenderer[] = [
+  new ASTRenderer(),
+  new MarkdownRenderer(),
+  new LaTeXRenderer(),
+];
```
agoose77: Expressions are currently _more_ opinionated than outputs. Both Expressions and Outputs will change in future. We want to parse as much as possible, but we also need to acknowledge / handle the problem of export-dependent render hierarchy.

packages/myst-cli/src/transforms/rendermime.ts
```diff
@@ -0,0 +1,221 @@
...
+    parseMyst: MystParser,
+    opts: MimeRendererOptions,
+  ) {
+    return singleParagraphChildren(
+      (await this.renderBlock(contentType, value, vfile, parseMyst, opts)) as GenericParent[],
+    );
+  }
+}
+
+export function findRenderer(rankedRenderers: MimeRenderer[], mimeBundle: Record<string, any>) {
```
agoose77: This function assumes that a single renderer wins out.

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -25,6 +26,118 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+    if (!node.result) {
+      continue;
+    }
+    const result = node.result as IExpressionResult;
+    if (result?.status !== 'ok') {
+      continue;
+    }
+    const mimeTypes = [...Object.keys(result.data)];
+    // Find the preferred mime type
+    const preferredMimeRenderer = renderers
```
agoose77: TODO: change this from `preferredMimeRenderer` to a loop over candidate MIME renderers.

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -25,6 +26,118 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+        node,
+        ruleId: RuleId.inlineExpressionRenders,
+      });
+      break;
+    }
+    const [contentType, renderer] = preferredMimeRenderer;
+
+    // Pull content from cache if minified
+    const content = (result.data as IMimeBundle)[contentType!];
+    const children = await renderer.renderPhrasing(
```
agoose77: TODO: Change `renderPhrasing` to return a "parent" that aggregates phrasing outputs. Possibly a paragraph? Ideally not though -- it's just a placeholder. Probably a "variant".

We want this so that we can tag the variant with the output type.

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -25,6 +26,118 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+      case 'stream': {
+        break;
+      }
+      // TODO: take latest by ID
+      case 'execute_result':
+      case 'update_display_data':
+      case 'display_data': {
+        // Find the preferred mime type
+        const mimeTypes = [...Object.keys(jupyterOutput.data as IMimeBundle)];
+        const preferredMimeRenderer = renderers
```
agoose77: See above.

packages/myst-cli/src/transforms/outputs.ts
```diff
@@ -25,6 +26,118 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+            node,
+            ruleId: RuleId.inlineExpressionRenders,
+          });
+          break;
+        }
+        const [contentType, renderer] = preferredMimeRenderer;
+
+        // Pull content from cache if minified
+        const content = (jupyterOutput.data as IMimeBundle)[contentType!];
+        (node as any).children = await renderer.renderBlock(
```
agoose77: TODO: Change `renderBlock` to return a "parent" that aggregates block outputs. 

We want this so that we can tag the variant with the output type.

stevejpurves:
Ha I was assuming a markdown link style image here but it s a figure!! so `Markdown` is not just a markdown renderer it is already MyST Markdown 🚀

image

stevejpurves:
@agoose77 is this PR the first time this `mimetype` has been exposed? `application/vnd.mystmd.ast+json;version=1` or do/can we already see it elsewhere? are `application/vnd.*` mimetypes just freeform or is there a convention for rendering them?

IANA knows about myst as a markdown flavor https://www.iana.org/assignments/markdown-variants/markdown-variants.xhtml

Outside of this PR is there something to do here for `application/vnd.mystmd.ast+json;version=x` IANA does register them https://www.iana.org/assignments/media-types/media-types.xhtml

stevejpurves:
The generated content output looks great 👏🏼 but there is something to fix on the normal figure outputs in read view in terms of:

* image sizes
* plotly outputs
* widgets outputs, maybe?

Were you seeing this @agoose77? 

image

I'll take a stab at that while reviewing the code

stevejpurves:
sorry, the comments above were all relevant to the theme PR here https://github.com/jupyter-book/myst-theme/pull/742 cc @agoose77

agoose77:
Yes!

For context, I'm still feeling a bit stuck about this PR, mainly in the execution.

Firstly, I think we need to ensure that each output renderer produces a single child (block), whereas I don't think this PR enforces that. I think it's the case that we won't need to massage any parse results - they always produce a block type e.g a paragraph. 

Then we can render all mime types (that we can), and tag them appropriately with their original mime type via a custom data field. 

However, it only just occurred to me that we can't easily do this for inline expressions, unless we wrap them in a `span`. It feels weird to use `span`s in a 'core way', which makes me wonder about a new type. But... That type would just be a `span`. So, I probably need to ignore my gut.

Here's, then, my suggestion for how I'll tweak this PR:

1. Change AST renderers to ensure they produce a single node. For `InlineExpression`, that means wrapping in a span (a paragraph is a block-level element, and shouldn't appear in the context of an inline expr).
2. Change AST rendering mechanism from "first-wins" to "all".
3. Change the `reduceOutputs` step to one that selects preferred mime types (and add same for site).

agoose77:
> The generated content output looks great 👏🏼 but there is something to fix on the normal figure outputs in read view in terms of:
> 
>     * image sizes
> 
>     * plotly outputs
> 
>     * widgets outputs, maybe?
> 
> 
> Were you seeing this @agoose77?
> image
> 
> I'll take a stab at that while reviewing the code

I think these are due to other changes made to myst theme around inline image styling recently.

packages/myst-cli/src/transforms/rendermime.ts
```diff
@@ -0,0 +1,233 @@
...
+    return result;
+  }
+}
+
+export class ImageRenderer extends MimeRenderer {
+  pattern = /^image\//;
+
+  async renderBlock(contentType: string, value: MultilineString) {
+    const phrasingNodes = await this.renderPhrasing(contentType, value);
+    const result: FlowContent[] = [{ type: 'paragraph', children: phrasingNodes }];
```
stevejpurves, line 187: this paragraph around the image is resulting in the following AST
image

which breaks the rendering in the theme and is causing the image size issues. Images should not be wrapped in paragraphs? @agoose77 is this related to your comment on `spans`?

stevejpurves: why do we always need to render a paragraph and not for example a plain block like a `div`, that does not have the same meaning as `p` but is still block level and both are classed as flow content? is `p` the only block level node we have to work with for our FlowContent type?

the heirarchy `output > p > image` seems strange to me, an image is classed as phrasing content when in a `p` as it is inline, afaik flow content includes phrasing content so `output > image` is valid from that perspective and image should be considered flow content in that context? we should not wrap it in a paragraph?

i.e. `output` results in a `div` when rendered, so that is our block level node and we should just render the image immediately below it.

I can see that the current FlowContent type does not include `image` and is pushing us to wrap it - should that be revisited though?

@agoose77 looking forward to a convo on this, and how tightly we are trying to adhere to the flow / phrasing content semantics that are suggested by the code.

stevejpurves: ok found the inline image styling change [here](https://github.com/jupyter-book/myst-theme/pull/710/files#diff-302c9eaa7ee4220daa9d34987583876a0a1f59c5855b8444f3bb79950fed7211) and we can probably modify that further to fix the styles if we went that way, but still the css rule makes sense i.e.

`p > img` - image is phrasing content whereas
`div > img` image is flow content (e.g. otherwise html float would not work)

the css change correctly targets inline images, our image renderer here enforces that the image is inline, even when it is block-level / flow content.

agoose77:
This PR now "renders" (or parses) out all known MIME types into AST. We have some important details to consider when making that statement:
1. Should we parse HTML into AST, or leave it as a terminal HTML node?
2. Should we parse raw Typst and LaTeX into raw nodes (or, into math?)

My current thinking is that we can't unilaterally decide this. Instead, we should rely on annotations from users. I think most of the time, HTML is HTML, and we should not parse it. The same goes for Typst, but not LaTeX — usually LaTeX is just mathematics. Let's use the `metadata` field to let users annotate outputs with context (and/or provide other mechanisms).

Now, parsing a single output's MIME bundle into multiple children is effectively an "alternation". We need to consider what happens when one of the alternation branches creates reference targets. My current thinking is twofold.

1. For outputs like Markdown and AST, _perhaps_ this is "canonical" AST and takes precedence over the other output types -- we should only render e.g. text/markdown from the bundle if it is present.
2. Or, we impose constraints like "lifting" up (and/or removing) targets from output children.

Whatever we do, I think this is solveable.

Now, all that is left to do (besides fixing the tests) is to update myst-theme to prefer a particular variant. We don't collapse variants when publishing to a site.

agoose77:
@stevejpurves I think this is now ready to be reviewed again. Here's the high level:

1. Now `mystmd` will populate the `output.children` with `root` nodes. These are tagged with some metadata in `data` that indicates the provenance MIME type (and whether they were e.g. a traceback stream)
2. `mystmd` will lift streams into `code` with `lang=ansi`. We don't render `ansi` with our existing code block renderer, so these outputs are handed off to the Jupyter renderer for now. In future, we'll use shiki that supports ANSI.
3. HTML is complicated. My current implementation is that the kernel transform that parses outputs should make a decision about whether the HTML is important or should be parsed into mdast. That is controllable by the user. As such, we check for naked `html` nodes that came from parsing, and where we see them, we again hand off to the Jupyter renderer. In future we'll want to support HTML rendering directly, but this didn't work out of the box with bokeh.
4. The legacy rendering should still exist, so old embed / xref content should work.

Here's an example of content preference handled by the theme, not the CLI:
image

choldgraf:
@agoose77 could I take a stab at adding documentation for this, to ensure that we are all aligned on the intended behavior? I'm nervous about merging a big PR like this without any documentation. This isn't just for users - it's a way for us to ensure that we all have the same _intent_ for what we want to make possible. 

My only other main question is about tests: I don't see a ton of new testing (maybe I am missing it though). We are adding a lot of new configurability and user outcome surface area, right? What are your thoughts on the balance of testing coverage we've got here? What would be the cleanest way to improve that?

Especially given the complexity of re-working the AST parsing, I think a test suite that did a bunch of different edge-cases and intended functionality would be a good way to convince ourselves this isn't going to do something unexpected or create user confusion.

As an aside: it would be useful if we had some way to view test coverage in PRs - I might be totally wrong about my analysis of what we're testing here, but it's hard to tell for example if the tests are touching all of these new renderer objects etc just by looking at the diff

agoose77:
> @agoose77 could I take a stab at adding documentation for this, to ensure that we are all aligned on the intended behavior? I'm nervous about merging a big PR like this without any documentation. This isn't just for users - it's a way for us to ensure that we all have the same _intent_ for what we want to make possible.

100%

I don't see this as ready to merge yet. We need more docs, examples, and tests. There is the example-outputs repo that serves as a visual interrogation (https://jupyter-book.github.io/example-outputs/, needs local testing), so we could look to enshrine that in our regro tests.


> My only other main question is about tests: I don't see a ton of new testing (maybe I am missing it though). We are adding a lot of new configurability and user outcome surface area, right? What are your thoughts on the balance of testing coverage we've got here? What would be the cleanest way to improve that?
> 
> Especially given the complexity of re-working the AST parsing, I think a test suite that did a bunch of different edge-cases and intended functionality would be a good way to convince ourselves this isn't going to do something unexpected or create user confusion.

Agree. (above)


> As an aside: it would be useful if we had some way to view test coverage in PRs - I might be totally wrong about my analysis of what we're testing here, but it's hard to tell for example if the tests are touching all of these new renderer objects etc just by looking at the diff

Agree. Vitest supports coverage, we should look into it.

agoose77:
I'm on the Eurostar back to London, and was rewriting the PR body now that it's closer to the end.

My PR style is to write my train of thought, which makes for long reading but I hope helps make it easy to follow my thinking.

I just tweaked this PR to support pulling out more than one MIME key from the MIME bundle and parse it into AST.

However, I'm not convinced whether this is necessary — I keep oscillating here. The _intention_ is to support ambiguous parse hierarchies — where MIME type A is preferred by exporter A, but MIME type B is preferred by exporter B. This might seem logical, but there's also a **pressure** for this to be _unambiguous_; anything we defer to export time (if we prefer different MIME types for exporters, this is the case) then we cannot do anything that modifies global state without imposing further constraints.

Most of the time the hierarchy is emergent as a function of complexity — the MyST AST in JSON form **always** wins. Meanwhile, plain text is pretty bad. The only argument I can immediately identify is that we might want to emit raw Typst **and** raw LaTeX for different export backends. But ... that feels both niche and messy. I don't like making such a big design choice in favour of a mechanism (`raw`) that is really an escape hatch.

In the back of my mind has been the idea of introducing a formal `alternation` node. This makes sense for user-driven alternations (like audience-dependent content variations), but my ideas for that mechanism (Sphinx `only`-like `expression`) feels ... _too_ general for this use:
````
```{only} supports_mime_text_latex
\section{foo}
```
````
is the ugliest thing I've ever written!

I think my original issue with automatically generating alternations is that it we don't always want them (see the above), and we probably also do not want to encode too much exporter knowledge into MyST — why _shouldn't_ someone support exporting to XML, for example?

But, maybe there's a sensible path that is simpler than both what this PR does at this moment in time, and the alternatives discussed above:

Perhaps we have a strict MIME hierarchy — anything we can parse, is ordered by richness. Where users want to break that heirachy, they can either specify a preferred MIME (in metadata, or by reducing the bundle), OR an alternation by defining the alternation in the metadata e.g.
````yaml
metadata:
  alternation:
    - mimeType: text/latex
      condition: export-latex
    - mimeType: text/vnd.typst
      condition: export-typst
````

Crucially, the _author_ is making this decision, not the engine! It's a curated alternation that _happens_ to switch on the exporter.

File: packages/myst-cli/src/transforms/outputs.ts [hunk 10]
```diff
@@ -25,6 +25,186 @@ function getWriteDestination(hash: string, contentType: string, writeFolder: str
...
+export async function liftOutputs(mdast: GenericParent, file: VFile, opts: LiftOptions) {
+  const renderers = opts.renderers ?? DEFAULT_MIME_RENDERERS;
+  for (const node of selectAll('output', mdast)) {
+    const jupyterOutput = (node as any).jupyter_data;
+
+    // Do we have a MIME bundle?
+    switch (jupyterOutput.output_type) {
+      case 'error':
+      case 'stream': {
+        console.log({ foo: jupyterOutput.traceback ?? jupyterOutput.text });
```
agoose77:
```suggestion
```

agoose77:
There is a subtle nuance to the above. Widgets are 'better' that static markdown or images. But, they're not AST representable, so they cannot automatically win. We must lift out the next best mime type. 

I think that's OK. Defining the best AST output is different to defining the preferred thing to render. I think we should think about exporters being free to either choose an unrepresentable mime type, or otherwise take the preferred (by complexity) rendered child. 

To support specialisation, we treat that as an opt-in user-guide feature, aka alternation.

agoose77:
I wrote up two design docs to hopefully disambiguate this design space:
- https://hackmd.io/c7Gqw53DRWa9cJdR1iQojA
- https://hackmd.io/kRGYNxv7Q9y29DF1fTZt9Q?both

haroldcampbell:
@agoose77 Do you have the design docs somewhere else?

I know this is a bit old, however, I was trying to access the design docs in the hackmd but they are returning 404.

agoose77:
The new documents are:
- [Content Alternations](https://hackmd.io/@jupyterbook/rJ6Sq_qvbg)
- [Universal AST](https://hackmd.io/Fc8ZWDI8SzW6jbTKLvkvFw)

There's also a discussion around [variants](https://hackmd.io/MhsvyKS7T_a6EUN5UjIk4A) in addition to alternatives, but it's immature.

Issue #1685, opened 2024-11-29
Add support for other social media handles in addition to twitter


thesamovar:
At the moment in the myst frontmatter you can add a twitter key to each author, it would be good to have support for bluesky, mastodon, linkedin, etc.

choldgraf:
Definitely something we want to support - here's the tracking issue for this in the html theme repository:

- https://github.com/jupyter-book/myst-theme/issues/362

thesamovar:
Great!

choldgraf:
I thought about it more and do think this is distinct from https://github.com/jupyter-book/myst-theme/issues/362 so I've updated the body of this issue and focused it on author metadata!

thesamovar:
I like the more general solution!

thesamovar:
Thanks Chris.

PR #1987, opened 2025-04-25
🦋 Add bluesky social links

rowanc1:
Adds bluesky, mastodon, x, threads, linkedin and allows these to be added to affiliations as well.

cc @thesamovar

See #1685

### DIFF
--- a/.changeset/curly-beers-study.md
+++ b/.changeset/curly-beers-study.md
@@ -1,0 +1,5 @@
+---
+"myst-frontmatter": patch
+---
+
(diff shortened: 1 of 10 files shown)


rowanc1:
There isn't validation at the moment, this has previously been done in the themes. Probably a good thing to add though.

I am going to merge this for now and then we can circle back to that after.

fwkoch (commented):
From a quick look, looks good - mostly frontmatter boilerplate... "no validation" means they are still validated as strings, but no regex validation. Seems fine for now, very easy to add later here or just handle gracefully on the theme side.

Issue #814, opened 2023-12-14
Allow pages to be hidden in a MyST site


fwkoch:
Currently, the myst site table of contents defines both (1) what pages are built and (2) what pages are visible in your site. It would be nice to enable pages to be built (and available as mdast on the content server) but not part of your site. This would allow embedding content from a source without necessarily including that source as a navigable page.

parmentelat:
I was expecting this to be auto-mentioned, but for the record this is related to
https://github.com/orgs/jupyter-book/discussions/1989

parmentelat:
my proposal on the quoted discussion was to have a `hidden` flag that could apply to TOC nodes like `file` or `pattern`

this is a feature that I would need badly and I would volunteer to at least look at it if there is a rough consensus on the relevance of that feature

fwkoch:
The first bit of work towards this

fwkoch:
What exactly does it mean if a page is `hidden`?

- It does not show up in the table of contents on the built site and it is not included in static exports.
- It is still built and, if applicable, executed.
- Its contents may be embedded and cross-referenced - this means the page JSON "data" is still published as part of the site.
- Its content is not numbered as to not impact the numbering of other pages.

Are there any other high-level implications of a page being hidden?

stevejpurves:
* ensuring that embedded content (embeds, figures, xrefs) are executable on page - I imagine that it is largely implicit in the embedding point above but maybe add it as a specific test case, as you would be able to embed without necessarily having the computable & dependency objects in order.
* Do citations, footnotes or parts need special consideration?

choldgraf:
I think that the combination of your two comments above feel right to me. To me, that's enough constraints to get started on an MVP implementation and then we can see what people ask out of this functionality once it's possible for them to start using it.

In my experience, the most common use-case for this is: I want to put executable content on an "invisible" page because it is huge and messy, and I want to *embed* that content elsewhere in my book so that it is part of my narrative without cluttering my prose with a bunch of python code.

In Sphinx, pages that aren't part of the TOC are _still built via HTML_, but they're treated as "orphan" pages that have no place (or links) in the toc tree.

That said, unless there are strategic doors we'd be closing with any of these decisions, IMO the assumptions you share above are good and we should just give this a shot at implementation.

parmentelat:
thanks @fwkoch for taking care of the changeset & prettier things !
[...]

PR #1993, opened 2025-04-28
🕵️‍♀️ Expose hidden attribute in TOC nodes

parmentelat:
this is in relation to https://github.com/jupyter-book/mystmd/issues/814

I am far from fluent with typescript
so this is a very preliminary work, and only partial

I'm primarily curious to have your feedback and check this is going in the right direction

### DIFF
--- a/.changeset/two-trains-invite.md
+++ b/.changeset/two-trains-invite.md
@@ -1,0 +1,6 @@
+---
+"myst-cli": patch
+"myst-toc": patch
+---
+
+expose hidden attribute in TOC nodes
--- a/packages/myst-cli/src/project/fromTOC.ts
+++ b/packages/myst-cli/src/project/fromTOC.ts
@@ -116,7 +116,7 @@ export function patternsToFileEntries(
   return entries
     .map((entry) => {
       if (isPattern(entry)) {
-        const { pattern } = entry as PatternEntry;
+        const { pattern, ...leftover } = entry as PatternEntry;
         // Glob matches, relative to `path`, ordered naturally
         const matches = globSync(pattern, { cwd: path, nodir: true, ...opts })
           .filter((item) => !ignore || !ignore.includes(item))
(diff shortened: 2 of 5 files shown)


fwkoch (commented):
Hey @parmentelat - thanks for getting a start on this! Sorry to be a bit slow reviewing.

So far, this looks great - it updates the `toc` validation to include `hidden` and makes sure this property persists when the toc is processed. (And has tests!)

Next steps get more complicated. A few places we need to consume this:
- When writing the site manifest, we want to exclude these pages, i.e. here: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/build/utils/projectManifest.ts#L16
- When enumerating, we want to skip these pages, I think. No numbering of hidden content: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/process/site.ts#L357
- I _think_ hidden pages still need `slug` even though the page isn't in the site - we still need the `slug.json` entry so the content can be cross-referenced on other pages. (Unless we don't even want cross-references... What would that leave us with? Only `embeds` of the hidden content...? I think we want cross-references.)
- I anticipate as we test this, we will discover more places we need to respect this flag. Everywhere we do anything with `project.pages`, we will need to decide if `hidden` matters.

As for this PR, I'm happy to merge it at its current scope. Two minor issues:
- no changeset (instructions for this are in the other comment on the PR)
- lint check is failing (just need to run prettier, either by installing the pre-commit hook with `npm run install-pre-commit` or just running `prettier` in your IDE/editor).
I'll go ahead and tidy these up for this one so we can get this PR landed.

We can continue the discussion about next steps on the issue here: https://github.com/jupyter-book/mystmd/issues/814

fwkoch (approved):

stevejpurves:
Hi @fwkoch, for a hidden page, imo hidden means hidden from the TOC/nav only:

* we want the ability to in browser compute
* we want the ability to cross reference (i.e. have a hidden listing of things, that can be cross referenced without having to include the big clumsy listing in the TOC)
* In the same way, hidden should be included in the search metadata?

Does this mean that implementing the `hidden` behaviour is actually a theme level change rather than build time change? These opinions align with my understanding of `hidden` from early requests and discussions - has that changed?

PR #2016, opened 2025-05-14
 Update links to example of landing pages

agoose77:
With a recent move from `myst-examples` repositories to the main `jupyter-book` organisation, we need to fix our links in our documentation.

This is tiny: self merging.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -176,11 +176,11 @@ Recall from the [architecture overview](#diagram-app) that `myst-theme` is a Rea
 
 #### Content server
 
-We need some example data to test our theme against, such as [the example landing page](https://github.com/myst-examples/landing-pages). Clone this example content repository and start the content server:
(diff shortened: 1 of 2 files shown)


choldgraf:
Just to give some context to others for this PR. We had a conversation at the scientific python submit about the downside in having multiple GitHub organizations for a single subproject.  That's partially for security and access reasons, but it's also because it makes it harder for external contributors in companies to make contributions.

Companies often give approval for upstream contributions at the GitHub organization level. So when we create multiple GitHub organizations, we create extra work for collaborators inside of companies to be able to contribute at all.

So for that reason, we thought it would be safer to just use a prefix and keep these examples inside of a single GitHub organization for the jb subproject

rowanc1:
Thanks for the context @choldgraf.

rowanc1 (approved):

PR #2021, opened 2025-05-14
💬 Add social media project frontmatter

kne42:
Extends the set of understood social identities to:
- `url`
- `github`
- `bluesky`
- `mastodon`
- `linkedin`
- `threads`
- `twitter`
- `youtube`
- `discourse`
- `discord`
- `slack`
- `facebook`
- `telegram`

Allows social links to be defined under the project frontmatter key `social` or `socials`:
```yaml
project:
  ...
  social:
    bluesky: @me.bsky.social
    youtube: https://youtube.com/c/me
```

- Potential related docs PR is here: https://github.com/jupyter-book/mystmd/pull/2050
 

### Tasks
- [x] Create follow up user-story for upgrade step to handle change in canonical representation of existing social links: https://github.com/jupyter-book/mystmd/issues/2074
- [x] Add tests to cover validation modes
- [x] Add documentation to cover new social kinds


rowanc1 (approved):
Code changes look awesome. Would be helpful to add to the documentation as part of this PR! Glad to see the additional social profiles (esp slack and discord) being added (hadn't thought about those in the context of a person).

packages/myst-frontmatter/src/utils/socialLinks.ts
```diff
@@ -34,35 +30,59 @@ export const SOCIAL_LINKS_ALIASES = {
...
+  const result = output ?? {};
 
+  // FIXME: normalize usernames into URLs
   if (defined(value.url)) {
-    output.url = validateUrl(value.url, incrementOptions('url', opts));
+    result.url = validateUrl(value.url, incrementOptions('url', opts));
   }
   if (defined(value.github)) {
-    output.github = validateString(value.github, incrementOptions('github', opts));
+    result.github = validateString(value.github, incrementOptions('github', opts));
```
fwkoch, line 55: Should we use `validateGithubUrl` for this?

packages/myst-frontmatter/src/utils/socialLinks.ts
```diff
@@ -34,35 +30,59 @@ export const SOCIAL_LINKS_ALIASES = {
...
-    output.linkedin = validateUrl(value.linkedin, incrementOptions('linkedin', opts));
+    result.linkedin = validateUrl(value.linkedin, incrementOptions('linkedin', opts));
   }
   if (defined(value.threads)) {
-    output.threads = validateString(value.threads, incrementOptions('threads', opts));
+    result.threads = validateString(value.threads, incrementOptions('threads', opts));
   }
   if (defined(value.twitter)) {
-    output.twitter = validateString(value.twitter, incrementOptions('twitter', opts));
+    result.twitter = validateString(value.twitter, incrementOptions('twitter', opts));
```
fwkoch, line 70: `validateString` has a regex option so if we want to be a bit more strict, we could do something like:

```suggestion
    result.twitter = validateString(value.twitter, {
      ...incrementOptions('twitter', opts),
      regex: '^@',
    });
```

fwkoch (commented):
Regarding the docs, there's extending the list here: https://github.com/jupyter-book/mystmd/blob/main/docs/frontmatter.md#social-links and adding "social" to available frontmatter fields: https://github.com/jupyter-book/mystmd/blob/main/docs/frontmatter.md#available-frontmatter-fields

Other than that and running prettier so linting is happy, this looks great!

packages/myst-frontmatter/src/utils/socialLinks.ts
kne42: It's a bit tricky actually since the documentation specifies for this field to be a username and NOT a url.

I was thinking maybe for some of the socials we could normalize the usernames to urls but we use the handles for the hover state like in [here](https://github.com/jupyter-book/myst-theme/blob/8283e4505fdb418355ca25ae114ba7bea3cec956/packages/frontmatter/src/Authors.tsx#L58).

I was thinking about normalizing urls to usernames instead but that might be weird too in the case of the site footer where you would be linking to a project instead of a user/org perhaps. Maybe we should not be too strict on the validation here and instead leave that to the React render, what do you think? We could also have different validators for authors than for the project-level but I don't think that's ideal either.

agoose77: @kne42 I'm here with @fwkoch and we discussed this -- let's treat the `project.github` as a _distinct_ validator to `project.social.github` and `project.author.github`. The social `github` may accept `ORG` on its own, whereas the `project.github` must be either a URL OR `ORG/REPO`. In both cases, we should output a URL. 

What do you think about this change? You don't need to make it here, though -- we can follow up and for now make the theme do more work if that's easier.

kne42: So if I understand correctly, we want different validators for project vs author social links then? Should we also consider making them accept different subset of social links (e.g. authors wouldn't accept `discourse` or `discord`)?

rowanc1: I am a fan of allowing any social links this to be specified by either the author or the project (I think this keeps complexity down, and opens up the possibility to say as an author: "if you want to contact me find me in the myst discord", etc.).

Very much also agree with @agoose77 that `project.github` is *not* a social link, that is the content of the repo, things like in-page execution and edit links hang off that value. The project social is probably the github org, and isn't used for those things.

packages/myst-frontmatter/src/utils/socialLinks.ts
rowanc1: I do think we should probably take a stance on if these are usernames or urls, and normalize between them. That could come in a different PR though.

I could make arguments for either direction (URL or username). Once we make a choice and a validator, that logic could be shared in the theme for backwards compatibility.

I would prefer if we don't _require_ starting with `@` — makes the yaml harder to write!

agoose77: When I last spoke to @kne42 about this, I'd suggested accepting either usernames or URLs for most social fields. I looked at this PR again to try and move it forward, and I feel more inclined to be a bit more restrictive here and accept (in some cases) only usernames:

1. Parsing URLs is fiddly, and there are some cases that are nontrivial — Mastodon doesn't guarantee that `foo.com/@bar` is a valid URL. 
3. Mapping from URL to username sometimes requires a fetch or is not well defined (mastodon)

agoose77: I've pushed a commit that adds validators for mastodon (only `@username@domain`) and bsky (`@domain` or `https://bsky.app/profile/domain`)! I'll pick this up tomorrow to visit the rest.

agoose77:
I've made another pass here. My thinking behind the validation results is:

1. We want to reduce the business logic to consume frontmatter
2. We want to make frontmatter easy to write
3. We don't want to lose information during validation
4. We want to be as strict-as-possible up-front to avoid things failing silently later on (we don't know how people will consume our frontmatter)

At present, we have:
| **Name**  | **Normalisated form**                                    |
|-----------|------------------------------------------------------|
| Mastodon  | @USER@DOMAIN                                         |
| Bluesky   | @DOMAIN                                              |
| Facebook  | https://facebook.com                     |
| Twitter   | USERNAME                                             |
| YouTube   | HANDLE or https://youtube.com                        |
| GitHub    | @USERNAME or ORG/REPO or https://github.com/orgs/foo |
| LinkedIn  | https://linkedin.com                                 |
| Threads   | USERNAME                                             |
| Discourse | https://                                             |
| Discord   | https://                                             |
| Slack     | https://                                             |


I think it's useful for the frontmatter consumer to be able to tell "this is a GitHub user" without doing any URL parsing -- they can trust our parsing, and we avoid downstream bugs. But, sometimes we can't usefully do much parsing e.g. Facebook, where some users do not have useful UIDs. My hunch is that we probably need to take a social view on this:

- Twitter really emphasises importance of the handle
- Facebook prioritises readable names
- LinkedIn prioritises readable names

Where we prioritise names, we have to lean in on links because that metadata isn't readily available at frontmatter validation time.

Where we prioritise handles, we can pass those through.

Any thoughts @rowanc1 @kne42?

packages/myst-frontmatter/src/socials/validators.ts
```diff
@@ -0,0 +1,224 @@
...
+  else if ((match = value.match(TWITTER_REGEX))) {
+    return match[1];
+  } else {
+    return validationError(
+      `Twitter social identity must be a valid URL starting with https://twitter.com/, https://x.com/, or a valid username: ${value}`,
+      opts,
+    );
+  }
+}
+
```
agoose77, line 109: Twitter usernames are _not_ themselves `@foo`, they're `foo`. But people are used to typing `@foo`, so we can strip that off.

choldgraf:
I gave a quick look at this, and think that it looks good to me once others are happy above. I read through our authorship social media docs to understand this, and it inspired me to make a little improvement to the docs to modularize our authorship metadata documentation a bit better so this is more discoverable. I didn't want to clutter up this PR if it wasn't relevant so I made a different one here:

- https://github.com/jupyter-book/mystmd/pull/2050

It wasn't clear to me whether this PR will impact the author-level social media metadata as well. If it doesn't, then it shouldn't impact this PR. If it does, then we'll need to update that PR when this one is done. If you all like, I'm happy to keep this PR focused on the implementation if somebody is willing to iterate with me on https://github.com/jupyter-book/mystmd/pull/2050 to get the docs cleaned up for this PR as well.

rowanc1:
Only thing left (that I do also think is important for these) are tests (@agoose77 mentioned above). If this is changing either github/twitter's main representation the theme needs to be able to consume that and upgrade. It would be nice to show this in the author popover as well, but that can come in a later pr.

We could also put this into an upgrade step (e.g. twitter handles are now always usernames or something). I am also fine not to do that step.

agoose77:
Let me update this PR with the various to-do notes!

agoose77:
@kne42 just checking in with you -- I've moved this PR forward in terms of some policy about what the validated forms of each social identity should be. Your input is welcome if you have any feedback or suggested alternatives.

Otherwise, are you still keen to work on this PR? We need to write up docs and add some test cases. There's also an upgrade mechanism that we can define so that older MyST projects can be upgraded to "valid" frontmatter.

kne42:
I’m happy to add some docs and test cases, I can probably get to it tomorrow afternoon or early next week. As for the upgrading, someone more experienced with the process should probably handle that.

agoose77:
I've done a first-pass adding upgrade/downgrade looking at these parts of the frontmatter:
```
funding.recipients
funding.investigators
funding.awards.recipients
funding.awards.investigators

authors
contributors
reviewers
editors
```

agoose77:
Hmm, on second thoughts I'm tempted to pull that upgrade/downgrade work out -- it's going to be a bit hefty as we need to vendor the validators, and I don't want to block this PR. Given that no-one is using myst-migrate right now, and it's intended that people always consume the latest version, I'm happy to decouple the upgrade work and this PR.

choldgraf:
I suggest that our next actions should be:

- Get this PR to a state where it's safe to try (e.g., no obvious bugs, reasonable tests)
- Merge it
- I can then update my docs PR about authorship and resolve merge conflicts if need be.
- @agoose77 opens an issue describing the upgrade work that's needed

And we can prioritize next steps according to all the other stuff we wanna do as well :-)

kne42:
Hi folks, today I plan to work on the following:
- update social links docs
- add unit tests for new validation funcs
- update e2e (? idk what to call these tbh but all the `.yml` stuff) tests

artoftheblue:
Hello! I'm very excited that this PR is coming together. Is there a possibility to add more social media links? I'm specifically interested in [Telegram](https://telegram.org), which is [widely used by people all over the world](https://www.demandsage.com/telegram-statistics/), justifying the addition.

>  In India, 45% of the population regularly uses Telegram.
> 
> According to a survey by Statista, nearly half of the people in India used Telegram. Other countries with high Telegram usage include Brazil and Mexico, where 38% and 34% of people use the app, respectively. In contrast, only 9% of people in the United States reported using Telegram regularly. 
> 
> Meanwhile, another survey conducted in March 2024 found that around 51% of Russian citizens use the platform.

agoose77:
@artoftheblue I've added Telegram to the socials category, on the basis that it feels more like social media than, say, an SMS number (you can have profiles & groups).

agoose77:
OK, I think this is good to ship! I pinged Rowan for another review before remembering that the major features in this PR have already been signed off (apart from the new Telegram inclusion).

Let's merge this, follow up with upgrade and theme work!

Thanks @kne42 for a detailed PR, it's clear you put a lot of time and effort in here!

Thanks to @choldgraf and @rowanc1 for review help!

PR #2026, opened 2025-05-17
Document CORS tip

choldgraf:
This adapts and documents some useful CORS tips that @rowanc1 posted in the following issue:

- https://github.com/the-turing-way/the-turing-way/pull/4156

### DIFF
--- a/docs/website-metadata.md
+++ b/docs/website-metadata.md
@@ -115,6 +115,8 @@ Enabling all origins in CORS enables your site to:
 - Support MyST cross-references (e.g., xref links, see docs) from other sites that use your website as a citation source.
 - Enable content embedding, linking, or automated metadata access (e.g., for previews or API consumers) without authentication or same-origin constraints.
 
+For an example of what it looks like to update CORS settings, see [this GitHub PR updating CORS settings for Netlify](https://github.com/the-turing-way/the-turing-way/pull/4156).
+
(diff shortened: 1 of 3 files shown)


docs/website-metadata.md
```diff
@@ -104,3 +104,27 @@ references
...
+
+By setting CORS headers to allow all origins (*), you make it possible for external tools and sites to:
+
+- Reference your content directly via structured and stable links.
+- Preview or embed sections of your site from another page (with attribution).
+- Use it in federated, cross-site knowledge systems (like MyST Markdown references in external books or educational hubs).
+
+If your MyST site is public and does not require authentication, allowing all origins does not pose a security risk.
+
+_Note, this was adapted from [this GitHub issue from @rowanc1](https://github.com/the-turing-way/the-turing-way/pull/4156)_.
```
rowanc1, line 130: Can remove this! Thanks for bringing it in.

Or maybe change it to a link as a "for example implementation"?

rowanc1 (approved):

docs/website-metadata.md
choldgraf: oh good idea - will do that

choldgraf:
I updated the reference to the PR so that it is used as an example, and also added a link in Netlify. Thanks for the suggestion! I'll merge since this already has a +1

Issue #2044, opened 2025-05-21
Deployment of CryoCloud website is broken for index route


agoose77:
## Description
Recently, the CryoCloud team tried to deploy a new version of their site built with MyST. Unfortunately, it threw errors at runtime (https://github.com/CryoInTheCloud/CryoCloudWebsite/issues/129) and they had to revert it.

## Proposed solution

It seems like this is another bug caused by the way that we do HTML exporting. I have identified that renaming the `intro.md` file to `index.md` fixes this, so I'm going to recommend that they do that until we fix the bug.

I identified the fix, for those interested, by observing that only the index route was affected. I know that we special case that, so I tested the hypothesis that it might be related to the file naming.

This issue tracks the underlying bug.

agoose77:
OK, looking at the `window.__remixContext` (which I know from previous encounters with this kind of bug is a good place to start), the `url` field changes from `/index` under `intro.md` to `/` under `index.md`. I think this is the right place to start looking.

Here's the script I used (with `ripgrep` and `sd` instead of `grep` and `sed`):

```shell
curl http://localhost:3000/  | rg -o 'window.__remixContext(.*)' | sd '(.*)</script>.*' '$1' | npx prettier --parser typescript
```
Or, more POSIX-y
```shell
curl http://localhost:3000/  | grep -o -E 'window\.__remixContext(.*)' | sed -E 's/(.*)<\/script>.*/\1/' | npx prettier --parser typescript
```

PR #2045, opened 2025-05-21
🐌 Fix slug handling for MyST projects using a ToC

agoose77:
In #2044, a bug with slug handling was identified whereby:
- projects whose index file had a non `index` slug 
- that contained another file called `index.md`

would write the `index.html` with two different contents. This leads to a hydration error where the two conflict.

This PR makes a simple fix by ensuring that the project root file is always given the `index` slug, such that it's reserved when other files named `index.md` are encountered. 

> [!Note]
> We currently choose to write the root project file to `index.html` because it's a standard amongst many static web servers to resolve `/` against `/index.html`. Although encoding this behaviour into `mystmd`'s slug generator is exposing these semantics to a non-webserver application, our choice of slugs is arbitrary and thus this is likely harmless.

Closes https://github.com/jupyter-book/mystmd/issues/2044


rowanc1 (commented):
Is this testable from an integration test?

agoose77:
I think so — I have added a simple integration test to verify that the slug for a root file `homepage.md` becomes `index`, not `homepage`!

PR #2073, opened 2025-06-01
Fix broken syntax

artoftheblue:
Trying to reproduce previous syntax returned an error. Additionally, it was unclear personally to see urls without /

### DIFF
--- a/docs/website-navigation.md
+++ b/docs/website-navigation.md
@@ -73,8 +73,10 @@ There are a few types of entries you can define:
   * ```yaml
     - title: Dropdown title
       children:
-      - url: pageone
-      - url: pagetwo
+      - title:
+        url: /pageone
+      - title:
+        url: /pagetwo
     ```
   * Becomes a dropdown with other entries inside.
 ````


choldgraf (approved):
Just to confirm, you're saying that putting `url: mypage` does not work, but `url: /mypage` does?

just a quick question: can you double-check that this also works if you're on a sub-page? e.g. if you're on `pageone/subpageone` do those nav buttons still work?

artoftheblue:
No, it does not work. 

## Examples

The following is fully functional:

```
    - title: Articles
      children:
      - title: First article
        url: /article1
      - title: Second article
        url: /article2
    - title: Information
      children:
      - title: Availability
        url: /info/availability
      - title: Showcase
        url: /info/showcase
```

Whereas if I were to remove the `/` in the beginnings of url-s, the corresponding pages disappear from the nav bar

Similar error from console: 
image

Another:
image

Example when no urls are shown in the nav bar:

```
    - title: Articles
      children:
      - title: First article
        url: article1
      - title: Second article
        url: article2
    - title: Information
      children:
      - title: Availability
        url: info/availability
      - title: Showcase
        url: info/showcase
```

--- 

Perhaps it would be great to add support for the syntax originally described in the docs, or at least support for urls without `/`-s

artoftheblue:
I might have accidentally broken the PR by deleting the branch, thinking that the changes have been merged. I'll republish the same branch with the changes above and contribute to the docs some more.

PR #2091, opened 2025-06-10
📦 Bump turbo repo

fwkoch:
This tiny PR just updates `turbo` from the previous 2-year-old version. This does not impact any of the packages, just the monorepo build, so hopefully it's non-controversial.


agoose77:
Fab! I am in favour of merging — I've tested newer versions of `turbo` with MyST before and it seems similarly fine!

choldgraf:
Merging because I see an approval and this seems like a good idea to me too! Thanks @fwkoch for the maintenance work!

Issue #2071, opened 2025-06-01
Adding `site.parts.footer` stops `project.plugins` from loading silently


artoftheblue:
## Description

If both site.parts.footer and project.plugins are present in the frontmatter, then no plugins load. The frontmatter is structured as follows:

```
version: 1
project:
  ...
  plugins: 
    - picsum.mjs


site:
  template: book-theme
  ...

  parts:
    footer: footer.md
```

`picsum.mjs` is the example plugin used in the documentation [here](https://mystmd.org/guide/javascript-plugins#defining-a-new-directive). Having followed the myst.yml structure in [this example project](https://github.com/jupyter-book/example-landing-pages/blob/main/myst.yml), adding the plugins option does not cause the plugins to load and does not output any error logs like `plugin has not loaded`.

Instead, the first error message is about an unknown directive. 

![Image](https://github.com/user-attachments/assets/01a6ee35-68f7-4071-b05e-cfa897f66b19)

Perhaps there is a way to make this work already, but as discussed in #1920, there is no documentation on footers yet.

## Proposed solution

In case there is already a simple way to structure the frontmatter differently to support both these functionalities, then just add a corresponding error message when the frontmatter is structured as above. Otherwise, fix plugins not loading in this case.

EDIT: These issues might be related: #1813, #1835

agoose77:
Likely a duplicate of https://github.com/jupyter-book/mystmd/pull/1925!

agoose77:
The cause of this bug is that the logic for parsing `parts` executes before the configuration that loads plugins. We need to separate the two!

stefanv:
https://github.com/jupyter-book/mystmd/pull/1925 does contain a fix, although it may not be an optimal fix.

fwkoch:
Taking a look at this - `loadPlugins` returns an empty list if the site/project config is not loaded: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/plugins.ts#L39

This truthy empty list gets cached on our session and never recalculated: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/session/session.ts#L166

Since introducing file-named `parts` to the site/project config, loading the config loads the file, e.g. here: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/config.ts#L435 - this calls `loadFile` which in turn calls `loadPlugins`, caching the empty list before the config is done loading.

[...]

PR #2093, opened 2025-06-10
🔌 Load plugins earlier, before parsing frontmatter parts

fwkoch:
This addresses https://github.com/jupyter-book/mystmd/issues/2071 - Previously, we were processing `parts` files before `plugins` were loaded on the `session`. This ended up caching an empty list of `plugins` which was then never repopulated. Plugins were just silently gone!

@stefanv had a first pass at addressing this here https://github.com/jupyter-book/mystmd/pull/1925 - that PR prevented caching the empty list; however, `plugins` were still not loaded before `parts` processing, so referencing a plugin in a part would fail.

In this PR, I now call `loadPlugins` earlier, while processing project fronmtatter. This happens before any `parts` or other files are processed, so now `plugins` are available during `part` processing and are never cached as an empty list.

### DIFF
--- a/packages/myst-cli/src/config.ts
+++ b/packages/myst-cli/src/config.ts
@@ -430,6 +430,7 @@ async function resolveProjectConfigPaths(
         }
       }),
     );
+    await session.loadPlugins(resolvedFields.plugins);
   }
   if (projectConfig.parts) {
     resolvedFields.parts = await loadFrontmatterParts(
--- a/packages/myst-cli/src/plugins.ts
+++ b/packages/myst-cli/src/plugins.ts
@@ -1,7 +1,6 @@
 import fs from 'node:fs';
 import { pathToFileURL } from 'node:url';
 import type { ISession } from './session/types.js';
-import { selectors } from './store/index.js';
 import { RuleId, plural, type MystPlugin, type ValidatedMystPlugin } from 'myst-common';
 import type { PluginInfo } from 'myst-config';
 import { addWarningForFile } from './utils/addWarningForFile.js';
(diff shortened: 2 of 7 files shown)


packages/myst-cli/src/config.ts
```diff
@@ -430,6 +430,7 @@ async function resolveProjectConfigPaths(
         }
       }),
     );
+    await session.loadPlugins(resolvedFields.plugins);
```
fwkoch, line 433: We are explicitly waiting for the plugins to load prior to resolving the `parts` on the next line.

packages/myst-cli/src/plugins.ts
```diff
@@ -12,37 +11,30 @@ import { loadExecutablePlugin } from './executablePlugin.js';
...
-  const plugins: ValidatedMystPlugin = {
+export async function loadPlugins(
+  session: ISession,
+  plugins: PluginInfo[],
+): Promise<ValidatedMystPlugin> {
+  const loadedPlugins: ValidatedMystPlugin = session.plugins ?? {
     directives: [],
     roles: [],
     transforms: [],
+    paths: [],
```
fwkoch, line 22: We keep track of `paths` to the loaded plugins so we do not duplicate loading. This means multiple loads should be ok, if that happens.

packages/myst-cli/src/session/session.ts
```diff
@@ -159,13 +160,8 @@ export class Session implements ISession {
...
 
-  _pluginPromise: Promise<ValidatedMystPlugin> | undefined;
-
-  async loadPlugins() {
-    // Early return if a promise has already been initiated
-    if (this._pluginPromise) return this._pluginPromise;
-    this._pluginPromise = loadPlugins(this);
-    this.plugins = await this._pluginPromise;
+  async loadPlugins(plugins: PluginInfo[]) {
+    this.plugins = await loadPlugins(this, plugins);
```
fwkoch, line 164: We are no longer caching the `_pluginPromise` for early return. Previously this was required since `loadPlugins` was called during `loadFile` and `loadFile` was called multiple times in parallel.

Since we've moved this to project frontmatter loading, it should only be called once.

(Also, we can't really cache the promise anymore anyway, since `loadPlugins` could theoretically be called with different arrays of input `plugins`.)

agoose77:
@fwkoch thanks for the speedy work on this! Like we discussed elsewhere, this feels like a quick fix that we can rework down the road. Please merge this, or I will tonight!

Then we should make a release of `mystmd`.

choldgraf (approved):
I think that we should merge, this is an improvement IMO. I won't merge because it seems like y'all are actively working on this, but I'll "just do it" in a few days if it hasn't had movement :-)

As a user, I'd expect parts to behave no differently from "main content" (maybe with some exceptions, like cell execution). So to me this aligns the expectations of a typical user with our tech implementation, and will lead to less confusion.

fwkoch:
Thanks for the reviews - I will merge this now and make a release after!

PR #2097, opened 2025-06-12
🫥 Exclude hidden entries from site manifests

parmentelat:
this is a followup to #1993

it addresses the first set of changes that @fwkoch was listing in https://github.com/jupyter-book/mystmd/pull/1993#pullrequestreview-2837944257

namely: hidden entries are excluded from the manifest

with this change, I already get what I need i.e. the ability to define entries in the toc that are built but not mentioned in the actual navigable toc


parmentelat:
@fwkoch 
I tried to take care of the changeset thing but was not allowed to go through
here's what I did
```
npx @changesets/cli
npx @changesets/cli version
# went well up to here, but for running this
npx @changesets/cli publish
# I get
# 🦋  error npm error 403 403 Forbidden - PUT https://registry.npmjs.org/mystmd -
 You do not have permission to publish "mystmd". Are you logged in as the correct user?
```

so instead I just committed the files modified as a result of the 2 first commands

stevejpurves:
Hi @parmentelat, 

You're going to have to remove the 6d873f5 commit, and force push an update. Basically, you've just bumped the versions in the mystmd packages and attempted to also publish to NPM from your local machine!

The point of using `changesets` is that each PR contributes a small changeset file into the code base (in the `.changesets` folder), many of these are then rolled up into a multi-package version bump + changelog at release time.

All you have to do on your PR is run `npx changeset` and follow the prompts to -- in this case -- add a `patch` level change to the `myst-cli` package. Once you remove 
6d873f5 and commit that the changeset bot should be happy.

parmentelat:
hey @stevejpurves 

thanks for the heads up, I have taken out the problematic commit

and apologies for any mishap; let me just explain how I came to that, in case this happens to others:

- first time around in pr #1993 I got a bot answer here
  https://github.com/jupyter-book/mystmd/pull/1993#issuecomment-2836181979
- that referred me to this page here
  https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

and this is what I - stupidly enough - tried to follow

please let me know what now, if needed

stevejpurves:
Hi @parmentelat, just double checked for my own sanity and yes `npx @changesets/cli` should get you what you need.

after running `npx @changesets/cli` do you not see sets of prompts guiding you though (i) choosing a change level (patch|minor|major) (ii) choosing packages that have changed (or perhaps the other way around) (iii) finally adding a change message?

parmentelat:
yes it does
but it does not alter the repo in any way

it ends up with this message
```
🦋  If you want to modify or expand on the changeset summary, you can find it here
🦋  info .../mystmd/.changeset/thick-seahorses-pump.md
```
but that's not added nor anything...

stevejpurves:
@parmentelat so strange. It should add a simple, randomly named markdown file in the `.changesets` folder, like the one on this PR: https://github.com/jupyter-book/myst-theme/pull/602/files. You can short circuit this by just adding a file manually following that format.

stevejpurves:
hmmm, also you need to run the command from the root folder, probably you are though.

parmentelat:
yeah that file was produced by `npx @changesets/cli`
what I'm saying is, following the doc above a bit too stupidly, there was no mention of having to add/commit this file manually, and instead there were the other commands (version&publish) that I misinterpreted as, precisely, do the right thing with this outcome file

so can you please confirm I just need to add/commit/push that one randomly named file and I'm done ?

stevejpurves:
Ah ok! so yes, confirmed, just commit & push the changeset file as part of your PR.

(and do the same to any PR you do 👍🏼, then the changesets CI actions take care of rolling up releases, like in [this autogenerated release PR](https://github.com/jupyter-book/mystmd/pull/2075))

parmentelat:
@stevejpurves
one hopefully last comment about the changeset thingy:
I realize I have not opted for the `patch` level as you instructed, but for `minor` instead, as this does adds a feature and has an impact on how the tool behaves
let me know if that's ok

parmentelat:
@stevejpurves
one hopefully last comment about the changeset thingy:
I realize I have not opted for the `patch` level as you instructed, but for `minor` instead, as this does adds a feature and has an impact on how the tool behaves
let me know if that's ok

stevejpurves:
Oh @parmentelat I was so focussed on the helping with the changeset, I did not pay attention to the actual change you were making! sorry - as I understand it this change will actually remove hidden pages from the site produced altogether! 

Which I don't think is what we want, we only want to have hidden files flagged as hidden and not shown in the TOC, but still available on the site. The use case I see there is for in-browser compute, where notebooks can be used to create embedded figures / outputs but can be marked as hidden too, so not shown on the site - however for computation to work they still need to be deployed as part of the build.

pinging @fwkoch to confirm my understanding i.e. that this change would remove them from the site build altogether.

PS I am happy to get us started on a change in myst-theme to do what you want.

stevejpurves:
Hmm I also see @fwkoch comment in https://github.com/jupyter-book/mystmd/blob/efcdaccce483788463afd5d4152f4273ceacf46c/packages/myst-cli/src/process/site.ts#L357 pointing you in this direction. I'll need @fwkoch to weigh in here.

parmentelat:
Hey @stevejpurves 
thanks for your input
as far as I can tell from my manual tests,  pages marked as hidden
- do not appear in the toc
- but do get built, and I can access them when I write the proper URL in the browser
so I thought this could be considered for merging
is there anything else I need to do to that end ?

.changeset/thick-seahorses-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"myst-cli": minor
```
fwkoch: I'd still consider this a `patch` change - explanation is buried in the docs here: https://mystmd.org/guide/developer#versioning-and-changesets

```suggestion
"myst-cli": patch
```

parmentelat:
added a word in the docs as per @fwkoch's request
sorry for the noise and all the force-pushes, just trying to keep things tidy....

fwkoch (approved):
Docs update looks great - thank you!

Not sure why the `release` action ran and created a commit. I reverted that; we just want to land this PR with the `.changeset` file - the release will be created in a separate PR.

packages/myst-cli/src/process/site.ts
```diff
@@ -337,7 +337,11 @@ export function selectPageReferenceStates(
   const cache = castSession(session);
   let previousCounts: TargetCounts | undefined;
   const pageReferenceStates: ReferenceState[] = pages
-    .map(({ file }) => {
+    .map(({ file, hidden }: { file: string; hidden?: boolean }) => {
+      if (hidden) {
+        // Skip hidden files
+        return undefined;
```
fwkoch: I think we just need to be less aggressive here... if `hidden`, skip the ~enumeration~ count-incrementing steps, but don't skip everything.

fwkoch (dismissed):
Oh man, sorry to come back and reverse my approving review, but testing it out a bit more, I realized cross-references/embeds/etc of hidden content do not resolve. I think we do want this to work, e.g. I have a notebook that generates a figure, I want to embed the figure elsewhere but I don't want the notebook as part of the toc.

I'll take a quick pass at trying to address this...

fwkoch:
Ok @parmentelat - you mind taking a look at this commit? https://github.com/jupyter-book/mystmd/pull/2097/commits/4e87d3ceb50c7feba4873dba2fc6342688ce5a9b

It keeps numbering disabled but re-enables cross-referencing.

parmentelat:
hi @fwkoch 

for the record, just repeating what I said on discord
I have tested your commit and was able to assess that we now have the following deisred properties
- hidden notebook does not show up in the toc
- it does get built, and can be seen in full through its url
- also labeled cells that it contains can be referred to from other notebooks in the document; I've tested only from non-hidden notebooks mind you, but...

as an aside, and still for the record: I just created #2120 to document the other issue that I cam across when testing for this

fwkoch:
Thanks for re-testing and following up with that issue!

PR #2102, opened 2025-06-14
👽 Handle external URL specified in TOC

### LINKED ISSUE #1445: URL entries in Table of Contents
It is helpful to support URL entries in the Table of Contents in case users want to directly point to an internal or external site from the TOC. This is usually most helpful with web exports, but may also be useful for PDF exports as well.

Our documentation already notes this feature, but says they are ignored:

CleanShot 2024-08-10 at 13 04 58

### LINKED ISSUE #2098: Open ToC URLs in new tabs
## Proposal
@parmentelat made a feature request to open URL entries in new tabs, when defined in the ToC. The URL feature doesn't exist yet, but when it does, we should remember this issue!

> with jb1 and sphinx I had the ability to simply inject some custom JS into the JB output with `html_js_files` as described here https://docs.readthedocs.com/platform/latest/guides/adding-custom-css.html
> 
> I need something similar with jb2; my use case is twofold:
> 
>     * to be able to define keyboard actions - nothing that would mess with the DOM, but typically some cheatcode for the teacher to access a solution notebook whose url is derived from the current one
> 
>     * also historically my jb1 JS addition was tweaking the left-hand-side TOC so that url-defined entries would open in a separate tab
>       this second use case, which indeed requires messing with the DOM a bit, is not currently important given that those url-defined entries are not yet supported as far as  [URL entries in Table of Contents mystmd#1445](https://github.com/jupyter-book/mystmd/issues/1445) has it
> 
> — @parlementat

## Next actions

- [ ] wait for URL entries to be supported in the TOC 
- [ ] implement this feature

stefanv:
Requires https://github.com/jupyter-book/myst-theme/pull/611

Closes #2098 and closes #1445

TODO:

- [x] Add ability to choose whether external link opens in new tab or not
- [x] Document feature

### DIFF
--- a/docs/table-of-contents.md
+++ b/docs/table-of-contents.md
@@ -52,6 +52,15 @@ project:
       title: Google
 ```
 
+By default, URLs open in a new tab. You can change that with the
+`open_in_same_tab` option:
+
+```{code} yaml
+- url: 'https://google.com'
+  title: Google
+  open_in_same_tab: true
+```
+
 ### Glob pattern matching
 
 You can specify glob-like patterns in the TOC with the `pattern` key.


packages/myst-cli/src/project/types.ts
```diff
@@ -24,6 +24,12 @@ export type LocalProjectPage = {
   implicit?: boolean;
 };
 
+export type ExternalURL = {
```
fwkoch, line 27: I think to follow the pattern of where this is slotting in, maybe it should be called `LocalProjectUrl`? (That name is a little funny on it's own, but I think these urls are present alongside `LocalProjectPage | LocalProjectFolder` on `pages` below.)

stefanv: These are distinct from LocalProjectPage and LocalProjectFolder in that they do *not* link to local files. But perhaps I misunderstood and LocalProject meant something different? Happy to change it, but would be helpful if we could provide a hint that this is *external*---i.e., something like https://google.com.

fwkoch: Hmm yeah. I was thinking of the naming as `LocalProject.page`, `LocalProject.folder` - so then `LocalProject.url` doesn't imply "local url" - but that's all just in my head.

If that's confusing for you and others, then I'm ok leaving as you've defined it.

stefanv: I believe you have a much clearer view of the overarching architecture, so I will defer to you on naming that is not confusing in the LocalProject context, and also clearly communicates that it is an external URL.

It may even be that I attached the information in the wrong place. Let me know, and I'm happy to change it.

docs/table-of-contents.md
```diff
@@ -41,11 +41,6 @@ project:
 
 URLs can be defined in the TOC. These URLs are links to external references within your table of contents. URLs are ignored in non-web exports.
 
-:::{warning} Work in progress
-Currently these URLs are also ignored in MyST sites.
-Follow https://github.com/jupyter-book/mystmd/issues/1445 for this enhancement and please provide feedback.
-:::
-
```
mfisher87, line 48: :raised_hands:

packages/myst-cli/src/project/types.ts
mfisher87: As someone who doesn't have a clear view of the architecture, having `external` in the name is a very useful hint. I shared Stefan's reading of these names, i.e. `LocalProjectPage` and `LocalProjectFolder` refer to artifacts that exist materially in the local project, so I found `ExternalUrl` a good name that differentiates clearly from a `LocalProjectPage` or `LocalProjectFolder`.

I see that `LocalProject.pages: Array<(LocalProjectPage | LocalProjectFolder | ExternalURL)>` makes `ExternalURL` a part of the `LocalProject`, so perhaps for consistency we could use `LocalProjectExternalURL`?

packages/myst-cli/src/build/utils/projectManifest.ts
```diff
@@ -43,6 +43,17 @@ export async function manifestPagesFromProject(session: ISession, projectPath: s
...
+        const { title, url, open_in_same_tab, level } = page;
+        const externalURL: ManifestProject['pages'][0] = {
+          title,
+          url,
+          open_in_same_tab,
+          level,
+        };
+        return externalURL;
+      }
+      // Not a file or URL, probably a sub-entry; try and bubble down
```
choldgraf, line 56: bubble down is a great verb lol

packages/myst-cli/src/project/types.ts
choldgraf: `LocalProjectExternalURL` feels the most explicit and consistent to me. +1 on just being a bit verbose and doing that.

choldgraf (approved):
IMO once the naming thing is resolved this feels good to me :-)

rowanc1 (changes requested):
Only comment is on the naming of the option. The default nav action is to open in the same tab, so changing this would be adding a new option. I would prefer the new option to have the `true` as the value, as it is less confusing. This would mean flipping the name to be: `open_in_new_tab: true`. That way undefined and false mean the same thing.

fwkoch:
@rowanc1 - I had a kinda opposite comment here: https://github.com/jupyter-book/myst-theme/pull/611#discussion_r2153814104

I think the question is what the default behavior should be: I was thinking `internal` toc entries are same tab, `external` URL entries are new tab. And for now this option is only applicable to the `external` entries. Then `same_tab` is `false` by default, which matches undefined. This is also how it's currently implemented (only applying to URL entries): https://github.com/jupyter-book/myst-theme/pull/611/files#diff-9be3b8c49cad78bc714a928e7e3548397033173fa72ccaf796d600bccc24953cR93

If this option applies to all toc entries, then I would agree `same_tab: true` is the default, and `new_tab` would be the correct name.

rowanc1:
Ok, didn't realize this was only for external urls. Then the `false === undefined`, and all is good.

packages/myst-config/src/site/types.ts
```diff
@@ -52,6 +52,9 @@ type ManifestProjectItem = {
   bannerOptimized?: string;
   tags?: string[];
   enumerator?: string;
+  // For external URLs
+  url?: string;
+  open_in_same_tab?: boolean;
```
rowanc1, line 57: Should this be open urls in same tab then?

fwkoch: I had this same thought, but then I realized this is no longer a global config option. It's on the toc item itself:
```
toc:
  - url: https://example.com
    title: Example
    open_in_same_tab: true
```
In that case `open_urls` is redundant.

Currently if you use this on a non-url entry:
```
toc:
  - file: example.md
    open_in_same_tab: true
```

It should just give a validation warning and the option will be ignored.

@stefanv - please correct me if my current understanding is off!

stefanv: That's right; here's the [corresponding documentation](https://github.com/jupyter-book/mystmd/pull/2102/files#diff-a459d5b8270fdbf605c531e3e973331bf618ee847320b67415338266552e4e4dR55)

docs/table-of-contents.md
```diff
@@ -54,6 +49,16 @@ project:
   toc:
     - file: root.md
     - url: 'https://google.com'
+      title: Google
+```
+
+By default, URLs open in a new tab. You can change that with the
```
fwkoch, line 55: 👍

fwkoch (approved):
I tested this out - works great. Urls render in the toc, same tab behavior is respected, and I get a console warning when I try to set `open_in_same_tab` on non-url toc entry.

For variable naming, I support Chris's suggestion of `LocalProjectExternalURL`.

The only other thing I noticed was this needs support in the `{toc}` directive. I created an issue here: https://github.com/jupyter-book/mystmd/issues/2119 - we can either address as part of this PR or separately.

choldgraf:
I just want to signal-boost the practice of opening an issue to track some nagging behavior that we need to improve, but that doesn't necessarily need to block this PR. I think that's great :-D 

@stefanv want to decide whether you wanna tackle the `{toc}` thing or just focus on merging current functionality?

fwkoch:
I'm gonna land this so it isn't lingering too long. We can tackle `{toc}` separately!

stefanv:
@fwkoch Did you want me to submit a change from `ExternalURL` to `LocalProjectExternalURL`? [code](https://github.com/jupyter-book/mystmd/pull/2102/files#diff-9d43afdfd1cd05cc3c9c95a41f41045c9de1033802e28f803d6a5be176d5a7bfR18)

fwkoch:
@stefanv - sure, if you make that change we can merge it right away! I just didn't want this PR to be stuck on internal API naming.

PR #2108, opened 2025-06-16
📝 Add a note about our release processes

agoose77:
This follows from some discussion on Discord.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -382,10 +382,8 @@ https://api.mystmd.org/templates/typst
 
 ## How to make a release
 
-:::{important} Informing other developers
-Before making a release, it is good practice to announce your intention in the [`#release_coordination`](https://discord.com/channels/1083088970059096114/1384242935645737141) Discord channel. This gives the rest of the team an opportunity to coordinate around the release, e.g. to ensure that features are not half-shipped. 
-
-A good rule of thumb is to give people a "day" to see a release notice and engage with the announcement. Due to the range of time-zones that the Jupyter Book team are subject to, this effectively means a 24h wait. This is a subjective rule, and your best judgement should be used to determine whether you need to hear back from the rest of the team before pushing the release. For example, releases consisting only of critical bug-fixes may be made immediately if time-sensitive. 
+:::{important} Release Coordination
+We use the [`#release_coordination`](https://discord.com/channels/1083088970059096114/1384242935645737141) Discord channel to coordinate and collaborate on releases. For example, to ensure that features are not half-shipped, additional PRs can make it into a release, or other release responsibilities that can be shared among the team (e.g. release notes, theme deployments, social media posts, documentation updates).
 :::
 
 ### Make a release of `mystmd`


docs/developer.md
```diff
@@ -380,8 +380,12 @@ And to see a list of available templates that can be resolved for Typst, you can
 
 https://api.mystmd.org/templates/typst
 
-
 ## How to make a release
+
+:::{important} Informing other developers
+Making a release of MyST components may require other stakeholders to update their deployments of MyST. In order to minimise disruption, contributors should inform other developers of their intention to make a release in the [`#development` Discord channel](https://discord.com/channels/1083088970059096114/1177359316886503464). This step may be skipped if the benefits of making an unannounced release outweigh the risks, such as mitigating a security vulnerability or preventing data loss.
```
rowanc1: I think this can't be skipped on the informing side. "Making a release now." Is not hard to inform! If this should not be public, then coordinate in private channels.

It would be good to coordinate: "planning to make a release today to .., any one have anything else they are planning to get in planning this by the end of my day in a few hours?"

rowanc1 (changes requested):

choldgraf:
Thanks for kicking this discussion off @agoose77 - I have a quick question: what's the downside risk we are worrying about, and our overall goals for releases? For example:

### Example goals

- Make releases as often as possible so that the latest released version is not too far from `main`.
- Make releases as reliably as possible so that users know when to expect new functionality, and we know when to dedicate time to releases.
- Bias towards "quality over quantity" (or, vice versa)

### Example downsides

- Are we worried that making a release will "surprise" consumers of the MyST stack in a negative way?
- Are we worried that making a release will not include enough features to justify it because they weren't merged yet?
- Are we worried that making too many releases will generate too much noise for consumers, or toil for maintainers?

docs/developer.md
agoose77: I think it's worth separating this into "notice" vs "synchronisation". The latter can be skipped (or, if we can wait a degree, with a time-box e.g. close of business)

choldgraf:
Also a quick reference: the JupyterLab team uses this Zulip thread for their releases:

[#Release Coordination](https://jupyter.zulipchat.com/#narrow/channel/469762-jupyterlab/topic/Release.20coordination/with/523774156)

It might be worth having a dedicated channel (or thread within Discord? Is that a thing?) for discussion around releases. The `#Developer` channel can get a bit noisy in my experience.

agoose77:
@rowanc1 echoing Chris' thoughts, it would be good to learn what your main concerns are here -- is it about missing that a release has happened, or wanting to prepare contributions to a release?

My preference is for us to ship often and with low cost, so the main benefit is letting people know that we've made a release over synchronising with other developers. But, I'm curious if there's an aspect to this that doesn't work for you?

choldgraf:
Personally, I think it's a generally good practice to have a loop like the following for any action:

- Signal intent.
- Wait for objections.
- If no objections, take action.
- If objections, define a new intent to act and repeat the process.

As long as the project has a culture where objections are rare and only in strategically critical circumstances, it doesn't slow things down too much and makes everything more predictable. Personally, if I had more visibility in "I'm planning to make a release today", it would give me more time to plan my own capacity for things like "merging that last PR I've been procrastinating on", or "making some time to make a GitHub release and post about it on social media".

That said, I also think we want to have a team culture that biases ourselves towards action and momentum, and I recognize that any "slow down" signal runs the risk of becoming bureaucratic and cumbersome, so I want to be sensitive to that too.

docs/developer.md
choldgraf: It feels like an easy and safe to try start would be a team policy like:

- Before making a release for any repository, make a post to `#development` stating your intent to do so.
- Use your best judgment about whether to wait any amount of time before taking action.

And then if we run into friction because we released too quickly, then we can add stronger guidelines for waiting for feedback. Or if it's too noisy in `#development` and we miss these messages, we can discuss adding a new channel.

rowanc1:
Not worried about making many small releases fast, however, I would like us to **coordinate** on releases. That is not a high burden nor impacts velocity:

- image
- "Planning a release later today, any one have anything else they want in."

That is much more healthy than just hitting merge on a PR with zero conversation.

We aren't stopping or slowing down in general: we are increasing visibility to the rest of the team and have some lite coordination on review (e.g. missed changesets, coordination on themes, downstream implications, "that requires a release to... ", "hold off on docs deploy until...", etc.). We have hit minor issues with all of these in the last month.

However, there are also changes which are going to require more coordination and not be released without some effort. The upcoming changes on outputs, for example, absolutely has downstream implications and production systems hanging off that. We should get into a habit of checking in and coordinating.

docs/developer.md
rowanc1: Made a `release_coordination` channel.

agoose77:
I understand that there are two threads here. I'm comfortable with the idea that we need to be aware that releases are happening, and it's best-practice for us to announce those ahead of time and/or as they happen (according to need).

The questions I had above are about avoiding the anti-pattern of blocking.

With both of these in mind, I've updated the PR body. Does this feel right to you both?

rowanc1:
Can we please move to a mindset of **coordination**, **collaboration** and **planning**, and not viewing that as "blocking"?

docs/developer.md
```diff
@@ -107,7 +108,7 @@ MyST has multiple renders, themes, and templates that allow it to transform MyST
 - describe the render loop, and how render blocks are registered
 - explain the ThemeProvider
 - explain styling
-:::
+  :::
```
choldgraf: Is this an intentional thing?

choldgraf (approved):
This looks safe to try to me. My main concern is that asking for a full day to wait is a lot of time. It means a person can't close this loop within their same work day. I suggest we revisit in a few months and decide if the day timeframe is too cumbersome

docs/developer.md
agoose77: My damn LSP linter. Let me tweak.

docs/developer.md
```diff
@@ -380,8 +380,14 @@ And to see a list of available templates that can be resolved for Typst, you can
 
 https://api.mystmd.org/templates/typst
 
-
 ## How to make a release
+
+:::{important} Informing other developers
+Before making a release, it is good practice to announce your intention in the [`#release_coordination`](https://discord.com/channels/1083088970059096114/1384242935645737141) Discord channel. This gives the rest of the team an opportunity to coordinate around the release, e.g. to ensure that features are not half-shipped. 
+
+A good rule of thumb is to give people a "day" to see a release notice and engage with the announcement. Due to the range of time-zones that the Jupyter Book team are subject to, this effectively means a 24h wait. This is a subjective rule, and your best-judgement should be used to determine whether you need to hear back from the rest of the team before pushing the release. For example, releases consisting only of critical bug-fixes may be made immediately if time-sensitive.
```
stefanv:
```suggestion
A good rule of thumb is to give people a "day" to see a release notice and engage with the announcement. Due to the range of time-zones that the Jupyter Book team are subject to, this effectively means a 24h wait. This is a subjective rule, and your best judgement should be used to determine whether you need to hear back from the rest of the team before pushing the release. For example, releases consisting only of critical bug-fixes may be made immediately if time-sensitive. 
```

agoose77: Nice catch!

agoose77:
@rowanc1 I think we're on the same page -- I'm trying to do a few things at once, and my messages have been consequently shorter than usual.

I am being explicit that we want processes that don't sync by default. Your point
> an we please move to a mindset of coordination, collaboration and planning, and not viewing that as "blocking"?

That _isn't_ blocking. Notifying people that things are happening is async-friendly. Planning and collaboration I think can be left to the judgement of the author, but this is the detail that we need to figure out.

Maybe this would be easier to discuss in a sync meeting, but here's my view:

- I would like for us to be able to make "cheap releases" that don't require filing a self-todo to come back later and merge. These are restricted to non-breaking (new exports, bugfixes).
- I would like us to block on releases that need coordination (breaking or multi-PR changes).

Does this make sense?

docs/developer.md
```diff
@@ -380,8 +380,14 @@ And to see a list of available templates that can be resolved for Typst, you can
...
 https://api.mystmd.org/templates/typst
 
-
 ## How to make a release
+
+:::{important} Informing other developers
+Before making a release, it is good practice to announce your intention in the [`#release_coordination`](https://discord.com/channels/1083088970059096114/1384242935645737141) Discord channel. This gives the rest of the team an opportunity to coordinate around the release, e.g. to ensure that features are not half-shipped. 
+
+A good rule of thumb is to give people a "day" to see a release notice and engage with the announcement. Due to the range of time-zones that the Jupyter Book team are subject to, this effectively means a 24h wait. This is a subjective rule, and your best judgement should be used to determine whether you need to hear back from the rest of the team before pushing the release. For example, releases consisting only of critical bug-fixes may be made immediately if time-sensitive. 
+:::
```
rowanc1: Can we try this as a statement of coordinating and collaborating — rather than defining this as a "rule" like "wait 24 hours before action X", let's start by just stating our intent and goals as a team to collaborate in a certain place around release. This will be simpler, and we can revisit more specific guidelines or rules if we still miss the mark or need more guard rails.

```suggestion
:::{important} Release Coordination
We use the [`#release_coordination`](https://discord.com/channels/1083088970059096114/1384242935645737141) Discord channel to coordinate and collaborate on releases. For example, to ensure that features are not half-shipped, additional PRs can make it into a release, or other release responsibilities that can be shared among the team (e.g. release notes, theme deployments, social media posts, documentation updates).
:::
```

Asking for a conversation and collaboration around a release:

- "Planning a release tomorrow after .. PR526 gets in, working on the PR now and really excited about it"
- "I have some time to do release notes!"
- "Just looked at the PR ... looks like a changeset is missed. Can you add one?"
- "Does this need a docs website update.. I don't think we have done that yet?"
- "On it. Yeah, I can handle that"
- "super excited to see these features go out!"

There are downstream changes that sometimes the release does take 4 hours to chase everything through (themes, updates, website, docs, social, release notes, etc.), and we shouldn't be blocked on waiting to hit a merge button on each step. We should prioritize visibility and collaboration that this is happening and quick conversation days in advance so people know what is going on or where people are jumping in.

stefanv: > or other aspects of a release can be shared responsibility with the team

Small typo there, maybe:

> or other release responsibilities can be shared among the team

Otherwise, looks good to me as a way to agree upon the general principle.

rowanc1: Updated the text. 🚀

choldgraf: This sounds good to me - I like the focus on lightweight team practices and guidelines rather than hard rules

agoose77:
@rowanc1 @choldgraf I've accepted Rowan's changes above, which drop the time-box and add emphasises coordination.

@rowanc1 I want it to be clear that we're on the same page about the primary _motivation_ for this process note, which follows from your observation on Discord about being surprised by releases (let me know if I'm paraphrasing that wrong!)

The original form of this PR also addressed a second issue -- what do maintainers do when we don't receive any engagement on these kinds of coordination posts? Ultimately, I want us to distribute the ability to make these kinds of decisions if we don't have the ability to make them as a team.

I don't perceive that to be the _main_ issue right now — and that nuance might have been lost above. My proposal is that we take your wording for now, and if it feels like the process isn't working quite right, we revisit this conversation then.

Given that this is @rowanc1's edit, and @choldgraf has approved the PR, and I also approve the changes, I'm comfortable merging this one in.

Thanks for the lively discussion everyone!

PR #2110, opened 2025-06-17
Update documentation about in-browser execution

choldgraf:
This refreshes the page about in-browser execution with Thebe. It does a few major things:

- Updates the documentation to be consistent in how we recommend configuration.
- Adds some examples of what the UI looks like to control execution, and what is possible.
- Streamlines the sections to follow the `how to` section title structure we've been using throughout.
- Generally reduce the amount of words and focus on action sentences and bullets.
- Activates in-browser computation for our documents so that we can

### Next steps

- [ ] Review the content again to make sure it is all correct
- [ ] Figure out why page-level over-rides aren't working. I tried this for just one page, but couldn't get it to work, which is why I switched to a project-level config.


choldgraf:
@stevejpurves would you be willing to give this one a look when you have a moment? It's not urgent but I think you know the most about MyST config and in-browser execution. Lemme know what you think

docs/in-browser-execution.md
```diff
@@ -0,0 +1,351 @@
...
+* - `provider`
+  - optional
+  - Tells `thebe` how to form urls for requesting binder services. Can be one of `github`, `gitlab` or `git`. See [](#binder-option-repoprovider).
+  - `github`
+```
+
+```{attention} Be aware of multiple keys
+There are two possible locations for `binder` keys:
+
+1. `project.binder` is for [launch button configuration](./website-launch-buttons.md), and will take the user to a new page with a Binder session.
```
agoose77: I love that you're disambiguating these Chris! To add a little bit of context - the top level binder property predates the launch work, and I chose to use it for reasons.


I would really like us to simplify all this, and there is a semi refined initiative to tackle it!

docs/in-browser-execution.md
```diff
@@ -0,0 +1,351 @@
...
+subtitle: Connecting Jupyter Kernels to your MyST website
+description: MyST allows you to connect a website directly to a Jupyter Kernel, enabling interactive computation on your page.
+thumbnail: thumbnails/integrating-jupyter.png
+aliases:
+  - integrating-jupyter
+---
+
+In-browser execution allows you to connect a live Jupyter kernel to your website, allowing you to add interactive visualizations and interactivity that is powered with a live kernel. This allows your readers to reproduce the outputs of your document in real-time, and to add more complex interactive content via tools like [`ipywidgets`](https://ipywidgets.readthedocs.io/) or [Panel](https://github.com/holoviz/panel).
+
+There are a variety of back-ends that can provide the computation for in-browser execution, such as **Binder**, **JupyterHub**, or **WASM-based kernels from JupyterLite**. See [](#thebe-backend-options) for more information.
```
agoose77: I am on my phone in sunny Belgium, so I can't derefence this label: I don't think we naively support JupyterHub - there's no auth support IIRC (see the above comment)

stevejpurves: yes there is no way to request / negotiate a server start & connection with a JupyterHub unless it has a BinderHub service installed.

docs/in-browser-execution.md
stevejpurves: imo the `project.binder` is legacy and should be dropped, it would be better if the launch button and in-browser drew from the same set of options (any reason why not?), independent of the initiative to reinvent the options layout.

stevejpurves:
@choldgraf I think the per page options not overriding is probably a theme limitation, as in it is not implemented in the core themes (yet?).

stevejpurves (approved):
Looks great @choldgraf - see comments above

docs/in-browser-execution.md
choldgraf: Sounds good - where's the issue where we can discuss the right config to use? Either way I'll leave this content as-is since it's the current reality.

choldgraf:
I've addressed the comment about JupyterHub - we should def improve that workflow so people can use their hubs to power their docs, but that can be a follow up enhancement. I've made one change, which is to change `in-browser-execution` to `in-page-execution`, since technically _everything_ in Jupyter interfaces is in-browser-execution and this disambiguates it a bit :-) 

I'll merge this one in, thanks all!

PR #2111, opened 2025-06-18
🗾 HTML Image Width

rowanc1:
Fixes ProjectPythia/cookbook-template#226


rowanc1:
Merging this (very safe!), has tests, changeset and tested locally that images work and can now be controlled by width/height.

PR #2115, opened 2025-06-19
🧮 Improve inline math symbol replacements

### LINKED ISSUE #1604: Math expressions are rendered in different fonts for inline vs display
## Description

Using inline versus whole line equations, some characters look different, e.g. gamma in this example:

Image

This has the potential to be quite confusing I think. This is from MyST version 1.2.9.

Incidentally those theta symbols are supposed to be bold ``\mathbf\theta`` but this might not be a MyST issue?

## Proposed solution

No idea.

rowanc1:
Fixes #1604

### DIFF
--- a/.changeset/light-pillows-love.md
+++ b/.changeset/light-pillows-love.md
@@ -1,0 +1,6 @@
+---
+"myst-transforms": patch
+"myst-cli": patch
+---
+
+Improve symbol replacements.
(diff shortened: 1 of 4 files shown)


rowanc1:
This is an improvement to recent issues users are facing - isolated change, have added tests, changesets and tested in CLI. Symbols are no longer replaced, numbers still are. Merging it in.

PR #2129, opened 2025-06-26
⨱ Underline in html is parsed correctly

rowanc1:


packages/myst-transforms/src/html.ts
```diff
@@ -163,6 +163,11 @@ const defaultHtmlToMdastOptions: Record<keyof HtmlTransformOptions, any> = {
     summary(h: H, node: any) {
       return h(node, 'summary', all(h, node));
     },
+    u(h: H, node: any) {
+      // The default is emphasis
```
rowanc1, line 167: For some reason the default in hast is `emphasis`!

PR #2132, opened 2025-06-27
🧜‍♀️ Create images of mermaid diagrams for static exports

### LINKED ISSUE #2006: Support export mermaid diagrams to docx as images.
Title says it: I would like to export mermaid diagrams to docx as images.

Image

❯ myst -v
v1.3.26

## Additional notes

- maybe this helps
- https://www.npmjs.com/package/mermaid
- https://www.npmjs.com/package/markdown-mermaid-exporter
- https://www.npmjs.com/package/@mermaid-js/mermaid-cli

rowanc1:
Fixes #2006

Requires that you have `mmdc` installed, the CLI warns you and I have added some docs. Works with typst, tex, docx and jats.

Image

image

Replaces https://github.com/jupyter-book/mystmd/pull/1452/

### DIFF
--- a/.changeset/stale-ghosts-drum.md
+++ b/.changeset/stale-ghosts-drum.md
@@ -1,0 +1,5 @@
+---
+'myst-spec-ext': patch
+---
+
+Added `label`, `identifier`, and `html_id` to `Image` type
--- a/docs/diagrams.md
+++ b/docs/diagrams.md
@@ -37,3 +37,19 @@ npm install -g @mermaid-js/mermaid-cli
 ```
 
 :::
+
+:::{note .dropdown} For CI Environments
+
+The Mermaid CLI uses Puppeteer which may require special configuration in CI environments. MyST automatically handles this by detecting the `CI` environment variable or you can manually control it:
+
+```bash
+# Automatic detection (recommended for CI)
+CI=true npm run build
+
+# Manual control
+MERMAID_NO_SANDBOX=true npm run build
+```
+
+This automatically creates a Puppeteer configuration file with `--no-sandbox` flag to resolve sandbox issues in Linux CI environments.
+
+:::
--- a/docs/diagrams.md
+++ b/docs/diagrams.md
@@ -24,3 +24,16 @@ flowchart LR
 :::{note}
 Both GitHub and JupyterLab ([#101](https://github.com/jupyter/enhancement-proposals/pull/101)) support the translation of a code-block ` ```mermaid ` to a mermaid diagram directly, this can also be used by default in MyST.
 :::
+
+## Rendering for Static Exports
+
+MyST supports static rendering of Mermaid diagrams for static export formats (PDF, Word, LaTeX, Typst). This feature converts Mermaid syntax to base64-encoded SVG or PNG images during the build process, ensuring consistent rendering across all static output formats.
+
+:::{important} Prerequisites
+To use static Mermaid rendering, you need the Mermaid CLI installed:
+
+```bash
+npm install -g @mermaid-js/mermaid-cli
+```
+
+:::
--- a/packages/jtex/tests/download.spec.ts
+++ b/packages/jtex/tests/download.spec.ts
@@ -4,57 +4,53 @@ import { TemplateKind } from 'myst-common';
 import MystTemplate, { downloadTemplate, resolveInputs, Session } from 'myst-templates';
 import { renderTemplate } from '../src';
 
-describe(
-  'Download Template',
-  () => {
-    it('Download default template', async () => {
-      const session = new Session();
-      const inputs = resolveInputs(session, { buildDir: '_build', kind: TemplateKind.tex });
-      await downloadTemplate(session, {
-        templatePath: inputs.templatePath,
-        templateUrl: inputs.templateUrl as string,
-      });
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.zip')).toBe(true);
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.yml')).toBe(true);
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.tex')).toBe(true);
+describe('Download Template', { timeout: 15000 }, () => {
+  it('Download default template', async () => {
+    const session = new Session();
+    const inputs = resolveInputs(session, { buildDir: '_build', kind: TemplateKind.tex });
+    await downloadTemplate(session, {
+      templatePath: inputs.templatePath,
+      templateUrl: inputs.templateUrl as string,
     });
-    it('Bad template paths to throw', async () => {
-      const jtex = new MystTemplate(new Session(), {
-        template: 'not-there',
-        kind: TemplateKind.tex,
-      });
-      expect(() => jtex.prepare({} as any)).toThrow(/does not exist/);
+    expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.zip')).toBe(true);
+    expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.yml')).toBe(true);
+    expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.tex')).toBe(true);
+  });
+  it('Bad template paths to throw', async () => {
+    const jtex = new MystTemplate(new Session(), {
+      template: 'not-there',
+      kind: TemplateKind.tex,
     });
-    it('Render out the template', async () => {
-      const jtex = new MystTemplate(new Session(), {
-        kind: TemplateKind.tex,
-        template: `${__dirname}/example`,
-      });
-      renderTemplate(jtex, {
-        contentOrPath: `${__dirname}/test.tex`,
-        outputPath: '_build/out/article.tex',
-        frontmatter: {
-          title: 'test',
-          description: 'test',
-          date: new Date(2022, 6, 22).toISOString(),
-          authors: [
-            { name: 'Rowan Cockett', affiliations: ['Curvenote'] },
-            { name: 'Steve Purves', affiliations: ['Curvenote'], orcid: '0000' },
-          ],
-        },
-        options: {
-          keywords: '',
-        },
-        parts: {
-          abstract: 'My abstract!',
-        },
-      });
-      expect(fs.existsSync('_build/out/article.tex')).toBe(true);
-      const content = fs.readFileSync('_build/out/article.tex').toString();
-      expect(content.includes('Volcanic Archipelago')).toBe(true);
-      expect(content.includes('Rowan Cockett')).toBe(true);
-      expect(content.includes('My abstract!')).toBe(true);
+    expect(() => jtex.prepare({} as any)).toThrow(/does not exist/);
+  });
+  it('Render out the template', async () => {
+    const jtex = new MystTemplate(new Session(), {
+      kind: TemplateKind.tex,
+      template: `${__dirname}/example`,
     });
-  },
-  { timeout: 15000 },
-);
+    renderTemplate(jtex, {
+      contentOrPath: `${__dirname}/test.tex`,
+      outputPath: '_build/out/article.tex',
+      frontmatter: {
+        title: 'test',
+        description: 'test',
+        date: new Date(2022, 6, 22).toISOString(),
+        authors: [
+          { name: 'Rowan Cockett', affiliations: ['Curvenote'] },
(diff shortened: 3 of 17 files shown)


packages/jtex/tests/download.spec.ts
```diff
@@ -4,57 +4,53 @@ import { TemplateKind } from 'myst-common';
...
-      const inputs = resolveInputs(session, { buildDir: '_build', kind: TemplateKind.tex });
-      await downloadTemplate(session, {
-        templatePath: inputs.templatePath,
-        templateUrl: inputs.templateUrl as string,
-      });
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.zip')).toBe(true);
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.yml')).toBe(true);
-      expect(fs.existsSync('_build/templates/tex/myst/plain_latex/template.tex')).toBe(true);
+describe('Download Template', { timeout: 15000 }, () => {
+  it('Download default template', async () => {
```
rowanc1, line 8: This is just putting the timeout as the second arg.

docs/diagrams.md
```diff
@@ -24,3 +24,32 @@ flowchart LR
...
+
+The Mermaid CLI uses Puppeteer which may require special configuration in CI environments. MyST automatically handles this by detecting the `CI` environment variable or you can manually control it:
+
+```bash
+# Automatic detection (recommended for CI)
+CI=true npm run build
+
+# Manual control
+MERMAID_NO_SANDBOX=true npm run build
+```
```
rowanc1, line 51: CI was way harder to get going than I thought. Should be automatic now though!

.changeset/stale-ghosts-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'myst-spec-ext': patch
+---
+
```
agoose77, line 4: I noticed this bug this week! :tada:

packages/myst-cli/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@jupyterlab/services": "^7.0.0",
+    "@mermaid-js/mermaid-cli": "^10.9.1",
```
fwkoch, line 46: I think this dependency is only here to install `mmdc` CLI, not ever imported in the codebase. Is that correct?

If so, I think we should remove it as a dependency for all users. We already have installation instructions in the error message when mermaid conversion fails.

packages/myst-cli/src/process/mdast.ts
```diff
@@ -112,6 +118,8 @@ export async function transformMdast(
     imageExtensions?: ImageExtensions[];
     watchMode?: boolean;
     execute?: boolean;
+    transforms?: TransformSpec[];
```
fwkoch, line 121: This small change is actually a pretty substantial change to the API - It's a completely new way to specify transforms within the code. I think it is non-controversial and for-the-better: matching how external plugins are defined and executed is great.

But worth noting this alone is a new feature, entirely independent of the mermaid stuff.

packages/myst-cli/src/build/docx/single.ts
```diff
@@ -98,6 +99,7 @@ export async function runWordExport(
     projectPath,
     imageExtensions: DOCX_IMAGE_EXTENSIONS,
     extraLinkTransformers,
+    transforms: [createMermaidImageMystPlugin],
```
fwkoch, line 102: Injecting this as `transforms` for each export type is interesting.

The current pattern for "we need to transform something so it works in a static export" is put it in `finalizeMdast` under the flag `simplifyFigures: true`, e.g. https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/process/mdast.ts#L441

Using a transform instead is maybe (probably...?) better: the interface is cleaner and follows existing plugin work, and we are not just dumping more random stuff in the `finalizeMdast` function. However, this now splits transforms with similar purpose ("simplify images for print") across 2 ways of doing things.

Probably fine to leave as-is, and maybe start moving the `finalizeMdast` steps to `transforms` (although that's not trivial - many of those steps need to happen as the very last thing during processing and right now transforms are only run at specific, earlier points). (Also... that begs the question - should the mermaid transform be run later during processing? Probably not...? But there could be edge cases where a mermaid transform is added later and never converted to an image.)

fwkoch (changes requested):
Tested this out - it works very smoothly! First I did not have `mmdc` installed - my diagram rendered with `myst start` but the PDF export reported an error with install instructions. Then I installed `mmdc` and tried the export again and the diagram conversion worked:
image

My only requested change is removing `@mermaid-js/mermaid-cli` from our dependencies; to me, it feels heavy and unnecessary - it's easy for users to install the CLI separately.

Other than that - there's a few high-level aspects that gave me pause (new `transforms` API for `processMdast` functions; different way of finalizing images for "static" output outside of `finalizeMdast` function). I made some comments and am happy to discuss further, but I don't think those discussions need to block landing this feature.

docs/diagrams.md
```diff
@@ -24,3 +24,32 @@ flowchart LR
...
+
+```bash
+npm install -g @mermaid-js/mermaid-cli
+```
+
+:::
+
+:::{note .dropdown} For CI Environments
+
+The Mermaid CLI uses Puppeteer which may require special configuration in CI environments. MyST automatically handles this by detecting the `CI` environment variable or you can manually control it:
```
stefanv, line 43: What is being handled here?

The example is a bit confusing, because `CI` is *already* set on both GH and GitLab, so user would never set it themselves.

docs/diagrams.md
stefanv: Nice!

docs/diagrams.md
```diff
@@ -24,3 +24,32 @@ flowchart LR
...
+
+```bash
+# Automatic detection (recommended for CI)
+CI=true npm run build
+
+# Manual control
+MERMAID_NO_SANDBOX=true npm run build
+```
+
+This automatically creates a Puppeteer configuration file with `--no-sandbox` flag to resolve sandbox issues in Linux CI environments.
```
stefanv, line 53: OK, this answers part of my question from above. Does generating the config equate to disabling the sandbox? If so, maybe something like:

```
By default, Mermaid runs Puppeteer in a sandbox. However, in CI it needs to run directly in the current session. When the CI environment variable is set (which is the case for GitHub and GitLab), then myst will disable the sandbox by generating a suitable Puppeteer configuration file. You can also control this behavior explicitly with `MERMAID_DISABLE_SANDBOX=true/false`.
```

`X_NO_Y` namings don't read correctly in my head, but could be just me. Should this be `PUPPETEER_DISABLE_SANDBOX` or `MERMAID_DISABLE_SANDBOX`?

packages/myst-cli/src/transforms/mermaid.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+      type: 'mermaid' as const,
+      value: 'flowchart LR\n  A --> B',
+      identifier: 'test-diagram',
+    };
+
+    const result = await renderMermaidDiagram(file, mermaidNode);
+
+    expect(result.type).toBe('image');
+    if (result.type === 'image') {
+      expect(result.url).toMatch(/^data:image\/svg\+xml;base64,/);
```
stefanv, line 19: :heart:

packages/myst-cli/package.json
stefanv: How are dependencies determined for the overall mystmd package? Presumably, you *could* have `mmdc` as a runtime dependency?

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+import { createTempFolder } from '../utils/createTempFolder.js';
+import pLimit from 'p-limit';
+
+const execAsync = promisify(exec);
+
+// Create a limiter with concurrency limit
+const limitConnections = pLimit(5);
+
+function isMermaidCliAvailable(): boolean {
+  return !!which.sync('mmdc', { nothrow: true });
```
stefanv, line 22: Took a double-take on that `!!` before realizing what it does :D

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+  theme = 'default',
+  format: 'png' | 'svg' = 'svg',
+): Promise<string> {
+  // Create unique temporary output file name
+  const hash = crypto.createHash('md5').update(mermaidCode).digest('hex');
+  const tempFolder = createTempFolder();
+  const outputFile = path.join(tempFolder, `mermaid-output-${hash}.${format}`);
+
+  if (!isMermaidCliAvailable()) {
+    throw new Error('Mermaid CLI is not available');
```
stefanv, line 49:
```suggestion
    throw new Error('Mermaid CLI is not available: try installing using `npm install -g @mermaid-js/mermaid-cli`');
```

If not this, then point users to the docs page where it is discussed?

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  try {
+    // Create puppeteer config file for CI environments
+    // See https://github.com/mermaid-js/mermaid-cli/blob/master/docs/linux-sandbox-issue.md
+    const useNoSandbox = process.env.MERMAID_NO_SANDBOX === 'true' || process.env.CI === 'true';
```
stefanv, line 55: In the docs, it says the user gets to control the behavior. It sounded like "regardless of whether in CI or not". So, what happens when `MERMAID_NO_SANDBOX === 'false'` and `CI === true`? Maybe the condition should be:

```
const useNoSandbox =
  process.env.MERMAID_NO_SANDBOX !== undefined
    ? process.env.MERMAID_NO_SANDBOX === 'true'
    : process.env.CI === 'true';
```

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (useNoSandbox) {
+      const puppeteerConfigPath = path.join(tempFolder, 'puppeteer-config.json');
+      const puppeteerConfig = { args: ['--no-sandbox'] };
+      await fs.writeFile(puppeteerConfigPath, JSON.stringify(puppeteerConfig, null, 2));
+      puppeteerConfigFlag = ` -p "${puppeteerConfigPath}"`;
+    }
+
+    // Render using Mermaid CLI with stdin input
+    await execAsync(
+      `echo '${mermaidCode.replace(/'/g, "'\"'\"'")}' | mmdc -i - -o "${outputFile}" -t ${theme} -b transparent${puppeteerConfigFlag}`,
```
stefanv, line 67: Yikes, this shell escaping is hard to parse.

Perhaps we need a utility function that promisifies `spawn` and accepts standard input, a-la https://stackoverflow.com/a/72863312?

stefanv: Also, is the escaping of `-o` sufficient for strange filenames?

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+async function renderMermaidToBase64(
+  mermaidCode: string,
+  theme = 'default',
+  format: 'png' | 'svg' = 'svg',
+): Promise<string> {
+  // Create unique temporary output file name
+  const hash = crypto.createHash('md5').update(mermaidCode).digest('hex');
+  const tempFolder = createTempFolder();
```
stefanv, line 45: Would be nice to have these clean up after themselves, like in Python. Using something like https://github.com/Mcsavvy/contextlib or a custom implementation:

```
withTempDir(callback)
```

packages/myst-cli/src/transforms/mermaid.ts
```diff
@@ -0,0 +1,161 @@
...
+      html_id: node.html_id,
+    };
+
+    return imageNode;
+  } catch (error) {
+    const message = `Failed to render Mermaid diagram: ${(error as Error).message}`;
+    fileError(file, message, {
+      ruleId: RuleId.imageFormatConverts,
+      node,
+      note: 'To install the Mermaid CLI, run `npm install -g @mermaid-js/mermaid-cli`',
```
stefanv, line 124: Is this the correct error here? We already throw above if CLI not found.

stefanv (commented):
This is neat, @rowanc1! :rocket: Also a super helpful example for me to learn from :pray:.

My comments are just that, feel free to disregard. Also curious to learn the answer to Franklin's question.

PR #2133, opened 2025-06-27
🧮 Add a `typst` option in math nodes

rowanc1:
This allows for fixing typst math without a release.

### DIFF
--- a/.changeset/breezy-rivers-heal.md
+++ b/.changeset/breezy-rivers-heal.md
@@ -1,0 +1,5 @@
+---
+"myst-parser": patch
+---
+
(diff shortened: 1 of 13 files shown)


fwkoch (approved):
This is great! Now if we encounter a new symbol or macro that `tex-to-typst` does not yet support, we can do a quick fix in the document by explicitly adding the typst math, then open an issue and address in `tex-to-typst` later. (Previously, if there was an unsupported symbol/macro, we had to fix it in `tex-to-typst`, release that, and consume the release in MyST just to get it working.)

Typst (top) and tex (bottom) pdf from same MyST source:
image

PR #2138, opened 2025-06-30
🚸 Take users directly to the GitHub editing interface when clicking "Edit this page"

### LINKED ISSUE #2137: Edit link currently takes the user to the source of the given page
## Description

Impacts: Main branch

Currently, the "edit this page" link takes the user to the source of the given page. E.g., from the homepage of the MyST guide, clicking "Edit this page" goes to: 

<https://github.com/jupyter-book/mystmd/blob/main/docs/index.md>

The user needs to then click an edit (📝) icon for a second time to actually edit!

## Proposed solution

I feel it should take the user directly to the code editor, which will guide the user through the process of creating a fork, branch, and PR:

<https://github.com/jupyter-book/mystmd/edit/main/docs/index.md>

This appears to be a trivial change, I'll open a PR very soon :)

## Additional information

Probably shouldn't have opened this as "bug" as the [documentation explains two paths to editing](https://mystmd.org/guide/website-navigation#use-the-edit-this-page-button) after clicking "edit this page".

However, I believe users of the webpage shouldn't have to read the docs of the webpage's build tool to know what to do next; they just want to edit! I think we should drop them straight into an editor.

I feel "view source" should be a separate feature from "Edit this page". Currently, on a given page, the GitHub icon links to the root of the GitHub repository.

mfisher87:
Resolves #2137 

Currently, "edit this page" takes the user to the the GitHub "view" interface, and the user needs to click an edit button a second time to enter the editing flow. 

The docs explain what to do from the "view" interface to edit, but the user doesn't know to read the MyST documentation, they just know they want to edit.

I really like the approach of having separate "view source" and "edit this page" links for each page. Perhaps also a "report an issue" link for users who don't feel comfortable making edits themselves!

My "Math macros" test failed locally, I'm assuming because I don't have something in my environment. I expect automated tests to pass, let's see ;)

### DIFF
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -205,6 +205,12 @@ The following table lists the available frontmatter fields, a brief description 
 * - `github`
   - a valid GitHub URL or `owner/reponame`
   - page can override project
+* - `edit_url`
+  - URL to edit the page source. If this value is unset but `github` is specified, MyST will attempt to compute the specific GitHub URL for the page. You may disable this behavior by explicitly setting `edit_url` to `null`.
+  - page can override project
+* - `source_url`
+  - URL to view the page source. If this value is unset but `github` is specified, MyST will attempt to compute the specific GitHub URL for the page. You may disable this behavior by explicitly setting `source_url` to `null`.
+  - page can override project
 * - `binder`
   - any valid URL
   - page can override project
@@ -261,12 +267,6 @@ The following table lists the available frontmatter fields, a brief description 
   - page only
 * - `skip_execution`
   - opt-out of execution for a particular document (see [](./execute-notebooks))
-  - page only
-* - `edit_url`
-  - URL to edit the page source. If this value is unset but `github` is specified, MyST will attempt to compute the specific GitHub URL for the page. You may disable this behavior by explicitly setting `edit_url` to `null`.
-  - page only
-* - `source_url`
-  - URL to view the page source. If this value is unset but `github` is specified, MyST will attempt to compute the specific GitHub URL for the page. You may disable this behavior by explicitly setting `source_url` to `null`.
   - page only
 ```
 
--- a/packages/myst-frontmatter/src/page/types.ts
+++ b/packages/myst-frontmatter/src/page/types.ts
@@ -14,8 +14,6 @@ export const PAGE_FRONTMATTER_KEYS = [
   'enumerator',
   'content_includes_title',
   'skip_execution',
-  'edit_url',
-  'source_url',
 ];
 
 export type PageFrontmatter = ProjectAndPageFrontmatter & {
@@ -33,7 +31,4 @@ export type PageFrontmatter = ProjectAndPageFrontmatter & {
   content_includes_title?: boolean;
   /** Site Options, for example for turning off the outline on a single page */
   site?: Record<string, any>;
-  // URLs to edit or view the current source
-  edit_url?: string | null;
-  source_url?: string | null;
 };
--- a/packages/myst-frontmatter/src/page/validators.ts
+++ b/packages/myst-frontmatter/src/page/validators.ts
@@ -72,16 +72,6 @@ export function validatePageFrontmatterKeys(value: Record<string, any>, opts: Va
     // At this point, they just need to be an object
     output.site = validateObject(value.site, incrementOptions('site', opts));
   }
-  if (value.edit_url === null) {
-    output.edit_url = null;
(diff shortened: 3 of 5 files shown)


choldgraf (approved):
I think this makes sense - we call it `edit this page` and so the verb of the button makes sense to do that. +1 from me, also this inspired this issue for enhancing edit buttons in general that shouldn't be too hard to implement!

- https://github.com/jupyter-book/mystmd/issues/2141

I'll not merge just yet because this is a change in behavior and might be surprising to users, and want to give a moment for others to object if they wish. but IMO this is still safe to try.

rowanc1:
@fwkoch can you take a quick look?

I think keeping the source URL somewhere would be good. This could also be done in the frontend by replacing the blob with edit or github.dev etc. or #2141.

packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -25,7 +25,7 @@ export async function addEditUrl(session: ISession, frontmatter: PageFrontmatter
     const getGitRoot = makeExecutable('git rev-parse --show-toplevel', gitLog);
     const gitRoot = (await getGitRoot()).trim();
     if (gitBranch && gitRoot && file.startsWith(gitRoot)) {
-      frontmatter.edit_url = `${frontmatter.github}/blob/${gitBranch}${file.replace(gitRoot, '')}`;
+      frontmatter.edit_url = `${frontmatter.github}/edit/${gitBranch}${file.replace(gitRoot, '')}`;
```
rowanc1, line 29: Perhaps we could add a new frontmatter field that is source_url?

```suggestion
      frontmatter.source_url = `${frontmatter.github}/blob/${gitBranch}${file.replace(gitRoot, '')}`;
      frontmatter.edit_url = `${frontmatter.github}/edit/${gitBranch}${file.replace(gitRoot, '')}`;
```

This allows us to have both URLs, and treat them differently in the future.

mfisher87: I like it. Maybe we should rename this module to `addGitHubUrls.ts`?

choldgraf:
Definitely agree there's value in keeping the source URL in the build system!

agoose77:
This _is_ a change in behaviour, but the feature is new and I think the name implies quite strongly what it should do! Echoing the other comments in this PR, like the direction of multiple URLs.

agoose77:
@fwkoch when fixing up some types, I noticed that `edit_url` is a page-and-project frontmatter entry. I can't see how it makes sense to have this at the project level, so I've included a commit to fix that (and update the docs). Can you see any issue with that? Right now, we only populate it for individual pages anyway. The only benefit is to explicitly set it to null at the project level, which disables the feature for all pages.

fwkoch:
Hey @agoose77 - the reason this is at the project level is exactly as you say in your last sentence. If we take `edit_url` out of project frontmatter, and someone wants to set `github` but does not ever want `edit_url` auto-populated, they will have to set it to null on every page, rather than once.

I'd be inclined to leave it as `project` frontmatter for this reason alone. I don't really see a downside - outside of the `null` case, someone might find it weird to set a single `edit_url` for the entire project. But then again, maybe there's some other use case where a single, say, google doc is the "source of truth" for the entire MyST project, and the authors want all pages to direct to that single doc for editing...? Slightly farfetched, but seems like a similar use-case could be possible.

agoose77:
> Hey @agoose77 - the reason this is at the project level is exactly as you say in your last sentence. If we take `edit_url` out of project frontmatter, and someone wants to set `github` but does not ever want `edit_url` auto-populated, they will have to set it to null on every page, rather than once.
> 
> I'd be inclined to leave it as `project` frontmatter for this reason alone. I don't really see a downside - outside of the `null` case, someone might find it weird to set a single `edit_url` for the entire project. But then again, maybe there's some other use case where a single, say, google doc is the "source of truth" for the entire MyST project, and the authors want all pages to direct to that single doc for editing...? Slightly farfetched, but seems like a similar use-case could be possible.

Thanks Franklin! Could you possibly revert the commit before the merge that makes this change? I am keen for us to merge this!

fwkoch (approved):
This looks good - we still have `source_url` and now `edit_url` points to a different, more sensible place.

I made the changes discussed in the PR comments to keep both these fields as `project` and `page` frontmatter. Even though the "source" is very much a page-specific thing, we can set either of these to `null` at the project level to disable them across all pages.

PR #2139, opened 2025-07-01
🎁 For JATS output, use <sup> to wrap parenthetical citation

### LINKED ISSUE #2136: Use numeric XML element `<sup>` group citation like Cell & Nature instead of parentheses to group citations
This enhancement request is for JATS XML exports that use the numeric `<sup>` JATS XML element like Cell & Nature to group citations instead of parentheses. The main reasoning behind this is to minimize how much citation style formatting is being set in the XML and only save information that serves a dual role of being both:

* presentation XML (for reader implementations that do nothing more than echo the numeric superscript style)
* and/or minimal semantic information (for reader implementations that can re-style citations).

This is related to issue [#1462](https://github.com/jupyter-book/mystmd/issues/1462#issuecomment-2957211347).

Currently, markdown such as:
```
There are examples [@jats_authoring; @ref-enwikiu003ajats]
```
results in a JATS XML export like
```
<p>There are examples (<xref ref-type="bibr" rid="ref-1">U.S. National Library of Medicine (NLM), 2024</xref>; <xref ref-type="bibr" rid="ref-2">contributors, 2023</xref>).</p>
```
So with this enhancement the JATS XML export would instead be:
```
<p>There are examples<sup><xref ref-type="bibr" rid="ref-1">1</xref>,<xref ref-type="bibr" rid="ref-2">2</xref></sup>.</p>
```

Both the existing implementation's use of parentheses and this enhancement's use of `<sup>` are used by JATS XML in the PubMed Central Open Access Subset. The numeric `<sup>` style matches the Citation Language Style of
```
<style xmlns="http://purl.org/net/xbiblio/csl" class="in-text" version="1.0">
...
  <citation>
    <sort>
      <key variable="citation-number"/>
    </sort>
    <layout vertical-align="sup" delimiter=",">
        <text variable="citation-number"/>
    </layout>
  </citation>
...
```
which are found in the `cell.csl` and `nature.csl` files of the <https://www.zotero.org/styles> and <https://github.com/citation-style-language/styles>.

Using the above CSL and style is supported by `pandoc` and will output the JATS XML in this enhancement request. This is how <https://gitlab.com/perm.pub/baseprinter> uses 'pandoc'.

This enhancement will also increase interoperability with the <https://gitlab.com/perm.pub/epijats> JATS XML parser and related tools. Currently, `epijats` will recognize `<sup><xref ...` as a citation group but is not as smart as npm `jats-xml` to parse the parenthetical style semantically.

It is too bad JATS XML does not have a citation group XML element. The defacto closest to one is a rather hacky `<sup>` in the style described in this enhancement.

stefanv:
Closes #2136

Note: I don't know *whether* #2136 should be handled in the way suggested in the issue, but this does what is requested.

### DIFF
--- a/.changeset/young-eggs-work.md
+++ b/.changeset/young-eggs-work.md
@@ -1,0 +1,5 @@
+---
+'myst-to-jats': minor
+---
+
+Wrap parenthetical citations in <sup> tag to group citations
--- a/packages/myst-to-jats/src/transforms/citations.ts
+++ b/packages/myst-to-jats/src/transforms/citations.ts
@@ -4,7 +4,7 @@ import { selectAll } from 'unist-util-select';
 import type { GenericNode, GenericParent } from 'myst-common';
 
 /**
- * Add parentheses and separator text to citeGroup nodes as children
+ * Add <sup> tag and separator text to citeGroup with nodes as children
  *
  * This allows us to simply render children and get all the correct formatting.
  */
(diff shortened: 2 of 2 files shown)


rowanc1:
This should likely be based on a flag (numbered_citations or something in the theme). We are mostly using this for non-numerical citations (APA etc) so those are **not** in a sup.

If this is an option, and not the default, then I am good with the direction!

PR #2140, opened 2025-07-01
🔗 Fix URLs in table of contents directive

### LINKED ISSUE #2119: URL entries show up as plain text in `toc` directive
#2102 added support for URLs in the toc. However, when rendering a toc with the `{toc}` directive, they show up as plain text, rather than a link.

I think this will be a pretty quick fix here: https://github.com/jupyter-book/mystmd/blob/main/packages/myst-transforms/src/toc.ts#L31 - pull out `url` and if that field is defined, create a `link` node, similar to this one https://github.com/jupyter-book/mystmd/blob/main/packages/myst-transforms/src/toc.ts#L39 - except `internal` will not be `true`.

brianhawthorne:
- closes #2119

### DIFF
--- a/packages/myst-transforms/src/toc.ts
+++ b/packages/myst-transforms/src/toc.ts
@@ -35,7 +35,7 @@ function listItemFromPages(pages: ProjectPage[], projectSlug?: string) {
     value: `${enumerator ? `${enumerator} ` : ''}${title}`,
   };
   let child;
-  if (!!url) {
+  if (url != null) {
     // Link to an external site if url is given
     child = {
       type: 'link',


packages/myst-transforms/package.json
```diff
@@ -13,6 +13,7 @@
     "clean": "rimraf dist",
     "test": "vitest run",
     "test:watch": "vitest watch",
+    "test:debug": "vitest --inspect-brk --pool threads --poolOptions.threads.singleThread --no-file-parallelism run",
```
brianhawthorne: Maybe want to take this out before merge (or add same to every separate package.json in the project).  
Or better to live as a debugging recipe in the developer guide?

packages/myst-transforms/src/toc.ts
```diff
@@ -11,8 +11,16 @@ type ProjectPage = {
   enumerator?: string;
 };
 
+function makeList(children?: ListItem[]): List {
```
stefanv: Given that we already have a `List` TypeScript type, is this necessary? E.g., could we do the following instead?

```
const lst : List = {children};
```

Not a TS expert, so asking, not saying!

brianhawthorne: Also still getting up to speed on TS, but I _think_ there would be two problems with this:

- The List type requires a non-optional `.type` property with constant value "list", and the code above would produce an object missing that property (I think!  Unless the TS compiler is super smart about adding non-optional object props with known constant value...).
- This would produce an object with `.children` property having the value of local variable `children`, which could be null or undefined; but the List type definition says `children` is required and must be an array of ListContents.

I'm consulting this reference for the definition of the List type :point_right: https://mystmd.org/spec/myst-schema#list

Still trying to get jump-to-definition working for these myst-spec types in vim... but they appear to be defined not in any primary TS source code, but in a .ts file _generated_ for distribution here: https://github.com/jupyter-book/myst-spec/blob/7b2e431d46ac11baba9d433b113449598815189d/index.ts#L280

If you don't want to generate this file yourself (clone myst-spec, npm install, npm run build), then you can find the definitions eg under your installed mystmd `node_modules/myst-spec/dist/index.d.ts`.  Search for "export type List" to get to the definition around line 2160.

Still, that source definition was not quite helpful to me, not being super TS expert yet, involving apparently nested intersection types (via [ampersand](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types)), so I'm not exactly sure how that def shakes out to be equivalent to the simpler definition described in the online type reference.

stefanv: I did a bit of reading now, and I think your factoryMethods are the correct way to go.

packages/myst-transforms/src/toc.ts
```diff
@@ -22,38 +30,52 @@ function listFromPages(pages: ProjectPage[], projectSlug?: string): List {
...
+  const page = pages[0];
+  const child = listItemChildFromPage(page, projectSlug);
+  const item = makeListItem(child);
+  if (pages[1] && pages[1].level > page.level) {
+    item.children.push(listFromPages(pages.slice(1), projectSlug));
+  }
+  return item;
+}
+
+function listItemChildFromPage(page: ProjectPage, projectSlug?: string): Text | Link {
```
stefanv: Confusing function name they used here, took me a while to figure out what's going on :upside_down_face:

packages/myst-transforms/src/toc.ts
```diff
@@ -22,38 +30,52 @@ function listFromPages(pages: ProjectPage[], projectSlug?: string): List {
...
 }
 
-function listItemFromPages(pages: ProjectPage[], projectSlug?: string) {
+function listItemFromPages(pages: ProjectPage[], projectSlug?: string): ListItem {
   if (pages.length === 0) return;
-  const { title, slug, enumerator, level } = pages[0];
+  const page = pages[0];
+  const child = listItemChildFromPage(page, projectSlug);
+  const item = makeListItem(child);
+  if (pages[1] && pages[1].level > page.level) {
```
stefanv: Looks like nesting is happening here; may be worth pointing out in a comment.

packages/myst-transforms/src/toc.ts
```diff
@@ -11,8 +11,16 @@ type ProjectPage = {
   enumerator?: string;
```
stefanv, line 12: This `ProjectPage` definition doesn't have `url` :thinking:

packages/myst-transforms/src/toc.ts
```diff
@@ -63,13 +85,24 @@ function listFromHeadings(headings: Heading[]): List {
...
+  const heading = headings[0];
+  const child = listItemChildFromHeading(heading);
+  const item = makeListItem(child);
+  if (headings[1] && headings[1].depth > heading.depth) {
+    item.children.push(listFromHeadings(headings.slice(1)));
+  }
+  return item;
+}
+
+function listItemChildFromHeading(heading: Heading): Text | Link {
```
stefanv: Do we have a way of documenting these? I think even a simple one sentence description would be quite helpful for future readers!

stefanv (commented):
OK, I took a bit more time to read this, and the changes LGTM!

I'm curious whether I just missed something about the `ProjectPage` type, whether we need to get that type definition from elsewhere, or whether we should extend the type definition here.

stefanv:
Note some linting errors too; perhaps try fixing with `npm run format` or `npm run lint:format`.

packages/myst-transforms/src/toc.ts
brianhawthorne: Yah, just realizing I have some type errors.  I had been testing with `npm run test` until I got it working, but didn't separately run `tsc`.  Since it was executing the updated TS source code to run the tests I assumed (silly me) that it was also type checking...

packages/myst-transforms/src/toc.ts
brianhawthorne: I added a small bit of documentation.  The mutually recursive function pairs here are turning me a little bit cross-eyed honestly.  Since I don't fully understand the logic, I'm afraid if I add too much by way of explaining, it might end up misleading.

I did note that the two mutually recursive function pairs (`listFromPages`/`listItemFromPages` and `listFromHeadings`/`listItemFromHeadings`) are largely identical modulo the details of the base-case transformation and the choice of `.level` vs `.depth` as the property that controls the recursion.  If they were to settle on either level or depth for both kinds of TOC trees, then the four could be combined into a single mutually recursive pair parametrized by a node transformer function.  I wasn't sure if the choice to use two different property names to represent what looks like the same thing in these two cases was a design decision or just incidental... maybe an opportunity to refactor, but not something I'd be inclined to take on just yet :grin: .

stefanv: Probably something @fwkoch would know!

rowanc1:
Will get this in shortly - @brianhawthorne thanks for the contribution!!

PR #2152, opened 2025-07-04
🎬 More whitelabelling for `myst init`

agoose77:
I noticed that running `jupyter book init --gh-pages` refers to myst. This PR fixes that under Jupyter Book 2, so that users install the proper package.

This still isn't perfect — the `--curvenote` option uses a Curvenote action, which theoretically doesn't know about Jupyter Book. But I think that's a problem for another day.

### DIFF
--- a/packages/myst-cli/src/init/gh-actions/index.ts
+++ b/packages/myst-cli/src/init/gh-actions/index.ts
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 18.x
-      - name: Install ${readableName()} Markdown
+      - name: Install ${readableName()}
         run: npm install -g ${npmPackageName()}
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/init/gh-actions/index.ts
```diff
@@ -57,10 +58,10 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 18.x
-      - name: Install MyST Markdown
-        run: npm install -g mystmd
+      - name: Install ${readableName()} Markdown
```
rowanc1: This would be `Jupyter Book (via myst) Markdown`

```suggestion
      - name: Install ${readableName()}
```

agoose77: Yes, much better!

PR #2159, opened 2025-07-06
Add `ipynb` format option in validators

kp992:
Fixes the validation error when used `ipynb` format option while building.

cc @agoose77 @mmcky


agoose77:
LGTM! For other readers, I'm merging this because it's being merged into a feature branch.

PR #2171, opened 2025-07-14
Support exercise and solution node in `myst-to-md`

kp992:
This PR tries to add exercise and solution node in `md` converter. The error looks like:
```
⛔️ exports/a.ipynb Unsupported node type: exercise
⛔️ exports/a.ipynb Unsupported node type: solution
```

cc @agoose77 @mmcky

### DIFF
--- a/packages/myst-to-md/package.json
+++ b/packages/myst-to-md/package.json
@@ -44,6 +44,7 @@
     "myst-frontmatter": "^1.7.6",
     "unist-util-select": "^4.0.3",
     "vfile": "^5.3.7",
-    "vfile-reporter": "^7.0.4"
+    "vfile-reporter": "^7.0.4",
+    "myst-ext-exercise": "^1.0.9"
   }
 }


kp992:
The diff fixes the error but the output is still empty. So I believe I might need to fix `directiveHandlers`, can you please review @agoose77?

Thank you.

kp992:
@agoose77 I have updated the PR with some tests. Can you please review and share your suggestions and improvements?

packages/myst-to-md/package.json
```diff
@@ -44,6 +44,7 @@
     "myst-frontmatter": "^1.7.6",
     "unist-util-select": "^4.0.3",
     "vfile": "^5.3.7",
-    "vfile-reporter": "^7.0.4"
+    "vfile-reporter": "^7.0.4",
+    "myst-ext-exercise": "^1.0.9"
```
rowanc1, line 48: PR looks good in general! I am trying to think of a way to not have the extension as a dependency?

kp992: Thanks for the review @rowanc1. Hmm, do you have any suggestions for that?

PR #2175, opened 2025-07-15
🏷️ Add NPM binary name to whitelabelling

agoose77:
This sets the scene to fix https://github.com/jupyter-book/jupyter-book/issues/2387, by adding an explicit way for JB to say "the binary for the NPM package XXX is called YYY".

The existing `binaryName` field remains apt to describe the actual current binary. 

In the longer term, we're probably going to do something more simple with whitelabelling -- this was _an_ approach at the time that may no longer be the most sensiblem

### DIFF
--- a/packages/myst-cli/src/init/gh-actions/index.ts
+++ b/packages/myst-cli/src/init/gh-actions/index.ts
@@ -5,8 +5,8 @@ import chalk from 'chalk';
 import type { ISession } from 'myst-cli-utils';
 import { writeFileToFolder } from 'myst-cli-utils';
 import { getGithubUrl } from '../../utils/github.js';
-import { checkFolderIsGit, checkAtGitRoot } from '../../utils/git.js';
 import { npmBinaryName, readableName, npmPackageName } from '../../utils/whiteLabelling.js';
+import { checkFolderIsGit, checkAtGitRoot } from '../../utils/git.js';
 
 function createGithubPagesAction({
   defaultBranch = 'main',


stefanv:
Could you explain a bit more what this PR does?

Looks like you allow setting the path to `myst`, for if you want to call it something different in JupyterBook.

I find the `npmBinaryName` rather confusing: to me, that would be "we are replacing `npm` with `yarn`". Perhaps `mystBinaryAlias` or something?

agoose77:
We have a whitelabelling mechanism in myst to allow JB to reuse myst code directly. This is was implemented before we knew exactly how JB and myst would share code, etc. There was a thought that we might actually reuse the MyST binary and simply set whitelabelling.

As it stands I'm less certain of that now, but nothing has changed.

The existing labelling allows JB to say "I'm called Jupyter book, and my binary is invoked via 'jupyter book'" . It also let's us say "my npm package is called jupyter-book", which was added for the purpose of the gh pages template. We didn't also add a way to say "my npm binary name is different, it's jupyter-book".

This PR adds that last part.

stefanv:
OK, cool, that gels with what I derived from your PR. I think `mystBinaryAlias` or `mystBinaryName` or something similar would be more intuitive, given that this change is happening *inside of myst*. Unless there's a special concept here "npm binary" that is different from simply "executable name", which is what I'm guessing it is.

agoose77:
The idea is that the binary name for JB is different on pypi to npm, so keeping that in the name is important - the other binaryName field represents the intended name on the host system, e.g 'jupyter book' via pypi.

stefanv:
I'm trying to decode: "the binary name for JB is different on pypi to npm".

Do you:

(a) Rename mystmd to something else,
(b) rebuild mystmd as something else, or
(c) call mystmd from `jupyter book`?

From wanting to name the npm package / binary, it sounds like (b)?

agoose77:
The idea is (b) that the binary for Jupyter Book _is_ (mostly) the same binary as mystmd (as of writing).

When JB is run, e.g `jupyter book init --gh-pages`, it runs the MyST routine to create the workflow. MyST consults the whitelabelling logic to learn that it's actually called jupyter book, etc. because Jupyter Book has set these environment variables in its own entrypoint.

rowanc1:
Was a bit confused at the scope of this PR as well and why it is necessary.

We do not install jupyter by default on the github runners, so `jupyter` is not a command that is available but `jupyter-book` is. This means that if we install the npm package, we use the npm binary.

agoose77:
I assume that you're not still confused given that you've merged it. I'll add further context -- `jupyter book` invokes `jupyter-book` under the hood (it's a `jupyter` binary feature). That binary name is made available by the PyPI package. Meanwhile, the NPM package installs a `jupyter-book` binary (so that we don't have to create our own "fake" jupyter binary to dispatch to `jupyter-book`, which would clash with the real `jupyter` binary). This PR ensures that Jupyter Book can tell the MyST code which binary the NPM package installs, which is `mystmd` for MyST and `jupyter-book` for JB2.

rowanc1:
Yes, figured it out. The context of this failing in github actions was the main thing I needed to understand why this was necessary.

PR #2188, opened 2025-07-22
Automate GitHub Release Creation with Reusable Workflow

choldgraf:
This adds a reusable workflow for publishing GitHub Releases with automatically generated release notes, it runs each time the `release.yml` action completes. It means we won't have to manually create GitHub releases, though we can always edit the release notes ourselves if we want to clean them up.

It defines this in a new action so that we can call it via a workflow_dispatch if we want to - it will let us insert the tag to use as the "start date" for releases.

So the new workflow for releases would be to just merge the changesets PR, wait for the workflow to complete, and a new release should now exist on GitHub.

### DIFF
--- a/.github/workflows/publish-github-release.yml
+++ b/.github/workflows/publish-github-release.yml
@@ -1,0 +1,76 @@
+name: Publish GitHub Release
+
+on:
+  workflow_call:
+    inputs:
+      tag_name:
+        required: true
+        type: string
+      release_title:
+        required: false
+        type: string
+        default: ''
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Tag name for the release (e.g. v1.2.3)'
+        required: true
+        type: string
+      release_title:
+        description: 'Optional release title (will be ignored, see workflow logic)'
+        required: false
+        type: string
+        default: ''
+
+jobs:
+  publish-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      # Get the date of the latest GitHub release
+      - name: Get latest release date
+        id: latest_release
+        run: |
+          latest_release_json=$(gh release list --limit 1 --json publishedAt --jq '.[0]')
+          if [ -z "$latest_release_json" ]; then
+            echo "No previous GitHub release found. Cannot generate release notes."
+            exit 1
+          else
+            date=$(echo "$latest_release_json" | jq -r '.publishedAt')
+            echo "date=$date" >> $GITHUB_OUTPUT
+          fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - name: Install github-activity
+        run: pip install github-activity
+
+      # Generate release notes since the last release
+      - name: Generate release notes with github-activity
+        run: |
+          github-activity --since "${{ steps.latest_release.outputs.date }}" --repo "$GITHUB_REPOSITORY" > release-notes.md
+
+      # Strip leading 'v' from tag for the release title
+      - name: Prepare release title
+        id: version
+        run: |
+          TAG="${{ github.event.inputs.tag_name || inputs.tag_name }}"
+          VERSION=${TAG#v}
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+      # Create the GitHub Release
+      - name: Create GitHub Release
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,75 @@
...
+        required: false
+        type: string
+        default: ''
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Tag name for the release (e.g. v1.2.3)'
+        required: true
+        type: string
+      release_title:
```
stefanv: Why is this here?

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  publish-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      # Get the date of the latest GitHub release
+      - name: Get latest release date
```
stefanv:
```suggestion
      - name: Get previous release date
```

Or something to suggest that we're not talking about the current release; it confused me, trying to figure out how you can already read the release date before it's been created!

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          python-version: '3.x'
+
+      - name: Install github-activity
+        run: pip install github-activity
+
+      # Generate release notes since the last release
+      - name: Generate release notes with github-activity
+        run: |
+          github-activity --since "${{ steps.latest_release.outputs.date }}" --repo "$GITHUB_REPOSITORY" > release-notes.md
```
stefanv: Do we need to worry about overlap by one? If so, shouldn't we grab the changes by hash instead of date?

.github/workflows/publish-github-release.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      # Strip leading 'v' from tag for the release title
+      - name: Prepare release title
+        id: version
+        run: |
+          TAG="${{ github.event.inputs.tag_name || inputs.tag_name }}"
```
stefanv: This is to handle the dispatch vs workflow?

stefanv: Could just export tag name along with version, to make life a bit simpler below.

stefanv (approved):
Looks good, thanks @choldgraf!

Left a few comments for you to consider or not :)

.github/workflows/publish-github-release.yml
choldgraf: I can't remember how exactly `github-activity` works, but if I recall, it converts hashes to dates anyway? Is this something we could either:

- Fix in `github-activity` directly (e.g. making sure that its end date is non-inclusive or something)
- Clean up by having a human look at the release notes after this action runs?

.github/workflows/publish-github-release.yml
choldgraf: yeah I am still not entirely confident in how to use variables and workflow_dispatches etc, but yes this is the goal haha

.github/workflows/publish-github-release.yml
choldgraf: This was from a previous iteration - let's just remove it, I got a little ahead of myself trying to make this configurable via workflow_dispatch but feel like for now we should just get the minimal functionality in

.github/workflows/publish-github-release.yml
choldgraf: yeah I agree that'd be simpler, if I recall the versions were a little bit confusing `mystmd@` rather than `mystmd v` but this is probably unnecessarily complex

choldgraf:
I cleaned it up so that we use the tag for the release title now - that's a bit simpler.

Things that could be better but IMO we could clean up in follow-ups if we decide we like this workflow:

- Make sure the dates don't do overlaps
- Improve bot detection so that we don't include bots in the changelog notes
- Improve the UX for defining dates/tags/etc (maybe w/ workflow dispatch)

stefanv:
> Make sure the dates don't do overlaps

It seems like you will almost certainly include any merges from the date of release N-1 this way. Easiest may be to start the next day, but then of course you could miss some.

choldgraf:
ok the latest commit uses the `tagName` of the latest release, rather than the `date`. For now this means we'll accidentally include things, but in the future we can fix this by resolving this issue:

- https://github.com/executablebooks/github-activity/issues/79

stefanv (approved):

choldgraf:
ok I'm gonna try this out and lets see if it works haha. I think this is safe to merge because it only adds a CI/CD workflow that is only triggered when we make a release (or workflow dispatch)

PR #2194, opened 2025-07-23
🔧 Ensure hidden pages are built for HTML exports

fwkoch:
@parmentelat brought up an issue where `hidden` pages do not exist in static HTML builds. This is a quick pass to try and fix - just for demonstration purposes for now.

- Related to https://github.com/jupyter-book/mystmd/issues/2225


parmentelat:
thanks @fwkoch for taking care of this !
I can see the actions include a deploy-preview step; 
once this passes, is there a way to try things out from another repo's actions ?

right now my action looks like this:
https://github.com/ue12-p25/numerique/blob/main/.github/workflows/myst-to-pages.yml#L47

fwkoch:
Hey @parmentelat - Ah, the `deploy-preview` step in CI is just a docs preview, not a code release preview. Still, I think what you want to do is possible.

First, I'd just test things locally. Run `myst build --execute --html` and make sure the hidden pages do not show up in your `_build/html` folder. Then, check out this branch, run `npm install && npm run build` then try `myst build --execute --html` again, and see if the hidden pages now do show up in `_build/html`.

Then, if you want to test deploying to github pages via your action, you can install by cloning the repo rather than via npm. It means in your action instead of this:
```yaml
- name: Install MyST Markdown
  run: npm install -g mystmd
```
you will have something more like
```yaml
- name: Checkout MyST repo
  uses: actions/checkout@v4
  with:
    repository: jupyter-book/mystmd
    ref: feat/prototype-hidden-in-html
    path: mystmd
- name: Install MyST
  working-directory: ./mystmd
  run: npm install && npm run build
```

Does that make sense...?

parmentelat:
hey @fwkoch 
thanks for the feedback

so let me clarify one point: I would really prefer being able to check things locally; 
however the reason I was planning on checking the GH-published version is that I am not exactly confident with local tests
(first time around, precisely because I was using the local build, I had wrongly concluded the hidden pages were exported with 1.5.0, while obviously they're not)

this being said, thanks to your instructions I made the test locally again, and indeed I can see the hidden pages show up under `_build/html`

I will try to setup the gh-hosted side too when I have a chance, just to be even more certain it works as expected

parmentelat:
ok so this angle does not pan out, I am getting this build error
https://github.com/ue12-p25/numerique/actions/runs/16530197539/job/46753445878

with `myst: command not found`

even though it's been installed from source as per your suggestion

fwkoch:
@parmentelat - Ok, I had a chance to actually test this. It should work if you replace the `myst` command with the actual path to the executable. In your case, I think this should be `./mystmd/node_modules/mystmd/dist/myst.cjs`.

Here's the relevant lines in my little repo where I got this working: https://github.com/fwkoch/test-action-myst-install/blob/main/.github/workflows/install-myst.yml#L29

parmentelat:
@fwkoch 
thanks for the tip, this works now for my sample repo on github:
hidden pages do get built, are excluded from the TOC, and can now also be retrieved if one knows their URL

thanks a lot ! let me know if that change has a chance to make it to the vanilla myst ?

parmentelat:
I am just seeing something odd though
this is possibly totally unrelated to the PR, but ..

on the gh pages published site, the left-hand-side toc behaves oddly

when a specific page is opened, e.g.
https://ue12-p25.github.io/numerique/numpy-introduction-nb/
then all the sections in the TOC menu are closed 
but this page is the first notebook in the first section so I would expect to see the first section "opened"

what I'm saying is I would expect this - that's what my local build gives me
image

and on the gh site I see this instead
image

again this is likely unrelated to the current PR but I have no way to convince me of that

stefanv:
This (menu not unfolding) is a known bug with the static builds currently.

parmentelat:
Hey, thanks for confirming that :)

fwkoch:
Mostly for the sake of cross-references in github, here's a link to the menu bug: https://github.com/jupyter-book/myst-theme/issues/635

@parmentelat - as for getting this into MyST for real, yes I think that should be fine! Thanks for testing it with your content - I'll tidy up the PR.

packages/myst-cli/src/build/html/index.ts
```diff
@@ -3,32 +3,37 @@ import path from 'node:path';
...
+      if (!localProj) return [];
       const projSlug = proj.slug ? `/${proj.slug}` : '';
       // We need to get the index from a slug page to make remix happy
       // If this gets from the index, then the site will trigger the wrong render path
       // And then hydration does not match
-      const siteIndex = baseurl ? `/${proj.index}` : '';
-      const pages = proj.pages.filter((page) => !!page.slug);
+      const siteIndex = baseurl ? `/${localProj.index}` : '';
+      const pages = localProj.pages.filter(
+        (page): page is LocalProjectPage => !!(page as any).slug,
```
stefanv, line 35: Why is the as any cast necessary here?

packages/myst-cli/src/build/html/index.ts
```diff
@@ -3,32 +3,37 @@ import path from 'node:path';
...
 ): Promise<{ url: string; path: string; binary?: boolean }[]> {
-  const manifest = await getSiteManifest(session, opts);
-  return (manifest.projects ?? [])
+  const state = session.store.getState();
+  const siteConfig = selectors.selectCurrentSiteConfig(state);
+  return (siteConfig?.projects ?? [])
     ?.map((proj) => {
+      if (!proj.path) return [];
+      const localProj = selectors.selectLocalProject(state, proj.path);
+      if (!localProj) return [];
```
stefanv, line 28: Is this the crux of the change? If so, a comment would be helpful. If not, where is the logic, since the word "hidden" doesn't occur anywhere.

parmentelat: this PR actually is a follow-up on top of 
#2097 
which it self was a follow-up of 
#1993

parmentelat:
I just realize this has not made it into 1.6.1
is there anyting I can do to get it merged ?

stefanv:
I think Chris last asked for a test. There's also one unresolved inline comment. But overall LGTM!

mfisher87:
>If you don't have time for that IMO we shouldn't block it because I consider this a bug

Would it be safe to assume that @fwkoch doesn't have the time for this task? :pray: Sorry, I'm just eager to see this released ;)

fwkoch:
Sorry - have the tab open to get this merged tonight!

packages/myst-cli/src/build/html/index.ts
fwkoch: @stefanv - the crux of the change in this PR is that we are iterating over pages in the `siteConfig` projects instead of the pages in `manifest` projects. The `siteConfig` has all pages present, the `manifest` is already filtered to remove `hidden`.

packages/myst-cli/src/build/html/index.ts
fwkoch: The `localProject.pages` can be `LocalProjectPage | LocalProjectFolder | ExternalURL` (not all of which have `slug` - see https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/project/types.ts#L42) whereas `manifestProject.pages` is only `ManifestProjectItem` (see https://github.com/jupyter-book/mystmd/blob/main/packages/myst-config/src/site/types.ts#L65).

mfisher87:
:raised_hands: Thanks everyone! Excited to update the GeoJupyter team compass with this change :)

PR #2216, opened 2025-07-30
🐦 Change social link from Twitter to BlueSky

agoose77:
Caught by @bsipocz on Discord: https://discord.com/channels/1083088970059096114/1126886232060870678/1399562247600214057

This is the smallest possible change to fix this _now_ -- I don't have the cycles to dig deeper and parametrise.

Is this a workable change for Curvenote (see the change to the CN init) @rowanc1?

### DIFF
--- a/.changeset/lazy-eyes-remember.md
+++ b/.changeset/lazy-eyes-remember.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


rowanc1:
Yeah, we don't use this upgrade message on our side! Thanks for checking.

Long overdue change.

Issue #1687, opened 2024-11-30
Absolute file paths in links, images, are not resolved relative to the project root


agoose77:
[MEP 002](https://mep.mystmd.org/mep-0002) defines the behaviour of absolute links (and implicitly URLs to other resources) as to be resolved relative to the project root, i.e. if `/content/index.md` contains
```markdown
[](/content/bar.md)
```
the link should resolve to `./bar.md`.

c.f. https://discord.com/channels/1083088970059096114/1177359316886503464/1312159994170310709

agoose77:
CC @jnywong

fperez:
Quick check on whether there's any progress on this? For now I'm manually fixing a bunch of links in our GeoJupyter source files to be relative, but it would be much better to have this correctly resolve. For static assets that live at the top (e.g. `/assets/images/foo.jpg`), and are routinely used from everywhere in a site, it's a pain to have to track how many `../` you have to prepend all the time, and end up with paths looking like `../../../../assets/...`.

I wish I could help with the code right now :)

mforbes:
Do I understand correctly that PR #2217 works for `![Image](/abs/path.svg)` references, but not `:::{image} /abs/path.svg` references, and this issue would make the latter work?  (This is what I seem to be experiencing.)

My current workaround is to symlink my top-level `images` folder as needed, then refer to everything locally.  Then I can refer to images like

```markdown
![Image](images/path.*)

:::{image} images/path.*
...
```
These seem to resolve properly through the symlink, and the first resolves to the top-level `images` folder.  Keeping a bunch of symlinks seems better than managing `../../../../` references and maybe this will just work after this is fixed (or maybe I need to replace `images/` -> `/images/` once later while removing the symlinks.

PR #2217, opened 2025-07-30
🔗 Resolve links and image paths relative to project root if they starts with `/`

fk128:
If the image path starts with `/` then it uses as root path not the file path but the project source path , e.g.

Files:
```
project-dir/content/subdir/index.md
project-dir/media/photo1.png
```

and `project-dir/content/subdir/index.md` contains

```markdown 
![](/media/photo1.png)

```

then it should resolve it as `project-dir/media/photo1.png` instead of `project-dir/content/subdir/media/photo1.png`

relates to https://github.com/jupyter-book/mystmd/issues/1687

### DIFF
--- a/packages/myst-cli/src/transforms/links.spec.ts
+++ b/packages/myst-cli/src/transforms/links.spec.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it, beforeEach, vi } from 'vitest';
 import memfs from 'memfs';
-import { fileFromRelativePath, checkLink } from './links';
+import path from 'path';
+import { fileFromRelativePath, fileFromSourceFolder, getSourceFolder, checkLink } from './links';
 import type { ISession } from '../session/types.js';
 
 vi.mock('fs', () => ({ ['default']: memfs.fs }));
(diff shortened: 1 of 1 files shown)


agoose77:
I'm just passing by -- we should extend the scope of this PR to also handle absolute paths to files, too.

fk128:
> I'm just passing by -- we should extend the scope of this PR to also handle absolute paths to files, too.

Would that just be in the include directive, or are there other places, too?

agoose77:
> > I'm just passing by -- we should extend the scope of this PR to also handle absolute paths to files, too.
> 
> Would that just be in the include directive, or are there other places, too?

This would be the include directive _and_ things like links to files, e.g. `[](/foo/bar)`. That might be a heavier lift than you feel comfortable tackling alone, so don't worry if you leave it -- we can fold this in to a bigger unit of work!

fk128:
> This would be the include directive _and_ things like links to files, e.g. `[](/foo/bar)`. That might be a heavier lift than you feel comfortable tackling alone, so don't worry if you leave it -- we can fold this in to a bigger unit of work!

I'll try to have a go

mfisher87:
Sooooo excited for this :star_struck:

rowanc1:
Thanks @fk128 for your work on this. Took a quick look, and it is looking solid. Let us know when you think it is good to go and I or others can do a bit more in depth testing.

fk128:
@rowanc1 I think it's good to go for now, but I would appreciate someone having a look to provide further input, as I don't know if I've missed something.

packages/myst-cli/src/transforms/links.ts
```diff
@@ -82,27 +82,38 @@ type LinkInfo = {
...
+export function getSourceFolder(link: string, sourceFile: string, projectFolder: string): string {
+  return link.startsWith('/') ? projectFolder : path.dirname(sourceFile);
+}
 /**
  * Compute link node file path relative to site root (currently, only the working directory)
  *
+ * @param pathFromLink Path from link node URL relative to file where it is defined
+ * @param file File where link is defined
+ */
+export function fileFromRelativePath(pathFromLink: string, file?: string): string | undefined {
```
fwkoch, line 102: I don't think this function is used anywhere in the codebase anymore - just kept in case any external libraries are dependent on it.

Could mark as `@deprecated`?

packages/myst-cli/src/transforms/links.ts
```diff
@@ -82,27 +82,38 @@ type LinkInfo = {
 
 export type LinkLookup = Record<string, LinkInfo>;
 
+export function getSourceFolder(link: string, sourceFile: string, projectFolder: string): string {
+  return link.startsWith('/') ? projectFolder : path.dirname(sourceFile);
```
fwkoch: Since we are using `path` utilities in these functions, should this use `path.sep` instead of `'/'`?

```suggestion
  return link.startsWith(path.sep) ? projectFolder : path.dirname(sourceFile);
```

Since most of our development / testing is on unix, I don't have a great feel for Windows support, but this small change makes sense to me.

fwkoch: Hmm, also looking at https://mep.mystmd.org/mep-0002#paths - the path separator is documented to be `'/'` always. However, this is definitely not the case with the current implementation. We will need to take a significant pass, replacing `path.*` with `path.posix.*` where appropriate...

For now, I still think consistency with the codebase is best.

fwkoch (commented):
Hey @fk128 - finally got some time to look into this. Although the change is small, it has some fundamental implications around our treatment of absolute paths.

For `images`, absolute paths simply didn't work before. (We just did a `join` on the folder and the image path, no `resolve`.) Now, "absolute" paths resolve relative to the project root. 👍

However, for `include` and `embed`, absolute paths were supported as system-level absolute paths. The changes in the PR modify that behavior in a backwards-incompatible way: absolute paths now resolve to project-level, not system-level.

Now, it turns out, system-level absolute paths (i.e. the current implementation) _do not_ comply with the current MyST spec, which states that absolute paths are from project root: https://mep.mystmd.org/mep-0002#paths - there's an open issue about this inconsistency here: https://github.com/jupyter-book/mystmd/issues/1687 - Therefore, I think this "backwards-incompatible" change is actually a bug fix, and anyone using system-level absolute paths will need to update their content. (I doubt very many people use these absolute paths, though. Most workflows involve building a project in multiple contexts (locally, github actions, etc), so actual absolute paths are useless. We _could_ continue to support system-level paths for now, and warn users if these are encountered... but my inclination is that is not worth the effort...)

Additionally, this PR does not cover all file paths. Particularly in `myst.yml` absolute file paths still resolve to system-level. (Now, this seems like less of a problem: since `myst.yml` is always at the project root, you can just always use relative paths...)

Anyway, aside from my two small comments (which I'm happy to make the changes for), I think we should get this in! It's just important to realize (1) some users could see this as a bad change that breaks their projects, and (2) this does not fully address all the path issues around POSIX separators and project-root absolute paths.

packages/myst-cli/src/transforms/links.ts
fk128: Hey @fwkoch, 

thanks for getting the time to look into this. 

Feel free to make the changes. Otherwise, I'll have time this weekend. I think `fileFromRelativePath` is still used in a test file. Probably, for completeness, we need to also add a few test cases for the new function.

packages/myst-cli/src/transforms/links.ts
```diff
@@ -82,27 +82,38 @@ type LinkInfo = {
...
+}
 /**
  * Compute link node file path relative to site root (currently, only the working directory)
  *
+ * @param pathFromLink Path from link node URL relative to file where it is defined
+ * @param file File where link is defined
```
fwkoch, line 99:
```suggestion
 * @param file File where link is defined
 * @deprecated Use getSourceFolder and fileFromSourceFolder to correctly resolve paths
```

fwkoch:
Good to hear from you @fk128 - I committed my suggested changes (and a `changeset` file).

Any additional test cases would be great, if you have the time over the weekend! Regardless, I will plan to merge this Monday.

fk128:
That's great, @fwkoch . I rebased the branch off main and added a few test cases.

fwkoch:
Amazing, thank you @fk128 - will merge after tests pass on CI!

Issue #1817, opened 2025-01-25
Support a custom path for `myst.yml` file


fperez:
This is somewhat related to #1816, and the use case is basically the same: easy batch handling of md/ipynb conversion where users don't modify their source files.

Having a custom `myst.yml` file will make it easier to build tools (possibly thin wrappers for myst) that produce templated `myst.yml` files in known, possibly temporary locations and can deliver the generated output to the user.

In our case at Berkeley (described in #1816) we could for example wrap that into small server-side tools triggered by a user menu or button in JupyterLab.

## Proposal

The CLI should allow for a custom path to the myst.yml file. I suggest `--config=` for the option name (with `-c` as short option).

My take is that if this which if given, would override the search and only work if the given location is valid. This will allow also for overriding the local yaml file when desired, without disturbing the project (useful in cases where one is experimenting with more than one yaml file for various output scenarios).

## Additional notes

[Original discussion on discord](https://discord.com/channels/1083088970059096114/1126886232060870678/1332030370639777887).

yanlisa:
+1 to this feature request—a lot of GitHub pages repos host from a subfolder; including the current jupyter-book [`next` branch](https://github.com/jupyter-book/jupyter-book/blob/next/docs/myst.yml).

### Current workaround

After making the subfolder (e.g. `docs`), test local build using `jupyter book start` within subfolder. For web deployment, manually edit GitHub Actions deployment to include paths to `docs` subfolder.

### Additional Notes

[Additional discussion on discord](https://discord.com/channels/1083088970059096114/1126886232060870678/1393271381109641386)

brianhawthorne:
:hand: I'll take a look at this one.

fperez:
Huge thanks @brianhawthorne @fwkoch for tackling this!! 🎉

PR #2219, opened 2025-08-02
:camel: Enable alternate YAML build config

brianhawthorne:
Add `--config` option to the `myst build` command to override the default `myst.yml` config file.

Closes https://github.com/jupyter-book/mystmd/issues/1817

### DIFF
--- a/packages/myst-cli/src/build/build.ts
+++ b/packages/myst-cli/src/build/build.ts
@@ -32,6 +32,7 @@ type FormatBuildOpts = {
   all?: boolean;
   force?: boolean;
   output?: string;
+  config?: string;
 };
 
 export type BuildOpts = FormatBuildOpts & CollectionOptions & RunExportOptions & StartOptions;
@@ -220,6 +221,12 @@ function extToKind(ext: string): string {
 
 export async function build(session: ISession, files: string[], opts: BuildOpts) {
   const { site, all, watch, writeDOIBib } = opts;
+
+  // Override default myst.yml if --config option is given.
+  if (opts.config !== undefined) {
+    session.configFiles[0] = opts.config;
+  }
+
   const performSiteBuild = all || (files.length === 0 && exportSite(session, opts)) || writeDOIBib;
   const exportOptionsList = await collectAllBuildExportOptions(session, files, opts);
   // TODO: generalize and pull this out!
--- a/packages/myst-cli/src/config.spec.ts
+++ b/packages/myst-cli/src/config.spec.ts
@@ -1,9 +1,4 @@
-import { describe, expect, it, beforeEach, vi } from 'vitest';
-import memfs from 'memfs';
-import { Session } from '../session';
-import { projectFromPath } from './fromPath';
-import { pagesFromSphinxTOC, projectFromSphinxTOC } from './fromTOC';
-import { tocFromProject } from './toTOC';
+import { describe, expect, it } from 'vitest';
 import { handleDeprecatedFields } from './config';
 import { VFile } from 'vfile';
 
--- a/packages/myst-cli/src/cli/build.ts
+++ b/packages/myst-cli/src/cli/build.ts
@@ -20,6 +20,7 @@ import {
   makeMaxSizeWebpOption,
   makeDOIBibOption,
   makeCffOption,
+  makeConfigOption,
 } from './options.js';
 import { readableName } from '../utils/whiteLabelling.js';
 
@@ -50,6 +51,7 @@ export function makeBuildCommand() {
     .addOption(makeCheckLinksOption())
     .addOption(makeStrictOption())
     .addOption(makeCIOption())
-    .addOption(makeMaxSizeWebpOption());
+    .addOption(makeMaxSizeWebpOption())
+    .addOption(makeConfigOption());
   return command;
 }
--- a/packages/myst-cli/src/cli/options.ts
+++ b/packages/myst-cli/src/cli/options.ts
@@ -175,3 +175,11 @@ export function makeTemplatesOption() {
     'Delete the _build/templates folder where downloaded templates are saved',
   ).default(false);
 }
+
+export function makeConfigOption() {
+  return new Option(
+    '--config <config-file>',
+    'Use this YAML config file, instead of the default myst.yml',
+  ).default(undefined);
+}
+


stefanv:
This LGTM, Brian! I am not at a computer right now, but wanted to check: is this generic to all the myst commands, or only build?

brianhawthorne:
> This LGTM, Brian! I am not at a computer right now, but wanted to check: is this generic to all the myst commands, or only build?

Thanks for taking a look Stefan!  Still working on this PR, and haven't tested the change yet, so it's not quite ready for review.  But as written now, it's only an option to the `build` command.

Whether it should be a global option instead was my first question.  I wasn't sure if the `myst.yml` file pertained to any of the other commands?

Do you think it makes more sense as a global option?  (the implementation would be simpler as the session loads before it gets into sub-command code, so we could skip that reload) 

EDIT:  Just realized this comment would have made more sense if I'd remembered to commit the reload... :dizzy_face: https://github.com/jupyter-book/mystmd/pull/2219/commits/dd07157c63662b8a594926a5d9ddb31c997839f6

@rowanc1 , maybe you have some thoughts on whether this new option should most naturally belong to the `build` sub-command or be a global option to the main `myst` command?  After some time matching execution paths of the sub-commands against spots where it looks like the config comes into play, I _think_ none of the others use it besides `build`, but it's a big codebase...

EDIT:  Looks like the `start` sub-command also does a build.  Since at least two sub-commands need to load the config, it should properly be a global option, so that's this question answered.  :white_check_mark:

brianhawthorne:
A second question/comment:  As is, it won't work given a full file path, only a naked file name (basename vs dirname) or local path relative to the project directory; but as an end-user presented with this config option, I'd probably expect it to also work given a full path to a config file unrelated to the project dir.

Looks like that would be a deeper cut though since there are a lot of spots where this value gets joined onto the project directory to produce a full path to the file.  So I thought this would be a sufficient first pass to minimally satisfy the use-case.

brianhawthorne:
A third and more general question (maybe for the original coders) is why use Redux for state management, in particular for the session configuration, in a single process server-side command line application?

For example, when loading the Session from config file, a bunch of stuff gets stashed in the redux store, necessitating a reload if anything changes.  Would it not be simpler for those values to just live directly as properties on the session object?  Maybe Redux adds leverage in other aspects of the program I haven't seen yet?

packages/myst-cli/src/config.spec.ts
```diff
@@ -1,9 +1,4 @@
-import { describe, expect, it, beforeEach, vi } from 'vitest';
```
rowanc1, line 1: Thanks for cleaning these unused imports up.

packages/myst-cli/src/build/build.ts
```diff
@@ -220,6 +221,12 @@ function extToKind(ext: string): string {
 
 export async function build(session: ISession, files: string[], opts: BuildOpts) {
   const { site, all, watch, writeDOIBib } = opts;
+
+  // Override default myst.yml if --config option is given.
+  if (opts.config !== undefined) {
+    session.configFiles[0] = opts.config;
```
rowanc1: Possibly put this at the top of the list, which means there is a fallback?

```suggestion
    session.configFiles = [opts.config, ...session.configFiles];
```

or this, if we don't want fallbacks:

```suggestion
    session.configFiles = [opts.config];
```

Creating a new list also means that this (for sure) won't effect other sessions. (I think that list is probably shared with all the instances?)

I think another thing that might be necessary to do is to add this to the `session.clone` as well. In a build that has multiple PDFs, we clone the session and run them in parallel, so this would need to be added as well to that side.

rowanc1 (commented):
This is looking good, haven't tested yet.

Is there something we need to do to validate the config (e.g. doesn't have `/` or `\` in it, ends in yml)? I haven't tested, so I don't know if this is giving a general config name, or a specific file?

rowanc1:
@brianhawthorne regarding the use of redux. This area of the code is pretty old, and was originally built in a web-app that was using redux/react, not a CLI (that was in 2021-ish). This area is certainly in need of an overhaul, but it would be a bit of a rewrite, without many guardrails in testing, etc. — we haven't done that yet..!

packages/myst-cli/src/build/build.ts
brianhawthorne: Thanks for the suggestions @rowanc1, and for cluing me in on multi-session builds !  This is exactly the sort of feedback I hope to get from someone with a broader context on the codebase.

Regarding fallbacks -- most places where I found direct references to `session.configFiles` only access the first item in that list, so I was curious why it is a list at all.  I wasn't sure if supporting a stack of increasingly specific configs, eg with latters overriding formers where they overlap, was a planned future feature or if it being a list was just a remnant of an earlier iteration of the tool?

brianhawthorne:
> Is there something we need to do to validate the config (e.g. doesn't have `/` or `\` in it, ends in yml)? I haven't tested, so I don't know if this is giving a general config name, or a specific file?

This expects a naked file name or file path relative to project directory, not a full file path.  It was the easiest to implement this way since the default configuration is "myst.yml".  There are a bunch of places where that value gets joined onto the project path and it would have been a mess to try to change all that to optionally support full path config.

Regarding validation, the config loading code already borks with an error message if the file doesn't exist or is not yaml, so I think no need to add more validation there.

brianhawthorne:
Hi @stefanv and @rowanc1 , this PR is ready to roll as far as I know.  I'd love it if someone could put eyes on it for final review and merge!

packages/mystmd/src/index.ts
```diff
@@ -37,5 +37,9 @@ program.addCommand(makeCleanCLI(program));
 program.addCommand(makeTemplatesCLI(program));
 program.version(`v${version}`, '-v, --version', `Print the current version of ${readableName()}`);
 program.option('-d, --debug', 'Log out any errors to the console');
+program.option(
```
fwkoch, line 40: Using a global option here is great. All commands require the config - `init` will write to the alternative config file, `clean` looks at the alternative config to find exports to delete.

packages/mystmd/src/clirun.ts
```diff
@@ -27,7 +28,9 @@ export function clirun(
   return async (...args: any[]) => {
     const opts = program.opts() as SessionOpts;
     const logger = chalkLogger(opts?.debug ? LogLevel.debug : LogLevel.info, process.cwd());
-    const session = new sessionClass({ logger });
+    // Override default myst.yml if --config option is given.
```
fwkoch, line 31: I think this is good - there's some subtle edge cases around nested folders, where the MyST CLI stops traversing folders if it sees another config file. Since we are overwriting `myst.yml`, traversing will only stop when another config file with the different name is encountered.

I think this case probably never actually comes up, and, if it does, I think the new behavior makes the most sense anyway!

packages/myst-cli/src/session/session.ts
```diff
@@ -97,9 +97,9 @@ export class Session implements ISession {
     return this.$logger;
   }
 
-  constructor(opts: { logger?: Logger; doiLimiter?: Limit } = {}) {
+  constructor(opts: { logger?: Logger; doiLimiter?: Limit; configFiles?: string[] } = {}) {
     this.API_URL = API_URL;
-    this.configFiles = CONFIG_FILES;
+    this.configFiles = (opts.configFiles ? opts.configFiles : CONFIG_FILES).slice();
```
fwkoch, line 102: `slice` here is just to make a shallow copy? 👍

brianhawthorne: I thought it would be surprising if mutating a particular session's configFiles list (or the default CONFIG_FILES list) affected other sessions, so it seemed safer for each session to get a fresh copy of the list.

packages/mystmd/src/index.ts
```diff
@@ -37,5 +37,9 @@ program.addCommand(makeCleanCLI(program));
...
+  '--config <config-file>',
+  'Use an alternate YAML config file, named relative to the project directory',
```
brianhawthorne, line 42: @fwkoch , since you added support for absolute file paths, we might want to tweak this option description so it doesn't explicitly say it must be a relative path.

brianhawthorne: @fwkoch , should these instances also be changed from `join` to `resolve` in order to better support absolute config path?

* https://github.com/jupyter-book/mystmd/blob/e060c9f1dfef5e91afabd6902948895989107f75/packages/myst-cli/src/project/load.ts#L47
* https://github.com/jupyter-book/mystmd/blob/e060c9f1dfef5e91afabd6902948895989107f75/packages/myst-cli/src/project/fromTOC.ts#L219
* https://github.com/jupyter-book/mystmd/blob/e060c9f1dfef5e91afabd6902948895989107f75/packages/myst-cli/src/project/fromTOC.ts#L253
* https://github.com/jupyter-book/mystmd/blob/e060c9f1dfef5e91afabd6902948895989107f75/packages/myst-cli/src/project/fromPath.ts#L49

fwkoch: @brianhawthorne - Hmm, good points, but I'm not sure about all these changes. Going through my testing last night, playing with `--config` absolute paths, originally it hung for a long time, then gave a bad error about calling `lstat` on system-level `/var` files... This is because it could not find the config at all (due to the incorrect `join` paths), so it traversed up the file tree, trying to find a config file, until it got all the way to the root. This hang-then-lstat-error seemed really nasty, more like a crash than an expected exception.

With the changes I made, it is able to find the config so it doesn't have the same error state. However, there is another layer of problems: If you have an (1) absolute path config with (2) explicit toc with subfolders, the absolute path will still cause issues... Basically to the code it "looks like" the config is in every folder (since the absolute path can be loaded from every folder). This means relative paths do not resolve correctly.

As we "improve" the `join` -> `resolve`, these problems will get even worse... For example, in your 4th bullet point, we are building an implicit TOC and looking at subfolders to see if they include a config; if they do, we stop traversing, assuming that's a separate MyST project. If we change that to a `resolve` and leave config as an absolute path, again it will "look like" the config is in every subfolder and the implicit TOC will never include any subfolder...

---

After thinking through all this, I have a few conclusions:

- Absolute paths for `--config` should still not be used. My changes made it so they _might_ work in simple cases (as opposed to crashing every time), but in many cases they will still not really work. Therefore the docstring should stay the same. (Maybe we should even say "config filename" instead of "config file" to lower expectations all the way...)
- Replacing `join` with `resolve` in more places could lead to unwanted consequences. (This is particularly the case for your 4th bullet - the other 3 bullets I _think_ would be ok, but they all look like code paths where the `join` is just a backup, and really we are loading the config from the `selectors`.)
- There is another possible simple-ish step - we could just resolve absolute paths to relative paths immediately at the CLI, before passing to other myst functions. I haven't given this a ton of thought - I'd expect there are still relative path issues if you are pointing to a file that is not in the current directory.
- Finally, the TOC resolution logic is just too complex. MyST was originally built for (a) 1:many site:project ratio, which means potentially multiple parallel config files, (b) git-like ability to run myst cli anywhere inside a project, not just root, (c) discovery and resolution of nested projects, and (d) constant handling of possibly undefined `projectPath`. None of this is necessary! We've documented (a) here https://github.com/jupyter-book/mystmd/issues/1308 (and #1103), the git workflow (b) is total overcomplexity, nested projects (c) are a silly edge case we can just not support, and (d) `projectPath` should always exist and (by eliminating (b)) should always just be the current directory when running the CLI...

Ugh, sorry for the big dump. All that boils down to (in my opinion) leave things as they are for now and make _lots_ of fundamental simplifications later... 🙃

brianhawthorne: Sounds good to me - and I appreciate the detail!   It helps me to understand the trajectory of the project :+1:   

(and thanks again for the final review and getting this merged!)

PR #2241, opened 2025-08-12
🐛 mdast lists need to be in paragraphs

rowanc1:
This transforms listItems after parsing with markdown-it, to ensure that they are mdast compliant. This is important in formatting (e.g. #1948). Without this each inline element of the list item will be on a different line.

Fixes #1957

PR #2245, opened 2025-08-12
🎨 Markdown formatting improvements

rowanc1:
This improves the markdown formatting output.

- [x] merge #2241
- [ ] Have a path forward on how to store parsed directive/option/body data on the nodes
- [ ] Fix parser tests

### DIFF
--- a/packages/myst-ext-card/tests/card.spec.ts
+++ b/packages/myst-ext-card/tests/card.spec.ts
@@ -11,7 +11,7 @@ describe('card directive', () => {
         {
           type: 'mystDirective',
           name: 'card',
-          args: 'Card Title',
+          args: [{ type: 'text', value: 'Card Title' }],
           value: 'Header\n^^^\n\nCard content\n+++\nFooter',
           position: {
             start: {
@@ -142,7 +142,7 @@ describe('card directive', () => {
     const output = mystParse(content, {
       directives: [cardDirective],
     });
-    expect(output).toEqual(expected);
+    expect(output).toMatchObject(expected);
   });
   it('card directive parses with options', async () => {
     const content =
@@ -153,11 +153,11 @@ describe('card directive', () => {
         {
           type: 'mystDirective',
           name: 'card',
-          args: 'Card Title',
+          args: [{ type: 'text', value: 'Card Title' }],
           options: {
-            header: 'Header',
-            footer: 'Footer',
-            link: 'my-url',
+            header: [{ type: 'text', value: 'Header' }],
+            footer: [{ type: 'text', value: 'Footer' }],
+            url: 'my-url',
           },
           value: 'Card\n^^^\ncontent',
           position: {
@@ -280,7 +280,7 @@ describe('card directive', () => {
     const output = mystParse(content, {
       directives: [cardDirective],
     });
-    expect(output).toEqual(expected);
+    expect(output).toMatchObject(expected);
   });
   it('card directive parses with minimal content', async () => {
     const content = '```{card}\nCard content\n```';
@@ -343,7 +343,7 @@ describe('card directive', () => {
     const output = mystParse(content, {
       directives: [cardDirective],
     });
-    expect(output).toEqual(expected);
+    expect(output).toMatchObject(expected);
   });
 });
 
--- a/packages/myst-parser/src/directives.ts
+++ b/packages/myst-parser/src/directives.ts
@@ -3,7 +3,6 @@ import type {
   DirectiveData,
   DirectiveSpec,
   DirectiveContext,
-  ParseTypes,
   GenericParent,
 } from 'myst-common';
 import { RuleId, fileError, fileWarn } from 'myst-common';
@@ -94,6 +93,7 @@ export function applyDirectives(
         if (argSpec.required && data.arg == null) {
           validationError = true;
         }
+        node.args = data.arg;
       }
     } else if (argNode) {
       const message = `unexpected argument provided for directive: ${name}`;
@@ -104,6 +104,7 @@ export function applyDirectives(
     // const options: Record<string, ParseTypes> = {};
     const { valid: validOptions, options } = parseOptions(name, node, vfile, optionsSpec);
     data.options = options;
+    node.options = options;
     validationError = validationError || validOptions;
 
     // Handle body
@@ -124,6 +125,7 @@ export function applyDirectives(
           `body of directive: ${name}`,
           RuleId.directiveBodyCorrect,
         );
+        node.body = data.body;
         if (bodySpec.required && data.body == null) {
           validationError = true;
         }
--- a/packages/myst-to-md/src/directives.ts
+++ b/packages/myst-to-md/src/directives.ts
@@ -5,6 +5,7 @@ import type { GenericNode } from 'myst-common';
 import { toText, fileError, RuleId } from 'myst-common';
 import { select, selectAll } from 'unist-util-select';
 import type { VFile } from 'vfile';
+import { longestStreak } from 'longest-streak';
 import type { NestedState, Parent, Validator } from './types.js';
 import { incrementNestedLevel, popNestedLevel } from './utils.js';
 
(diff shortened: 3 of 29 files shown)


packages/myst-ext-card/tests/card.spec.ts
```diff
@@ -142,7 +142,7 @@ describe('card directive', () => {
     const output = mystParse(content, {
       directives: [cardDirective],
     });
-    expect(output).toEqual(expected);
+    expect(output).toMatchObject(expected);
```
rowanc1, line 145: Tests in the extension code are made a bit less strict here. No changes to the code, so no need to bump the packages.

packages/myst-parser/src/directives.ts
```diff
@@ -94,6 +93,7 @@ export function applyDirectives(
         if (argSpec.required && data.arg == null) {
           validationError = true;
         }
+        node.args = data.arg;
```
rowanc1, line 96: This is the biggest change is that we put the arg, options and body of roles and directives onto the `mystDirective` node. This allows us to process them afterward in a formatting package - or transform a figure into a directive etc. rather than doing that in the markdown writer.

This shouldn't matter for the duplication as we remove these nodes early in the processing for most things. But it is really helpful to have them for the formatting / writing of markdown.

We might want to change it to `node.argsParsed` or something so we have the pre/post parsing? Or put it on the data field.

@fwkoch or @agoose77 if you have opinions here, that would help.

rowanc1 (commented):
Highlighting changes below. We need to get the parsing data onto the mystDirective/Role nodes, so that we have an easier time in formatting. Would love some thoughts on direction here.

packages/myst-parser/src/directives.ts
fwkoch: Ok - seeing this, my first impression was "we already stash this stuff on the `mystDirective` nodes, don't we...?" - That's not entirely true.

On the initial parse pass, we do save `args` and `options` to the `mystDirective` node; these are the raw, unparsed values. I don't _think_ these fields are used anywhere. Additionally, we create `mystDirectiveArg/Option/Body` nodes, where the raw value is saved to be parsed, in this `applyDirectives` function. Within this function we are building a `data` object of type `DirectiveData` which contains all the parsed arg/option/body content. Critically, this `data` object is transient; it's just passed directly to the directive plugins.

So, the change in this PR is taking the parsed, previously transient arg/option/body values and saving them directly to the `mystDirective` node. I agree this is a good change.

---

Getting back to your specific question @rowanc1 - are we putting this in the right place? The biggest issue is that we are clobbering raw `arg/option` values with parsed `arg/option` values. I think this is fine; we really should only ever be dealing with the parsed values. The only argument I can think of for keeping the raw values is for round-tripping documents _exactly_ as written. But that's not useful at all - MyST does not care about this (I don't think!), we just don't want to lose any data, and for that, parsed structured data is better. (This is especially true if we are thinking about converting between formats - in that case, any raw value is totally useless.)

I guess the other argument against clobbering the raw values with parsed values is simply backwards compatibility. This is a backwards incompatible change that I don't _think_ will impact anyone, but there could be something somewhere using these raw `mystDirective` fields. (Maybe we make this a minor version bump?)

Regarding the alternatives you suggest:

1. I don't think we should do `argsParsed`... that feels excessive; we are keeping track if `node.processed === false`, so if that's not the case (i.e. the node is processed), `args` will be parsed. It's also a little silly with `bodyParsed` since we are not saving the raw `body`...
2. Stashing `args/options/body` on `node.data` would be fine - this is just taking the `DirectiveData` object and saving it as "directive data." I somewhat like that, but at the same time, the `data` field is usually used for `metadata` - in the case of `mystDirective` nodes, args/options/body are the fundamental content, not metadata. 🤷

So maybe it's just fine as-implemented? (We could potentially even stop creating the `data` object, and pass the `node as DirectiveData` to the `run` function... But I think for now this is overthinking it...)

---

TL;DR: 👍 to your change, maybe minor version bump.

packages/myst-to-md/src/directives.ts
```diff
@@ -74,12 +75,45 @@ function writeStaticDirective(name: string, options?: DirectiveOptions) {
...
+    const optLabel = label ? `#${label}` : '';
+    const optClass = className
+      ? className
+          .split(' ')
+          .filter((c: string) => c.trim() !== '')
+          .map((c: string) => `.${c.trim()}`)
+          .join(' ')
+      : '';
+    const optOther = Object.entries(otherOptions)
+      .map(([key, value]) => `${key}=${value}`)
```
fwkoch, line 93: Writing all options this way feels risky. Should we check if `value` has spaces?

packages/myst-to-md/src/directives.ts
```diff
@@ -74,12 +75,45 @@ function writeStaticDirective(name: string, options?: DirectiveOptions) {
...
+      ? state.containerFlow({ type: 'root', children: node.body }, info)
+      : node.value;
+
+    const nesting = popNestedLevel('directive', state);
+    const markerChar = markdownBody ? ':' : '`';
+    const markerLength = Math.max(longestStreak(content, markerChar) + 1, nesting + 3);
+    const marker = markerChar.repeat(markerLength);
+    const directiveInline = [node.name, directiveOpts].filter(Boolean).join(' ');
+    const args = Array.isArray(node.args)
+      ? state.containerPhrasing({ type: 'heading', depth: 1, children: node.args }, info)
```
fwkoch, line 110: I don't get this - Are we injecting this `heading` just to prepend `#`...? I don't think we always want to do that? But I might just be completely misunderstanding!

packages/myst-to-md/tests/mdast/directives.yml
```diff
@@ -119,18 +119,11 @@ cases:
       type: root
       children:
         - type: math
-          value: |-
-            5+5
-             ````{abc}
-            ````
-            print("hello world")
+          value: Ax=b
```
fwkoch, line 122: Is there a reason this `value` was simplified? I guess maybe backticks no longer matter so much in `math` since we are using `$$` instead of math directive.

But what about the equivalent issue - what if there is a `$$` inside the math value...? Are we handling that correctly?

packages/myst-to-md/tests/mdast/directives.yml
```diff
@@ -960,3 +945,82 @@ cases:
...
+        - type: paragraph
+          children:
+            - type: text
+              value: This paragraph introduces a directive with arguments and options.
+        - type: mystDirective
+          name: admonition
+          args: Important Note
+          value: |-
+            :class: my-class
+            :name: my-note
```
fwkoch, line 1009: I don't think this test is quite correct. By the time we have a `mystDirective` node, I _think_ these options should already be stripped off into a separate field. (Now the test is just treating these as part of the body)

packages/myst-to-md/src/roles.ts
```diff
@@ -103,14 +104,25 @@ function citeGroup(node: any, _: Parent, state: NestedState, info: Info): string
...
 
+/**
+ * Inline math handler
+ */
+function inlineMath(node: InlineMath, _: Parent, state: NestedState, info: Info): string {
+  if (node.label || node.typst) {
+    return writeStaticRole('math')(node, _, state);
+  }
+  const tracker = state.createTracker(info);
+  return tracker.move(`$${node.value}$`);
```
fwkoch, line 115: Why does this have two `$` at the beginning?

packages/myst-to-md/src/directives.ts
```diff
@@ -74,12 +75,45 @@ function writeStaticDirective(name: string, options?: DirectiveOptions) {
...
 
 /**
- * Generic MyST directive handler
- *
- * This uses the directive name/args/value and ignores any children nodes
+ * Handler for a raw directive
  */
-function mystDirective(node: any, _: Parent, state: NestedState, info: Info): string {
-  return writeStaticDirective(node.name, { argsKey: 'args' })(node, _, state, info);
+function writeDirective(options?: DirectiveOptions) {
```
fwkoch, line 80: Ok, I'm feeling a little confused about this.

First, I don't think we need this wrapper "factory function" - `options` is unused. We can just define the function directly.

Second, we are only using `writeDirective` when we encounter `mystDirective` nodes. These only persist in the tree when they have an unknown directive type. In that case, in the `applyDirectives` function, we are deleting the `children` and writing the `options` back to the body. It never even gets to the new code where `node.args`/`node.options`/`node.body` are set. For known directives, the `mystDirective` nodes get lifted out.

packages/myst-to-md/tests/roundtrip/roundtrip.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      And another paragraph
+      ```
+
+      Paragraph after directive
+    result: |-
+      (my-label)=
+      # Heading
+
+      :::{admonition #my-label .dropdown} Admonition title
```
fwkoch, line 48: I'm extra confused why this works - it's a known directive type but it's using the new `writeDirective` code, which only applies to `mystDirective` nodes, not `admonition` nodes.

packages/myst-to-md/tests/roundtrip/roundtrip.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+casesList.forEach(({ title, cases }) => {
+  describe(title, () => {
+    test.each(cases.map((c): [string, TestCase] => [c.title, c]))(
+      '%s',
+      (_, { markdown, result: expected }) => {
+        const mdast = mystParse(markdown, { mdast: { hoistSingleImagesOutofParagraphs: false } });
+        const file = unified()
+          .use(mystToMd)
```
fwkoch, line 35: Ohhhh... now I understand. This roundtrip test is checking the `myst-parse -> myst-to-md` roundtrip. It is _not_ using any of the `myst-cli` pipeline. Therefore, it's skipping all the basic transforms which are just applied by default in `myst-cli`.

If we were to try to round-trip via `myst-cli`, I don't think we'd see the same results.

fwkoch (commented):
Ok - I responded to your question @rowanc1 - but after getting to understand this PR a little more, I think we need slight clarity on the direction:

The tests for these improvements, especially around directive/role handling, are focused on a narrow scoped pipeline that only includes `myst-parser` and `myst-to-md`. It ignores all the transforms and additional processing included in the full `myst-cli` pipeline.

For me to feel comfortable with these changes we need to do one of three things:
1. Clarify the scope of this PR is isolated to standalone usage of `myst-parser` and `myst-to-md`; we should not expect to see the same improvements when using the full `myst-cli` pipeline.
2. Modify the `myst-cli` pipeline for md export to perform fewer transforms (including no lifting of `mystDirective` nodes, no propagating `mystTargets`, etc).
3. Refactor the changes here so they provide the same improvements to post-processed `myst-cli` mdast as they do to unprocessed `myst-parser` mdast.

Happy to discuss all this with you more!

PR #2281, opened 2025-09-09
🔐 Improve security for localhost

### LINKED ISSUE #2254: `myst start` with `article-theme` and `book-theme` binds to all interfaces instead of just localhost
## Description

```shell-session
$ myst -v
v1.6.0
```

When I run `myst start` to preview my documentation via a local web server (as suggested at <https://mystmd.org/guide/quickstart#preview-your-myst-site-locally>) I find that the web server is actually listening on all interfaces.

```shell-session
$ myst start
```

```shell-session
$ netstat -f inet -anp tcp | grep -E '(^Proto|3000)'
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)    
tcp46      0      0  *.3000                 *.*                    LISTEN
```

Based on the MyST MD documentation and the output of `--help`, I would have expected the server to listen only on localhost by default.

```shell-session
$ myst start --help
Usage: myst start [options]

Start the current project as a website

Options:
  --execute                      Execute Notebooks (default: false)
  --keep-host                    The HOST environment variable is changed to "localhost" by default. This flag
                                 uses the original environment variable. (default: false)
  --headless                     Run the server in headless mode, with only the content server started
                                 (default: false)
  --port <port>                  Run the application server from the specified port number (env: PORT)
  --server-port <server-port>    Run the content server from the specified port number (env: SERVER_PORT)
  --template <path-to-template>  Use this template file, instead of the one specified in the myst.yml manifest
  --max-size-webp <size>         Max image size to convert to webp format in MB (default: 1.5)
  -h, --help                     display help for command
```

Is there any way to force the server started by `myst start` to only listen on `localhost`?

rowanc1:
Pass the host name through correctly to the theme and ensure that the default configuration will only serve on localhost unless a host is specified and passed through explicitly with `--keep-host`.

See also: jupyter-book/myst-theme#646
Fixes #2254

### DIFF
--- a/.changeset/light-coins-draw.md
+++ b/.changeset/light-coins-draw.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Ensure host is only for local development and is not exposed on local network.
--- a/packages/myst-cli/src/build/site/logger.ts
+++ b/packages/myst-cli/src/build/site/logger.ts
@@ -2,7 +2,10 @@ import chalk from 'chalk';
 import type { LoggerDE } from 'myst-cli-utils';
 import type { ISession } from '../../session/types.js';
 
-export function createServerLogger(session: ISession, ready: () => void): LoggerDE {
+export function createServerLogger(
+  session: ISession,
+  opts: { host: string; ready: () => void },
+): LoggerDE {
   const logger = {
     debug(data: string) {
       const line = data.trim();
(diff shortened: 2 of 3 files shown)


rowanc1:
This is good to go from my perspective, and has been tested. See #2254 for results before and after.

The major change is to add a `host` to the express app and pass that through to the theme server in the theme:

- https://github.com/jupyter-book/myst-theme/pull/646

I will leave it up for another day or so to have any comments and then merge.

PR #2309, opened 2025-10-02
Document static theme development

brianhawthorne:


brianhawthorne:
@stefanv ping

Not sure what I did to get all the --config commits appearing in this PR... but it only shows the single `developer.md` file as changed (which is correct).  Any suggestion about how to avoid this in the future will be appreciated!

docs/developer.md
```diff
@@ -311,13 +311,17 @@ When building on Windows, use either WSL or a unix-like shell (such as Git Bash
 
 The [`myst-theme` README](https://github.com/jupyter-book/myst-theme/) provides a more detailed overview of the components of that package.
 
+A theme may be deployed locally for development as a dynamic theme server, _or_ a site may be statically built against a local theme.  Development procedures will be different depending on which deployment scenario you are targeting.
```
stefanv:
```suggestion
For local development, there are two ways to deploy the theme (AKA the website):
 as a server, or as a static build. Development procedures are slightly different, depending on which approach you take.
```

I hope my rephrasing is accurate 🤔 

Here, we could also add a short sentence as to why you might use which approach. E.g., something like: "The static build approach is typically used when..." where it might be "you need a React debug build" or "you want to emulate a GitHub Pages deploy" or ... ?

I just re-read the whole thing, and realized that matters are further complicated, because additionally you can build the theme into a JS package, or access the theme locally. It feels like, in this PR, we may be conflating the two axes: static/non-static build, pre-built or non-pre-built theme server.

docs/developer.md
```diff
@@ -311,13 +311,17 @@ When building on Windows, use either WSL or a unix-like shell (such as Git Bash
...
+
+### Dynamic Site
```
stefanv:
```suggestion
### Approach 1: Theme server
```

"Approach" may not be the right word, but it would be good to highlight the two paths in the titles.

docs/developer.md
```diff
@@ -367,6 +371,129 @@ npm run dev
 
 Note that you can run `npm run dev` from within any folder if you'd like to watch individual packages instead of the entire directory structure.
 
+### Static Site
```
stefanv:
```suggestion
### Approach 2: Static build
```

docs/developer.md
```diff
@@ -367,6 +371,129 @@ npm run dev
...
+
+No content or theme server is required for a static site build.  Steps are:
+
+1. Build the theme into a production deployment package
```
stefanv: perhaps "into a package, as would be deployed to npm" or similar. Just trying to think how we can make it obvious what we're building here. "production deployment package" may not be immediately intuitive.

stefanv: Also, this is the step I wasn't sure about—in the sense of: you can do a static build against the theme server too.

stefanv: I'm going to pause my review until we've discussed the axes mentioned above (static/non-static, pre-built/non-pre-built), and I have a better sense of what this guidance should look like.

brianhawthorne: I think we determined in our external discussion that static build against a manually hosted theme server is not possible.  But I do agree the word "package" is confusing, so I've just avoided it entirely here.  No need to even mention github archive zipfile download and unpacking.

**Side note:**  Though `myst-theme` repo does publish everything under its `packages` directory as a JS package to npmjs registry, the themes themselves do _not_ get published to npmjs registry (or I couldn't find them), despite being structured as JS packages.  When used, they get downloaded as a github zip archive (of main:HEAD by default) from a github repo under the `myst-templates` github org.  (not to be confused with the `@mystmd/myst-templates` npm package..)

So "publishing" a theme means committing its built `.deploy` directory as the contents of the corresponding repo under myst-templates org.

choldgraf:
Could I suggest we get this PR to "safe to try" and just merge it? THere's a lot of good content in here and we can always iterate on it!

brianhawthorne:
Thanks for the merge, @stefanv !

PR #2323, opened 2025-10-08
🛠️ Bring MarkdownIt version consistently in-line internally

agoose77:
When bundling the latest MyST components in `jupyterlab-myst`, I observed the following:

- We have mixed `markdown-it` versions supported by our plugins (`myst-parser` expects `^12` whilst `markdown-it-myst` expects `^13`)
- MarkdownIt should not be a dependency (it should be a peer or not at all).
- The `types` for MarkdownIt plugins should be dependencies.

I've made the following changes:
- Updated `myst-parser` to pull in `markdown-it` 13.
- Removed `markdown-it` from MarkdownIt plugins defined in this monorepo -- we don't want to install a copy of MarkdownIt inside a plugin.
- Added MarkdownIt types to the `dependencies` field of our MarkdownIt plugins -- we should be able to export types for plugins.

I don't know if this will fully resolve things in `jupyterlab-myst`, but it feels like the right step.


rowanc1:
All other packages in this repo, all `@types/*` packages are in devDependencies. As they are compile time only and the type definitions are only needed during development and compilation, not at runtime.

I think that should also be the case here, right?

Looks like the tests are failing on `text_collapse`, I haven't taken a look at the release notes yet to know what to do with that change.

agoose77:
I think this is a policy we should change. We need to propagate transitive
types.
For example:
Package a exports an interface that has imported types in the signature
(e.g markdown it plugin with options).
Package b cannot infer the types of that signature without the user
manually installing and correcting the types.

I noticed that failure and moved this pr to draft - I will try to get to it
today!

rowanc1:
Types are not used in production, and **_most_** of the types we use are internal and not exposed, so that generally goes against the official recommendation here:
https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

If we are exposing types from another `@types/*` package (like this case), then they should go in deps. But this should be done case-by-case and really only if we have a problem downstream (like in this case).

---

i.e. the exception is one of our TypeScript libraries that:

- Exposes public types (e.g. in index.d.ts or .ts exports).
- Those types reference types from **another** library’s `@types/*` package (e.g. markdownit)
- The other library **doesn’t** ship its own types — they’re only available via `@types/some-lib`.

In this case, if we put `@types/some-lib` only in devDependencies, consumers get type errors, because our published `.d.ts` files reference symbols they (probably) don’t have installed.

agoose77:
> Types are not used in production, and **_most_** of the types we use are internal and not exposed, so that generally goes against the official recommendation here: https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file
> 
> If we are exposing types from another `@types/*` package (like this case), then they should go in deps. But this should be done case-by-case and really only if we have a problem downstream (like in this case).

I think we're talking cross-purposes here!

I am _not_ advocating for putting all types into `dependencies`. I'm rather pointing out that _these_ types should be in `dependencies` because they're transitive. Regarding policy, my point is that we shouldn't _always_ put types into `devDependencies` (for the reason that you give above).

I think you're on the same page as me there, so we're aligned!

rowanc1:
I thought you were advocating to move all types into deps! Perfect, we are on the same page.

(Was helpful for me to write / work through understanding what the exception was as well..!)

agoose77:
So I haven't added a rule for `text_special` because I'm expecting the `text_join` rule to convert these tokens to `text` tokens. I am not sure why that's not taking place.

agoose77:
Ah, it's because the existing `roles` rule is added after `inline`, which skips the `text_join` rule. I've just fixed that.

agoose77:
@rowanc1 is this GTG?

rowanc1:
Can I take a look this eve, and will merge.

rowanc1:
🚀

PR #2329, opened 2025-10-09
🤐 Show an error when the `myst.xref.json` fails to load

rowanc1:


packages/myst-cli/src/process/loadReferences.ts
```diff
@@ -180,9 +189,10 @@ export async function loadReferences(
       })
       .filter((exists) => !!exists),
   );
-  return Promise.all(
+  await Promise.all(
     references.map(async (ref) => {
       await loadReference(session, vfile, ref);
     }),
   );
+  logMessagesFromVFile(session, vfile);
```
rowanc1, line 197: We were just loosing the `vfile` and log messages here...!

PR #2330, opened 2025-10-09
✍️ Improvements for typst exports

fwkoch:
- Allow cross-references to implicit headings in typst (typst has a separate guard to stop _duplicate_ headings)
- Pass corresponding=false to templates (so templates can differentiate between corresponding=undefined and corresponding=false)
- bump tex-to-typst

### DIFF
--- a/packages/myst-templates/src/frontmatter.spec.ts
+++ b/packages/myst-templates/src/frontmatter.spec.ts
@@ -279,4 +279,55 @@ describe('extendFrontmatter', () => {
       },
     ]);
   });
+  it('corresponding=false is persisted in extended frontmatter', async () => {
+    const frontmatter: PageFrontmatter = {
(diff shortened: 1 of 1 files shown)


packages/myst-to-typst/src/index.ts
```diff
@@ -489,7 +490,7 @@ class TypstSerializer implements ITypstSerializer {
     const { math, ...otherOpts } = opts ?? {};
     this.options = { ...otherOpts };
     if (math) this.options.math = resolveRecursiveCommands(math);
-    this.data = { mathPlugins: {}, macros: new Set() };
+    this.data = { mathPlugins: {}, macros: new Set(), headingIdentifiers: [] };
```
rowanc1, line 493: Nice!

PR #2332, opened 2025-10-10
📖 Document error rules

rowanc1:
This adds descriptions for the error rules. These are also added to the docs pages for better discoverability. Currently you had to run in debug mode `myst build -d` to figure these out. Nicer if they are documented!!

image

cc @bsipocz


packages/myst-common/src/ruleids.ts
```diff
@@ -107,3 +107,122 @@ export enum RuleId {
   // File rules
   mystJsonValid = 'myst-json-valid',
 }
+
+/**
+ * Descriptions for each rule ID.
+ * TypeScript will enforce that every RuleId has a corresponding description.
+ */
+export const RULE_ID_DESCRIPTIONS: Record<RuleId, string> = {
```
rowanc1, line 115: This will fail typescript if a new rule is added without a description!!

PR #2333, opened 2025-10-10
📖 Add default severity for `RuleIds` to the docs

rowanc1:
This looks up the code and adds a default severity to myst-common.

image

cc @bsipocz

### DIFF
--- a/packages/myst-common/src/ruleids.ts
+++ b/packages/myst-common/src/ruleids.ts
@@ -41,7 +41,6 @@ export enum RuleId {
   encounteredLegacyTOC = 'encountered-legacy-toc',
   validTOCStructure = 'valid-toc-structure',
   validTOC = 'valid-toc',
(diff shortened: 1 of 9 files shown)


packages/myst-common/src/ruleids.ts
```diff
@@ -41,7 +41,6 @@ export enum RuleId {
   encounteredLegacyTOC = 'encountered-legacy-toc',
   validTOCStructure = 'valid-toc-structure',
   validTOC = 'valid-toc',
-  tocWritten = 'toc-written',
```
rowanc1, line 44: `tocWritten` and `pluginExecutionFailed` were not used in the code anywhere. This removes them.

rowanc1:
Nice - yeah I think that is better. And can come later. We will get build errors when the code changes, so can start to address it then and then throw away the script.

Issue #2317, opened 2025-10-04
using gh release to publish themes / beneficial for custom themes ?


parmentelat:
as of today, to the best of my knowledge, one can use an alternate/custom theme by specifying in `myst.md` e.g.

```yaml
site:
  template: https://github.com/myst-templates/book-theme/archive/de4c1e9c09c6cd12fc529c1f22e8abf8e7f35381.zip
```

in this example, the zip file is produced by essentially 2 pieces

- a gh action  
  https://github.com/jupyter-book/myst-theme/blob/main/.github/workflows/release.yml
- which in turn uses the topelvel Makefile  
  https://github.com/jupyter-book/myst-theme/blob/main/Makefile

the effect of which being to create a commit in each of the 2 'ghost' repos  
- executablebooks/myst-article-theme
- executablebooks/myst-book-theme

the content of that commit being then retrievable through github's `/archive/` route

---

the purpose of this issue is to propose the use of the GH releases instead, which would imho 
- make this workflow simpler for the official themes : in particular
  - no need for the ghost repos
  - and a clearer web interface to browse through available releases/versions
  - also get rid of Makefile and group all the code in the gh actions
- but also make it simpler for forked instances of the official theme - typically for people who simply to do small changes - to publish themselves, as they could just use the exact same action without having to change or provision anything

---
I have made some preliminary work along these lines that are rather encouraging here
https://github.com/flotpython/myst-theme/blob/main/.github/workflows/gh-release.yml
and the resulting artefacts are browsable here
https://github.com/flotpython/myst-theme/releases

so I'd love to hear from the core team to see if that could be of interest, and to hear about any pitfall that I have not identified at this point
I reckon my understanding of the theme fetch/download code is minimal; plus I have only considered book/article and have no idea about the other available themes

parmentelat:
an additional benefit, users would be able to refer to tagged template using the tag name as in
```
template: https://github.com/executablebooks/myst-theme/releases/download/the-tag/book.zip
```

parmentelat:
which btw raises another, partially unrelated question:
imo it would be nicer if a given version of mystmd came with a pinned version of its default themes, instead of relying on the latest published ones, that might just break stuff for older versions
but that's probably for another issue altogether..

rowanc1:
@parmentelat I think this would be awesome to move to this pattern of releases for the theme.

I think the myst-template resolver will need a bit of work to look for the zip file in a new place, but the URL looks a lot cleaner...!

[...]

PR #2344, opened 2025-10-11
support theme URLs that end in .release

parmentelat:
implement an idea described in https://github.com/jupyter-book/mystmd/issues/2317

specifically, this allows a theme endpoint to define a 'download' key as being e.g.

    "download": "https://github.com/jupyter-book/myst-theme/book.release"

and then rewrite it into

https://github.com/jupyter-book/myst-theme/releases/download/main/book.zip

### DIFF
--- a/packages/myst-templates/src/download.ts
+++ b/packages/myst-templates/src/download.ts
@@ -176,7 +176,7 @@ export async function downloadTemplate(
     }
     const urlBase = downloadUrl.substring(0, downloadUrl.length - 4);
     downloadUrl = `${urlBase}/archive/refs/heads/${branch}.zip`;
-  // this scheme is for using GH releases as a way to distribute stable templates
+    // this scheme is for using GH releases as a way to distribute stable templates
   } else if (downloadUrl.endsWith('.release')) {
     session.log.warn(`👷 Warning, using a .release link which may not be stable yet`);
     // xxx rather use a tag ???
@@ -185,7 +185,7 @@ export async function downloadTemplate(
       session.log.warn(`👷 Warning, using a branch: ${branch}`);
     }
     // the /releases/ endpoint makes sense under gh only
-    const re = /^https:\/\/github\.com\/([^/]+)\/([^/]+)\/([^.\/]+)\.release\/?$/;
+    const re = /^https:\/\/github\.com\/([^/]+)\/([^/]+)\/([^./]+)\.release\/?$/;
     const match = downloadUrl.match(re);
     if (!match) {
       throw new Error(`Problem with .release link "${downloadUrl}": Not a valid GitHub URL`);


parmentelat:
as an aside, it feels odd to test for node 18 and not for node 24...

parmentelat:
in order to get the first check to pass, I would add the 'enhancement' label if I were allowed

agoose77:
Hey @parmentelat, for some reason I keep writing replies and I can't find them! Probably getting stuck on spotty internet.

I think we should look to publishing releases on GitHub, and we should support consuming GitHub release URLs. However I don't like the idea of introducing a pseudo URL for an existing scheme (`https`). Normally, `https://` URLs are resolvable by HTTP clients, but this wouldn't be true for this scheme, right?

Is there a problem with just directly referring to the release artifact? e.g. https://github.com/jupyter-book/mystmd/archive/refs/tags/mystmd@1.6.7.zip

parmentelat:
hey @agoose77 

first off, this goes together with 
https://github.com/jupyter-book/mystmd/issues/2317
and
https://github.com/jupyter-book/myst-theme/pull/659
maybe that's what you had in mind ?

---

so, the approach here has been to mimick what was in place with the .git URLs (the lines just above)
because, as much as the .git URLs like `https://github.com/myst-templates/book-theme.git` would actually resolve and be fetchable by a regular http client, what the fetch would yield is just the landing page for that gh repo ! 
it looked like using this scheme for naming URLs was just a trick to refer to the content of a specific commit on github

and so, this PR is about providing a means for the API server to say, go fetch a release on gh instead of a commit
it's admittedly as artificial as the .git URLs routing :)

--
now, you seem to suggest that the API server could deliver a .zip URL directly - thus removing the need for the CLI to rewrite a .git / .release URL

I'm all for that, just wanted to stress that there seemed to be a reason for this extra logic in the CLI

--
specifically, what I'm trying to say is that right now 
```
❯ http https://api.mystmd.org/templates/site/myst/book-theme | yq .links.download
https://github.com/myst-templates/book-theme.git
```

iiuc, you're saying it could just as well answer instead
```
https://github.com/myst-templates/book-theme/archive/refs/heads/main.zip
```

except that this scheme does not allow for the CLI to inject a branch name
and I took it that was the whole purpose of adding that extra layer of logic in the CLI

--
to summarize, the proposed .release scheme was an attempt at being consistent with what is done with the .git URLs

in my experience I haven't been able to leverage the `branch` thing with the current code when I needed to pin a theme version, but that clearly would have been a plus ! 
so imho it's worth preserving that extra logic layer in the CLI so that it becomes easier in the future to pin themes

note that another way around would be to alter the API and to allow the CLI to specify a theme/template versioning data, but that sounds over the top..

let me know what you think

parmentelat:
@agoose77 - @rowanc1 

digging this matter a little further
I just came across this part of the puzzle
https://github.com/myst-templates/templates/blob/main/src/utils.ts#L33-L40

which is where the API builds the 'download' link that the CLI then consumes to actually fetch stuff

that's where the '.git' URL is made up - a rather fake one I was arguing - and that's the reason for these lines here - the ones I was trying to mimick
https://github.com/jupyter-book/mystmd/blob/main/packages/myst-templates/src/download.ts#L171-L178

now all this put together, it feels like the code would behave exactly the same if we removed
- the one-liner in the template that creates the .git url
- the special case for the CLI to handle downlaod URLs that end in .git

which means if I'm right that this very PR would be about removing code rather than adding some :)


I'll try to find a means to confirm this conjecture (that removing the .git URL right in the API doesn't break anything)
in the meanwhile I'd love to hear your input, because I take it at some point at least this .git URL hack must have beeen useful for something, right ?

agoose77:
@parmentelat I'm struggling to understand this discussion slightly -- why would we need to remove the `.git` handling?

parmentelat:
ok, understandable :)

so when the CLI asks the API about where to find e.g. book-theme, the answer currently is a .git URL; then in turn the CLI has this special case that when it's a .git URL it rewrites it, taking advantage of the /archive github route, into an actual .zip URL that is what gets actually fetched 

the only apparent interest is, maybe, for the CLI to be able to inject some variant (e.g. the branch); 

as you were saying yourself, this .git looks like a genuine one; but in reality it's feels more like a hack, imho the API should do the resolving and the CLI should just need to download that without this messy rewriting rule

there clearly are too many different open channels about this obviously (because PRs belong in their respective repos..)
but having stepped back a bit after the various discussions, it is my belief that one reasonable path could be to extend the API so that the CLI can pass it a them version, and possibly a CLI version, and for the API to perform all the resolutions needed at one given point in time.

I am eager to spell this out properly, and was proposing on discord to use the new MEP process for that, just waiting for some kind of green light to avoid polluting this new channel if some other media would be better

parmentelat:
closing this thread, please use 
https://github.com/jupyter-book/myst-enhancement-proposals/issues/34
for further discussion on this topic

Issue #2338, opened 2025-10-10
Capitalization not preserved when using `term` role


mfisher87:
## Description

If I have a glossary entry:

```
:::{glossary}
Salad
: A food that's only worth eating with peanut sauce 🥜
:::
```

And I reference it with 

```
I only enjoy {term}`salad` sometimes.
```

It renders as:

>I only enjoy [Salad](https://example.com) sometimes.

The link goes to the correct glossary entry, but the capitalization is not what I asked for in my term reference. The capitalization used in the term definition is used instead.

## Proposed solution

Continue matching terms in the glossary case-insensitively, but preserve the capitalization the author used in the `{term}` role when rendering the link. Should look like:

>I only enjoy [salad](https://example.com) sometimes.

rowanc1:
Yeah - this annoys me a lot. Work around at the moment:

```
{term}`See CapitaLiZation <capitalization>`
```

mfisher87:
Yay @agoose77 🎉 Thank you!

PR #2346, opened 2025-10-14
🔠 Always set term label from `{term}`

agoose77:
Fixes https://github.com/jupyter-book/mystmd/issues/2338 by _always_ setting the children of the `crossReference`, solving the problem of mis-matched capitalisation.

I'm not sure if there are any drawbacks to this approach — any objections welcome.

### DIFF
--- a/packages/myst-roles/src/term.ts
+++ b/packages/myst-roles/src/term.ts
@@ -25,7 +25,7 @@ export const termRole: RoleSpec = {
       type: 'crossReference',
       label,
       identifier: `term-${identifier}`,
-      children: [{ type: 'text', value: modified ? modified.trim() : identifier }],
+      children: [{ type: 'text', value: modified ? modified.trim() : rawLabel ?? body }],
     };
     return [crossRef];
   },


rowanc1:
I will aim to have a play around with this in the evening.

agoose77:
@rowanc1 did you get a chance to see this?

packages/myst-roles/src/term.ts
```diff
@@ -25,8 +25,8 @@ export const termRole: RoleSpec = {
       type: 'crossReference',
       label,
       identifier: `term-${identifier}`,
+      children: [{ type: 'text', value: modified ? modified.trim() : identifier }],
```
fwkoch: I think we want to use `label` instead of `identifier`? Or maybe `rawLabel ?? body`? If I do this:

```markdown
:::{glossary}
MyST
: An amazing markup language that supports glossaries
:::

You can use {term}`MyST` to create glossaries.
```

It renders as `You can use myst to create glossaries`; there is no `modified` so we are just putting in the `identifier`.

fwkoch:
```suggestion
      children: [{ type: 'text', value: modified ? modified.trim() : rawLabel ?? body }],
```

fwkoch (changes requested):

packages/myst-roles/src/term.ts
agoose77: Yes, absolutely!

Issue #2351, opened 2025-10-16
Parts Headings


alanlujan91:
I'm working on a latex template, where I am adding an appendix as a "part" of the document.

```
parts:
    appendix: appendix.md
```

This seems to work, but it always demotes the headers in the appendix; for example, it turns `# Appendix` to a subsection instead of a section in LaTeX. 

I would like it to be able to follow the headings hierarchy, where # Appendix is another top level section of the paper.

alanlujan91:
Reproduction on 1.10.1 and current `main`: `appendix.md` with `title: Appendix A` in frontmatter and

```markdown
# First appendix section
## Nested appendix subsection
```

referenced as `parts: {appendix: appendix.md}` comes out at depth 2 and 3. #2352 fixes it by skipping heading normalization for `Part` files.

One thing worth deciding there: preserving the levels puts a part's `# Heading` at depth 1, the same level as the page title, so HTML gets a second `<h1>`. That suits the LaTeX appendix case but may not suit the web.

PR #2352, opened 2025-10-16
Preserve heading levels for Part source files

alanlujan91:
Updated the mdast transformation to retain original heading levels for files of kind 'Part', ensuring structural sections like appendices are not altered. Other file kinds continue to adjust heading depth based on title and frontmatter.

### DIFF
--- a/.changeset/tidy-melons-cheer.md
+++ b/.changeset/tidy-melons-cheer.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Preserve heading levels in frontmatter part files
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -200,7 +200,11 @@ export async function transformMdast(
     .use(htmlPlugin, { htmlHandlers }) // Some of the HTML plugins need to operate on the transformed html, e.g. figure caption transforms
     .use(basicTransformationsPlugin, {
       parser: (content: string) => parseMyst(session, content, file),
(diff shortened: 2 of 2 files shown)


stefanv:
@alanlujan91 I'm sorry, I see your PR never got reviewed. Could you add a bit of context and an example that illustrates the need of the PR?

alanlujan91:
Thanks for the nudge, and sorry for the slow reply.

**Context:** part files loaded via frontmatter (`parts:`, `abstract:`) go through the same heading normalization as regular pages, so their headings are demoted one level. For a structural part like an appendix that breaks LaTeX appendix numbering: the demoted `\subsection` does not pick up the class's appendix prefix, and the PDF prints a malformed ".1" instead of "APPENDIX A:".

**Example.** `paper.md` with `parts: {appendix: appendix.md}`, and `appendix.md` containing frontmatter `title: Appendix A` plus:

```markdown
# First appendix section
## Nested appendix subsection
```

Built against current `main`, those headings come out at depth 2 and 3. With this PR they stay at 1 and 2, and the main body is unaffected (`# Introduction` is depth 2 either way).

I have merged current `main` into the branch and added a changeset. `myst-cli` tests pass (272/272).

One thing worth your judgment: preserving the levels means a part's `# Heading` lands at depth 1, the same level as the page title, so HTML gets a second `<h1>`. That is correct for the LaTeX appendix case but may not be what you want on the web. Happy to gate it to specific part kinds if you would prefer.

alanlujan91:
Context and reproduction are in #2351.

PR #2355, opened 2025-10-18
🐛 Log errors on watch

rowanc1:


packages/myst-cli/src/build/site/watch.ts
```diff
@@ -161,8 +161,9 @@ function watchProcessor(
           reloadProject: true,
         });
       }
-    } catch (err: any) {
-      session.log.error(`Error during reload${err.message ? `:\n${err.message}` : ''}`);
+    } catch (error: any) {
+      session.log.debug(`\n\n${(error as Error)?.stack}\n\n`);
```
rowanc1, line 165: Now shows the stack trace on debug! Helpful to track down #2356

PR #2356, opened 2025-10-18
🐛 Undefined URLs can cause errors

rowanc1:


packages/myst-transforms/src/links/plugin.ts
```diff
@@ -20,7 +20,7 @@ export function formatLinkText(link: Link) {
   if (link.children?.length !== 1 || link.children[0].type !== 'text') return;
   const url = link.children[0].value;
   // Add an exception for wiki transforms links.
-  if (url.length < 20 || url.match(/\s/)) return;
+  if (!url || url.length < 20 || url.match(/\s/)) return;
```
rowanc1, line 23: This is picked up in a different warning, but was causing a hard crash here. Just protecting the `.length` here.

PR #2358, opened 2025-10-18
📖 Formatting bug in docs, and improve admonition page

rowanc1:


docs/admonitions.md
```diff
@@ -192,7 +192,7 @@ You can also use a {myst:directive}`dropdown` directive, which provides a more c
 
 ## Simpler Admonitions
 
-Admonitions can additionally be styled as `simple`, and can optionally hide the icon using the `icon` option of the `{myst:directive}`admonition.class`.
```
rowanc1, line 195: `` `{myst:directive}`admonition.class`. `` Was the main error here. (an extra `` ` `` at the start!)

docs/admonitions.md
```diff
@@ -202,7 +202,7 @@ This is a magic cat. It casts a luck spell on you that lasts an hour. \
 :::
 ```
 
-Removing the icon from an admonition of a certain class allows using a custom emoji for style. 
+Removing the icon from an admonition allows using a custom emoji for style.
```
rowanc1, line 205: `icon` is not a class, so cleaning up the language.

Issue #2374, opened 2025-10-27
Clarify "Use of the {doc} role is not recommended" message


mfisher87:
This message could use more context! It reads:

```
Usage of the {doc} role is not recommended, use a markdown link to the file instead.
```

This doesn't give me enough information to reproduce the functionality I was trying to achieve with the `{doc}` role: Printing a link to a document and automatically populating the dialog title as the link text.

## Proposal

Update the message to explain to the user that they need to leave the markdown link's text empty to auto-populate the document link.

For:

```markdown
{doc}`path/to/my/doc.md`
```

Print:

```
Usage of the {doc} role is not recommended. Instead, use a markdown link to the file with empty text: [](/path/to/my/doc.md)
```

stefanv:
Yes! Nice catch @mfisher87

rowanc1:
This is actually in the `note` field of the warning, which is currently only shown on debug — we could change that behaviour to show the note, which would solve this and put that on a new line in the console.

<img width="1612" height="432" alt="Image" src="https://github.com/user-attachments/assets/74db2879-fb8c-41e3-8b87-6f2a5476956f" />

rowanc1:
Actually, that note is printed in the terminal?

```
⚠️  index.md Usage of the {doc} role is not recommended, use a markdown link to the file instead.
   For {doc}`./supporting_information.md` use [](./supporting_information.md)
```

@mfisher87 are you also seeing that second line?

It does not have the line information, that should be added to make it easier to find!

- #2375

mfisher87:
😆 You're right, it's there. I just totally missed it when I opened this issue. It's low-contrast with the background in my terminal color scheme and I wonder how much of it was the contrast and how much of it was me just not reading carefully 🤣 

Line information _would_ be wonderful :)

Closing as not planned! Sorry for the noise y'all.

PR #2375, opened 2025-10-27
⚠️ Improve `{doc}` role warning with node position

rowanc1:
Adds position information to more easily find the node.

image

See #2374

cc @mfisher87


mfisher87:
:star_struck:

PR #2382, opened 2025-10-31
🗺️ Add support for HTTP caching

agoose77:
> [!Important]
> This doesn't patch the `fetch` we use for template downloads, so you'll need to pre-fetch those by running `myst init`.

> [!Warning]
> This fails on node 18.

I often have good stretches of time to work on MyST whilst I'm traveling. This coincides usually with a dearth of connectivity. In order to partially address this, and possibly speed up local MyST server rebuilds, this PR adds an HTTP cache.

> [!Note]
> I looked into using a local caching proxy using `squid`. Unfortunately, to cache HTTPS connections, we have to use TLS bumping, which effectively requires us to MITM attack the connection. This is unfortunate, and I also was not successful in getting the cache to actually work. In the end, the complexity of this approach (along with the fact that it requires an MITM attack _and_ there is seemingly only one decent tool for the job) made me discontinue this line of attack.

This PR replaces `node-fetch` with `undici`. Undici is actually built-in to Node.js, so we should be using it, but this PR lets us target a specific version.

The following env vars are useful:
1. `MYST_HTTP_CACHE_DB`
2. `MYST_HTTP_CACHE_DEFAULT_AGE`
3. `MYST_HTTP_CACHE_FORCE`

(1) ensures that we use a persistent rather than in-memory cache.
(2) lets us forcibly cache ambiguous responses that would not normally be cached.
(3) lets us forcibly cache explicit no-cache (`no-cache`, `max-age`, and `must-revalidate`) responses. Where this applies, we take the aforementioned `DEFAULT_AGE`.

This PR doesn't use a cache for local requests. We could keep the cache and rely on these servers setting `no-cache` properly, except that the `CACHE_FORCE` mode explicitly ignores these headers.

In theory the way to use this is to trigger an initial build and set all three of these variables (with a large default age). I want to give more thought to whether we can somehow prevent cache expiry, etc.

Long term I think this is the right move, and we might want to us it to rethink our manual caching strategy (i.e. invalidate derived caches when the HTTP response is no longer cached).

I'm sharing this now so that people can use it during travel time!


agoose77:
I think this is failing on node18 because of a ipv6-happy eyeballs bug that doesn't exist in newer versions of node. It might be that there is a version of node 18 that's fine, I'm not sure.

stefanv:
Hi @agoose77, thanks for exploring this caching approach!

Overriding our `fetch` seems like the right place to implement this. I also wonder:

- Do we know all the places that external fetches are made?
- How many of those categories of fetches are strictly necessary to make a build succeed? E.g., the theme is downloaded, but is cached on the filesystem, so doesn't strictly require it.
- Are there simpler alternatives, such as ignoring image fetches e.g., or is this caching strategy the best way to make builds work offline?
- Can we come up with a default strategy so that users don't have to set variables to make it work for the most common scenarios? (E.g., can the cache be a fallback even when the cache expired?)

I suspect your PR is the right way to go, I just thought it'd be worth exploring the context a bit. Offline builds is a feature I'd really like to see!

agoose77:
@stefanv these are the right questions!

I'm thinking about caching in two ways: one, we can speed up regular builds if the network is slower than the HTTP cache. I suspect this is normally true. However, we would still want a way to empty this cache.

Two, derived state can also be cached, where the computation is not free. We also want to keep this caching mechanism. 

It feels like we want an explicit 'prepare' step that does its best to populate the cache and download the remote assets. In theory, we can exclusively use a large http cache for this if necessary. I like the elegance of it, but I worry about how fragile it will feel (or won't feel). We particularly need to be mindful of the aggressive cache strategy that ignores maxAge and no-cache directives. To complicate things, sometimes we don't hit the network at the moment because our derived state cache has an entry. So, it's hard to populate the http cache ahead of time, unless we ignore derived cache altogether. 

Then, building in offline mode, we want to support the lack of a cache. This means we avoid blocking on fetches with big timeouts, either pulling from cache, or skip optional code paths.

I also want to share cache between just projects, but I don't know if that can just be the http cache.

agoose77:
Given our current prioritisation and review capacity, I'm closing this PR for now. It needs a bit more planning and work that we have space for atm. Thanks @stefanv for the review.

Here are my next steps:
1. Bump minimum compile-target to Node 18.
1. Drop `node-fetch` in favour of `undici` or built-in Fetch (with Node 18+)
1. Take learnings from the `fetch` requirements so that we can think about what `session`-like object we need in our plugins.

PR #2390, opened 2025-11-03
🪧  Add message for using Jupyter Book < 1 to upgrade

choldgraf:
This adds a little message in the logs for "discovering" a Jupyter Book 1 book. It gives the command to *not* upgrade in case people want to pin the version.

### DIFF
--- a/.changeset/tender-mirrors-buy.md
+++ b/.changeset/tender-mirrors-buy.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Add message for using Jupyter Book < 1 to upgrade


packages/myst-cli/src/init/init.ts
```diff
@@ -134,7 +134,11 @@ export async function init(session: ISession, opts: InitOptions) {
 ‣ Rename any modified or unneeded files so that they are hidden
 
 `),
-            `Are you willing to proceed?`,
+            // Just for a few months while people get used to JB2. Remove after Feb 2026.
+            chalk.dim(`If you prefer to continue using Jupyter Book 1.x, run: ${chalk.blue('pip install jupyter-book<2')}
```
choldgraf:
```suggestion
            chalk.dim(`If you prefer to continue using Jupyter Book 1.x, run: ${chalk.blue('pip install "jupyter-book<2"')}
```

PR #2397, opened 2025-11-04
🐧Fix image paths using `posix` path submodule

agoose77:
This is a slightly more robust way to do things.

@rowanc1 thanks for flagging this before we ran the tutorial.


agoose77:
Nope, this isn't right!

We internally use platform paths, and serialise to POSIX.

PR #2410, opened 2025-11-07
🧪 Add node 24 to ci testing

fwkoch:
@parmentelat pointed out that [we should probably be testing against node 24](https://github.com/jupyter-book/mystmd/pull/2344#issuecomment-3484749629). We can also probably boot 18, but that requires a slightly larger change of updating docs, etc.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['18', '20', '22.4.x']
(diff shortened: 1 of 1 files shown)


agoose77:
@fwkoch I'd :+1: to bump >18 and update our package.json — it's a blocker for #2382

fwkoch:
@agoose77 - I just don't have a great feeling for what will happen to users on 18. Will they get nice messaging to "update node" when they run the myst upgrade command we suggest in the console logs?

agoose77:
They should do! But, let's put that in a separate PR.

Issue #2405, opened 2025-11-06
When extends: doesn't resolve its target properly the error message is hard to parse


choldgraf:
If the extends: config points to a URL that doesn't resolve, we get a hard-to-parse error message. Things that we could improve:

- Human error message at the very bottom instead of the top
- Move the broken URL to the bottom as well so that it's easy to spot (`https://raw.githubusercontent.com/jupyter-book/jupyter-book/refs/heads/next/docs/site.yml`)
- Explicitly say this is the result of `extends:` not pointing to the right file.

```
Cache file not found: /home/runner/work/blog/blog/_build/cache/config-item-ef6fa4542542f1c9f23f1e258817ad8f.yml
Fetching: https://raw.githubusercontent.com/jupyter-book/jupyter-book/refs/heads/next/docs/site.yml
Bad response from config URL: https://raw.githubusercontent.com/jupyter-book/jupyter-book/refs/heads/next/docs/site.yml
/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291472
    throw Error(`Cannot find config file: ${file}`);
          ^

Error: Cannot find config file: /home/runner/work/blog/blog/_build/cache/config-item-ef6fa4542542f1c9f23f1e258817ad8f.yml
    at loadConfigYaml (/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291472:11)
    at getValidatedConfigsFromFile (/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291532:35)
    at /opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291567:60
    at Array.map (<anonymous>)
    at getValidatedConfigsFromFile (/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291559:79)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async loadConfig (/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291611:46)
    at async findCurrentProjectAndLoad (/opt/hostedtoolcache/node/20.19.5/x64/lib/node_modules/mystmd/dist/myst.cjs:291781:5)
[...]

PR #2424, opened 2025-11-10
Improve error message for `extends`

agoose77:
Addresses #2405


agoose77:
Working on this PR surfaced something I've noticed implicitly — our story for error handling (and probably the TUI more generally) could benefit with being clearer. 

e.g. we have
- `fileWarn` for pure transforms
- `session.log` and `addWarningForFile` for session transforms
- thrown exceptions

In general, the core of MyST should be very tolerant to errors and warn where possible. But there are things like "broken config" that aren't really ignorable. We need a nicer way to handle such situations and early exit

PR #2428, opened 2025-11-11
🌍↔️🐍 Limit the number of simultaneous executions

agahkarakuzu:
Following discussions on #2413 and #1831, this PR introduces `--execute-concurrency <n>` only. 

@agoose77 I tested this locally and it seems to be working. I did not implement an `if (execute) {` conditional in `process::site`, which as @fwkoch noted is not really needed. The only subtlety is the difference between the “effective” and “apparent” number of pages to execute, but users are free to set <n> as low as 1 if desired, noted in the docs. 

I hope I got the changeset right 👀

### DIFF
--- a/packages/mystmd/tests/serial-execution/first.md
+++ b/packages/mystmd/tests/serial-execution/first.md
@@ -18,7 +18,7 @@ import pathlib
 # 1. F°         F'--S°        S'--  (serial)
 # 2. --S°        S'--F°         F'  (serial)
 # 3. F°--S      F'--S'             (concur)
-# Below, each notebook (F°→F', and S°→S') touches their start marker, does some work, and touches the stop marker. We're testing for case (3)
+# Below, each notebook (F°→F', and S°→S') touches their start marker, does some work, and touches the stop marker. We're testing that case (3) does _not_ occur.
 ################
 
 this_start_path = pathlib.Path.cwd() / "start-first"


packages/myst-cli/src/process/site.ts
```diff
@@ -579,21 +582,33 @@ export async function processProject(
     ...projectParts,
   ];
   const usedImageExtensions = imageExtensions ?? WEB_IMAGE_EXTENSIONS;
-  // Transform all pages
+  
+  const concurrency = executeConcurrency ?? 5;
```
agoose77: By default, can we limit this to either `os.availableParallelism()` or `os.cpus().length`? I think the latter, likely smaller, probably makes more sense, because execution performance won't appear to differ according to whether the notebook is IO or CPU bound.

agahkarakuzu: Good idea! I set it to Math.max(1, cpus().length - 1); both in options and site, works on my end:

```
Options:
  --execute                  Execute Notebooks (default: false)
  --execute-concurrency <n>  Maximum number of notebooks to execute concurrently (default: 13)
  ```

I just reserved one core for main event loops.

Comment (stefanv) [reply packages/myst-cli/src/process/site.ts]: Do most users know what concurrency is? Perhaps a simpler name like `--execute-jobs` or just `--jobs` will do?

Comment (agahkarakuzu) [reply packages/myst-cli/src/process/site.ts]: Good point, maybe `--execute-parallel` would hit the sweet spot between clarity and accuracy?

Comment (agoose77) [reply packages/myst-cli/src/process/site.ts]: I think concurency and job are equally cryptic. Let's try this suggestion?

Comment (stefanv) [reply packages/myst-cli/src/process/site.ts]: That works, thanks.

Comment (agahkarakuzu) [reply packages/myst-cli/src/process/site.ts]: Great! Done in https://github.com/jupyter-book/mystmd/pull/2428/commits/7f08b9d9ba5e73091691ab42a334ea3f792eea1e.

File: .changeset/cute-aliens-follow.md [hunk 2]
```diff
@@ -0,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Enhancement: Limit simultaneous executions
```
bsipocz: I think we can and should be more verbose in the changelog. What about something like this?

```suggestion
Enhancement: Making it possible to configure the maximum number of simultaneous executions
```

docs/execute-notebooks.md
```diff
@@ -39,6 +39,12 @@ If you enable execution with the `--execute` flag as above, the following conten
 In order to execute your MyST content, you must install a Jupyter Server and the kernel needed to execute your code (e.g., the [IPython kernel](https://ipython.readthedocs.io/en/stable/), the [Xeus Python kernel](https://github.com/jupyter-xeus/xeus-python), or the [IRKernel](https://irkernel.github.io/).)
 :::
 
+## Limiting simultaneous executions
+
+By default, executable files are run concurrently in batches of 5.
```
bsipocz: Where is this 5 hard-wired?

agoose77: This is no longer the case!
```suggestion
By default, up to {math}`N-1` executable files are run concurrently, where {math}`N` is the number of available CPUs.
```

agahkarakuzu:
@agoose77 that's indeed a cleaner solution and paves the way better for an orchestration approach. I'll give it a stab.

agoose77:
What a star! :star:

agahkarakuzu:
@agoose77 I added `semaphore.runExclusive` at the `kernelExecutionTransform` level as you suggested, and it appears to work. In the example below, the first notebook that sleeps for 10 seconds is correctly blocking the rest of the executable content.

image

I am just not really sure if [this](https://github.com/agahkarakuzu/mystmd/blob/0e6df0df07d071eeb3d61ec6b259eab82791500b/packages/mystmd/src/clirun.ts#L34) is elegant or acceptable: 

```java
const opts = { ...program.opts(), ...this.opts() } as SessionOpts;
```

But I needed a way to pass `--execute-parallel` to this session.

docs/execute-notebooks.md
```diff
@@ -39,6 +39,12 @@ If you enable execution with the `--execute` flag as above, the following conten
 In order to execute your MyST content, you must install a Jupyter Server and the kernel needed to execute your code (e.g., the [IPython kernel](https://ipython.readthedocs.io/en/stable/), the [Xeus Python kernel](https://github.com/jupyter-xeus/xeus-python), or the [IRKernel](https://irkernel.github.io/).)
 :::
 
+## Limiting simultaneous executions
+
+By default, up to {math}`N-1` executable files are run concurrently, where {math}`N` is the number of available CPUs.
+
+You can change this by using the `--execute-parallel <n>` option in your build command, where `<n>` sets the maximum number of executable documents to run at once. For example, using `--execute-parallel 1` will run the documents one after another.
```
choldgraf: Can you add a short "this is useful when___" type of sentence so users know why they might want to do this?

packages/myst-cli/src/cli/options.ts
```diff
@@ -60,6 +61,13 @@ export function makeExecuteOption(description: string) {
   return new Option('--execute', description).default(false);
 }
 
+export function makeExecuteParallelOption() {
+  const defaultParallelism = Math.max(1, cpus().length - 1);
+  return new Option('--execute-parallel <n>', `Maximum number of notebooks to execute in parallel`)
+    .argParser(parseInt)
```
choldgraf: will this ensure that `--execute-parallel foo` doesn't work? Should we test for that?

choldgraf: What happens if `--execute-parallel 0`?

agahkarakuzu:
Thanks for the feedback @choldgraf, I’ve pushed updates to address your comments.

Regarding testing, we could add three notebooks that each wait for about five seconds, then assert that the total build time is at least fifteen seconds when `--execute-parallel 1`, and around 5 seconds when `--execute-parallel 3`? 

Since the transformations run in a non-deterministic order, it is not possible to create a test case where the output of one serves as the input to another.

choldgraf:
hmmm, could we make that more like .5 seconds each? I don't want to just auto-add 15 seconds to the test suite 😅

If nobody has better ideas for how to test this, I'd also be fine just leaving it and seeing if users complain about it or not

agoose77:
I think the fastest robust way to do this is to use concurrency primitives like barrier and mutex. The easiest thing is probably to create these in a python process, and share their pickled representations via the filesystem or env vars. This would require a small addition to the test harness. 

We can then ensure that 
A) only a single document can hold a lock at once. If the lock is busy, we fail. 
B) all files grab the lock before execution can proceed for any of them. If the barrier times out, we fail.

agoose77:
Testing this logic is a bit awkward because the implementation is at the end-to-end (E2E) level, rather than within the core execution package.

This makes sense from a design standpoint, because the concept of dependencies and ordering is a higher level one than "execute this file". Unfortunately, to then test this is fiddly because we need to do so in a mostly stateless way. 

For now, it's possible to implement some good-enough tests that use the filesystem and delays (although these are sensitive to kernel startup time). 

Once we have dependency ordering, it should simplify our tests but we might wish to implement a helper such as a socket server that enforces resource counting (e.g. ensure this lock counter never exceeds 1, or ensure this lock counter reaches 2).

agoose77:
I've run out of time to debug this further today, so I'd welcome anyone to take it over the line. I'm not immediately sure why the tests pass locally but fail on CI.

choldgraf:
A quick thought: if we're hitting test timeouts, perhaps the answer for now is just to increase the timeout? It could be that resource differences on a CI runner vs. a local machine are enough to change the test times. Then a future PR could optimize the test infra to allow for faster times in general...

bsipocz:
> A quick thought: if we're hitting test timeouts, perhaps the answer for now is just to increase the timeout?

Yes, please, especially if that would unlock getting in PRs such as this faster. I mean we're not talking about letting CI run for hours, so it shouldn't really matter anyway.

choldgraf:
OK I poked around a little bit and pushed a few changes that seemed to improve things locally, I broke it up into 3 commits. Let's see if that makes the tests happy.

EDIT: Pushed another commit to bump teh timeout of the E2E tests because I think those need to be raised at the same time as the execution-specific timeout, but we will see...I don't know why those would be taking so long so this is kind of a shot in the dark...

choldgraf:
Hey that finally made tests happy..man I wish it were easier to replicate hardware setups on CI.

I added comments to try and clarify how the test execution timeouts are related to one another because that was def not obvious haha. @agoose77 I want to give it a look through to make sure the changes are sensible?

agahkarakuzu:
Thank you so much @choldgraf for helping out with this! I look forward to a sanity check by @agoose77 to ensure that the semaphore implementation is as expected.

packages/mystmd/tests/serial-execution/first.md
```diff
@@ -0,0 +1,48 @@
...
+import pathlib
+
+# Ensure that new files are not created by a sibling process during execution
+# Let's assume `first` and `second` take about the same time. We can introduce a delay before and after `B`'s core routines'. This ensures that if first and second execute concurrently, they are well overlapped, and if they execute serially, the boundary effects are negligible
+# There are three possibilities, where `-` indicates sleep, `°` indicates start, and `'` indicates stop:
+#
+# 1. F°         F'--S°        S'--  (serial)
+# 2. --S°        S'--F°         F'  (serial)
+# 3. F°--S      F'--S'             (concur)
+# Below, each notebook (F°→F', and S°→S') touches their start marker, does some work, and touches the stop marker. We're testing for case (3)
```
agoose77:
```suggestion
# Below, each notebook (F°→F', and S°→S') touches their start marker, does some work, and touches the stop marker. We're testing that case (3) does _not_ occur.
```

agoose77:
LGTM. We'll likely revisit this when we change our extension interfaces, so I'm keen to ship something ASAP!

choldgraf:
@agoose77 I think you're the one to click the green button if you like - please do so if you're comfortable with the PR as is

choldgraf:
@agoose77 you're sending slightly conflicting signals because you've commented lgtm but your review still says it's requesting changes. So I'll assume your latest comment takes preference and merge this, thanks @agahkarakuzu !

PR #2436, opened 2025-11-13
Add support for iframe title for accessibility.

ryanlovett:
My institution uses Siteimprove to evaluate websites for accessibility compliance. It spotted that our iframes (for [embedded youtube videos](https://mystmd.org/guide/figures#youtube-videos)) were missing a "text alternative". `<iframe>` doesn't have an `alt` attribute, but it does use `title` for this purpose apparently.

 - https://help.siteimprove.com/support/solutions/articles/80001051799-accessibility-rule-inline-frame-without-a-text-alternative-explained
 - https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#accessibility

The content of the iframe directive is used as a caption. I'm not sure if it would be better to re-use the `caption.body` as the iframe title, re-use the iframe title for the caption, or let them be separate.

### DIFF
--- a/docs/figures.md
+++ b/docs/figures.md
@@ -226,7 +226,7 @@ Get up and running with MyST in Jupyter!
 :::
 ```
 
-You can find this URL when clicking share > embed on various platforms. You can also give the {myst:directive}`iframe` directive {myst:directive}`iframe.width` and a {myst:directive}`caption <iframe.body>`.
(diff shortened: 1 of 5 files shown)


ryanlovett:
I was planning on renaming the branch for the label check, but accidentally closed the PR. I made a new one.

Issue #2419, opened 2025-11-09
Plugin API improvements


agoose77:
We've hit an extensibility limit with the existing plugin APIs for MyST. In order to unblock people, we should build out greater functionality in our plugin APIs. Here are some unsorted ideas that we can ultimately refine.

yshavit:
How about addition of assets?

I'm trying to write a plugin that generates an SVG and imports it as `{"type": "html", ...}`. It's not working (the SVG seems to just get stripped away?), so I was thinking storing it as a separate asset might help.

I can store it as a new file and then have my role reference that as an image, but then I need to also gitignore it. If it were a separate asset, it could be more drop-in for users.

One way to do this might be to create an empty temp dir and then pass that path in via an env variable.

yshavit:
Another suggestion: pass in (via an env var?) the markdown file that called the plugin. That lets people specify args to the plugin in paths relative to the markdown.

agoose77:
> How about addition of assets?
> 
> I'm trying to write a plugin that generates an SVG and imports it as `{"type": "html", ...}`. It's not working (the SVG seems to just get stripped away?), so I was thinking storing it as a separate asset might help.
> 
> I can store it as a new file and then have my role reference that as an image, but then I need to also gitignore it. If it were a separate asset, it could be more drop-in for users.
> 
> One way to do this might be to create an empty temp dir and then pass that path in via an env variable.

This _should_ work. Perhaps you can share an example?

agoose77:
> Another suggestion: pass in (via an env var?) the markdown file that called the plugin. That lets people specify args to the plugin in paths relative to the markdown.

This is also possible via, cryptically, the `path` attribute of the `vfile` argument to transforms, roles, and directives.

yshavit:
> > How about addition of assets?
> > I'm trying to write a plugin that generates an SVG and imports it as `{"type": "html", ...}`. It's not working (the SVG seems to just get stripped away?), so I was thinking storing it as a separate asset might help.
> > I can store it as a new file and then have my role reference that as an image, but then I need to also gitignore it. If it were a separate asset, it could be more drop-in for users.
> > One way to do this might be to create an empty temp dir and then pass that path in via an env variable.
> 
> This _should_ work. Perhaps you can share an example?

Very simple one is at https://github.com/yshavit/mystmd-plugin-demo-svg

[...]

PR #2448, opened 2025-11-18
Lift plugin types to myst cli (refactoring) :arrow_heading_down: 

brianhawthorne:
1) In the interest of being able to add a Session instance into the plugin utils that get passed to a plugin transform, move the following types from `myst-common` into `myst-cli`:
* `PluginUtils`
* `PluginOptions`
* `TransformSpec`
* `MystPlugin`
* `ValidatedMystPlugin`

2) Organize plugin stuff in `myst-cli` into a proper `plugins` sub-package.

### DIFF
--- a/packages/myst-cli/src/build/utils/index.ts
+++ b/packages/myst-cli/src/build/utils/index.ts
@@ -5,3 +5,4 @@ export * from './getFileContent.js';
 export * from './localArticleExport.js';
 export * from './resolveAndLogErrors.js';
 export * from './bibtex.js';
+export * from './projectManifest.js';
--- a/packages/myst-cli/src/index.ts
+++ b/packages/myst-cli/src/index.ts
@@ -3,7 +3,7 @@ export * from './cli/index.js';
 export * from './config.js';
 export * from './init/index.js';
 export * from './frontmatter.js';
-export * from './plugins.js';
+export * from './plugins/index.js';
 export * from './process/index.js';
 export * from './project/index.js';
 export * from './session/index.js';
--- a/packages/myst-cli/src/plugins/executable.ts
+++ b/packages/myst-cli/src/plugins/executable.ts
@@ -1,12 +1,7 @@
-import type { ISession } from './session/types.js';
-import {
-  type MystPlugin,
-  type DirectiveSpec,
-  type RoleSpec,
-  type TransformSpec,
-  type GenericNode,
-} from 'myst-common';
 import { spawn, spawnSync } from 'node:child_process';
+import type { ISession } from '../session/types.js';
+import type { DirectiveSpec, RoleSpec, GenericNode, } from 'myst-common';
+import type {MystPlugin, TransformSpec } from './types.js';
 
 type DirectiveJSONSpec = Omit<DirectiveSpec, 'run' | 'validate'>;
 type RoleJSONSpec = Omit<RoleSpec, 'run' | 'validate'>;
--- a/packages/myst-cli/src/plugins/index.ts
+++ b/packages/myst-cli/src/plugins/index.ts
@@ -1,0 +1,3 @@
+export * from './load.js';
+export * from './executable.js';
+export * from './types.js';
--- a/packages/myst-cli/src/plugins/load.ts
+++ b/packages/myst-cli/src/plugins/load.ts
@@ -1,10 +1,11 @@
 import fs from 'node:fs';
 import { pathToFileURL } from 'node:url';
-import type { ISession } from './session/types.js';
-import { RuleId, plural, type MystPlugin, type ValidatedMystPlugin } from 'myst-common';
+import { RuleId, plural } from 'myst-common';
 import type { PluginInfo } from 'myst-config';
-import { addWarningForFile } from './utils/addWarningForFile.js';
-import { loadExecutablePlugin } from './executablePlugin.js';
+import type { ISession } from '../session/index.js';
+import { addWarningForFile } from '../utils/addWarningForFile.js';
+import { loadExecutablePlugin } from './executable.js';
+import type { MystPlugin, ValidatedMystPlugin } from './types.js';
 
 /**
  * Load user-defined plugin modules declared in the project frontmatter
--- a/packages/myst-cli/src/plugins/types.ts
+++ b/packages/myst-cli/src/plugins/types.ts
@@ -1,0 +1,40 @@
+import type { Plugin } from 'unified';
+import type { GenericNode, GenericParent, DirectiveSpec, RoleSpec } from 'myst-common';
+
+type Select = (selector: string, tree?: GenericParent) => GenericNode | null;
+type SelectAll = (selector: string, tree?: GenericParent) => GenericNode[] | null;
+
+export type PluginUtils = { select: Select; selectAll: SelectAll };
+export type PluginOptions = Record<string, any>;
+
+export type TransformSpec = {
+  name: string;
+  doc?: string;
+  stage: 'document' | 'project';
+  // context?: 'tex' | 'docx' | 'jats' | 'typst' | 'site';
+  plugin: Plugin<
+    [PluginOptions | undefined, PluginUtils],
+    GenericParent,
+    GenericParent | Promise<GenericParent>
+  >;
+};
+
+/**
+ * Create MyST plugins that export this from a file,
+ * or combine multiple plugins to a single object.
+ */
+export type MystPlugin = {
+  name?: string;
+  author?: string;
+  license?: string;
+  directives?: DirectiveSpec[];
+  roles?: RoleSpec[];
+  transforms?: TransformSpec[];
+};
+
+export type ValidatedMystPlugin = Required<
+  Pick<MystPlugin, 'directives' | 'roles' | 'transforms'>
+> & {
+  paths: string[];
+};
+
--- a/packages/myst-cli/src/process/mdast.ts
+++ b/packages/myst-cli/src/process/mdast.ts
@@ -1,6 +1,6 @@
 import path from 'node:path';
 import { tic } from 'myst-cli-utils';
-import type { GenericParent, IExpressionResult, PluginUtils, References } from 'myst-common';
+import type { GenericParent, IExpressionResult, References } from 'myst-common';
 import { fileError, fileWarn, RuleId, slugToUrl } from 'myst-common';
 import type { PageFrontmatter } from 'myst-frontmatter';
 import { SourceFileKind } from 'myst-spec-ext';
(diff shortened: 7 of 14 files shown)


stevejpurves:
@brianhawthorne @fwkoch great to see work on this! 

What is the motivation for moving types out of `myst-common` and into `myst-cli` here? `myst-cli` is a node dependent package, and I am always hesitant to move code out of a general JS package and into one with node dependencies, even if it is just types. (node deps. can really mess with browser only bundlers)

(I get that there might be a desire to group the types and code so that relevant changes are more contained, but the monorepo solves that anyways and all packages change version in step)


BUT having said that, is the idea that the `plugins` subpackage could be split out to a package in it's own right and one with no `node:` dependencies? just thinking that then that would make plugins accessible to general JS contexts, e.g. imagine being able add plugins to the myst-demo and have them parsing in the browser. If plugins remain tightly coupled to `myst-cli` it removes the potential for browser only plugins.

agoose77:
@brianhawthorne to add to @stevejpurves' points, which are useful in setting some of the context for how we think about the internal design of MyST — I'd like us to bring together people looking to build plugins in order to plan how we build out these APIs. You're entirely correct that it's not easy to do many things at the moment because of the limited context available to plugins.

@stevejpurves mentions the browser, and it's a pertinent question. The `jupyterlab-myst` extension (due some love) is currently single-document. We need to build out the multi-document features, and there are several ways that this could happen. We could lift as much "engine" responsibility into an abstraction, and rebuild the abstraction for JupyterLab. Or, we could connect the JupyterLab renderer with a server-like backend. Both options have merits, though the latter won't easily deploy to JupyterLite.

@brianhawthorne are you coming to the collaboration cafe today?

stefanv:
Maybe we can start with a simpler question: is it reasonable to expose the Session as part of the plugin interface? It will solve one of Brian's problems, so it would be good to know if there's any reason we can't do that.

Brian, why are these type moves required to make that happen? 

And then just noting that Steve's suggested plugins packages grouping may also help simplify things.

fwkoch:
I agree with Steve's comments about the types - `myst-cli` is a bit of a mess (in addition to it's `node` dependence); anything we can keep in other narrow-scoped, lighter-weight packages is better for reuse and separation of concerns.

@stefanv - regarding exposing `Session` to plugins: I agree this would enable a ton of things. The main reason _not_ to do this is the `Session` object _will_ change, and we don't really want these internal changes to be breaking. (Also, the session object is overly complex, since it's just grown organically during early development.) This means early adopters who write plugins will (1) have a challenging time deciphering how to use the `session`, and then (2) their plugins will more-likely-than-not break relatively soon.

I certainly think there's an argument for not worrying about this and exposing the `session` anyway; it's better to enable people to build cool things. (And there's not really any technical blocker.) But I think we need to be really clear in our messaging so early adopters don't feel resentful when things change.

If we aren't exposing the session, we need to figure out a way to quickly iterate on requests like "I need X for my plugin" - do we continue to expand `pluginUtils` https://github.com/jupyter-book/mystmd/blob/main/packages/myst-cli/src/process/mdast.ts#L88 (and risk creating something as hacked-together as the current session)? Or we could start building a simpler "pseudo-Session" (which could maybe even evolve into the next version of the actual `Session` MyST uses everywhere)?

brianhawthorne:
Thanks for all the feedback on this!

> What is the motivation for moving types out of myst-common and into myst-cli here?

My motivation is to give plugin transforms access to the session, which would involve changing this:
```typescript
export type PluginUtils = { select: Select; selectAll: SelectAll };
```
to this:
```typescript
export type PluginUtils = { select: Select; selectAll: SelectAll; session: ISession };
```

But `ISession` is defined in `myst-cli` and `PluginUtils` is defined in `myst-common`, which `myst-cli` depends on.  So we don't want to create a dependency cycle by adding the reverse dependence.

I agree moving these types out of `myst-common` feels like "petting the cat in the wrong direction".  I also hadn't considered limiting the radius of node dependence - that makes sense.

An alternative might be to go the other way and push the `ISession` declaration down into `myst-common`.  But it requires types from `redux` and `@jupyterlab/services`, and those seem like pretty heavyweight dependencies we wouldn't want to add to `myst-common`.

We could also ignore typing and do this:
```typescript
export type PluginUtils = { select: Select; selectAll: SelectAll; session: any };
```
That feels lazy but it would unblock me :grin:

stefanv:
Could we do the selective exposure of some parts of Session (only the parts we know are stable) into a new type, that we can cast a Session into?

Is this the Session we are talking about?

<details>

```md
- API_URL: string
- configFiles: string[]
- store: Store<RootState>
- $logger: Logger
- doiLimiter: Limit
- proxyAgent?: HttpsProxyAgent<string>
- _shownUpgrade: boolean
- _latestVersion?: string
- _jupyterSessionManagerPromise?: Promise<SessionManager | undefined>
- get log(): Logger
- constructor(opts?: { logger?: Logger; doiLimiter?: Limit; configFiles?: string[] })
- showUpgradeNotice(): void
- reload(): Promise<this>
- fetch(url: URL | RequestInfo, init?: RequestInit): Promise<Response>
- plugins?: ValidatedMystPlugin
- loadPlugins(plugins: PluginInfo[]): Promise<ValidatedMystPlugin | undefined>
- sourcePath(): string
- buildPath(): string
- sitePath(): string
- contentPath(): string
- publicPath(): string
- _clones: ISession[]
- clone(): Promise<ISession>
- getAllWarnings(ruleId: RuleId): (BuildWarning & { file: string })[]
- jupyterSessionManager(): Promise<SessionManager | undefined>
- private createJupyterSessionManager(): Promise<SessionManager | undefined>
- dispose(): void
```

Curious which part of that is needed.

</details>

Can someone also explain the server side rendering consideration to me?

brianhawthorne:
@stefanv 
> Curious which part of that is needed.

For my use case, potentially almost _all_ of it, since I'd be using it to parse an arbitrary chunk of myst markdown, which could end up invoking almost any part of the system's functionality (excepting the framing stuff like writing build artifacts to `_build` or running a content server).

brianhawthorne:
@fwkoch 
> Or we could start building a simpler "pseudo-Session"

This an intriguing idea.

agoose77:
### Responses
> Maybe we can start with a simpler question: is it reasonable to expose the Session as part of the plugin interface? It will solve one of Brian's problems, so it would be good to know if there's any reason we can't do that.

I don't think so (which was what prompted the topic at our JH/JB cafe)!

For example, we already know that our execution abstraction will change with future work on ordering, and there will be other parts too.

> Could we do the selective exposure of some parts of Session (only the parts we know are stable) into a new type, that we can cast a Session into?

This is where I was moving towards. Either we design an aspirational API, or we built iteratively an API that _will be_ broken down the road. Both are fine if we message appropriately.

> since I'd be using it to parse an arbitrary chunk of myst markdown, 

We do have `parseMyst` but it sounds like you also want to run subsequent transforms on it too? This requires a bit more work and care than e.g. making referencing state available to plugins.

### Next steps
> (which could maybe even evolve into the next version of the actual Session MyST uses everywhere)?

This is my hunch. The Sphinx API has two rough parts:
- The core application state
- The hooks into that state

Our core state is much less nice to work with (@fwkoch it's fair to say?) and our hooks are fewer. https://github.com/jupyter-book/mystmd/issues/2419 is a small initiative to improve the hooks side, but I think this thread is much more about the kind of state that a plugin can work with, which is the core.

It feels to me like we need to spend some async/sync cycles on figuring out what state needs to be accessible, and what kinds of things need to be possible, so that we can either:

1. Build an interface over our existing state that is nicer to work with than the session
2. Rework the session etc. directly

(1) is "easier" in many ways, but doesn't deal with the problem. (2) is "harder" but likely also makes the code easier to work with, too.

> [!Important]
> Do @jupyter-book/core-team agree that it feels like the time for some design work rather than piecemeal iterative development here?  
My feeling is yes — the API is "more important" than the internal implementation details to get right(ish) first time, _and_ has knock-on consequences for our internals.

stevejpurves:
@brianhawthorne are you able to share more on your actual use case? maybe you have already (in an issue) somewhere?

>For my use case, potentially almost all of it, since I'd be using it to parse an arbitrary chunk of myst markdown, which could end up invoking almost any part of the system's functionality (excepting the framing stuff like writing build artifacts to _build or running a content server).

Just curious whether session exposure is the only solution

brianhawthorne:
> @brianhawthorne are you able to share more on your actual use case? maybe you have already (in an issue) somewhere?

Yes (and no issue yet).  I'm trying to implement a template directive in an external plugin that will render an inline chunk of nunjucks template into (myst) markdown for inclusion in the document (working name of `{nunjucks}`, but could be `{template}` or something).  Almost identical to the built-in `{include}` directive, but instead of including a chunk of markdown read in from a file it would include the chunk produced by rendering an inline nunjucks template, very similar to Curvenote's `{template:listing}` directive.

Eg, a team gallery rendered from team member info in a local data file:

`team.md`:
```jinja
...
"""{nunjucks} team.toml
{% for m in members %}
:::{card}
  :footer: {{ m.name }}
  :link: {{ m.github }}
  ![{{ m.name }}]{{ m.avatar }})
:::
{% endfor %}
"""
...
```

`team.toml`:
```toml
[[members]]
name = "Jake Bowhay"
github = "https://github.com/j-bowhay"
avatar = "https://avatars.githubusercontent.com/u/60778417?u=6272842e4c0946308604a5f54cdaa8e6505adcbe&v=4"

[[members]]
name = "Eric Larson"
github = "https://github.com/larsoner"
avatar = "https://avatars.githubusercontent.com/u/2365790?v=4"

etc...
```

So what I really need is a `parseMarkdown` function, _not_ direct access to the whole session object.  The built-in `{include}` directive is able to parse the included markdown because its transform is given a bespoke parsing function among its options (deviating from the declared `PluginUtils` options type).

It's clear from this discussion that stuffing session into `PluginUtils` is a nonstarter, so I'm going to close this PR.  Thanks everyone for helping hash this out!

brianhawthorne:
Closing as a "won't do".  This is not the way.

akhmerov:
The application seems cool! Right now, in principle you could generate the AST directly in the plugin, right? This is less convenient for a user, but works if you only have a few custom listings.

rowanc1:
If you are parsing the toml in the directive run function (not in a transform), then you do have access to the `ctx` as the third (currently undocumented) argument that has a `parseMyst` command:

https://github.com/executablebooks/mystmd/blob/main/packages/myst-common/src/types.ts#L103

Used as an example in the CSV table importer (which is pretty similar to what you are trying I think):
https://github.com/executablebooks/mystmd/blob/main/packages/myst-directives/src/table.ts#L189

brianhawthorne:
> you do have access to the ctx as the third (currently undocumented) argument that has a parseMyst command

Excellent - thank you, @rowanc1 !!  :pray:  That's the link I was missing.  I had explored possible extra args to the directive `run` function, just to see what was getting passed in; but to see the values I used `console.log`, which prints "{}" for the third arg, so I thought it was being passed an empty object :dizzy_face: .  I just tried `console.dir` instead and got "{ parseMyst: [Function: parseMyst] }".  Perfect!

PR #2453, opened 2025-11-19
♻ Revert skip execution changes

stefanv:


stefanv:
More nuanced solution in https://github.com/jupyter-book/mystmd/pull/2454

Issue #1040, opened 2024-03-29
Expose raw html / `allowDangerousHtml` to the MyST CLI


choldgraf:
In the MyST parser library, there's an option called `allowDangerousHtml` that allows the parser to include "raw" HTML that isn't a formal MyST node. This is mentioned on this page:

https://mystmd.org/guide/commonmark#valid-html

![CleanShot 2024-03-29 at 09 01 17@2x](https://github.com/executablebooks/mystmd/assets/1839645/8d294b2e-94af-4262-b58a-e40a291815b3)

However, it is unclear in the documentation how to set `allowDangerousHtml`, and I think it is actually not possible from a MyST configuration + CLI system, but only at the command line.

It would be useful if this could be exposed to users so that they can embed their own HTML, JS, CSS, etc. This is useful if you want to do things like:

- Embed your own styles on a page
- Embed the HTML embed code of a widget from some other service
- Generate some custom structure that you don't know how to accomplish with MyST
- Add some JavaScript to give the page custom behavior.

### Where is this implemented?

This is an educated guess, but I believe that the `myst-transform` that builds the HTML outputs doesn't use this parameter and thus this functionality is not exposed to users:

https://github.com/executablebooks/mystmd/blob/ffb239a921ee90aa8e55c067e243e4213767c6fe/packages/myst-transforms/src/html.ts#L160-L180

lorenzo-rovigatti:
I'd love to have this feature, as I currently cannot use myst for my project because I cannot find a way of adding raw HTML to the page :(

choldgraf:
### Could this be done with a `{raw:html}` directive?

In this PR we added support for `{raw:typst}` and `{raw:latex}` directives:

- https://github.com/jupyter-book/mystmd/pull/1442

Could that be extended to allow for `{raw:html}`?

choldgraf:
Another use case for this that I'd like to start doing more of: embedding social media posts.

BlueSky has a nice little [web app to generate post embed scripts](https://embed.bsky.app/) that create nicer-looking previews. 

```
[...]

PR #2462, opened 2025-11-21
Update documentation for MyST HTML option

akhmerov:
Clarify the requirement for setting 'allowDangerousHtml' in the MyST parser. Knowing that this isn't accessible to a user would have saved me some time digging.

### DIFF
--- a/docs/commonmark.md
+++ b/docs/commonmark.md
@@ -137,6 +137,7 @@ any _text_
 ### Valid HTML
 
 Any valid HTML may also be included in a MyST document and rendered to HTML. However, you must set the option `allowDangerousHtml: true` in the MyST parser.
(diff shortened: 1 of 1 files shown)


stefanv:
@agoose77 @choldgraf This is something we discussed earlier in the week; what was the mechanism decided on to do it? A raw directive?

akhmerov:
The relevant issue is: https://github.com/jupyter-book/mystmd/issues/1040. I don't remember hearing a definite answer on discord or during the collaboration cafe, so I'm not sure there's a final decision.

I think making docs a bit less confusing is a useful step while there's no clear way forward.

stefanv:
Agreed, and approved.

PR #2463, opened 2025-11-21
Document how to use options with dashes in their names

stefanv:
Also add missing `file` type

I removed the recommendation to use underscores, because (a) that goes against what typst uses itself and (b) is also not a common pattern in YAML. So, I think we can let template authors decide for themselves if they want to deal with the dashes or not?


choldgraf (approved):

choldgraf:
that em-dash hack looks complicated! 😅

stefanv:
Ah, but it's not an em-dash 😅

PR #2468, opened 2025-11-22
Update site options docs for clarity

choldgraf:


docs/website-templates.md
```diff
@@ -96,14 +96,15 @@ site:
 
 ### Page Options
 
-Depending on the option, these can also be controlled in the frontmatter on each page under the `site` key.
+Any option from `site.options` can also be set per-page in the frontmatter under the `site` key.
```
stefanv:
```suggestion
Any option from `site.options` can be overridden per-page in the frontmatter under the `site` key.
```

And then something like:

So, if you have, in `myst.yml`:

```
site:
  options:
    hide_toc: true
```

Override that in your page as:

...

PR #2482, opened 2025-11-24
add support for rendering mermaid files to typst output

AbelHeinsbroek:
I've added a Typst handler for the mermaid directive based on [mermaid-cli](https://github.com/mermaid-js/mermaid-cli)


AbelHeinsbroek:
Nevermind, I just found PR #1452

PR #2492, opened 2025-11-25
🔑 Support `keys` for DOI links

### LINKED ISSUE #2489: Support keys for the `doi-link-valid` error rule
## Proposal

Currently, I can control the `doi-link-valid` error rule by adjusting the level of _all_ errors that fall under that rule. I only want to handle cases where I'm trying to link to a webpage on DOI.org, not to an actual DOI (https://github.com/jupyter-book/mystmd/issues/828).

Rowan was providing some support and found that this rule doesn't support keys:

https://github.com/jupyter-book/mystmd/blob/23bf540110b08d42bc9b049d9479562d3f750bcd/packages/myst-transforms/src/links/doi.ts#L29

I'd like to use a key that corresponds exactly to the URL I'm using in my code, e.g.:

```yaml
  error_rules:
    - rule: "doi-link-valid"
      key: "https://doi.org/the-identifier/what-is-a-doi/"
      severity: "warn"
```

i.e. I would prefer that the key _not_ look like `doi:the-identifier/what-is-a-doi/`!

## Additional notes

Thanks, y'all are awesome ;)

rowanc1:
Fixes #2489

image

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -102,6 +102,8 @@ project:
 
 The `severity` of each rule can be set to `ignore`, `warn`, or `error`. If the rule is triggered, then the severity listed will be adopted rather than the default log message severity. The default severity for rules included in the list is `ignore`, which means that simply listing the rule IDs as strings will ignore those rules. To discover the rule ID, run myst in debug mode to get the error (and optional key) printed to the console. For example, the above configuration updates will no longer warn on `math-eqnarray-replaced` and will also ignore the two links when running `myst build --check-links --strict` in continuous integration.
 
+Some error rules support a `key` field that identifies specific instances of the error. This allows you to target particular cases rather than all instances of a rule. For example, the `link-resolves` rule uses the URL as the key, allowing you to ignore specific broken links while still checking others. Similarly, the `doi-link-valid` rule uses the DOI value as the key, so you can ignore specific invalid DOIs while still validating others. When a rule supports keys, you can provide a list of keys (or key patterns) in the `keys` field to match multiple specific instances.
+
 :::{seealso .dropdown} List of Error Rules
 
 The full list of errors and warnings used across MyST with their defaults shown.


stefanv:
LGTM. Shall we update the docs accordingly?

rowanc1:
Where in the docs need updating?

stefanv:
The docs don't explain `key` at all: https://mystmd.org/guide/settings#error-rules

Maybe the example is enough, but one sentence of description might be helpful.

Also, this is labeled as enhancement, so presumed this is a new feature, but looks like a bugfix maybe.

mfisher87:
Thanks Rowan! :star_struck:

packages/myst-cli/src/transforms/dois.ts
```diff
@@ -212,6 +213,7 @@ export async function getCitation(
         node,
         ruleId: RuleId.doiLinkValid,
         note: `Citation data from ${doiString}:\n\n${JSON.stringify(data)}\n`,
+        key: doiString,
       },
```
rowanc1, line 217: @stefanv this is the "enhancement" part! Adding "key" to the `doi-link-valid`.

packages/myst-common/src/utils.ts
```diff
@@ -13,12 +13,15 @@ export type MessageInfo = {
...
+  /** This key can be combined with the ruleId to suppress a warning */
+  key?: string;
 };
 
-function addMessageInfo(message: VFileMessage, info?: MessageInfo) {
+function addMessageInfo(message: VFileMessage & { key?: string }, info?: MessageInfo) {
   if (info?.note) message.note = info.note;
   if (info?.url) message.url = info.url;
   if (info?.ruleId) message.ruleId = info.ruleId as string;
+  if (info?.key) message.key = info.key;
```
rowanc1, line 24: This is an infra-level change, which allows for other `fileError` and `fileWarn` to also pass keys. (also an enhancement!)

docs/settings.md
```diff
@@ -102,6 +102,8 @@ project:
...
 
+Some error rules support a `key` field that identifies specific instances of the error. This allows you to target particular cases rather than all instances of a rule. For example, the `link-resolves` rule uses the URL as the key, allowing you to ignore specific broken links while still checking others. Similarly, the `doi-link-valid` rule uses the DOI value as the key, so you can ignore specific invalid DOIs while still validating others. When a rule supports keys, you can provide a list of keys (or key patterns) in the `keys` field to match multiple specific instances.
```
rowanc1, line 105: Added some docs!

rowanc1 (commented):
@stefanv some improvements, reasoning for "enhancement" (I don't think this was a bug fix, other than conceptually it should have already been there...) and some docs.

docs/settings.md
stefanv: Perfect, thanks very much!

stefanv (approved):

docs/settings.md
mfisher87: :100: 

TIL you can list keys. That will make config for one of my sites much simpler :)

mfisher87: >or key patterns

TIL again ;)

It'd be great to have a bit more detail about key patterns. Is it a full regex syntax, full glob syntax, or only `*` that's supported?

docs/settings.md
rowanc1, line 126: @mfisher87 patter matching info is right below!

mfisher87: Thanks, Rowan! :dancers:

PR #2494, opened 2025-11-25
Provide explicit debug command in error rules documentation

mfisher87:
My first time reading this doc, I struggled a bit because I assumed `--debug` would be e.g. an argument to `myst build`, and it didn't show up in `myst build --help`. I figured it out soon after, but I figure for folks less used to CLIs it would be good to explicitly provide an example.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -100,7 +100,7 @@ project:
         - 'https://*.example.com/**'
 ```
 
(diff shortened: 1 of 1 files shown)


stefanv:
Thanks, Matt!

mfisher87:
Thanks, Stefan :heart:

PR #2498, opened 2025-11-26
🌍 Use `undici` instead of `node-fetch`

agoose77:
> [!Important]
> This doesn't add `undici` to `myst-to-docx`. I think right now that package will work in a browser context, so we'll want to spend a little bit of time injecting that dependency rather than importing top level. A future PR.

This PR replaces `node-fetch` with `undici`. Undici is actually built-in to Node.js, so we should be using it, but this PR lets us target a specific version.

The intention is to make it easier to add support for features like caching in the future.


agoose77:
I'm closing this. We should perform this kind of systematic maintenance work, but I want to focus on nearer-term cleanups for now and we don't have the review bandwidth.

PR #2499, opened 2025-11-26
🤖 Update our TS config and bundler versions

agoose77:
## Summary

> [!Note]
> This PR description is long because I'm self-documenting my understanding of this very complex web of things to worry about. The PR itself is tiny.

We recently bumped the minimum CI version of Node.js to Node 18, and we have long required Node 18 in our `mystmd` Python package.

This PR attempts to enforce consistency with this internally:

- Node.js minimum version 18.0.0
- TypeScript minimum version 5.9.3
- ESM target ES2022
- Module Resolution: node16

This is a bit confusing, and I'm no expert. I've added some detail in the file-view as review comments.

## Why?
This PR helps us with the following:
1. Catch bugs like "this only works in browser" (e.g. `myst-to-docx`) by being stricter about the types we expect -- this PR turns off DOM types. 
2. Bring us in line with our own declared runtime minimum version of Node.js 18
3. Permit us to leverage newer features from ES2022 (vs ES2021).

### Choosing `target`
I think what this PR is explicitly surfacing is that we may require some people to use a transpiler/bundler to use `mystmd` packages in the browser, if they want need browser support than ES2022. I think that's OK. 

### Disabling `DOM`
Disabling the DOM is a quick way to catch usages of browser specific APIs. It does make writing platform-specific code harder, though, e.g. different implementations that target browser vs node. 

I think when it comes to supporting multiple runtimes, you have two options:
1. Implement your library in such a way that the caller passes in runtime adaptors, e.g. "how to write to the filesystem"
2. Define a browser-specific entry point to your application via `browser`. 

When it comes to typing, I'm pretty sure that TypeScript generally does not support isomorphic APIs, so (2) is somewhat hard unless your external API type signature is consistent for all platforms. In light of that, the preferred way to do this is probably to have three packages:
- `foo`
- `foo-web-adaptor`
- `foo-node-adaptor`

where the (external) type information varies between `foo-web-adaptor` and `foo-node-adaptor`. In this PR I'm not worrying about that, which just means that our type hints may sometimes be slightly wrong (`Buffer | Blob` is only `Buffer` for Node, e.g.)

## Enforcing Cross-Platform APIs
I don't think it's easy/possible to tell TypeScript to ban e.g. `node:fs` imports. I think only bundlers handle that at the moment, and we don't use bundlers in our `packages/`. So, whilst we can prevent the built-in DOM types from being available, there's nothing preventing us from importing `node:` modules

### DIFF
--- a/packages/myst-to-docx/src/utils.ts
+++ b/packages/myst-to-docx/src/utils.ts
@@ -86,13 +86,18 @@ export function getImageWidth(width?: number | string, maxWidth = MAX_DOCX_IMAGE
   return (lineWidth / 100) * maxWidth;
 }
 
+/**
+ * Get the dimensions of an image using the browser's `Image` object.
+ * @param file - The image file.
+ * @returns The dimensions of the image.
+ * @note This is a browser only function, server side this function should return undefined
+ * and the `buffer-image-size` package should be used to get the dimensions.
+ */
 async function getImageDimensions(
   file: Blob | Buffer,
 ): Promise<{ width: number; height: number } | undefined> {
   // @ts-expect-error Image is not defined
-  if (typeof Image === 'undefined') {
-    return Promise.resolve(undefined);
-  }
+  if (typeof Image === 'undefined') return undefined;
   return new Promise((resolve, reject) => {
     // @ts-expect-error Image not defined
     const img = new Image();


packages/tsconfig/base.json
```diff
@@ -5,20 +5,20 @@
     "composite": false,
     "declaration": true,
     "declarationMap": true,
-    "target": "ES2019",
+    "target": "es2022",
```
agoose77, line 8: `target` defines what kind of syntax TypeScript transpile to. In this case, ES2022. 

This is not supported by _all_ browsers, but given that all of our packages have dependencies, this usually doesn't matter -- you'll probably use a bundler _anyway_, and bundlers normally support transpiling. 

We don't want to go higher than ES2022, because Node.JS supports a particular version (or older), and that's ES2022 for Node.18. So that defines `<=ES2022` as a bound here, and we might as well aim as high as possible.

packages/tsconfig/base.json
```diff
@@ -5,20 +5,20 @@
...
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "inlineSources": false,
     "isolatedModules": true,
-    "module": "Node16",
-    "moduleResolution": "Node16",
+    "lib": ["es2022"],
```
agoose77, line 13: `lib` defines an escape hatch to tweak `target`, in the event that you customise your runtime environment. Here, I duplicate "es2022" deliberately, and remove DOM types -- it seems that the default, implied by `target`, pulls in the `full` variant that includes the DOM types. I *think* that we *don't* want DOM types here, because we dual publish our packages to the web. 

For example, if you target version `X` of JS, but want to use newer code features that only exist in version `Y` of JS (`X < Y`), you can ask TypeScript to allow these types by adding them to `lib`. This doesn't actually add the runtime code to support this, though, so you will need to add the polyfills to your runtime library. For example, supporting `Promise` in old versions of JS. As we don't need the DOM in `mystmd`, we disable those types.

packages/tsconfig/base.json
```diff
@@ -5,20 +5,20 @@
...
+    "module": "node18",
+    "moduleResolution": "node16",
```
agoose77, line 15: `module` and `moduleresolution` — `module` defines the kind of output module type (ESM, CJS, etc) we support. `moduleresolution` defines how imports are resolved. This is pretty gnarly, but I'm fairly confident that to target _both_ browser and Node.js, we need `nodenext` `nodenext`, or equivalent - in this case, `node18` and `node16` respectively.

packages/myst-to-jats/src/transforms/containers.ts
```diff
@@ -1,5 +1,5 @@
 import type { Plugin } from 'unified';
-import type { Blockquote, Legend, Caption, FlowContent } from 'myst-spec';
+import type { Node, Blockquote, Legend, Caption, FlowContent } from 'myst-spec';
```
agoose77, line 2: I think this was a bug.

packages/myst-to-docx/src/utils.ts
```diff
@@ -86,8 +86,14 @@ export function getImageWidth(width?: number | string, maxWidth = MAX_DOCX_IMAGE
   return (lineWidth / 100) * maxWidth;
 }
 
-async function getImageDimensions(file: Blob | Buffer): Promise<{ width: number; height: number }> {
+async function getImageDimensions(
```
agoose77, line 96: This is a browser-specific implementation. I don't have the time to gracefully polyfill this on Node.js, but I think that's OK -- it's only used in the web atm anyway, so I make the dimension field a noop on Node.js.

agoose77:
@rowanc1 I've pinged you for review because this affects a fairly low level part of the stack (tsconfig).

Do you have capacity to:
1. Validate my decisions around the various tsconfig fields
2. Cast your eye to the "conversation" around packaging for Node and Browser, to validate my understanding matches yours?

Thanks!

packages/myst-to-jats/src/transforms/containers.ts
rowanc1: I have picked this up in #2541.

rowanc1:
I think everything makes sense. Pulled the type bug into a different PR and added some more docs around the browser-only word stuff. Works locally and in CI - we can merge and see what happens...!

Thanks for the comprehensive notes on the PR.

PR #2508, opened 2025-11-27
Add documentation to `{myst}` directive

mfisher87:
Hooray, documentation! :)


mfisher87:
I think since this only affects the documentation, there's no need for a changeset, correct?

agoose77:
And technically, we do want a changeset for this because it's code. I'll square it!

mfisher87:
> If you use `doc-` or `doc/` (etc., some boundary after `doc`) for the prefix of your branch, it will auto-label for you!

:heavy_check_mark: TIL, thanks :) 

> And technically, we do want a changeset for this because it's code. I'll square it!

Understood for the future :) Thanks Angus!!

PR #2513, opened 2025-11-27
chore: add changeset

agoose77:
Finishes up #2508


stefanv:
@agoose77 as a matter of procedure, do we need changesets for docs-only changes?

rowanc1:
No, only for things that require bumps in packages. This docs change was in code, so a changeset is preferable.

PR #2515, opened 2025-11-27
Add plugin use case and debugging docs

### LINKED ISSUE #2507: Move/copy some plugin documentation from JupyterBook site -> mystmd.org
I found the JupyterBook documentation [Generate MyST AST with plugins](https://jupyterbook.org/stable/plugins/ast/) super useful today, but it took me some time to find it because I was looking on mystmd.org, not jupyterbook.org.

I'd love to see mystmd.org as the source of truth for this information, and the JupyterBook website pointing at it!

Same for the "Debug plugins" documentation :) https://jupyterbook.org/stable/plugins/debug

### LINKED ISSUE #2509: Document a common plugin use-case: DRYing a document
I recently played with this as I wrote a document that had repeated callouts which needed to have a consistent style. I wrote [two directives that provide that consistent appearance and prose](https://github.com/geojupyter/workshop-open-source-geospatial/blob/f50fdc7b2c57ac5ff94139a7a4f292adde95b044/plugin-custom-directives.mjs#L40-L60) and was pleased with how easy (and fun 😉) this process of deduping my doc was with the help of the MyST Sandbox.

I'd love to take this issue on but I have a question: I wanted to write this plugin so that I only dealt with "PRE" AST as it's more concise than "POST" AST. I found that I had to write "POST" AST to get this working. Is there a way, i.e. a function, that will enable the plugin author to only write "PRE" AST?

mfisher87:
The DRY doc is new, written by me. The rest are copied from the Jupyter Book docs. I also edited those docs for clarity based on my experience reading them. All of the edits are in separate commits after the initial copy commit.

I decide to create a new subsection in the plugin section for use case docs, as I think we'll likely have more in the future.

Resolves #2509
Resolves #2507

### DIFF
--- a/docs/plugins-ast.md
+++ b/docs/plugins-ast.md
@@ -19,8 +19,10 @@ Here's an example of using the `parseMyst` function within a directive plugin to
 const myDirective = {
   name: "mydirective",
   doc: "My new directive!",
+  // Here the body is passed as a string
   body: { type: String, doc: "To be parsed into MyST AST." },
   run(data, vfile, ctx) {
+    // Here we parse the body into MyST AST
     const ast = ctx.parseMyst(data.body);
     return ast.children[0];
   },
@@ -40,6 +42,7 @@ const myDirective = {
   arg: { type: String, doc: "To be parsed into MyST AST." },
   body: { type: String, doc: "The body of the directive." },
   run(data, vfile, ctx) {
+    // Here we parse a string that defines a multi-line card directive
     const ast = ctx.parseMyst(`:::{card} ${data.arg}\n${data.body}\n:::`);
     return ast.children[0];
   },
@@ -60,6 +63,7 @@ const myDirective = {
   arg: { type: String, doc: "To be parsed into MyST AST." },
   body: { type: String, doc: "The body of the directive." },
   run(data, vfile, ctx) {
+    // By removing the indentation white-space the string is correctly parsed
     const ast = ctx.parseMyst(`
 :::{card} ${data.arg}
 ${data.body}
@@ -82,6 +86,7 @@ In this case, `data.body` will be a list of AST nodes instead of text:
 const myDirective = {
   name: "mydirective",
   doc: "My new directive!",
+  // By defining type: 'myst', the directive body will automatically be parsed into MyST AST
   body: {
     type: 'myst',
     doc: "Pre-parsed and immediately usable as MyST AST.",
--- a/docs/plugins-dry.md
+++ b/docs/plugins-dry.md
@@ -1,5 +1,5 @@
 ---
-title: "Reusable document elements with plugins"
+title: "Reusable and custom document elements with plugins"
 ---
 
 Sometimes our documents require repetitive elements that must be consistent across a
@@ -13,7 +13,7 @@ Deduplicating our documents can, as a side effect, make it much easier to alter 
 appearance of repetitive elements by making a change in only one place!
 
 
-## An example of a repetitive element
+## An example of a repetitive element: a custom admonition
 
 If you're writing a tutorial, you may want to use the Diátaxis principle
 ["point out what the learner should notice"](https://diataxis.fr/tutorials/#point-out-what-the-learner-should-notice)
--- a/docs/plugins-debug.md
+++ b/docs/plugins-debug.md
@@ -35,7 +35,7 @@ You should see the contents of `data` printed in your MyST build log.
 ### Show the full depth of a MyST object in JavaScript logs
 
 `console.log` will only show you the first layer of depth, but you may want to inspect _all_ layers of a MyST object or AST node.
-To do so, use a JavaScript call like the following:
+This command will display all layers of an AST node at once:
 
 ```javascript
 console.dir(node, { depth: null, colors: true });


docs/myst.yml
```diff
@@ -155,12 +155,15 @@ project:
         - file: creating-word-documents.md
         - file: creating-jats-xml.md
         - file: creating-citation-cff.md
-    - title: Extensions
```
mfisher87, line 158: This title was being overridden by the document title

docs/plugins-ast.md
```diff
@@ -0,0 +1,128 @@
...
+  body: { type: String, doc: "To be parsed into MyST AST." },
+  run(data, vfile, ctx) {
+    const ast = ctx.parseMyst(data.body);
+    return ast.children[0];
+  },
+};
+```
+
+`ctx.parseMyst` returns a `root` node, which contains the parsed MyST.
+This function takes the first child of the root node, which is usually better for inserting into the document.
```
mfisher87: > This function takes the first child of the root node, which is usually better for inserting into the document.


I find this confusing! I'd like to improve it with your help!

What is meant by "This function"? `ctx.parseMyst`? It doesn't accept a node as an argument, it accepts text. Does the reader know what's meant by "node"? If we're keeping that language, let's say "AST node"? What is meant by "better for inserting into the document"? How? What's the alternative?

mfisher87:
@agoose77 I didn't get the automatic label! What did I do wrong? :face_with_peeking_eye:

mfisher87:
Rebased to credit original authors :)

docs/plugins-ast.md
choldgraf:
```suggestion
This function then uses the **first child** of the "root node" that is produced, which is usually better for inserting into a document.
```

docs/plugins-dry.md
```diff
@@ -0,0 +1,106 @@
+---
+title: '"DRY" your document with plugins'
```
choldgraf: What do you think about renaming this "Custom admonitions with a plugin"? That's a bit more tangible than the more abstract "DRY" (which I agree is important and maybe we can note this throughout?) and "how do I create a custom admonition type ?" Is one of the most common questions out there

docs/plugins-dry.md
```diff
@@ -0,0 +1,106 @@
...
+:::
+
+```{code} javascript
+:filename: plugin-you-should-notice.mjs
+:linenos:
+
+const youShouldNoticeDirective = {
+  name: 'you-should-notice',
+  doc: 'Renders a consistent callout when the learner should notice something.',
+  body: {type: 'myst'},
```
choldgraf, line 80: I had no idea this works! So if you set type to myst then it is auto parsed? Can we add a snippet how-to just about that somewhere else as well? Eg a section in one of the above docs, titled something like "How to parse directive arguments and body as MyST" and use that to introduce both this pattern and the parseMyst function?

docs/plugins-dry.md
```diff
@@ -0,0 +1,106 @@
...
+  run(data) {
+    return [{
+      type: 'admonition',
+      kind: 'important',
+      icon: false,
+      class: 'simple',
+      children: [
```
choldgraf, line 87: It might be useful (perhaps in the above how-to I recommended adding) to show how the same thing could be accomplished with raw strings and then a call to parseMyst rather than hand-rolling the AST. That way users could compare the two ways of doing things.

Random thought: I think you could even do intermixing of both. Eg hand craft some MyST AST and then run parseMyst on a gnarly admonition call that you give as a string because you can't remember how to hand craft MyST AST. Then insert it into the original AST you had created.

docs/plugins-dry.md
mfisher87: I'm not sure about this. I do think that the title needs some work because it's kind of jargony. But I like that the use case it presents is generic and broadly applicable (write less repetitive and more maintainable documents), but has a very clear and concrete implementation method and value proposition, using admonitions as an example. "Custom admonitions" could be interpreted differently, e.g. as creating a new admonition "kind" or customizing the styling of existing admonitions with CSS, or it could be dismissed as not applicable to other repeated elements users might want to "refactor" out of their document.

What are your thoughts on this? Do you have any ideas for a title that doesn't use the "DRY" jargon? E.g. "Reusable elements with plugins", "Extract repetitive content as plugins", "Eliminate repetition with plugins"?

I think "Reusable elements with plugins" may be the best of these, but I still don't think it does a good enough job of differentiating from the kind of repetition that's solved with `{include}`. :thinking:

docs/plugins-dry.md
mfisher87: Me too! Thanks @rowanc1 ;)

>Can we add a snippet how-to just about that somewhere else as well? Eg a section in one of the above docs, titled something like "How to parse directive arguments and body as MyST" and use that to introduce both this pattern and the parseMyst function?

Love it. Thinking about where this should go!

docs/plugins-ast.md
mfisher87: I'm [struggling](https://github.com/jupyter-book/mystmd/pull/2515#discussion_r2569678151) with this sentence as a whole. Does "this function" refer to `ctx.parseMyst`? What do we mean by "uses" or "takes"? I interpret that to mean "accepts as an argument", but isn't the argument text? What is meant by "better for inserting into a document"? Better than what alternative?

choldgraf: I think the main point is that the parsing returns a "root mode" and we throw away that root and only take the children inside because the root is an unnecessary level to add to the AST of a page that already has its own root node. Sure that make sense?

docs/plugins-dry.md
mfisher87: 32f2d366 is a start, what do you think? The `plugins-ast` document seemed like the natural place to cover both the explicit and implicit parsing as separate methods. I removed the term "easiest"; one could argue that implicit is easier, but it depends on the use case ;)

docs/plugins-dry.md
choldgraf: Where does "don't repeat yourself " fit in with the diataxis framework? It feels more abstract than a focused how-to. Maybe it would be best as a tutorial that shows a few different examples of DRY with plugins? I think either easy we should not key this bog us down so I'm fine just using the title you just suggested is you like

docs/plugins-dry.md
choldgraf: That's an improvement! I'd use more precise language than "explicitly and implicitly". How about "parse with a function" and "automatically parse directive arguments" ?

choldgraf: That's an improvement! I'd use more precise language than "explicitly and implicitly". How about "parse with a function" and "automatically parse directive arguments" ?

mfisher87: >how-to

One thing I'm thinking as I'm working through these docs changes is how the docs could be updated to better reflect the [different kinds of audiences](https://diataxis.fr/) of these docs. These use case docs are "how-tos" in my mind, but there are how-tos that are not "use cases". Should we rename the header these docs live under from "Use cases" -> "How tos"? And perhaps break them in to smaller chunks?

The diataxis paradigm is IMO, useful as always, but MyST has an extra axis that makes this a bit harder: The type of user could be a document author, a plugin author, or a core code contributor, or some mix of those. This makes it _really_ hard to structure and navigate the docs.

## Thinking out loud about the big picture

What about breaking the "Guide" in to 3+ separate guides with 3 separate TOCs? We've talked about the potential of having separate TOCs with navigation between TOCs at the top, and I think the MyST docs could benefit from that -- it would provide documentation authors more clear and intuitive structure to work within, and provide users a more clear and intuitive structure to navigate.

E.g.: User guide; Plugin developer guide; Theme developer guide; MySTMD developer guide. Each of these can have tutorials, reference, explanation, and how-tos clearly delimited. And we can link between the guides where the lines blur to provide onramps to different kinds of contributions; e.g. document authors can easily become plugin developers when they see a need to DRY their document. Plugin authors may be interested in core contributions when they find they need an additional extension point or that they need additional data as input. Etc.

docs/plugins-ast.md
mfisher87: OK that makes a lot more sense, thank you. I'd suggest changing the wording a bit:

```suggestion
`parseMyst` returns a {term}`"root" node <Root AST node>`. We're grabbing `ast.children[0]` (the first child of the root node) in the plugin example below because inserting a root node as a child would result in a malformed document!
```

mfisher87: fc439657 ?

docs/plugins-dry.md
mfisher87: I think of this document right now as a mix of explanation and how-to, which I'd like to avoid. "Don't repeat yourself" is explanation (we can remove it and link to an external explanation), but "How to reduce repetition in your document" is a how-to.

docs/plugins-dry.md
mfisher87: >That's an improvement! I'd use more precise language than "explicitly and implicitly". How about "parse with a function" and "automatically parse directive arguments" ?

I like that. I'll take another pass.

mfisher87: af2e17d0

Getting better?

choldgraf: I think the refactor is definitely worth discussing - want to capture those ideas in a new issue?

docs/plugins-dry.md
mfisher87: What do you think of the title "Reusable elements with plugins"? I think that's the best option I've considered so far; it fits the category of "use case", and we can revisit the categorization of the docs later. That's a really hard problem.

docs/plugins-dry.md
choldgraf: IMO there's enough improvement here that we could try to merge this one soon in the name of iteration

mfisher87: Good with me! :)

docs/plugins-dry.md
choldgraf: Yes! I like that title

choldgraf: Yes!

choldgraf:
@mfisher87 let me know when you are ready for another review or merge!

mfisher87:
Ready for merge if you are! (is it possible this PR broke the tests? :sweat_smile: )

choldgraf:
I'm pretty sure the tests are not related to this - there seems to be a timeout in myst-execute

PR #2516, opened 2025-11-28
Add npm verb to quickly demo docs

choldgraf:
I have been playing around with git worktrees, which means I have to keep re-running `npm run build`, `npm run link` etc. I think that this is a little npm command that one-shots all of that in one go:

```
npm run docs
```

This should let you just get started quickly w/ a new myst clone, worktree, etc. I tried it locally and it seemed to work well - I'm definitely not a JS expert but this would make my life a tiny bit easier at least, let me know if there's a better way to do this!

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build": "turbo run build",
     "dev": "turbo run dev --parallel",
     "docs": "npm-run-all -l build --parallel dev docs:run",
-    "docs:run": "cd docs && node ../packages/mystmd/dist/myst.cjs start .",
+    "docs:run": "cd docs && node ../packages/mystmd/dist/myst.cjs start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,6 +7,8 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev --parallel",
+    "docs": "npm-run-all -l build --parallel dev docs:run",
+    "docs:run": "cd docs && node ../packages/mystmd/dist/myst.cjs start .",
```
rowanc1: Does it need the '.'?

```suggestion
    "docs:run": "cd docs && node ../packages/mystmd/dist/myst.cjs start",
```

choldgraf:
Oops ya I think that was iteration cruft. Now I'm wondering why that worked 😅

PR #2520, opened 2025-12-01
📖 Make the myst-transform doc smoother and more legible

parmentelat:
very superficial changes for this part
https://mystmd.org/myst-transforms
just to make the content a little more edible


rowanc1:
🚀 Thank you!

Issue #2523, opened 2025-12-01
Superscript and subscript myst roles get misformatted when markdown includes footnotes


kaminski-pawel:
### Example without footnotes (correct)

Superscript and subscript markup gets wrapped in a separate `<p>` tag when a text includes footnotes.

For example,

```md
Paragraph before.

H{sub}`2`O. 1{sup}`st` Dec.
```

is transformed to the following html when no footnotes are present:

```html
<p>Paragraph before.</p>
<p>H<sub>2</sub>O. 1<sup>st</sup> Dec.</p>
```

<img width="1145" height="425" alt="Image" src="https://github.com/user-attachments/assets/48a87f12-30f6-4e8b-bf5e-5513c50493c0" />

### Example with footnotes (error)

However, introducing footnotes changes how superscript and subscript are rendered. The same example with a footnote:

```md
Paragraph before[^1].

H{sub}`2`O. 1{sup}`st` Dec.
```

is transformed to:

```html
<p>Paragraph before<span id="fnref-gkHdh3Wjr7" class="" data-state="closed">
<sup class="hover-link"><a class="no-underline text-inherit hover:text-inherit font-normal hover:underline" href="#fn-1" title="Link to Footnote" aria-label="Link to Footnote">[1]</a>
</sup></span>.</p>
<p>H<sub><p>2</p></sub>O. 1<sup><p>st</p></sup> Dec.</p>
<!--      ^    ^ cause of the wrong formatting -->
```

<img width="1145" height="539" alt="Image" src="https://github.com/user-attachments/assets/4c7032c9-c3fe-4209-8bce-48d2ad6fe32f" />

Note the nested `<p>` tag inside the `<sub>` and `<sup>`. Removing the inner `<p>` restores expected formatting.

Tested on mystmd==1.6.6

## Proposed solution

A workaround is to use the `<sub>` and `<sup>` tags directly in markdown. The same example with those tags:

```md
H<sub>2</sub>O. 1<sup>st</sup> Dec.
```

gets transformed to expected html, with or without footnotes appearing in the markdown:

```html
<p>H<sub>2</sub>O. 1<sup>st</sup> Dec.</p>
``` 

[...]

PR #2525, opened 2025-12-01
🐛 Footnote and role parsing bug

rowanc1:
Fixes: #2523

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15310,6 +15310,9 @@
         "js-yaml": "^4.1.0",
         "vfile": "^5.3.7"
       },
+      "devDependencies": {
+        "markdown-it-footnote": "^4.0.0"
+      },
       "peerDependencies": {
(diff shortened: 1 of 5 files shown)


packages/markdown-it-myst/src/nestedParse.ts
```diff
@@ -53,6 +53,30 @@ export function nestedCoreParse(
...
+  // These footnotes are still in the state, but should not be returned here
+  if (
+    tokens[0].type === 'paragraph_open' &&
+    tokens[1].type === 'inline' &&
+    tokens[2].type === 'paragraph_close' &&
+    tokens[3].type === 'footnote_block_open' &&
+    tokens[tokens.length - 1].type === 'footnote_block_close'
+  ) {
+    return [tokens[1]];
+  }
```
rowanc1, line 79: This is the fix. The rest is testing updates!

PR #2532, opened 2025-12-01
🐛 Do not raise errors on non-accessible sub-trees

### LINKED ISSUE #2530: code-block directives parse their content as MyST and log an error if it's not valid content
The Jupyter Book docs have a persistent warning like this:

```
  ⛔️ plugins/plugins.md:505 unknown directive: mydirective
```

And I think it's a bug in how the `{code-block}` and `{code}` directive parses its contents. For example:

This works with no error:

`````
````
```{nope}
```
````
`````

This returns the error though the contents are still output properly:

`````
````{code-block}
```{nope}
```
````
`````

[Sandbox link for example](https://mystmd.org/sandbox?tab=demo&myst=VABoAGkAcwAgAHcAbwByAGsAcwA6AAoACgBgAGAAYABgAAoAYABgAGAAewBuAG8AcABlAH0ACgBgAGAAYAAKAGAAYABgAGAACgAKAFQAaABpAHMAIABkAG8AZQBzAG4AJwB0ACAAdwBvAHIAawA6AAoACgBgAGAAYABgAHsAYwBvAGQAZQAtAGIAbABvAGMAawB9AAoAYABgAGAAewBuAG8AcABlAH0ACgBgAGAAYAAKAGAAYABgAGAA)

### PR DESCRIPTION
**Pull request by rowanc1**
Fixes #2530

### DIFF
--- a/packages/myst-parser/src/directives.ts
+++ b/packages/myst-parser/src/directives.ts
@@ -3,18 +3,18 @@ import type {
   DirectiveData,
   DirectiveSpec,
   DirectiveContext,
-  ParseTypes,
   GenericParent,
 } from 'myst-common';
 import { RuleId, fileError, fileWarn } from 'myst-common';
 import { selectAll } from 'unist-util-select';
 import type { VFile } from 'vfile';
-import { contentFromNode } from './utils.js';
+import { contentFromNode, markChildrenAsProcessed } from './utils.js';
 import type { Directive } from 'myst-spec';
 import { parseOptions } from './inlineAttributes.js';
 
 type MystDirectiveNode = GenericNode & {
   name: string;
+  processed?: false;
 };
 /**
  * Apply directive `run()` methods to build directive ASTs.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/myst-parser/src/directives.ts [hunk 1]
```diff
@@ -47,12 +47,20 @@ export function applyDirectives(
...
     const spec = specLookup[name];
+    if (spec && spec.body && spec.body.type !== 'myst') {
+      // Do not process the children of the directive that is not a myst block
+      // Doing so would raise errors (for example, if the parent directive is a code block)
+      // See https://github.com/jupyter-book/mystmd/issues/2530
+      markChildrenAsProcessed(node);
+    }
+    // Re-check if the directive has been processed
+    if (node.processed !== false) return;
+    delete node.processed; // Indicate that the directive has been processed
```
rowanc1, line 63: This is the fix @choldgraf, if the spec doesn't have a myst body, we mark all children as processed and then they will not raise errors.

choldgraf (approved):
So if I understand correctly, this basically says "if the content of a directive isn't marked as `myst` content it will not be processed". That sounds like the correct behavior to me! Thanks for the tests as well~

PR #2541, opened 2025-12-03
🐛 Incorrect type import

rowanc1:
From @agoose77 in #2499.


packages/myst-to-jats/src/transforms/containers.ts
```diff
@@ -1,5 +1,5 @@
 import type { Plugin } from 'unified';
-import type { Blockquote, Legend, Caption, FlowContent } from 'myst-spec';
+import type { Node, Blockquote, Legend, Caption, FlowContent } from 'myst-spec';
```
rowanc1, line 2: This is referenced below, but is not the right thing!

Issue #2477, opened 2025-11-23
Feature Request: Add configuration to disable or restrict dollar sign math parsing


MaxGhenis:
## Problem

Documents with frequent currency values (e.g., policy analysis, financial reports) trigger unintended math parsing when using dollar signs. For example:

```markdown
Taxpayers earning $25,000 to $34,000 pay tax...
```

This renders incorrectly as math mode, requiring escaping every dollar sign (`\$25,000`), which is tedious for documents with hundreds of currency mentions.

## Current Workaround

Manually escape every dollar sign: `\$25,000`

## Proposed Solution

Add configuration options to `myst.yml` similar to MyST-Parser (Sphinx):

### Option 1: Disable dollarmath entirely
```yaml
parse:
  math:
    dollarmath: false
```

### Option 2: Restrict dollar math parsing (preferred)
```yaml
parse:
  math:
    allow_digits: false  # Don't parse $25,000
    allow_space: false   # Don't parse $ 25,000
```

These options exist in MyST-Parser as:
- `myst_dmath_allow_digits`
- `myst_dmath_allow_space`
- Removing "dollarmath" from `myst_enable_extensions`

See: https://github.com/executablebooks/MyST-Parser/blob/master/docs/configuration.md

## Use Cases

- Financial documents
- Policy analysis reports
- Economic research
- Any document discussing monetary values

## Additional Context

MyST-Parser v0.17.0 made dollarmath disabled by default for Sphinx users, but mystmd appears to always have it enabled with no configuration option.

Related discussions:
- https://github.com/executablebooks/jupyter-book/issues/1185
- https://github.com/executablebooks/MyST-Parser/pull/507

rowanc1:
Thank you for the detailed report, and sorry we didn't get to this before your 150 changes to unescaped dollar signs...!

I have added option 1 in #2542, and am going to leave this open for more in-depth ways to configure the parser (digits, spaces, etc.).

PR #2542, opened 2025-12-03
💰 Dollarmath settings for markdown parsing

rowanc1:
See #2477

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -60,6 +60,31 @@ output_matplotlib_strings
     - `"remove-warn"` (default) or `"remove-error"`: remove all matplotlib strings in outputs, and log a warning or error
     - `"warn"` or "error": log a warning or error if matplotlib strings in outputs
 
+## Markdown Parsing Settings
+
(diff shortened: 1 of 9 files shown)


docs/settings.md
```diff
@@ -60,6 +60,31 @@ output_matplotlib_strings
...
+
+Adding an object of `parser` to the settings will allow you to control various parser behaviors.
+
+```{code-block} yaml
+:filename: myst.yml
+project:
+  settings:
+    parser:
+      dollarmath: false
+```
```
rowanc1, line 73: This is the config that I have added, it is in the `project.settings.parser` right now just enabling/disabling dollarmath, but many other options could easily come in here as well.

docs/settings.md
```diff
@@ -60,6 +60,31 @@ output_matplotlib_strings
...
+
+:::{warning}
+Markdown parsing settings currently only work on the `myst.yml` project settings, not the page frontmatter.
```
rowanc1, line 76: Right now this is only in the project, in future we could have there settings in the frontmatter of the files, but that would require pre-parsing that part, and we aren't setup for that.

PR #2546, opened 2025-12-03
⏪ Support `sort: ascending` order on patterns in the table of contents

### LINKED ISSUE #2545: Support reverse order for glob pattern matching in table of content
This feature request will be used by JupyterHub's `team-compass` and possible Jupyter Book's `team-compass`.

## Use case

MyST is used to archive notes of meetings, e.g.

- January 2026 as `2026-01-01.md`
- February 2026 as `2026-02-01.md`
- March 2026 as `2026-03-01.md`
- ...

Visitors of the MyST powered web site are probably more interested in the latest meeting (e.g. `2026-03-01.md`) instead of the first ever holding meeting (e.g. `2026-01-01.md`). For this, it is nice to be able to list the archive notes in chronological reverse order.

## Proposal

Expand [Glob pattern matching](https://mystmd.org/guide/table-of-contents#glob-pattern-matching) to accept a option `reverse`. For example, the pattern-matching is

```
version: 1
project:
  toc:
    - file: root.md
    - pattern: '*.md'
      reverse: true
```

and the no pattern-matching is

```
version: 1
project:
  toc:
    - file: root.md
    - file: 2026-03-01.md
    - file: 2026-02-01.md
    - file: 2026-01-01.md
```

rowanc1:
Fixes #2545

### DIFF
--- a/docs/notebook-configuration.md
+++ b/docs/notebook-configuration.md
@@ -22,6 +22,7 @@ Using Jupyter Lab, you can use the `Property Inspector` builtin extension to man
 Using this tool you can add key-value pairs to the notebook metadata. For
 example, to skip execution of a notebook, you would add the following key-value
 pair inside the notebook metadata dictionary:
+
 ```json
 "tags": [
         "skip-execution"
@@ -41,7 +42,6 @@ kernelspec:
 skip_execution: true
 ---
 ```
-
 
 (notebook-cell-tags)=
 
@@ -72,7 +72,7 @@ You can use the `Property Inspector` builtin extension in Jupyter Lab to manage 
 
 ### Markdown notebooks
 
-In addition, [MyST also supports the {myst-directive}`code-cell` directive](#code-cell), and here's an example of adding the `raises-exception` tag on such a code cell:
+In addition, MyST also supports the {myst:directive}`code-cell` directive for [markdown notebooks](#code-cell); here is an example of adding the `raises-exception` tag on such a code cell:
 
 In Markdown of a jupyter notebook these look like:
 
--- a/docs/myst.yml
+++ b/docs/myst.yml
@@ -90,12 +90,12 @@ project:
         - file: quickstart-jupyter-lab-myst.md
     - title: Configuration and setup
       children:
-      - file: configuration.md
-      - file: frontmatter.md
-      - file: settings.md
-      - file: table-of-contents.md
-      - file: document-parts.md
-      - file: authorship.md
+        - file: configuration.md
+        - file: frontmatter.md
+        - file: settings.md
+        - file: table-of-contents.md
+        - file: document-parts.md
+        - file: authorship.md
     - title: Authoring
       children:
         - file: typography.md
@@ -161,7 +161,7 @@ project:
         - file: plugins-distribute.md
         - file: executable-plugins.md
         - file: plugins-debug.md
-        - title: "Use cases"
+        - title: Use cases
           children:
             - file: plugins-ast.md
             - file: plugins-dry.md


rowanc1:
I need a bit more testing on this before merging.

packages/myst-toc/src/toc.ts
```diff
@@ -152,7 +152,13 @@ export function validatePatternEntry(
   }
 
   const commonEntry = validateCommonEntry(intermediate, opts);
-  return { pattern, ...commonEntry };
+  const output: PatternEntry = { pattern, ...commonEntry };
+
+  if (defined(intermediate.reverse)) {
+    output.reverse = validateBoolean(intermediate.reverse, incrementOptions('reverse', opts));
```
rowanc1: I think we should warn here and ignore if pattern is not defined.

docs/table-of-contents.md
```diff
@@ -103,6 +103,22 @@ project:
...
+:filename: myst.yml
+version: 1
+project:
+  toc:
+    - file: root.md
+    - pattern: '*.md'
+      reverse: true
+```
+
+This will list files in reverse chronological order (newest first), which is particularly useful for meeting notes, blog posts, or other time-based content.
```
rowanc1:
```suggestion
If your files are named with a date at the start, for example, `2025-12-03-reverse-toc.md`, this will list files in reverse chronological order (newest first), which is particularly useful for meeting notes, blog posts, or other time-based content.
```

mfisher87:
Thoughts on `sort: descending` (default `ascending`) instead of `reverse`? ([reasoning](https://github.com/jupyter-book/mystmd/issues/2545#issuecomment-3608077026))

choldgraf:
Carrying some conversation from https://github.com/jupyter-book/mystmd/issues/2545 into here:

- I think this is reasonable to support, `glob:` is useful because it means users don't have to re-commit `myst.yml` every time a file changes inside.
- I think it is reasonable to support "inverse glob" because many people use numbers in their filenames to denote dates, *and* they often want the latest dates to show up first. We could implement a more "clever" way of "removing" the date before doing the sorting etc, but IMO that's more complex and "magic" than just letting a user explicitly say "do this in reverse".
- I kinda like `sort: ascending` vs `descending` as a more explicit config than `reverse: true`, but don't feel strongly. I think @rowanc1 should just decide if he wants to change it or not
- One risk is that we do open ourselves up to users saying "you let us reverse, now let us define arbitrary functions to sort contents of a folder" 🤷

stefanv:
I think `sort:` is more intuitive. What does the user read when they see "glob the files, then reverse". I wouldn't *assume* that that sorts chronologically (that depends on OS, file system, etc.).

So, `sort: time-asc`, `sort: time-desc`, `sort: filename-asc`, `sort: filename-desc` or some scheme like that would immediately give a sense of what is going on. Alternative: `sort: newest-first`, `sort: oldest-first`, `sort: filename`, `sort: filename-reverse`.

Or do we sort by titles? I don't think so, but this would also clarify that :)

choldgraf:
just noting that GitHub has a standard sort syntax that is similar to what @stefanv describes:

https://docs.github.com/en/search-github/getting-started-with-searching-on-github/sorting-search-results

`sort:comments` implies `sort:comments-desc` and users can also specify `sort:comments-asc`.

That said, I don't think we need to grow our config surface area just yet, and I'd be fine with `sort:ascending` etc...but that would be a natural extension point if users really wanted it in the future

stefanv:
Problem with `reverse` and `sort: ascending` is that you're not actually telling the user what you're sorting by.

agoose77:
My preference is to only sort by filename -- support for things like `mtimes` etc. vary across platforms and deployment strategies. As such, I'm leaning towards `sort: ascending | descending` and assuming that it's evident enough (plus docs) that this is filename ordering.

I do prefer that to `reverse`, and it feels like that's a majority of us? @rowanc1 I'm happy to rework the PR to use `sort` if you're needing to down tools.

stefanv:
> As such, I'm leaning towards `sort: ascending | descending`

That fully closes the door on any future changes to sort scheme, though :thinking: Does it hurt to call it `sort: filename` and `sort: filename-desc` or similar, so the `sort` keyword indicates the sort field?

agoose77:
> > As such, I'm leaning towards `sort: ascending | descending`
> 
> That fully closes the door on any future changes to sort scheme, though 🤔 Does it hurt to call it `sort: filename` and `sort: filename-desc` or similar, so the `sort` keyword indicates the sort field?

It feels like a tradeoff between present awkwardness and future extension. FWIW, in future nothing precludes us from having `ascending` and `ascending-XXX` if we identify an `XXX`. 

This is an aesthetics conversation imo, and I don't feel super strongly. Happy to roll with your suggestion @stefanv.

choldgraf:
IMO this isn't strictly true:

> That fully closes the door on any future changes to sort scheme, though

In the future we can always introduce new keys, and say that `sort: ascending` now maps onto `sort: filename-asc` 🤷 . i think for now going with `ascending` and `descending` is fine. I also think `sorf: filename-asc/desc` is also fine, so I if @stefanv or @rowanc1 have stronger preferences than me or @agoose77 just go with your gut IMO

mfisher87:
> So, `sort: time-asc`, `sort: time-desc`, `sort: filename-asc`, `sort: filename-desc` or some scheme like that would immediately give a sense of what is going on. Alternative: `sort: newest-first`, `sort: oldest-first`, `sort: filename`, `sort: filename-reverse`.

Are you imagining `time-asc` would be sorting by the `date` field in the frontmatter or the modified time of the file? :thinking: 

Not sure how I feel about a compound string value, it could be limiting in the future. If we move towards allowing sorting by different attributes/fields, we could support e.g. sorting by arbitrary document frontmatter fields with a simple structure:

```yaml
sort:
  field: "my-metadata-field"
  order: "ascending"
```

And the simple value of `sort: ascending` could always mean sorting by filename.

To be explicit, we could start with type:

```typescript
'ascending' | 'descending'
```

And evolve towards a type like:

```typescript
'ascending' | 'descending' | {field: String, order: 'ascending' | 'descending'}
```

stefanv:
```
    - pattern: '*.md'
      sort: ascending
```

Looking at this again in the context of how it will be used, I suppose `sort: ascending` will  be OK, given that it clearly applies to the glob result, and nothing else.

We'll need something more sophisticated for listings, e.g., but we'll cross that bridge when we get to it.

choldgraf:
Agree w/ @stefanv - I think that listings are where we'll need to provide more flexibility for `sort` rather than the table of contents structure of the project.

rowanc1:
I have changed this to `sort: ascending` and `sort: descending` and will get it in later today.

stefanv:
Great, thanks Rowan!

docs/notebook-configuration.md
```diff
@@ -72,7 +72,7 @@ You can use the `Property Inspector` builtin extension in Jupyter Lab to manage
 
 ### Markdown notebooks
 
-In addition, [MyST also supports the {myst-directive}`code-cell` directive](#code-cell), and here's an example of adding the `raises-exception` tag on such a code cell:
+In addition, MyST also supports the {myst:directive}`code-cell` directive for [markdown notebooks](#code-cell); here is an example of adding the `raises-exception` tag on such a code cell:
```
rowanc1, line 75: this was raising an error and has a pop up inside a link, which isn't valid anyways. rephrased.

docs/table-of-contents.md
```diff
@@ -103,6 +103,22 @@ project:
 ::::
 :::::
 
+#### Sort order
+
+You can control the sort order of files matched by a pattern by adding `sort: 'ascending'` or `sort: 'descending'` to the pattern entry. By default, files are sorted in ascending order. The `descending` option is useful for archives where you want the most recent files listed first:
```
rowanc1, line 108: Have now fully tested, and works as expected!

mfisher87: Thanks, Rowan! This will be immediately useful for me :D

choldgraf:
amazing, thanks all for the discussion and @rowanc1 for the implementation

PR #2551, opened 2025-12-04
🚛 migrate xrefs

stevejpurves:
This PR simply applies the migration functionality that we already have in Myst migrate to incoming content from xrefs.

We had made a decision not to upgrade previous content automatically within the theme. However, when it comes to xrefs, the lack of this results in a DOM, in an AST that contains basically invalid nodes or nodes that don't conform to the overall AST version being advertised at the root level.Which is particularly bad.

Tested using xrefs to V2 content from a V3 AST.

### DIFF
--- a/packages/mystmd/tests/notebook-fig-embed/outputs/index.json
+++ b/packages/mystmd/tests/notebook-fig-embed/outputs/index.json
@@ -18,8 +18,6 @@
     "content_includes_title": false,
     "github": "https://github.com/jupyter-book/mystmd",
     "keywords": [],
-    "source_url": "https://github.com/jupyter-book/mystmd/blob/feat/migrate-external-content/packages/mystmd/tests/notebook-fig-embed/index.md",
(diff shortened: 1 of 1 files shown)


stevejpurves:
The test failure is real, will address if we want to bring this in cc @agoose77

stevejpurves:
My last commit updated the test snapshot which i expected to fix the failure, but it has not. I am not clear on why.

packages/mystmd/tests/notebook-fig-embed/outputs/index.json
```diff
@@ -185,37 +200,45 @@
...
           },
           {
             "type": "container",
             "kind": "figure",
             "children": [
               {
-                "type": "output",
-                "data": [
+                "type": "outputs",
+                "children": [
```
rowanc1, line 219: this looks correct!

stefanv:
Thanks @rowanc1 for educating me on the upgrade infra on the community call. This now makes a lot of sense, and happy that it provides some future-proofing!

PR #2554, opened 2025-12-04
Configure API URL

parmentelat:
previously the code had https://api.mystmd.org hardwired
this can now be redefined by setting the API_URL environment variable

this comes in handy when you need to run myst against a locally running API

### DIFF
--- a/packages/myst-cli/src/session/session.ts
+++ b/packages/myst-cli/src/session/session.ts
@@ -102,7 +102,7 @@ export class Session implements ISession {
     this.API_URL = process.env.API_URL ?? DEFAULT_API_URL;
     // trailing slashes will cause issues
     this.API_URL = this.API_URL.replace(/\/+$/, '');
-    console.log(`building myst-cli session with API URL: ${this.API_URL}`);
+    console.debug(`building myst-cli session with API URL: ${this.API_URL}`);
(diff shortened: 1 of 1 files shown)


parmentelat:
note that there is some duplication between the 2 session.ts files
my limited ts skills have not allowed me to factor these..

packages/myst-cli/src/session/session.ts
```diff
@@ -98,7 +98,11 @@ export class Session implements ISession {
   }
 
   constructor(opts: { logger?: Logger; doiLimiter?: Limit; configFiles?: string[] } = {}) {
-    this.API_URL = API_URL;
+    // use env variable if set
+    this.API_URL = process.env.API_URL ?? DEFAULT_API_URL;
+    // trailing slashes will cause issues
+    this.API_URL = this.API_URL.replace(/\/+$/, '');
+    console.log(`building myst-cli session with API URL: ${this.API_URL}`);
```
agoose77:
```suggestion
    console.debug(`building myst-cli session with API URL: ${this.API_URL}`);
```

agoose77:
I'm in favour of _not_ overthinking this. Although we want to possibly remove the API server as-is in future, let's defer that decision. Thanks @parmentelat!

PR #2555, opened 2025-12-04
Add support for `--readthedocs` on `init`

jtpio:
Landing here after trying to publish a Jupyter Book site to ReadTheDocs.

There is already some documentation for GitHub Pages here: https://jupyterbook.org/stable/get-started/publish/#github-pages. And while ReadTheDocs is mentioned in the docs, there didn't seem to be the equivalent for ReadTheDocs.

So this PR adds a new `--readthedocs` flag so it's easier to get the config right. Especially as the last line (`rm -rf _build`) seems to be important on ReadTheDocs.

### DIFF
--- a/packages/myst-cli/src/init/readthedocs/index.ts
+++ b/packages/myst-cli/src/init/readthedocs/index.ts
@@ -23,6 +23,7 @@ function createReadTheDocsConfig({
 version: 2
 
 # Set the OS, Python version, and Node.js version
+# Note: Python is included for executing code in notebooks or using Jupyter Book
 build:
   os: ubuntu-22.04
   tools:
@@ -88,7 +89,7 @@ export async function readTheDocsAction(session: ISession) {
   const prompt = await inquirer.prompt([
     {
       name: 'pythonVersion',
-      message: 'What Python version would you like to use?',
+      message: 'What Python version? (For executing notebooks or using Jupyter Book)',
       default: '3.12',
     },
     {


packages/myst-cli/src/init/readthedocs/index.ts
```diff
@@ -0,0 +1,123 @@
...
+
+✅ ${chalk.bold.green('Next Steps')}
+
+1. Create an account on ${chalk.blue('https://readthedocs.org')} if you don't have one
+2. Import your project from GitHub/GitLab/Bitbucket
+3. Commit and push the ${chalk.yellow(READTHEDOCS_FILENAME)} file to your repository
+4. Read the Docs will automatically build your documentation on each push
+5. 🎉 Your documentation will be available at ${chalk.blue('https://YOUR-PROJECT.readthedocs.io')}
+
+For more information, see: ${chalk.blue('https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html')}
```
mfisher87:
```suggestion
For more information, see: ${chalk.blue('https://docs.readthedocs.com/platform/stable/intro/mystmd.html')}
```

packages/myst-cli/src/init/readthedocs/index.ts
```diff
@@ -0,0 +1,123 @@
...
+# This file was created automatically with \`${npmBinaryName()} init --readthedocs\`
+
+# Required
+version: 2
+
+# Set the OS, Python version, and Node.js version
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "${pythonVersion}"
```
choldgraf, line 30: let's add a note that python is just for if we're using jupyter book or if the execution needs python

PR #2557, opened 2025-12-04
Fix empty button role

### LINKED ISSUE #2061: Button role with no link shows as an empty button
When you use a button role, but don't give a link in addition to the text content, the displayed button has no text inside it. For example see this image:

Image

choldgraf:
Now when buttons don't have a link specified (e.g. `` {button}`Text` ``) it emits a `span` with `.button` class instead of a link button that is empty.

- Adds more explicit parsing to button roles so it's easier to understand
- Adds a test for a few different button role conditions
- Fixes #2061 

Right now this won't be styled so this PR adds a style to it:

- https://github.com/jupyter-book/myst-theme/pull/728

That PR is merged, and I confirmed locally (and in the tests) that the button now renders when it has an empty link:

CleanShot 2025-12-26 at 10 07 15@2x

### DIFF
--- a/packages/myst-ext-button/tests/button.spec.ts
+++ b/packages/myst-ext-button/tests/button.spec.ts
@@ -31,17 +31,6 @@ describe('Button component', () => {
     ]);
   });
 
-  it('should treat bare text (even if it looks like a URL) as a non-link button', () => {
-    const result = buttonRole.run({ name: 'button', body: 'http://example.com' }, new VFile());
-    expect(result).toEqual([
-      {
-        type: 'span',
-        class: 'button',
-        children: [{ type: 'text', value: 'http://example.com' }],
-      },
-    ]);
-  });
-
   it('should display body text with no link when no URL is provided', () => {
     const result = buttonRole.run({ name: 'button', body: 'Click me' }, new VFile());
     expect(result).toEqual([


packages/myst-ext-button/src/index.ts
```diff
@@ -13,16 +16,43 @@ export const buttonRole: RoleSpec = {
...
-    };
-    if (modified) node.children = [{ type: 'text', value: modified.trim() }];
-    return [node];
+    /**
+     * Behavior:
+     * - `{button}`text`` => button with text only (no link target).
+     * - `{button}`<text>`` => button links to `text`, shows `text`.
+     * - `{button}`text<label>`` => button links to `label`, shows `text`.
+     */
+    const bodyTargetMatch = BODY_TARGET_PATTERN.exec(body);
```
stefanv: Instead of two regexs, why not match on the standard pattern: `text<link>`, and then do the logic of figuring out what to display: if <link> is empty, show text, if text is empty, show link.

stefanv: I.e., two groups, one for text, one for link, including triangle brackets. Then can check if link == `<>` and in that case render nothing for the link.

choldgraf: I had split it into two just to be really explicit because the last REGEX I found pretty opaque, but then again I find all regexes opaque, so sounds good I will revert back to a one-liner :-)

choldgraf:
OK this is now two PRs - one for myst-theme, I updated the body to have links etc. I think the failing test is that flaky DOI test

packages/myst-ext-button/src/index.ts
```diff
@@ -1,28 +1,72 @@
 import type { RoleSpec, RoleData, GenericNode } from 'myst-common';
 import type { Link } from 'myst-spec-ext';
+import { fileWarn, RuleId } from 'myst-common';
 
-const REF_PATTERN = /^(.+?)<([^<>]+)>$/;
+// Matches "text<link>" capturing body text (group 1) and an optional "<link>" suffix (group 2).
+// Group 2 keeps the angle brackets; it can also be exactly "<>" to signal an empty target.
+const TEXT_LINK_PATTERN = /^([^<>]*)(<[^<>]*>)?$/;
```
choldgraf: This is too edge-casey, I am going to simplify this

packages/myst-ext-button/src/index.ts
```diff
@@ -1,28 +1,72 @@
...
 
 export const buttonRole: RoleSpec = {
   name: 'button',
-  doc: 'Button element with an action to navigate to internal or external links.',
+  doc: 'Button element with an action to navigate to internal or external links. See [](#ui:buttons).',
```
choldgraf, line 10: This is a little addition to make it easier to find the more verbose docs about this from the reference page. I opened an issue to do this for others here:

- https://github.com/jupyter-book/mystmd/issues/2609

packages/myst-ext-button/tests/button.spec.ts
```diff
@@ -19,15 +19,61 @@ describe('Button component', () => {
...
         type: 'link',
         class: 'button',
         url: 'http://example.com',
-        children: [],
+        children: [{ type: 'text', value: 'http://example.com' }],
+      },
+    ]);
+  });
+
+  it('should treat bare text (even if it looks like a URL) as a non-link button', () => {
```
choldgraf: I feel like this test isn't needed because it's not really _intended_ behavior but just not behavior that's implemented yet

packages/myst-ext-button/tests/button.spec.ts
```diff
@@ -19,15 +19,61 @@ describe('Button component', () => {
...
+    expect(result).toEqual([
+      {
+        type: 'span',
+        class: 'button',
+        children: [{ type: 'text', value: 'Click me' }],
+      },
+    ]);
+  });
+
+  it('should treat an empty link target as a non-link button', () => {
```
choldgraf: also feels excessive

packages/myst-ext-button/tests/button.spec.ts
```diff
@@ -19,15 +19,50 @@ describe('Button component', () => {
...
+    const result = buttonRole.run({ name: 'button', body: 'http://example.com' }, new VFile());
+    expect(result).toEqual([
+      {
+        type: 'span',
+        class: 'button',
+        children: [{ type: 'text', value: 'http://example.com' }],
+      },
+    ]);
+  });
+
```
choldgraf:
```suggestion
```

PR #2572, opened 2025-12-07
🧪 Remove short DOI test in latex

rowanc1:
This doi is picked up in a myst-transform test already, the key in latex is not consistent and is leading to failures.


rowanc1:
This should improve test flaky-ness with DOI.org.

PR #2577, opened 2025-12-10
🔑 Add key to avoid react error

rowanc1:
Multiple paragraph acknowledgements, for example, would raise a frontend react error complianing that the `<MyST />` component did not have keys.


packages/myst-common/src/extractParts.ts
```diff
@@ -76,7 +76,7 @@ function createPartBlock(
     removePartData?: boolean;
   },
 ) {
-  const block: GenericParent = { type: 'block', children };
+  const block: GenericParent = { type: 'block', key: createId(), children };
```
rowanc1, line 79: This is the only change, adding a key to the block. Necessary for frontend when we don't put these through the transforms again.

Tested in the app that it removes the error successfully.

Issue #2563, opened 2025-12-05
Migrate `myst-spec` to the `mystmd` monorepo


agoose77:
## Context
> [!Important]
> The MyST Spec means _many_ things, and we mean different things at different times. Listen out for the context!

Formally, the MyST Spec, defined at https://mystmd.org/spec, is an attempt to define the public API for MyST documents (the abstract syntax tree structure), and more importantly a commitment to the idea of the existence and benefit of such a spec. People can consume the spec in different ways, such as by writing MyST plugins, or by consuming the `.json` of a published MyST site. 

However, the spec itself is fairly old, and the tooling makes it hard to iterate on as we build out the MyST stack. JSON Schema is less rich than TypeScript in many ways, and so it's inconvenient to use it as a source of truth for our compile time types. It also doesn't play nicely with unist's newer typing conventions, which is why we're stuck on an old version of unist/mdast (that limits our ability to upgrade the tooling nicely), and why we have `GenericParent` all over the place. 

To work around this friction, the core team created `myst-spec-ext`, which is like a sandbox for us to modify our types in a low friction way. It's _these_ types that we consume in most places. Unfortunately, this means what we promise to publish is quite different from what we actually publish.

## Proposal

Publishing an out-of-date spec that we tell people to ignore is worse, in my view, than not publishing anything at all. I have two suggestions, one of them serious!

1. Delete `https://mystmd.org/spec` until we publish it properly
2. Address the root cause for the existence of `myst-spec-ext` and the atrophying of `myst-spec`.

As you can imagine, I am in favour of (2). Here's a technical proposal for how to achieve this:

1. We migrate the jupyter-book/myst-spec repository contents to jupyter-book/mystmd, adding a new package `packages/myst-spec` to this repo.
1. We convert myst-spec from JSON Schema to TypeScript, and stop publishing JSON Schema for now. This makes it easier to tweak our types (see https://github.com/jupyter-book/myst-spec/pull/66/files).
1. We deprecate `myst-spec-ext` after migrating its contents into `myst-spec`
1. We modify https://github.com/jupyter-book/myst-spec to pull from the `myst-spec` package and generate docs accordingly.

akhmerov:
Sounds excellent! To check about things not breaking: are there current consumers of the JSON schema? If yes, what will happen to them? Is it possible to publish a JSON schema that is strictly more generous than a typescript spec? That would still be a useful safeguard for data then.

agoose77:
[...]

PR #2579, opened 2025-12-10
Add `myst-spec` as TypeScript types

### LINKED ISSUE #2582: Convert `myst-spec` to use `mdast@3`
The `myst-spec` packages uses JSON Schema as a source of truth. Given that the primary consumers of `myst-spec` are our own TypeScript packages, and that JSON Schema cannot express the kinds of types that the `mdast` ecosystem uses, it would be nicer to rewrite this package as `mdast` types.

This work will not involve upgrading to `mdast@4`, which is a longer-term goal that will let us upgrade more of our stack.

agoose77:
Closes #2582, as a part of #2563.

### How `mdast` is typed
`mdast@^3` is defined in text on [the GitHub Repo](https://github.com/syntax-tree/mdast/tree/3.0.0), and in TypeScript in `@types/mdast` via [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mdast/v3/index.d.ts).

The type system is designed to be extensible through the use of mapping types, e.g.
```typescript
type BlockContent = BlockContentMap[keyof BlockContentMap];

interface BlockContentMap {
  paragraph: Paragraph;
  ...
}
```

The reason for this approach over directly specifying `type BlockContent = Paragraph | ...` is that this type set is open to extension. Third-party code can _extend_ the members of `BlockContentMap` by using declaration merging via module augmentation: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

## What this means for `myst`
Presently, we use two type families in MyST:
- `myst-spec-ext`, which defines our effective spec (rather than our published one)
- `myst-common`, which defines our `GenericParent` and `GenericNode` workarounds

As `mystmd` has been developed, we've worked around type problems with type casts, and our types are fragile. The goal with this PR is to set the foundations for us to be stricter with these types. I suspect future PRs will have to chip away at removing `GenericParent` altogether.


packages/myst-common/src/utils.ts
```diff
@@ -179,27 +179,32 @@ export function toText(content?: Node[] | Node | null): string {
...
+          if (!last.position) last.position = {} as Required<Node>['position'];
+          last.position.end = n.position.end;
+        }
+        if (!last.value) last.value = '';
+        if (n.value) last.value += n.value;
+        return c;
+      }, [] as Content[])
+    : [];
+  if (children) (node as Parent).children = children;
   return node;
```
agoose77, line 208: I suspect this should actually be generic, e.g. 
```typescript
export function mergeTextNodes<T extends Nodes>(node: T): T {
```
although that won't be a lower bound

packages/myst-spec-ext/src/types.ts
```diff
@@ -348,3 +48,5 @@ export type MystSearchIndex = {
   version: '1';
   records: SearchRecord[];
 };
+
+export type * from 'myst-spec';
```
agoose77, line 52: For compatibility.

packages/myst-spec/src/index.ts
```diff
@@ -0,0 +1,704 @@
...
+   * Count this enumerated object for numbering based on node type, kind, etc.
+   */
+  enumerated: boolean;
+  /**
+   * Resolved enumerated value for this enumerated object.
+   */
+  enumerator: string;
+}
+
+/// Types //////////////////////////////
```
agoose77, line 51: First we define the custom MyST types.

packages/myst-spec/src/index.ts
```diff
@@ -0,0 +1,704 @@
...
+  label?: string;
+}
+
+interface Underline extends Parent {
+  type: 'underline';
+  children: PhrasingContent[];
+}
+type Visibility = 'show' | 'hide' | 'remove';
+
+declare module 'mdast' {
```
agoose77, line 507: Then we define the customisations to the mdast types

agoose77:
Closing this for a smaller PR that _just_ merges the changes in and fudges the types.

PR #2596, opened 2025-12-17
DOC: removing deprecated frontmatter from docs

### LINKED ISSUE #2595: DOC: cleanup deprecated frontmatter from documentation
`skip_execution` has been deprecated, yet it's still in the docs; e.g. here: https://mystmd.org/guide/notebook-configuration

bsipocz:
This should close #2595; however I'm not certain if we still have the cell level tag or not? Looking at the source code I think those are still valid, but if I misunderstood something then more cleanup will be needed.

### DIFF
--- a/docs/notebook-configuration.md
+++ b/docs/notebook-configuration.md
@@ -39,7 +39,8 @@ kernelspec:
   name: python3
   display_name: Python 3
 
-skip_execution: true
+execute:
+  skip: true
 ---
 ```
 


agoose77:
Ah nice catch! This must have been missed in the original PR. I think we do still handle tags, but I'll check that for a regression!

choldgraf (approved):
Given that this improves what we've got, I'll just merge this, thanks @bsipocz ! @agoose77 in the future I suggest we just merge these kinds of straightforward PRs!

Issue #2599, opened 2025-12-19
`si` role ignores numbers before decimal point


jan-david-fischbach:
When writing
```
{si}`348.64 <\nano \meter>`
```

the generated AST reads as

``` yaml
type: root
children:
  - type: paragraph
    children:
      - type: mystRole
        name: si
        value: 348.64 <\nano \meter>
        children:
          - type: si
            number: '64'
            unit: nm
            units:
              - nano
              - meter
            alt: nano meter
            value: 64 nm

```

resulting a rendered:

```
64 nm
```

See [sandbox](https://mystmd.org/sandbox?tab=demo&myst=ewBzAGkAfQBgADMANAA4AC4ANgA0ACAAPABcAG4AYQBuAG8AIABcAG0AZQB0AGUAcgA%2BAGAA)

## Description
see above

## Proposed solution
correct parsing of decimal numbers in `si` role

PR #2600, opened 2025-12-19
Add support for parsing decimal numbers in si role

jan-david-fischbach:
Considers common decimal separators as part of the number.
Further I also added `e` and `-` to tolerate expressions like

```
{si}`348.9e-9 <\meter>`
```

Fixes https://github.com/jupyter-book/mystmd/issues/2599

### DIFF
--- a/.changeset/stale-apes-shave.md
+++ b/.changeset/stale-apes-shave.md
@@ -1,0 +1,5 @@
+---
+"myst-roles": patch
+---
+
+Add support for parsing decimal numbers in si role


jan-david-fischbach (commented):
One could come up with a more sophisticated regex to make sure the number format is correct. Seems like overengineering though

jan-david-fischbach:
fixes #2599

packages/myst-roles/src/si.ts
```diff
@@ -9,7 +9,7 @@ export const siRole: RoleSpec = {
   },
   run(data: RoleData): GenericNode[] {
     const value = data.body as string;
-    const match = value.match(/([0-9]+)\s?<([\\a-zA-Z\s]+)>/);
+    const match = value.match(/([0-9.,e-]+)\s?<([\\a-zA-Z\s]+)>/);
```
rowanc1: Maybe `E` as well?

jan-david-fischbach: Added

jan-david-fischbach:
@rowanc1 I incorporated your suggestion

rowanc1:
Thanks!

PR #2602, opened 2025-12-20
✨ Add `{anywidget}` directive and AST node

stevejpurves:
Upstreaming and adapting the AnyWidget directive code [started by Curvenote](https://github.com/curvenote/curvenote/tree/main/packages/any-widget), to provide a `widget` directive for MyST with first class AST support.

> [!Note]
> See jupyter-book/myst-theme#795 for the theme counterpart.

See the tracking issue: #2603

### DIFF
--- a/.changeset/grumpy-clowns-cheer.md
+++ b/.changeset/grumpy-clowns-cheer.md
@@ -1,0 +1,5 @@
+---
+'myst-frontmatter': patch
+---
+
+Updates to latest SPDX JLOB
--- a/.changeset/ripe-kings-crash.md
+++ b/.changeset/ripe-kings-crash.md
@@ -1,0 +1,7 @@
+---
+'myst-spec-ext': patch
+'myst-common': patch
+'myst-cli': patch
+---
+
+Cheanges to suport first version of anywidget


agoose77:
YAAAS! Thanks for sharing this, saves us having to reinvent the wheel!

packages/myst-directives/src/widget.ts
```diff
@@ -0,0 +1,110 @@
...
+  const notString = typeof field !== 'string';
+  const invalidValues = validValues && typeof field === 'string' && !validValues.includes(field);
+  if (notString) vfile.message(`Invalid ${fieldName} supplied.`);
+  if (invalidValues)
+    vfile.message(`Invalid ${fieldName} supplied must be one of (${validValues.join(' | ')}).`);
+}
+
+export const widgetDirective: DirectiveSpec = {
+  name: 'widget',
+  alias: ['any:widget'],
```
agoose77: Maybe drop the alias whilst we're not actually implementing the full anywidget MVC etc.

packages/myst-directives/src/widget.ts
```diff
@@ -0,0 +1,110 @@
...
+) {
+  const notString = typeof field !== 'string';
+  const invalidValues = validValues && typeof field === 'string' && !validValues.includes(field);
+  if (notString) vfile.message(`Invalid ${fieldName} supplied.`);
+  if (invalidValues)
+    vfile.message(`Invalid ${fieldName} supplied must be one of (${validValues.join(' | ')}).`);
+}
+
+export const widgetDirective: DirectiveSpec = {
+  name: 'widget',
```
agoose77: Maybe anywidget 

```suggestion
  name: 'anywidget',
```

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+export type AnyWidgetDirective = {
+  /** The type of the directive */
+  type: 'anywidget';
+  /** The ES module to import */
+  esm: string;
+  /** The JSON data to initialize the widget */
+  json: Record<string, unknown>;
+  /** URL to a css stylesheet to load for the widget */
+  css?: string;
+  /** Tailwind classes to apply to the container element */
```
choldgraf: this assumes that the renderer would be using tailwind, right? should we remove the mention of tailwind in case other renderers exist that don't use tailwind?

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (invalidValues)
+    vfile.message(`Invalid ${fieldName} supplied must be one of (${validValues.join(' | ')}).`);
+}
+
+export const widgetDirective: DirectiveSpec = {
+  name: 'anywidget',
+  doc: 'Embed a MyST Widget (AnyWidget) component with data in the body',
+  arg: {
+    type: String,
+    required: true,
```
choldgraf, line 25: do we _always_ need to load a JS module? Could somebody use this just to load CSS onto the page or something? My first thought was that this could be a way for a plugin to load custom CSS even if it didn't want to change stuff with JS.

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    return [
+      {
+        type: 'anywidget',
+        esm: data.arg as string,
+        json,
+        css: (data.options?.css ?? data.options?.styles) as string | undefined,
+        class: data.options?.class as string | undefined,
+        static: data.options?.static as string[] | undefined,
```
choldgraf: we don't do anything with these static files yet, right? this is just metadata in the AST. should we be bundling the static files with the AST (and thus putting the bundling logic in mystmd) vs. asking the theme to do it?

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+import JSON5 from 'json5';
+
+export type AnyWidgetDirective = {
+  /** The type of the directive */
+  type: 'anywidget';
+  /** The ES module to import */
+  esm: string;
+  /** The JSON data to initialize the widget */
+  json: Record<string, unknown>;
+  /** URL to a css stylesheet to load for the widget */
```
choldgraf: Why "URL"? Could this be a local path as well? Or would that be future functionality to enable?

If we require this to be a URL, doesn't this assume that the thing you're referring to already  follows the anywidget interface? So if you're a plugin author you'd have to publish your JS at a URL and then tell users to refer to that URL in their use of anywidget?

I think a useful workflow for plugin authors would be 
 
- they write logic that pulls in functionality from a library that doesn't know anything about anywidget
- they write their own anywidget interface js
- they create a directive or transform that uses their interface to do stuff in myst

rowanc1:
> Embed custom styling for the handles in the github-handle-links plugin without the [hacky abomination I'm using now](https://github.com/jupyter-book/myst-plugins/blob/8d98bb32ae54bfbb826fde2ac85934417e0ebb24/plugins/github-handle-links/src/index.mjs#L149-L158)

I don't think this PR is aimed at injecting styles in a general sense, but that is something that we should allow plugins to do somehow. This is for sure a hack — taking over the math node. :(

Any widget is mostly aimed at JS (e.g. your second example of the tables) — that could be an any widget + static document to a csv of something, or a map, or a dynamic chart, or basically anything. It skips the need for Jupyter a bit for the web-renderer, which is really nice — but it also can play with Jupyter as well.

Agree on getting this in and iterating, maybe with a beta flag in the docs when it shows up there.

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+    type: String,
+    required: true,
+  },
+  validate(data, vfile) {
+    validateStringOptions(vfile, 'arg', data.arg);
+    if (data.options?.css) validateStringOptions(vfile, 'css', data.options?.css);
+    // TODO?: validate existence of the ESM/CSS files
+    if (data.options?.class) validateStringOptions(vfile, 'class', data.options?.class);
+    if (data.options?.static) {
+      if (!Array.isArray(data.options?.static)) {
```
rowanc1: This is declared as a string, but is asserted as an array. So likely not possible to pass static properties in at the moment?

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+      json = JSON5.parse(body);
+    } catch (e) {
+      json = { error: 'Invalid JSON5/JSON supplied.' };
+    }
+
+    return [
+      {
+        type: 'anywidget',
+        esm: data.arg as string,
+        json,
```
rowanc1: Maybe we could call this something clearer? This is the data that the widget is initialized with?

packages/myst-directives/src/anywidget.ts
```diff
@@ -0,0 +1,109 @@
...
+    static: {
+      type: String,
+      required: false,
+      doc: 'A file path, folder path or glob pattern to static files to make available to the module',
+    },
+  },
+  body: {
+    doc: 'JSON5 (or JSON) object with props to pass down to the component',
+    type: String,
+    required: true,
```
rowanc1: Is a JSON body actually required? Could we have a widget that can be initialized without any starting data?

agoose77: I would like to make it not-required, and set a reasonable empty default.

packages/myst-directives/src/anywidget.ts
stevejpurves: yes, anywidget always loads a JS module. The JS and CSS are rendered within a [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) for isolation, so this is not a vehicle to get global CSS on the page. People can already do that via user style sheets.

packages/myst-directives/src/anywidget.ts
stevejpurves: yes really it should just be any class name, will update

packages/myst-directives/src/anywidget.ts
stevejpurves: future to enable? I'm not 100% clear on how to get the assets into the _build/site in a clear way, i.e. that rubs up against the filename randomisation & cache key discussion?

packages/myst-directives/src/anywidget.ts
agoose77: These are all remote files. We do want to add the same kind of support as template `file` options to load local files from disk. As for bundling remote files ... I hadn't thought of that yet.

stevejpurves:
@choldgraf yes a datatable widget is a nice example, albeit one that might be involved to create unless there is an existing out of the box js library you are looking at wrapping.

note it would be good to get examples in here: https://github.com/jupyter-book/example-widgets as a testing ground as well as more extensive docs

packages/myst-directives/src/anywidget.ts
stevejpurves: good call, will action it

packages/myst-directives/src/anywidget.ts
agoose77: > Why "URL"? Could this be a local path as well? Or would that be future functionality to enable?

Yes — it's more work to enable. We don't have an existing pattern for transforms to define new static files. It's definitely a wanted-on-roadmap feature, and as @stevejpurves points out, is related to several existing static assets conversations.

packages/myst-directives/src/anywidget.ts
stevejpurves: yes static is future / wishful thinking, I think connected to the enhancement to allow local files to also be used for the esm and css. i.e. if we were to want static assets bundled how would we get myst to load them into the _build/site folder and to provide predictable urls, e.g. those static assets urls might be out of our control as an anywidget author, who is wrapping an opinionated library.

is it better to remove this for now and extend later?

packages/myst-directives/src/anywidget.ts
stevejpurves: having the `esm` and `css` be an inline code string OR a url is more doable right now, is meaningful for smaller widgets and also aligns with AnyWidget's interface over in python. e.g. the Example on this page https://anywidget.dev/en/getting-started/

stevejpurves:
Thanks for all the feedback, I'll do another quick pass on this soon. After that I think some docs is the main blocker for merge - in like with @agoose77's definition of done in the issue.

agoose77:
In bef3fbf I made some tweaks:
- moved types to myst-spec-ext
- deleted the static files (it's not in use right now)

I also made the options for CSS and ESM path or URL. I'll explain why.

I've realised that artifacts from GH Actions are not CORS-friendly! This is a real pain! It means that people will have to hack around with GitHub Pages if they want to serve content that we can fetch from another page.

Instead, I think we should take this as a signal that local files are important — the same system for loading local ESM is useful in cacheing remote ESM at build time, which ignores CORS.

I have a prototype locally but it's ugly and needs a bit of work. I will push it here to unblock people (42b8ac44), but feel free to delete/redo it as needs be.

Long term I think we should use this as a cue to support `path` option types in directives etc, so that we can generalise this pattern (or, perhaps better, a utility to "declare" a static asset) in the directive. However, for now it's probably fine to special case widgets.

packages/myst-directives/src/anywidget.ts
choldgraf: Got it - so basically styling could be applied only to the DOM elements that are created by the directive.

> People can already do that via user style sheets.

To clarify, _users_ can, but plugins cannot.

packages/myst-directives/src/anywidget.ts
agoose77: I've resolved this.

packages/myst-directives/src/anywidget.ts
agoose77: I've renamed this to `model`.

packages/myst-directives/src/anywidget.ts
agoose77: I've removed it as you suggested Steve -- this simplifies things whilst we're not sure exactly how it will work.

packages/myst-directives/src/anywidget.ts
agoose77: I've changed this to support both, as discussed below.

packages/myst-directives/src/anywidget.ts
agoose77: Reworded to match the existing convention.

agoose77:
Here's a preview of the docs:
image


@choldgraf you've spent a lot of time in our docs recently. I didn't spend a huge amount of time here because I time-boxed to get the whole PR forward. Could I ask for a glance over?

@stevejpurves can you sign off the changes?

packages/myst-directives/src/anywidget.ts
```diff
@@ -59,23 +59,23 @@ export const widgetDirective: DirectiveSpec = {
...
-      json = {};
+      model = {};
     } else {
       const body = data.body as string;
       try {
-        json = JSON5.parse(body);
+        model = JSON5.parse(body);
       } catch (e) {
-        json = { error: 'Invalid JSON5/JSON supplied.' };
+        model = { error: 'Invalid JSON5/JSON supplied.' };
```
stevejpurves, line 72: I think this is a good step 👍🏼

packages/myst-directives/src/anywidget.ts
```diff
@@ -59,23 +59,23 @@ export const widgetDirective: DirectiveSpec = {
...
       }
     }
 
     return [
       {
         type: 'anywidget',
```
stevejpurves, line 78: should we just be going with `widget`?

packages/myst-directives/src/anywidget.ts
```diff
@@ -59,23 +59,23 @@ export const widgetDirective: DirectiveSpec = {
     return data;
   },
   run(data, _vfile, _opts) {
-    let json: ReturnType<typeof JSON5.parse>;
+    let model: ReturnType<typeof JSON5.parse>;
     if (data.body === undefined) {
-      json = {};
+      model = {};
```
stevejpurves, line 66: 👍🏼

stevejpurves: thanks for getting to this @agoose77

packages/myst-directives/src/anywidget.ts
agoose77: I'm leaning that way too, actually!

packages/myst-directives/src/anywidget.ts
agoose77: My pleasure!

stevejpurves:
@agoose77 feedback on changes:

* from string -> url/file: I think the file copying change is definitely what we want, perhaps there should be an option which is an escape hatch to allow it to remain remote only, when remote? This implementation is good enough for this cut (but also if generalised could solve the other "get static files into the _build folder" issues/requests).
* just related to that should we support inline esm/css like this? or the local files are good enough/better? I feel now like there is no need? question for next PR though!
image

stevejpurves:
I am testing and seeing issues when using the `jupyter-book/example-widgets` repo; 

image

I am using the latest from this branch https://github.com/jupyter-book/myst-theme/pull/795

@agoose77 did you get the examples repo to work as expected after your change?

agoose77:
@stevejpurves and I discussed this via DM on Discord — I forgot an `await`, which introduced a race race condition.

> from string -> url/file: I think the file copying change is definitely what we want, perhaps there should be an option which is an escape hatch to allow it to remain remote only, when remote? This implementation is good enough for this cut (but also if generalised could solve the other "get static files into the _build folder" issues/requests).

Yes, I definitely think this should be generalised. I want to tackle that separately, so that it doesn't block this work. See more here: https://github.com/jupyter-book/mystmd/issues/2714!

> just related to that should we support inline esm/css like this? or the local files are good enough/better? I feel now like there is no need? question for next PR though!


I think ESM and CSS in-document is an antipattern for us. I'm also thinking about whether the same is true for JSON. For simple widgets, inline JSON is probably preferable, but for bigger ones we don't want huge 3D model definitions _in_ the bundle! We can probably add a separate `model-file` directive argument for that and combine it with a transform that moves JSON to disk. I _think_ it's OK to hold off on that and introduce backwards compatibility down the road?

stevejpurves:
> Yes, I definitely think this should be generalised. I want to tackle that separately, so that it doesn't block this work. See more here: https://github.com/jupyter-book/mystmd/issues/2714!

💯 on that @agoose77 and thanks for the replies on DM

choldgraf:
I took a pass at the documentation to build on what @agoose77 has added (as a compliment to the docs work I added over in the myst-theme PR). I think overall this is good enough to ship as-is, especially now that we have local modules, which IMO makes this much more powerful because you can wrap other libraries in your own anywidget logic).

When I did that, I also updated the nomenclature to use the phrase `MyST Widgets` and then `Widgets` (since it seemed like both @agoose77 and @stevejpurves were in favor, and I also think this is a good idea). I **have not** changed any filenames or code, because I don't want to clobber somebody's work if y'all have working branches locally (I assumed the docs were safer to change).

IMO, this is good enough to ship right now, if we do:

- label it as an `alpha` feature with an admonition at the top (ie, no promises that the API or key functionality will be stable, but we quickly hope to converge on beta-level stability)
- (optionally) create an issue where people can provide comments etc, but IMO we could also just ask people to open issues if they hit problems

That way we can invite experimentation from power users without setting up the assumption of stability.

I think this is a really cool feature, and leans into our strategy of making myst a more hackable/pluggable platform for developers!

stevejpurves:
I think we can merge as soon as CI is green

stevejpurves:
I am approving this based on out of band convos with @agoose77

PR #2616, opened 2026-01-04
feat: users can specify static files to include in their sites

### LINKED ISSUE #1921: Allow users to specify static files to include in their MyST sites
It is often useful to be able to link static assets with your site. For example, if you want to link to a video file directly (rather than embedding it),  a data file like a CSV, or an asset like an image that you want others to access directly.

### User story

As an author, I want to be able to share static assets with my MyST site in a consistent way.

### How JB1 did this

There were three primary ways you can do this in JB1 / Sphinx

- The `{download}` role - you could point this to a local file, and when you built the site it would copy it over to a `_build/html/static` folder for a permalink and link it in the doc.
- The `_static` folder - you could configure a folder with static assets and they would all be copied over to the `_build/html/static` folder at build time.
- Manual linking with the Sphinx `app` object - behaved similar to `{download}` but was called within Python by extension (I think it was something like `app.add_static_file` but may not remember properly)

bryanwweber:
Using a new field in in the site config, users can specify static files that are copied as-is to the build output folder and served by the content server.

Closes #1921 

TODO:

- [ ] Specify folders in addition to single files
- [ ] Move the content to a dedicated subdirectory (perhaps `_build/site/static` to match the `myst-theme` route?)
- [ ] Copy to the URL that the user specifies (relative to the `static` folder) rather than keeping the same structure as the filesystem
- [ ] See if a new handler needs to be added in the content server for that folder
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/packages/myst-cli/src/build/site/manifest.ts
+++ b/packages/myst-cli/src/build/site/manifest.ts
@@ -3,7 +3,7 @@ import path from 'node:path';
 import { SPEC_VERSION } from '../../spec-version.js';
 import { hashAndCopyStaticFile } from 'myst-cli-utils';
 import { RuleId, TemplateOptionType } from 'myst-common';
-import type { SiteAction, SiteExport, SiteManifest } from 'myst-config';
+import type { SiteAction, StaticAsset, SiteExport, SiteManifest } from 'myst-config';
 import type { Download } from 'myst-frontmatter';
 import {
   EXT_TO_FORMAT,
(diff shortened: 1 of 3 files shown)


choldgraf:
A quick thought from me on implementation: 


Rather than doing all the new routes and such, could we do a combination of:

1) Extend hashAndCopyStaticFile to support query-string cache busting (e.g. `myfile.png?cache=[hash]`) as an option. Then we could re-use that function for more permanent and referenceable links for files that function is called on (like the `{download}` role):

    https://github.com/jupyter-book/mystmd/blob/2919d86aad079da4136724bd1394346470828a75/packages/myst-cli-utils/src/filesystem.ts#L24-L56
2) It's used in places like the StaticFileTransformer:

    https://github.com/jupyter-book/mystmd/blob/75576e34a04c3c3a1774879f579afae2528d493f/packages/myst-cli/src/transforms/links.ts#L186-L199

3) And for example is run on static files like the download role here: https://github.com/jupyter-book/mystmd/blob/75576e34a04c3c3a1774879f579afae2528d493f/packages/myst-roles/src/download.ts#L16-L20

4) Which generates a link AST node with `static: true` and this seems to be picked up by a transform that does the hashAndCopy work

We could then add configuration like `site.static:` that runs the same logic that `{download}` and the above workflow does, more-or-less.

That might be a way to piggy-back on the current infrastructure relatively simply without adding extra routes and such in the theme. Does that make sense?

choldgraf:
I'm closing this one, as I think it'll be superceded by this PR that's getting close to being done!

- #2770

bryanwweber:
Thanks @choldgraf sorry I couldn't come back to this!

Issue #2609, opened 2025-12-28
Add a link to the label of our how-to docs in role/directive docstrings


choldgraf:
In https://github.com/jupyter-book/mystmd/pull/2557 I played around with a pattern that seemed useful: in the docstring of the `{button}` role, I included a `See [](#label)` bit at the end, where `#label` pointed to the button section. This makes it easy for a person who lands on the `button` reference section to find the deeper docs on how it works.

I suggest we do this for the other roles and directives defined in `mystmd`. This would make it much easier to discover the right content, especially since the search bar often responds with the (less verbose) reference page.

I think the steps to do this would be:

1. For each role and directive we define, check if there's a section in the documents that describes how it works.
2. If there is, then if it doesn't have a label, then add a label (e.g. `(mylabel)=`).
3. Add a `See [](#mylabel).` to the end of each docstring in the role/directive definition.

Optionally, we could also _add documentation_ for roles/directives that are defined but not documented, but depending on how many there are this might be out of scope.

PR #2630, opened 2026-01-08
Add directives and roles labels

choldgraf:
This cleans up our reference documentation around directives and roles, with the following goals:

- Add a cross-reference to our docs in the docstring of each role/directive, so users have a quick place to look for more information.
- Add extra roles that we'd created but that weren't documented. First noted here: https://github.com/orgs/jupyter-book/discussions/2625#discussioncomment-15448678
- Alphabetize those lists.

### To do

- [ ] Double check all added directives and roles and make sure they're correct
- [ ] Double check that the labels for each are in the correct space
- [ ] Confirm it's OK to have cross-reference syntax in docstrings

### Refs

- closes https://github.com/jupyter-book/mystmd/issues/2609

### DIFF
--- a/docs/quickstart-jupyter-lab-myst.md
+++ b/docs/quickstart-jupyter-lab-myst.md
@@ -94,7 +94,7 @@ More Coming Soon™
 
 ---
 
-(role:inline-expressions)=
+(inline-expressions)=
 ## Inline Expressions
 
 JupyterLab MyST allows you to create and evaluate inline expressions using the {myst:role}`eval` role.
--- a/packages/myst-roles/src/inlineExpression.ts
+++ b/packages/myst-roles/src/inlineExpression.ts
@@ -4,7 +4,7 @@ import { addCommonRoleOptions, commonRoleOptions } from './utils.js';
 
 export const evalRole: RoleSpec = {
   name: 'eval',
-  doc: 'Evaluate code and insert its output inline. See [](#role:inline-expressions).',
+  doc: 'Evaluate code and insert its output inline. See [](#inline-expressions).',
   options: { ...commonRoleOptions('eval') },
   body: {
     type: String,
--- a/docs/cross-references.md
+++ b/docs/cross-references.md
@@ -105,28 +105,28 @@ There several roles that define various kinds of references in MyST.
 In many cases, the short-hand syntax described above are aliases for these roles.
 Below we describe the most commonly-used ones.
 
-(role:ref)=
+(ref-role)=
 
 ref
 : The {myst:role}`ref` role can be used to bring the title or caption directly in line, the role can take a single argument which is the label, for example, `` {ref}`reference-target` ``
 : You can also choose the reference text directly (not taking from the title or caption) by using, `` {ref}`your text here <reference-target>` ``.
 
-(role:numref)=
+(numref-role)=
 
 numref
 : The {myst:role}`numref` role is exactly the same as the above {myst:role}`ref` role, but also allows you to use a `%s` in place of the number, which will get filled in when the content is rendered. For example, ``{numref}`Custom Table %s text <my-table-ref>`.`` will become `Custom Table 3 text`.
 
-(role:eq)=
+(eq-role)=
 
 eq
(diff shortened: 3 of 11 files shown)


packages/myst-roles/src/inlineExpression.ts
```diff
@@ -4,6 +4,7 @@ import { addCommonRoleOptions, commonRoleOptions } from './utils.js';
 
 export const evalRole: RoleSpec = {
   name: 'eval',
+  doc: 'Evaluate code and insert its output inline. See [](#role:inline-expressions).',
```
choldgraf:
```suggestion
  doc: 'Evaluate code and insert its output inline. See [](#role:eval).',
```

docs/quickstart-jupyter-lab-myst.md
```diff
@@ -94,7 +94,7 @@ More Coming Soon™
 
 ---
 
-(inline-expressions)=
+(role:inline-expressions)=
```
choldgraf:
```suggestion
(role:eval)=
```

docs/creating-pdf-documents.md
```diff
@@ -289,6 +289,8 @@ This will not be in the pdf
 +++
 ````
 
+(directive:raw)=
```
agoose77, line 292: These slightly shadow the built-in `myst:directive` directive labels (and same for roles) that are of the form `directive-raw`. Probably fine, but worth thinking about.

packages/myst-directives/src/code.ts
```diff
@@ -224,7 +224,7 @@ export const codeDirective: DirectiveSpec = {
 
 export const codeCellDirective: DirectiveSpec = {
   name: 'code-cell',
-  doc: 'An executable code cell',
+  doc: 'An executable code cell. See [](#code-cell).',
```
agoose77, line 227: RE docstrings using internal references, this now couples `myst-directives` to the myst docs. I'm tempted to say this is find -- the convenience is worth it.

rowanc1: Given that this is added for everything, I would much rather pick this up in the plugin rather than put it in the docs so that the reference is added through code rather than md.

rowanc1:
For cross refs in doc strings: I think we should add that through the mjs plugin rather than the doc string; we should have all the information to do it programatically?

docs/creating-pdf-documents.md
choldgraf: Yeah I noticed that too - want to reserve this syntax structure for those autogenerated cases?

agoose77:
@rowanc1 I thought about that too. I am not sure whether we have an injective relationship here,though — does every directive have a docs section? Maybe that's a useful constraint to impose, though -- that we have every directive documented :laughing:

agoose77:
For robustness, we would easily add a new role that creates a `mystTarget` for directive docs, e.g.

```markdown
{directive-docs}`raw`
```

rowanc1:
I like the constraint that every directive is documented. :)

choldgraf:
I'm not sure what the recommendation is given the conversation you two have had above. What's the pathway to getting this ready to merge? Should I close this PR?

Check the suggestion below, and if that's better then I can close this PR and start from scratch to try and simplify this.

choldgraf:
Do directives or transforms have access to the full `xref` information of a myst site when they run? If so, then we could do something like:

- For each directive
- Search the xrefs data to get a list of every time `{myst:directive}` is used for it.
- At the bottom of its reference section, add a list of all the places in the myst docs where that directive is referenced

Then we could programmatically generate it. It would be noisier, but much easier to maintain.

Actually, if we had this approach then we wouldn't need to manually maintain a list of directives in `directives.md` or `roles.md` at all. Those pages could just have a single directive like `{all-directives}` that would generate the docstring for each one in alphabetical order, along with backreferences to where `` {myst:directive}`mydirective` `` was used in the docs.

If we don't have access to xref information, we could do another manual parse of all the pages in the docs at build time. That's hacky but might be our only option.

choldgraf:
closing this one because I don't have headspace to think about it right now, and it wasn't as simple as I thought it would be

PR #2633, opened 2026-01-09
📖 Clarify that subtitle is also available in project configuration

lucafrance:
`subtitle` is only available on pages, as specified in the `Titles` section


rowanc1:
Pretty sure it is available on the project. This can be checked in the article theme. So the other side of this reference should be changed!

lucafrance:
@rowanc1 Done 👌

PR #2640, opened 2026-01-13
🃏 Wrap card in a passthrough typst function, that can be redefined by the user

jan-david-fischbach:
Currently `cards` have no special styling in `typst` exports. To allow templates to apply stiling to cards, I propose to add a passthrough wrapper `#let card(content) = [#content]` around the content. As such a `typst` template can redefine `card` e.g. as follows:

```
#import "@preview/shadowed:0.2.0": shadowed
#let card(content) = shadowed(radius: 6pt, inset: 10pt, fill: rgb(247,247,247), content)
```

Which would add a rounded boxshadow around the card content.

image


packages/myst-to-typst/src/index.ts
```diff
@@ -454,13 +454,24 @@ const handlers: Record<string, Handler> = {
     state.write(')\n\n');
   },
   card(node, state) {
+    state.useMacro('#let card(content) = [#content]');
+    if (!(state.data.isInFigure || state.data.isInTable || state.data.isInBlockquote)) {
+      //FIXME: this is very inelegant. How to avoid # when within a function call?
```
rowanc1, line 465: Agree. :)

packages/myst-to-typst/src/index.ts
```diff
@@ -454,13 +454,24 @@ const handlers: Record<string, Handler> = {
     state.write(')\n\n');
   },
   card(node, state) {
+    state.useMacro('#let card(content) = [#content]');
```
rowanc1, line 463: This is a pretty great approach in general!

rowanc1 (approved):
This is great. Sorry it took so long to get to.

stefanv:
I think this is a solid pattern to use across the typst generator, which will allow for easier customization.

rowanc1:
Yeah - agreed. It is super extensible over there and allows people to customize pretty easily!

PR #2642, opened 2026-01-13
Improve edit URL branch logic

choldgraf:
This is a little experiment aimed at improving the "edit this url" logic, so that if the `rev-parse` branch doesn't properly resolve (e.g., to `HEAD`) that it tries to infer the branch from `origin` instead of generating an edit link with an invalid branch.

- closes https://github.com/jupyter-book/mystmd/issues/2234
- closes https://github.com/jupyter-book/jupyter-book/issues/2480
- closes https://github.com/jupyter-book/jupyter-book/issues/2546

I also removed the netlify `ignore:` command, because the docs build can be useful when we change mystmd functionality in order to preview what those changes do to the docs, even if we don't update the docs directly. These are pretty quick builds so I feel like it shouldn't be that big of a deal.

### How to test?

I think the easiest way to test this is to look at the "edit URLs" for our Netlify previews. Here is the edit URL for this PR:

- This PR edit link: https://github.com/jupyter-book/mystmd/edit/main/docs/website-templates.md (which is correct!)
- This [other PR](https://github.com/jupyter-book/mystmd/pull/2630): https://github.com/jupyter-book/mystmd/edit/HEAD/docs/index.md

### DIFF
--- a/packages/myst-cli/src/utils/addEditUrl.spec.ts
+++ b/packages/myst-cli/src/utils/addEditUrl.spec.ts
@@ -1,6 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
-// Mock git commands to avoid system dependencies.
+// Mock git commands to make these more lightweight unit tests and avoid system deps
 // TODO: Improve with integration tests against actual git repos (see PR #2642 discussion).
 vi.mock('which', () => ({ default: { sync: () => '/usr/bin/git' } }));
 
@@ -9,7 +9,9 @@ vi.mock('which', () => ({ default: { sync: () => '/usr/bin/git' } }));
 vi.mock('myst-cli-utils', async (importOriginal) => {
   const actual = await importOriginal<typeof import('myst-cli-utils')>();
   return {
+    // The logger breaks this little test so we are going to mock it
     silentLogger: () => ({ debug: () => {} }),
+    // Mocking makeExecutable so we can intercept git commands
     makeExecutable: (command: string, log: unknown) => {
       // Mock branch detection via git
       if (command.startsWith('git symbolic-ref')) {


packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -7,6 +7,36 @@ function gitCommandAvailable(): boolean {
   return !!which.sync('git', { nothrow: true });
 }
 
+/**
+ * Infer a git branch to use for edit URLs.
+ * Uses git rev-parse, but if this returns something that isn't a branch, it uses the `origin` default branch.
+ * This is because many CI environments check out the repo in a way that rev-parse doesn't return the branch.
```
rowanc1: I think GitHub CI actually has this in an env variable, which might be easier to look up:

`process.env.GITHUB_REF_NAME`

https://docs.github.com/en/actions/reference/workflows-and-actions/variables

rowanc1: I suppose that this is actually running in netlify even for our case though?

packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -7,6 +7,36 @@ function gitCommandAvailable(): boolean {
...
+        'git symbolic-ref --short refs/remotes/origin/HEAD',
+        gitLog,
+      );
+      const defaultBranch = (await getDefaultBranch()).trim();
+      if (defaultBranch) {
+        // origin/HEAD should resolve to origin/<branch>; strip the remote prefix for URLs.
+        gitBranch = defaultBranch.slice('origin/'.length);
+      }
+    } catch {
+      // Keep original gitBranch if we cannot resolve a default branch.
```
stefanv: Should we default to `main`?

packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -7,6 +7,36 @@ function gitCommandAvailable(): boolean {
...
+ * Uses git rev-parse, but if this returns something that isn't a branch, it uses the `origin` default branch.
+ * This is because many CI environments check out the repo in a way that rev-parse doesn't return the branch.
+ */
+async function getGitBranch(gitLog: ReturnType<typeof silentLogger>): Promise<string> {
+  // First step try rev-parse
+  const getBranchRevParse = makeExecutable('git rev-parse --abbrev-ref HEAD', gitLog);
+  let gitBranch = (await getBranchRevParse()).trim();
+  // Check for a few common failure-modes and get the origin branch if found
+  // Assume detached HEAD returns "HEAD", and PR checkouts look like "pull/<id>".
+  if (!gitBranch || gitBranch === 'HEAD' || gitBranch.startsWith('pull/')) {
```
stefanv: Perhaps worth mentioning here that this handles the (near) only failure mode of this command, which happens when there is a detached HEAD?

I'm not sure what how much luck we'll then have with `symbolic-ref`—what are the odds that you have a whole branch structure available? I guess it doesn't hurt to try.

netlify.toml
```diff
@@ -8,4 +8,3 @@
   echo "User-agent: *\nDisallow: /" > _build/html/robots.txt
   """
   publish = "/docs/_build/html/"
-  ignore = "git diff --quiet HEAD^ HEAD -- docs"
```
stefanv, line 11: This is fine; agree that it's worth checking the docs under some circumstances.

packages/myst-cli/src/utils/addEditUrl.ts
stefanv: Netlify also provides some metadata: https://docs.netlify.com/build/configure-builds/environment-variables/#git-metadata

Could be worth scouring for these, although again how likely is it that these will be available but the `rev-parse` fails?

choldgraf: My reasoning here was that doing it in a platform agnostic way would be less brittle than special casing each CICD system. Do you think it's likely this approach won't work or break unexpectedly?

packages/myst-cli/src/utils/addEditUrl.ts
choldgraf: IMO that's a reasonable idea! Maybe just emit a warning like "Could not infer branch for edit button, defaulting to `main`"?

packages/myst-cli/src/utils/addEditUrl.ts
stefanv: I think fine to start with git-only approaches and see how far we get.

packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -7,6 +7,36 @@ function gitCommandAvailable(): boolean {
   return !!which.sync('git', { nothrow: true });
 }
 
+/**
+ * Infer a git branch to use for edit URLs.
+ * Uses git rev-parse, but if this returns something that isn't a branch, it uses the `origin` default branch.
+ * This is because many CI environments check out the repo in a way that rev-parse doesn't return the branch.
+ */
+async function getGitBranch(gitLog: ReturnType<typeof silentLogger>): Promise<string> {
```
stefanv: The logic here felt a bit messy: try this, then if it fails try that, etc. And then Rowan had another potential way to find the branch name, and perhaps Netlify comes into the mix too. I thought perhaps we could try a bunch of different strategies and see whether any of them pan out. Here's how Gemini suggests that could be done—looks sensible enough but would have to double check:

```typescript
async function getGitBranch(gitLog: ReturnType<typeof silentLogger>): Promise<string> {
  const exec = (cmd: string) => makeExecutable(cmd, gitLog)();

  const strategies = [
    // Get local branch name; fails if HEAD is detached
    async () => (await exec('git symbolic-ref --short HEAD')).trim(),

    // Use CI variables; essential for editable refs in detached CI runners
    async () => (
      process.env.BRANCH ||
      process.env.GITHUB_REF_NAME ||
      process.env.GIT_BRANCH ||
      process.env.CI_COMMIT_REF_NAME
    ),

    // Find the nearest local or remote branch, explicitly excluding tags
    async () => {
      const name = await exec('git name-rev --name-only --refs="refs/heads/*" --refs="refs/remotes/*" HEAD');
      return name.trim().replace(/^remotes\/origin\//, '');
    },

    // Fallback to the remote's default branch (usually origin/main)
    async () => {
      const ref = (await exec('git symbolic-ref --short refs/remotes/origin/HEAD')).trim();
      return ref.replace(/^origin\//, '');
    },
  ];

  for (const strategy of strategies) {
    try {
      const result = await strategy();
      // Filter out generic HEAD labels or specific PR-state refs
      if (result && result !== 'HEAD' && !result.startsWith('pull/')) {
        return result;
      }
    } catch {
      // Continue until a strategy provides a valid branch
    }
  }

  return 'main';
}
```

choldgraf: Ooh I like it. Let me add this in. It feels the most complete. Will get to this tomorrow

stefanv: (You'll see I modified the git commands a bit, since I thought it wouldn't be too useful if we gave back tags—we really need a branch name for editing.)

choldgraf:
OK I think I more or less implemented @stefanv 's suggestion. I've got 3 checks that I'm running now:

1. Check for the GITHUB_ACTIONS environment variable[^1]
2. Use `git` based branch check
3. Fall-back to `main`

I also added a test for some of the logic in the editURL function and comments throughout (though, the test is mocking a bunch of stuff to avoid using actual git).

Unless this latest commit screwed something up or added too much complexity, could we merge this and iterate from there?

[^1]: I'm not checking CircleCI or Netlify because I could only find environment variables for the _current branch_, not the _base branch_. I'm not including GitLab, since  it feels excessive given that the edit buttons _only_ work on GitHub right now)

packages/myst-cli/src/utils/addEditUrl.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+vi.mock('which', () => ({ default: { sync: () => '/usr/bin/git' } }));
+vi.mock('myst-cli-utils', () => ({
+  silentLogger: () => ({ debug: () => {} }),
+  makeExecutable: (command: string) => async () => {
+    if (command.includes('--show-toplevel')) return '/repo\n';
+    if (process.env.MOCK_GIT_FAIL) throw new Error('git failed');
+    if (command.includes('origin/HEAD')) return process.env.MOCK_GIT_BRANCH ?? 'origin/main\n';
+    return '';
+  },
+}));
```
agoose77: This feels like a bit of a sledgehammer. `makeExecutable` is used to wrap calls to system binaries, including but not limited to Git. We shouldn't assume every call is to a git command. Similarly for `which`.

I _think_ I'd possibly prefer actually testing against a git repo? We could create some test repos in our org.

packages/myst-cli/src/utils/addEditUrl.ts
```diff
@@ -8,28 +8,77 @@ function gitCommandAvailable(): boolean {
...
+    const branch = await getDefaultBranch(session);
+    const filePath = file.replace(gitRoot, '');
+
+    // Only set URLs that aren't already defined (and aren't explicitly null)
+    if (!frontmatter.source_url && frontmatter.source_url !== null) {
+      frontmatter.source_url = `${frontmatter.github}/blob/${branch}${filePath}`;
+    }
+    if (!frontmatter.edit_url && frontmatter.edit_url !== null) {
+      frontmatter.edit_url = `${frontmatter.github}/edit/${branch}${filePath}`;
+      session.log.debug(`Added edit URL ${frontmatter.edit_url}`);
```
agoose77, line 79: https://github.com/jupyter-book/mystmd/pull/2650/ touches this part of the code as well. Should we merge the two PRs?

choldgraf: I considered that but decided not to in order to keep the diff relatively precise. If we merge this one I am happy to rebase or merge the conflict that'll arise with the other

packages/myst-cli/src/utils/addEditUrl.spec.ts
choldgraf: what would be the simplest way to do this ? I agree these tests are kinda ugly but I wasn't sure how to set up a proper E2E test. Could you point me in the right direction?

Also just to confirm my assumptions: the `mock` stuff happening here is *just* for this editUrl unit test file right? That's why I thought it was OK to make these assumptions

choldgraf:
Discussed this one in colab cafe: @stefanv could you make a recommendation on whether we should:

- block merging this on improving the tests
- add a comment w/ the need to add better tests for this and just merge

?

docs/website-navigation.md
```diff
@@ -165,6 +165,14 @@ To override this behavior and set a manual edit URL, use the `edit_url` field in
 
 To disable the `Edit this page` button, set the value of `edit_url` to `null`.
 
+**MyST will attempt to automatically detect the base branch to use**. Our goal is to detect the "trunk" branch that a pull request, CI/CD build, etc likely came from. Here's the rough logic we follow when doing so:
```
stefanv: Do people still know what a trunk is nowadays?

stefanv (commented):
This looks fine. I think mocking is fine for now since we don't necessarily have a repo to operate on during testing, but I agree that the way the mocking is done is a bit heavy-handed. Could we do a more specific mock that checks that the command starts with `git symbolic-ref` and otherwise falls back to `makeExecutable`?

docs/website-navigation.md
choldgraf: great question haha I didn't know what to use. Maybe if I add `` usually `main` for GitHub-based projects `` it'll be clearer?

choldgraf:
ok so my plan will be to:

- simplify the mocking
- add a comment that we want to improve this with better integration testing
- merge this before the release

docs/website-navigation.md
agoose77: I saw "trunk" on the jujutsu tut recently that made me laugh: https://steveklabnik.github.io/jujutsu-tutorial/sharing-code/remotes.html

choldgraf:
well that was a fun little rabbit hole down the world of mocking in typescript, thanks for nudging me to learn a little bit there 😅.

The latest commit changes the mock logic so that we import the original `makeExecutable` as well, and only use the "mock" version if we're testing the specific `git` commands that are relevant for this change. So I think that satisfies @stefanv and @agoose77 's concerns.

I will merge this in if tests are happy!

packages/myst-cli/src/utils/addEditUrl.spec.ts
```diff
@@ -1,17 +1,32 @@
...
-    if (process.env.MOCK_GIT_FAIL) throw new Error('git failed');
-    if (command.includes('origin/HEAD')) return process.env.MOCK_GIT_BRANCH ?? 'origin/main\n';
-    return '';
-  },
-}));
+
+// Partial mock: intercept specific git commands, pass everything else to real implementation.
+// Uses vitest's importOriginal to get the real module (https://vitest.dev/guide/mocking.html)
+vi.mock('myst-cli-utils', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('myst-cli-utils')>();
```
choldgraf, line 10: omg typescript is so ugly sometimes

PR #2645, opened 2026-01-14
Minor spelling fix

tylere:
MaterMatterMatter


stefanv:
mater = car from Pixar's Cars
matter = physics

choldgraf:
oh man I like this PR even more now

PR #2648, opened 2026-01-16
Add an embed role

### LINKED ISSUE #2646: Add an {embed} role for inserting labeled content inline
### Proposal

Add a role that is able to insert labeled content inline, instead of getting the nested <div> formatting generated by the directives.

For example, here is what the `{embed}` directive produces:

```html
<div data-name="output-decoration-notebook-output">
  <div data-name="outputs-container" data-mdast-node-id="PJTgFejfcgMcnWeVp1RUz" class="max-w-full overflow-y-visible overflow-x-auto m-0 group not-prose relative text-left mb-5">
    <div data-name="safe-output-stream">
      <div>
        <pre class="myst-jp-stream-output text-sm font-thin font-system">
          <code>
            <span>I just want to insert this content!</span>
          </code>
        </pre>
      </div>
    </div>
  </div>
</div>
```

As a role, the following syntax could work:
```
To summarize a complex calculation contained on another page, the result is **{embed}`my-result-calculated-elsewhere`**.
```

### Additional notes

*Originally started as discussion #2644.*

tylere:
Adds an embed role to include content inline.
Adds documentation.

Fixes #2646.

This is my first time trying to add a role, following the [Adding a New MyST Feature](https://mystmd.org/guide/contribute-add-feature) docs. Claude Code helped too. Suggestions appreciated. I realize this may not be the best way to implement the feature, especially given the work on #1961.


tylere:
This issue has a failing check because it is not labeled. It should be labeled as "enhancement", but I don't have sufficient permissions to add the label.

rowanc1:
This is awesome - very into this feature having a role that is equivalent to the directive makes a ton of sense. Haven't had a chance to look at it specifically yet though!

PR #2650, opened 2026-01-16
Define `edit_url` as page and project frontmatter

choldgraf:
This fixes some inheritance bugs for project -> page configuration, so that edit_url now correctly propagates to pages if it is `null`.

Here are a few of the major changes:

- Adds `edit_url` and `source_url` to the list of keys that will be inherited from `project` config.
- Improves logic so that `null` is treated as a distinct thing and not as `undefined`. This was preventing `key: null` values from propagating to the page settings because `null` was treated as undefined instead of "explicitly null".

I am pretty sure that's the behavior we'd want, so I fixed it in the underlying utility function that updates config from another source. If this _violates_ the assumption that `null` !== `undefined`, then I'd hope this would pop up in the tests but they pass on my machine at least!

Also updates the docs on this a bit to clarify that the edit_url should be in `project.` in site-wide configuration. Tries to clarify this for other frontmatter as well.

- closes https://github.com/jupyter-book/jupyter-book/issues/2546

### DIFF
--- a/packages/myst-frontmatter/src/page/validators.ts
+++ b/packages/myst-frontmatter/src/page/validators.ts
@@ -18,6 +18,7 @@ import { FRONTMATTER_ALIASES } from '../site/types.js';
 
 /**
  * Frontmatter keys that inherit from project to page when not set on the page.
+ * Add new keys here if you want them to propagate from project-level config to pages.
  */
 export const USE_PROJECT_FALLBACK = [
   'authors',
--- a/packages/simple-validators/src/validators.ts
+++ b/packages/simple-validators/src/validators.ts
@@ -1,5 +1,12 @@
 import type { KeyOptions, ValidationOptions } from './types.js';
 
+/**
+ * Check if a value is neither null nor undefined.
+ *
+ * NOTE: This returns false for null, which may not be what you want.
+ * Consider using `=== undefined` checks directly if you need null to propagate as a valid value.
+ * See: https://github.com/jupyter-book/mystmd/issues/2659
+ */
 export function defined<T = any>(val: T | null | undefined): val is T {
   return val != null;
 }
@@ -456,6 +463,10 @@ export function validateList<T>(
  * We treat `null` as intentional values, not "undefined".
  * `null` in base is respected and not overwritten.
  * Explicit `null` in filler is copied when base is undefined.
+ *
+ * NOTE: We intentionally use `=== undefined` checks instead of the `defined()` function
+ * because `defined()` treats `null` as "not defined", but we need `null` to propagate
+ * See: https://github.com/jupyter-book/mystmd/issues/2659
  */
 export function fillMissingKeys<T extends Record<string, any>>(
   base: T,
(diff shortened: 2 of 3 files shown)


packages/simple-validators/src/validators.ts
```diff
@@ -461,7 +464,7 @@ export function fillMissingKeys<T extends Record<string, any>>(
 ): T {
   const output: T = { ...base };
   keys.forEach((key) => {
-    if (!defined(output[key]) && defined(filler[key])) {
+    if (output[key] === undefined && filler[key] !== undefined) {
```
agoose77, line 479: This is quite a big change (ref)
> We treat `null` as intentional values, not "undefined".

It has the potential to break existing configs that use `null` interchangeably with `undefined`. I think I'm in favour of the objective (that `null` is semantic), but I raise this concern.

agoose77:
I've added a changeset. For now, the MyST packages mostly follow a `X.major.minor` scheme (effectively), so I've bumped the middle component of the CLI and validator packages.

packages/simple-validators/src/validators.ts
choldgraf: yeah - what isn't clear to me is if this is _documented_ behavior and thus this is fixing a bug, or if this was _intended_ behavior and we are fixing the behavior.

The reason I was thinking `null` is intended to be a special thing different from `undefined` is that we we have many documented places where we use it to turn off behavior:

CleanShot 2026-01-20 at 09 14 12@2x

That said, I don't think we've explicitly documented `null` as a special kind of value.

Maybe the solution is to figure out if some things effectively defaulted to `null`, and if so, explicitly set those defaults rather than implicitly treating undefined as null?

That said, can you think of a case where this will break things? The tests were happy and this made me think it is not as risky

stefanv:
I haven't reviewed the code here in detail, but I agree with the concept, and I also think being able to nullify a value inherited from project-wide settings is very useful.

packages/simple-validators/src/validators.ts
rowanc1: You should be able to leave it as is here.

We explicitly check against undefined, so a `null` will (should) pass through?

packages/myst-frontmatter/src/page/validators.ts
```diff
@@ -28,6 +31,8 @@ export const USE_PROJECT_FALLBACK = [
   'open_access',
   'license',
   'github',
+  'edit_url',
+  'source_url',
```
rowanc1, line 36: I would be curious to see if this change alone is sufficient....

choldgraf:
We had a conversation about this in Discord and I've actioned the follow-ups. Here's a quick summary:

- We agreed that we need the `===` checks here because we (confusingly) have two `defined` functions in mystmd and the one we were using is more like `notNullish` rather than `defined`.
- I've opened this issue to track that: https://github.com/jupyter-book/mystmd/issues/2659
- I've documented that issue in the two places in the code that is relevant

Will merge this one if tests are happy and we can iteratively improve from there

PR #2661, opened 2026-01-21
Only trigger documentation update when a release succeeds

choldgraf:
I realized that our documentation was being updated for every commit to `main` because the docs deploy workflow was being called when the release workflow _succeeded_ - that happens regardless of whether an actual release is made. So this updates the workflow to only run on success

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -107,3 +107,14 @@ jobs:
     with:
       tag_name: ${{ needs.get-tag.outputs.tag_name }}
     secrets: inherit
+
+  # Update our documentation if we've successfully released
+  deploy-docs:
+    name: Deploy Documentation
+    needs:
+      - release-npm
+    if: ${{ always() && needs.release-npm.result == 'success' }}
+    uses: ./.github/workflows/docs.yml
+    with:
+      published: ${{ needs.release-npm.outputs.published == 'true' }}
+    secrets: inherit
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,10 +1,13 @@
 name: Deploy Docs
 on:
   workflow_dispatch:
-  workflow_run:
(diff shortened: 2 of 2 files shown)


bsipocz:
OTOH, it would be really really nice to have `latest` docs from main -- and that wasn't the case before, but then got fixed. 

So I would like a review on this from other team members before merging.

bsipocz (changes requested):
My 2c:

- while theoretically this is the correct approach I would like this PR not to be merged until we have a way to have multiple docs versions.

My reasoning:

We still do have so many DOC fixing PRs that if there is only one docs version I very strongly think it should be from the main branch and not from a release.

choldgraf:
I'm fine with either approach - the danger of releasing all of our docs on `main` is that we'll have docs that don't match the actual behavior. In either case, we need to release more often.

If we do _intentionally_ release on main, then we need to re-name the logic because it's confusing to have a job that runs after `release.yml` succeeds if what we really mean is "run every time `main` is updated'. I think right now we have a bug because there's a difference between the documented intent and the actual behavior

.github/workflows/release.yml
```diff
@@ -107,3 +107,14 @@ jobs:
     with:
       tag_name: ${{ needs.get-tag.outputs.tag_name }}
     secrets: inherit
+
+  # Update our documentation if we've successfully released
+  deploy-docs:
+    name: Deploy Documentation
+    needs:
+      - release-npm
+    if: ${{ always() && needs.release-npm.result == 'success' }}
```
stefanv, line 116: Doesn't this "if" already take care of the conditional publishing? Why do we need to pass it in to the workflow as well?

stefanv:
I think until we have a version switcher, the safe thing to do is to push docs out with releases. It'd be pretty easy to host parallel dev docs and add a link to it from the main docs, we just need to set that up.

Issue #2538, opened 2025-12-02
URLs with @ symbols in them are parsed as citations


choldgraf:
When a URL has an @ symbol in it and is pasted in as raw text, MyST parses it as if it were a citation node. Instead, we should detect whether the @ is actually part of the URL and not treat it as a citation.

This is a common practice with hackmd shared links. For example: https://hackmd.io/@choldgraf/rJoImy6ZWg

### Example

For example: https://hackmd.io/@choldgraf/rJoImy6ZWg

[Example sandbox demo](https://mystmd.org/sandbox?tab=ast&myst=aAB0AHQAcABzADoALwAvAGgAYQBjAGsAbQBkAC4AaQBvAC8AQABjAGgAbwBsAGQAZwByAGEAZgAvAHIASgBvAEkAbQB5ADYAWgBXAGcA)

Is parsed as:

```yaml
type: root
children:
  - type: paragraph
    children:
      - type: link
        url: https://hackmd.io/
        children:
          - type: text
            value: https://hackmd.io/
      - type: cite
        identifier: choldgraf/rjoimy6zwg
        label: choldgraf/rJoImy6ZWg
        kind: narrative
```

### Workaround

Put `< >` around the link and it'll work as expected, e.g. `<https://hackmd.io/@choldgraf/rJoImy6ZWg>`

rowanc1:
For sure a bug (I am not sure when linkify is run, but clearly after the citation parse)! [Workaround](https://mystmd.org/sandbox?tab=ast&myst=PABoAHQAdABwAHMAOgAvAC8AaABhAGMAawBtAGQALgBpAG8ALwBAAGMAaABvAGwAZABnAHIAYQBmAC8AcgBKAG8ASQBtAHkANgBaAFcAZwA%2BAA%3D%3D) is to put chevrons around it `<link>`:

```
<https://hackmd.io/@choldgraf/rJoImy6ZWg>
```

PR #2684, opened 2026-01-30
Make citation parsing less aggressive

choldgraf:
This makes the parsing logic for `@citation` text a bit less aggressive, so that `chris@gmail.com` , `[@choldgraf](https://github.com/)`, and `https://hackmd.io/@choldgraf/foo` will all *not* be parsed as citations (I think right now, they all would be).

It adds two checks:

- Are we in a link? If so, then assume it's not a citation.
- Is the citation likely in a URL (e.g. `https://...@citation`)? If so, assume it's not a citation. 

I don't think this changes any *intended citation* behavior, but we should double check that the logic looks correct. Tests passed locally but let's see what happens in CICD.

The main places where we are hitting this are in pasting github usernames into the docs (`@choldgraf`) or in hackmd links, which is fairly common in community docs (including our own!)

### Todo

- [x] Double check that we aren't being too aggressive in the opposite direction.
- [x] Maybe add more testing examples for various edgecases

---

- fixes https://github.com/jupyter-book/mystmd/issues/2538

### DIFF
--- a/.changeset/purple-dolls-listen.md
+++ b/.changeset/purple-dolls-listen.md
@@ -1,0 +1,5 @@
+---
+"markdown-it-myst": patch
+---
+
+Make citation parsing less aggressive


packages/markdown-it-myst/src/citations.ts
```diff
@@ -27,10 +27,14 @@ export const citationsPlugin: PluginWithOptions = (md) => {
     citation: /^([^^-]|[^^].+?)?(-)?@([\w][\w:.#$%&\-+?<>~/]*)(.+)?$/,
     // Only allow a short [suffix] for in text citations (e.g. 50 characters)
     inText: /^@((?:[\w|{][\w:.#$%&\-+?<>~/]*[\w|}])|\w)(\s*)(\[([^\]]{1,50})\])?/,
-    allowedBefore: /^[^a-zA-Z.0-9]$/,
+    // Check for a simple boundary before @
+    // Allows separators like ',', ';', or '|' in case we have grouped cites like @citeA|@citeB
+    allowedBefore: /^[\s(*_,;|]$/,
```
rowanc1: Can we have some new explicit tests around brackets here? I think at least `` [` `` are missing, can also be the trailing end of a role/code (backticks).

This has gone from a **not** ABC etc. to a allow list.

e.g. `[@key]` should work, and a lot of times people use `(@key)` as well (probably covered).

The `` some `code`@cite `` is possible (especially with numbered citations).

For a directive, we could also have a citation ```` ```{directive}@cite ```` ([demo](https://mystmd.org/sandbox?tab=demo&myst=YABgAGAAewB0AGkAcAB9AEAAYwBpAHQAZQAKAEIAbwBkAHkACgBgAGAAYAA%3D)).

rowanc1: So adding:

- `[`
- `` ` ``
- `}`

And expanding the test suite a bit to cover those positive cases now more explicitly.

choldgraf: I decided to revert this back to the permissive logic it had before, and just added an explicit check for "are we in a URL". I added a few more test examples to make sure it actually has the behavior we want. Does that look sensible to you?

rowanc1: Yep, that works to. Examples will help lock it down!

packages/markdown-it-myst/src/citations.ts
```diff
@@ -121,3 +130,20 @@ export const citationsPlugin: PluginWithOptions = (md) => {
...
+// Skip parsing inside URL-like text such as https://.../@path or www.example.com/@path.
+// This lets us keep allowing `/` before citations, while still filtering out URLs.
+function isUrlContext(state: InlineState): boolean {
+  // Look at the current "word" before @ (from the last whitespace to the cursor).
+  const left = state.src.slice(0, state.pos);
+  const lastSpace = left.search(/\s\S*$/);
+  const wordStart = lastSpace === -1 ? 0 : lastSpace + 1;
+  const word = left.slice(wordStart);
+  // If the word already looks like a URL, treat @ as part of the URL.
+  return word.includes('://') || word.startsWith('www.');
```
rowanc1: Making a note perhaps that this approach would isn't going to catch `hackmd.com/@rowanc1` (i.e. without the protocol).

Thinking about a URL followed by punctuation, no space and a citation:

```
In https://mystmd.org,@scipy2025, the authors...
```

A bit contrived but have seen versions of this when people have numbered citations (less common when things are narrative citations, which is the default in myst). Maybe this posture is better: if it looks like a URL and you are being sloppy about defining it, then citations might not work, so put a `<>` around your bare link to be explicit!

A note like that could go in the docs as we make the link parser is a bit more liberal (i.e. this is the way out).

choldgraf: Maybe we can add some special cases for preceding text like ".com". Will see if that's too complex

choldgraf: latest push tries to account for this - WDYT?

docs/citations.md
```diff
@@ -134,6 +134,14 @@ As with a link to a DOI, you can also use the DOI directly instead of the BibTeX
 Citations that are embedded in links will _not_ be parsed as citations.
 For example: [@jupyter-book](https://github.com/jupyter-book) or https://hackmd.io/@turingway/B1OnVLe-3.
 
+:::{note} Mixing URLs and `@citations`
+If you have citations around URLs, add a space or use `<...>` around the URL to make the citation parse.
```
stefanv: Not clear what this is meant to cover: can you add an example?

stefanv: If it ties into the example below, make that more explicit.

docs/citations.md
```diff
@@ -134,6 +134,14 @@ As with a link to a DOI, you can also use the DOI directly instead of the BibTeX
...
+
+If you use bare URLs without a scheme (e.g. `hackmd.com/@user`), the `@user` part may be treated as a citation.
+Similarly, if a URL is followed immediately by a citation without a space (e.g. `https://mystmd.org,@scipy2025`), the `@scipy2025` will be treated as part of the URL and *not* a citation.
+Angle brackets `<...>` help avoid both outcomes.
```
stefanv: Show how.

packages/markdown-it-myst/src/citations.ts
```diff
@@ -145,5 +145,7 @@ function isUrlContext(state: InlineState): boolean {
   const wordStart = lastSpace === -1 ? 0 : lastSpace + 1;
   const word = left.slice(wordStart);
   // If the word already looks like a URL, treat @ as part of the URL.
-  return word.includes('://') || word.startsWith('www.');
+  if (word.includes('://') || word.startsWith('www.')) return true;
+  // Also treat common TLDs as URL-like (e.g. hackmd.com/@user).
+  return /\.(com|org|net|io|edu|gov)(\/|$)/.test(word);
```
stefanv: This regexp feels fragile.

Perhaps easier way is to use whatever url check we do elsewhere, by prepending `https://` to the word and seeing if it passes. Or just make the regexp robust.

Something like

```
const domainRegex = /^([a-z0-9|-]+\.)+[a-z]{2,}(\/.*)?$/i;
```

choldgraf: oh interesting....so you mean use a pattern like 'if we were to add https:// to this, would it be a valid URL? A "no" would be something like "https://foobar"?

choldgraf:
I think I incorporated your regex suggestion @stefanv - lemme know! Regexes are always a bit of a black box to me, but the tests are happy :-)

choldgraf:
I decided to add in explicit test cases for the conditions mentioned here, and that uncovered a little bug where `https://jupyter.org, @myst` would be treated as "in a url" so I updated the code to check for a space just before the label-like thing in the check, and now have tests for that too. I think we are good to go if that looks good!

docs/citations.md
```diff
@@ -131,6 +131,17 @@ As with a link to a DOI, you can also use the DOI directly instead of the BibTeX
   - Citation using a DOI directly
 ```
 
+Citations that are embedded in links will _not_ be parsed as citations.
+For example: [@jupyter-book](https://github.com/jupyter-book) or https://hackmd.io/@turingway/B1OnVLe-3.
+
+:::{note} Mixing URLs and `@citations`
+If you have citations around URLs, wrap the URL in angle brackets (`<...>`) or add whitespace to ensure the citation is parsed correctly.
+
+For example, `hackmd.com/@user` will treat `@user` as a citation because there's no scheme. To avoid this, write `<hackmd.com/@user>` instead.
```
stefanv: I thought the PR now catches this case!

stefanv:
Other cases to check / decide to ignore:

- URL in brackets: `my notes file (https://hackmd.io/@user)`
- `myemail@gmail.com`

docs/citations.md
choldgraf: good point! I remove that - it's in our tests as well

choldgraf:
ok I think I cleaned that up:

- Added a test for parentheses and URL only case
- Added a note that `[URL with @ in it]` will actually be parsed as a citation (I think) because it matches the square brackets syntax
- Removed the outdated note about the URL `< >` wrapping need

I think this is good to go

packages/markdown-it-myst/src/citations.ts
```diff
@@ -121,3 +130,29 @@ export const citationsPlugin: PluginWithOptions = (md) => {
...
+  if (left.length === 0) return false;
+  // If the character immediately before @ is whitespace, not in URL context.
+  const charBefore = left[left.length - 1];
+  if (/\s/.test(charBefore)) return false;
+  // Find the word directly before @ (from the last whitespace to the cursor).
+  const match = left.match(/(\S+)$/);
+  if (!match) return false;
+  const word = match[1];
+  // Autolink already closed: <...> so the @ after it is not part of the URL.
+  if (word.startsWith('<') && word.includes('>')) return false;
```
rowanc1, line 152: I added a new condition here that terminated auto-links should not consider the citation part of a URL.

packages/markdown-it-myst/tests/citations.yml
```diff
@@ -204,6 +204,67 @@ cases:
...
+          - type: text
+            content: '.'
+          - type: cite
+            meta:
+              label: nospace
+              kind: narrative
+      - type: paragraph_close
+  # Autolink then text with @: @ is not citation (no allowed char before it), so literal text.
+  - title: Autolink then letters@nospace is link and text
+    md: '<https://mystmd.org>letters@nospace'
```
rowanc1, line 282: This is an example in the docs that didn't actually work!

packages/markdown-it-myst/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
```
rowanc1: Added config that triggers test reruns based on yaml changes.

docs/citations.md
```diff
@@ -131,6 +133,17 @@ As with a link to a DOI, you can also use the DOI directly instead of the BibTeX
...
 
+Citations that are embedded in links will _not_ be parsed as citations.
+For example: [@jupyter-book](https://github.com/jupyter-book) or https://hackmd.io/@turingway/B1OnVLe-3.
+
+:::{note} Mixing URLs and `@citations`
+If you have a citation directly after a URL without whitespace, it may be treated as part of the URL rather than as a citation.
+
+For example, `https://mystmd.org,@scipy2025` will treat `@scipy2025` as part of the URL. To cite after a URL, add a space (`See https://mystmd.org, @scipy2025`) or use angle brackets (`See <https://mystmd.org>,@scipy2025`).
+
+**Avoid URLs with @ in square brackets**. A URL with `@` inside square brackets like `[https://hackmd.io/@user]` will be parsed as a citation, to avoid this you can escape the first bracket with a backslash `\[`.
```
rowanc1, line 144: Added tests for these and gave a workaround.

packages/markdown-it-myst/tests/citations.yml
```diff
@@ -204,6 +204,93 @@ cases:
...
+          - type: text
+            content: ' or '
+          - type: code_inline
+            content: '@code'
+          - type: text
+            content: ' or (https://hackmd.io/@user)'
+      - type: paragraph_close
+  # URL followed by space then citation - citation IS parsed (whitespace breaks URL context)
+  - title: Citation after URL with space is parsed
+    md: 'See https://mystmd.org, @scipy2025 for details or <https://mystmd.org>.@nospace'
```
rowanc1, line 230: This now works: explicit terminated links.

packages/markdown-it-myst/vitest.config.ts
rowanc1: This was making CI break for some reason, removing for now.

Issue #2127, opened 2025-06-25
Support URLs for specifying the location of parts in config


brian-rose:
## Proposal

`site.parts` expects a reference to a markdown file, e.g.
```
site:
  parts:
    footer: footer.md
```
([here](https://mystmd.org/guide/website-navigation#navigation-footer) in the docs])

Related to #2126, we would like to be able to define footer via a remote file. Eg:

```
site:
  parts:
    footer: https://myorg.com/footer.md
```

So the proposal here is to support full URLs to files for `site.parts`

brian-rose:
I looked for duplicate issues about this but didn't find any! Happy to be told otherwise.

Arielle-Bennett:
Hiya, I'd also be interested in this - I'm due to be updating documentation across two different JupyterBooks and so being able to reference just one location as a `site.part` and have it render directly in the other book as well will massively help with maintenance and versioning.

choldgraf:
@brian-rose and @Arielle-Bennett I think I've got a fix for this here!

- https://github.com/jupyter-book/mystmd/pull/2687

choldgraf:
ba da boom!

PR #2687, opened 2026-02-02
Add support for remote parts

choldgraf:
This adds an extra step to the part resolving logic that will check if a `part:` is a URL, and download it locally if it is. I added a few more comments to that section in general so it's a bit easier to follow, somebody should double-check my language there to make sure it is correct!

I added a little test and a mock so that we don't actually trigger a download, but I think it still tests most of the important bits of this. 

👉 [See the footer in the docs preview](https://deploy-preview-2687--mystmd.netlify.app/)

---

- closes https://github.com/jupyter-book/mystmd/issues/2127

### DIFF
--- a/.changeset/kind-rice-pull.md
+++ b/.changeset/kind-rice-pull.md
@@ -1,0 +1,6 @@
+---
+"myst-cli": patch
+"mystmd": patch
+---
+
+Add support for remote parts


choldgraf:
argh linting is unhappy because there's a circular dependency. I'll try moving the 'resolvepath' function into its own little module, and will see if I can write a test for taht while I'm at it, since I don't think it's directly tested (?)

agoose77:
Love this. I want to think about how we can possibly implement relative references, so that remote `myst.yml` can link to parts that aren't URLs, and other people can resolve them. This is a useful step towards making extends more useful!

PR #2690, opened 2026-02-04
Fix extends configuration with parts

### LINKED ISSUE #2126: extends ignores site.parts.footer
## Description

Currently we have a footer defined in a [remote yaml file](https://github.com/ProjectPythia/pythia-config/blob/main/pythia.yml) that we are reading in multiple repos that house individual myst sites via `extends`, e.g. [here]()

The intent is to have the same remotely-defined footer appear in multiple rendered sites.

Currently, the footer does not appear. E.g. there is no footer in [this site](https://foundations.projectpythia.org) which is rendering from [this myst.yml](https://github.com/ProjectPythia/pythia-foundations/blob/247115fbec2da4af2fd3643b2e5a975617f5aef0/myst.yml#L4) which `extends` the remote file I linked above.

## Proposed solution

Based on discussion on discord, this seems to be a bug in the handling of `extends`. 

I guess my proposal is that if `site.parts.footer` is fully specified in an upstream file with `extends`, then the footer should be built.

choldgraf:
I think I figured out what's going on with #2126 .

When we loaded configuration via `extends:`, we're loading the config from a cached folder, not the project folder. But the logic for grabbing parts        assumed that we just needed the parent folder of the config file, which is true if it's our project myst.yml but not if it's a cached config file we downloaded.

This was causing the parts from an "extends:" config to be missed in the final build. So this PR calculates the parent folder of the initial myst.yml file (the projectPath) and explicitly passes that to subsequent places where we expect "the path of the parent of the main myst.yml file".

I also updated the test to check for this case.

**I accidentally pushed and then reverted this change to `main`**. (That felt safer than force-pushing). The worktree in my vscode somehow got synced to `main` so when I clicked the "sync" button it pushed to main...that's the first time that has happened to me, sorry about that!

---

- fixes #2126

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -197,7 +197,7 @@ project:
     - format: pdf
 ```
 
-The final build will include **both** a `meca` and a `pdf` export.
+The final project configuration will include **both** a `meca` and a `pdf` export.
 
 :::{note}
 The fields `exports` and `downloads` are deduplicated by `id`, so if you wish to override a value from an inherited configuration you may assign it the same `id`. Other fields cannot be overridden; instead, shared configurations should be as granular and shareable as possible.


agoose77:
I don't think this PR necessarily fixes #2126.

In **this PR** we're introducing  / formalising an assumption that `parts` (and other resources, implicitly) in a `myst.remote.yml` should be resolved relative to the local project `myst.local.yml`. 

However, **#2126 implicitly expects the opposite** -- that a `part.remote.md` defined adjacent to some `myst.remote.yml` can be resolved when `myst.local.yml` extends from `myst.remote.yml`. That's why they've inlined the footer, so that it comes with the myst.yml from the upstream repo.

We have therefore a design choice: 

1. If we resolve relative to the _local_ myst.yml, we can use the extension mechanism in a template-like way.
2. If we resolve relative to the _remote_ myst.yml, we can re-use config in a pre-baked kind of way.

I'm not sure how reasonable it is to do (2). If we have 
```yaml
parts:
  footer: part.remote.md
```
in `myst.remote.yml`, it's obvious how to retrieve `part.remote.md` if we're linking to `myst.remote.yml`** by a local file path**:
```yaml
extends:
  - ../myst.remote.yml
```
— we simply resolve `part.remote.md` relative to `../myst.remote.yml`.

But, _can_ / _should_ we make the same assumption if `myst.remote.yml` is referred to by a URL? E.g.
```yaml
extends:
  - https://foo.com/site/myst.remote.yml
```

Is it reasonable to resolve `part.remote.md` relative to `https://foo.com/site/myst.remote.yml` — `https://foo.com/site/part.remote.md`? This is something I touched upon in https://github.com/jupyter-book/mystmd/pull/2687#issuecomment-3843117337.

choldgraf:
I think it makes sense to design the right pattern more thoughtfully. That said, IMO the current functionality is totally fine as a first step. I'd suggest that we initially focus on fixing this bug, and then just being really clear in the documentation that you cannot use relative paths for now, create an issue to track the feature, and link that issue in the docs. Does that make sense?

I've given a more fleshed out version of that in the issue you linked!

choldgraf:
I went ahead and tried to clarify this a bit in [`ebee196` (this PR)](https://github.com/jupyter-book/mystmd/pull/2690/commits/ebee1963dff21ea0e161c602f51a8ff2b299f86e) - let me know what you think @agoose77 !

docs/configuration.md
```diff
@@ -169,19 +169,87 @@ project: ...
...
+
+```{code-block} yaml
+:filename: myst.yml
 version: 1
+extends: ./small.yml
 project:
-  export:
-    format: meca
+  exports:
+    format: pdf
```
agoose77:
```suggestion
    - format: pdf
```

docs/configuration.md
```diff
@@ -169,19 +169,87 @@ project: ...
...
 ```
 
-Then, any `myst.yml` file that extends this file will have a `meca` export in addition to any other exports it defines. This behavior applies to the list fields: `tags`, `keywords`, `exports`, `downloads`, `funding`, `resources`, `requirements`, `bibliography`, `editors`, and `reviewers`. The fields `exports` and `downloads` are deduplicated by `id`, so if you wish to override a value from an inherited configuration you may assign it the same `id`. Other fields cannot be overridden; instead, shared configurations should be as granular and shareable as possible.
+The final build will include **both** a `meca` and a `pdf` export.
```
agoose77:
```suggestion
The final project configuration will include **both** a `meca` and a `pdf` export.
```

choldgraf:
thanks @agoose77 ! I'll merge this one in since you've approved and I think it's in a good enough state :-)

PR #2698, opened 2026-02-10
🏷️ Render span identifiers as Typst labels

luukfroling:
update span handler to include node identifiers as Typst tags, which allows {index} to render properly.


rowanc1:
This looks good! Could you add a few test cases as well, they should be in the `.yml` files in the test directory.

packages/myst-to-typst/src/index.ts
```diff
@@ -418,6 +418,9 @@ const handlers: Record<string, Handler> = {
   },
   span(node, state) {
     state.renderChildren(node, 0, { trimEnd: false });
+    if (node.identifier) {
+      state.write(` <${node.identifier}>`);
```
rowanc1: For a span, do we want the `#label("${node.identifier}")` syntax?

I am not sure...

https://typst.app/docs/reference/foundations/label

For example, if you have a labeled span in a header (silly, but maybe illustrative).

luukfroling:
@rowanc1 thanks for the review! I have added some test cases, and changes the typst syntax to #label() as I thought it made more sense. You were right about a labeled span in a header potentially breaking things, which is solved now. 

I'm still getting familiar with typst, so if you have any other comments please let me know!

FreekPols:
@rowanc1 can you check whether all comments have sufficiently be addressed? Thnx in advance! (Luuk has not much time left on working on the project, hence the request)

Issue #2655, opened 2026-01-20
DOCX references section is jumbled with incorrect formatting


VRConservation:
myst v1.7.1

## Description
Jumbled rendering of word documents for downloads

## Proposed solution
Please fix so that each reference has a hard return between citations. See below for how it currently renders

## Additional notes

<img width="799" height="429" alt="Image" src="https://github.com/user-attachments/assets/66d5e4a7-07c1-4c68-9eb2-f60209fb1280" />

cmarmo:
Hi @VRConservation , do you mind providing the markdown file you rendered? Or a smaller version of it, so it can be easier to reproduce for possible contributors?
Thanks!

VRConservation:
Hi @cmarmo, I haven't yet published the article that has the paper above, but [this](https://github.com/VRConservation/neca_jpa) repo has a similar md file with myst.doi.bib that will export a garbled references section if you run myst build --docx with it. Note that the docx in this repo is correct because I manually put in the hard returns separating the references. Thanks for looking into this.

cmarmo:
Hi @VRConservation I have opened a pull request, available  for you to check if it's doing what you expect (#2677).
The failing check is related to a label issue ... if I understand correctly.

VRConservation:
Hi @cmarmo, Just upgraded to the latest version of myst and ran the --docx export and still have the same issue. Here's how the references cited are rendered:

[...]

PR #2699, opened 2026-02-13
🖊️ Separate references into paragraphs when exporting to docx

### LINKED ISSUE #2711: DOCX references are sorted by HTML tag
When exporting to HTML, the references section is ordered by HTML tags, which disrupts the expected locale ordering:

Image

cmarmo:
Dear maintainers,
this pull request adds a line break to each html reference node, so the translation to docx keeps a hard return after every reference.
Fixes https://github.com/jupyter-book/mystmd/issues/2655 and closes #2711

This is the same as #2677 with the correct branch name so the label check is working.

A more aesthetic solution would probably have been to add one more level of nesting in the node structure before serializing... but this was above my code understanding.
Hope this can help anyway.

Thanks for considering this PR

### DIFF
--- a/packages/myst-cli/src/build/docx/single.ts
+++ b/packages/myst-cli/src/build/docx/single.ts
@@ -77,12 +77,9 @@ function defaultWordRenderer(
   if (referenceNodes.length > 0) {
     serializer.render(createReferenceTitle());
 
-    const breakNode = {
-      type: 'break',
-    };
     const referencesRoot = {
       type: 'root',
-      children: referenceNodes.map((node) => [node, breakNode]).flat(),
+      children: referenceNodes,
     };
     serializer.renderChildren(referencesRoot);
     serializer.closeBlock();


agoose77:
@cmarmo thank you for this PR! Although I slightly tweaked what we do here, it was super useful to immediately end up in the right spot!

I did as you described 
> add one more level of nesting in the node structure before serializing

and also fixed a bug I noticed with the reference sorting — now, emphasis doesn't affect the reference position.

image

agoose77:
Thanks for this fix @cmarmo!

agoose77:
I don't think the deploy preview is failing due to _this_ PR. Merging!

cmarmo:
>  thank you for this PR! Although I slightly tweaked what we do here, it was super useful to immediately end up in the right spot!

@agoose77 thank you for your nice review!

> I did as you described
> 
> > add one more level of nesting in the node structure before serializing

I am ready for vibe-coding...  😂

PR #2702, opened 2026-02-13
Fix URL generation bug in toc directive

### LINKED ISSUE #2701: Table of contents directive results in incorrect URL paths when using folder URLs
### Reproduce the behavior

- go to https://jupyterbook.org/stable/get-started/
- click on `Get started with executable code`
- it sends you to https://jupyterbook.org/stable/execution.execution which is a 404
- it should instead be `https://jupyterbook.org/stable/execution/execution` because the site is using folder URL patterns

choldgraf:
This applies `slugToURL` in the toc transform - it's what we use everywhere else but we weren't calling it here.

---

- closes #2701

### DIFF
--- a/.changeset/hot-tips-develop.md
+++ b/.changeset/hot-tips-develop.md
@@ -1,0 +1,5 @@
+---
+"myst-transforms": patch
+---
+
+Fix URL generation bug in toc directive


packages/myst-transforms/src/toc.ts
```diff
@@ -73,7 +80,7 @@ function transformPage(page: ProjectPage, projectSlug?: string): Text | Link {
   } else if (slug != null) {
     return {
       type: 'link',
-      url: `${projectSlug ? `/${projectSlug}` : ''}/${slug}`,
+      url: `${projectSlug ? `/${projectSlug}` : ''}/${slugToUrl(slug)}`,
```
rowanc1, line 83: 👍

PR #2704, opened 2026-02-14
Document folder and landing page behavior

choldgraf:
This documents behavior that we've already got but didn't mention: if a page is at `folder/index.md`, then the URL that's rendered becomes `/folder/`

It also de-duplicates some folder config instructions and points to a source of truth.

---

- closes https://github.com/jupyter-book/mystmd/issues/2372


choldgraf:
I'm gonna merge this one in because it's *just* documentation, and I think it just makes functionality easier to discover and understand.

Issue #2685, opened 2026-01-30
Allow `{toc}` directive to display child pages of the current section


choldgraf:
The most common project-level use I've seen for the `{toc}` directive is that it can serve like an "index" page list. For example if you have a landing page for a section, you want something like:

````
```{toc} Section contents
```
````

And it'll list all of the child pages that are underneath in the toc tree.

However, this isn't currently supported!

- You have `:context: project` which displays the TOCtree of the entire project
- You have `:context: section` which displays the headers of the _current_ section

But you don't have the equivalent of "display the TOCtree of only the current page and its children".

### Ideas for implementation

1. You could create a new `context` e.g. `:context: children` which would list the child pages of the current page (if any)
2. You could add a kwarg like `:root: current` which would set the "root" item from `:context:` to be the current page or section. In this case, you could remove the `:context: section` entirely and replace it with `:context: page`, `:root: current`

PR #2705, opened 2026-02-14
Add children context to toc directive

choldgraf:
This adds an extra argument for `:context: children`, so that the `{toc}` can show child pages of the current page. This is the most common use-case for `toctree` in Sphinx, so that you can define parent index pages that list their sub-pages. Once this is released I'll update the [get started guide](https://jupyterbook.org/stable/get-started/) so that it lists only the subpages in there and not all pages in the project.

Need to double-check the implementation logic but IMO it's good for review, it basically just mirrors a lot of the project-based context.

🖼️ demo: I used the [deploy websites section](https://deploy-preview-2705--mystmd.netlify.app/deployment/#services-for-static-web-hosting) to demonstrate this in our docs!

---

- closes https://github.com/jupyter-book/mystmd/issues/2685
- last issue identified via the docs refactor effort! https://github.com/jupyter-book/jupyter-book/issues/2528

### DIFF
--- a/packages/myst-transforms/src/toc.ts
+++ b/packages/myst-transforms/src/toc.ts
@@ -201,6 +201,7 @@ function transformChildrenTocs(
   }
 
   childrenTocs.forEach((toc) => {
+    // Filter out pages based on `depth` kwarg
     const filteredPages = toc.depth
       ? childPages.filter((page) => page.level - currentLevel <= toc.depth)
       : childPages;


packages/myst-transforms/src/toc.ts
```diff
@@ -173,8 +167,53 @@ function transformProjectTocs(
...
+  const currentLevel = pages[currentIndex].level;
+  // Extract child pages: pages after the current one with deeper level, stopping at equal/shallower
+  const childPages: ProjectPage[] = [];
+  for (let i = currentIndex + 1; i < pages.length; i++) {
+    if (pages[i].level <= currentLevel) break;
+    childPages.push(pages[i]);
+  }
+
+  childrenTocs.forEach((toc) => {
+    const filteredPages = toc.depth
```
choldgraf, line 205:
```suggestion
    // Filter out pages based on `depth` kwarg
    const filteredPages = toc.depth
```

PR #2716, opened 2026-02-25
:package: Bump Node.js version minimum

agoose77:
Node.js 20 is the "oldest" supported Node.js runtime. We should follow suit.

This was motivated by seeing CI failures in myst-theme, which uses experimental features in Node 18. We should keep the two projects in sync.

### DIFF
--- a/packages/mystmd-py/src/mystmd_py/main.py
+++ b/packages/mystmd-py/src/mystmd_py/main.py
@@ -8,8 +8,8 @@ import textwrap
 
 from .nodeenv import find_any_node, PermissionDeniedError, NodeEnvCreationError
 
-
-NODEENV_VERSION = "20.0.0"
+MIN_NODE_VERSION: int = 20
+NODEENV_VERSION: str = "20.0.0"
 
 
 def ensure_valid_version(node_path, node_env):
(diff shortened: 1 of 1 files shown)


packages/mystmd-py/src/mystmd_py/main.py
```diff
@@ -22,9 +22,10 @@ def ensure_valid_version(node_path, node_env):
     if major_version_match is None:
         raise SystemExit(f"MyST could not determine the version of Node.js: {_version}")
     major_version = int(major_version_match[1])
-    if not (major_version in {18, 20, 22} or major_version > 22):
+    # Require an LTS release
+    if major_version < 20:
         raise SystemExit(
-            f"MyST requires node 18, 20, or 22+; you are running node {major_version}.\n\n"
+            f"MyST requires Node.js >= 20; you are running node {major_version}.\n\n"
```
agoose77: @rowanc1 I implemented this check for LTS-like versions. 

My feeling is that rather than imposing too strict a check here, we just require a "new-enough" version, and handle any problems for specific non-LTS versions in our issue tracker. 

That will make it easier to keep updating this. Long-term, we can pull this from package.json

rowanc1: Yeah, I like this better.

Issue #2717, opened 2026-02-25
Strange ROR behaviour


JimMadge:
## Description

I think there are some problems with the links created from RORs,

- No link text is automatically generated
- Hovering over a link results an error (using `myst start`)

## Additional notes

Minimal example,

```Markdown
# Hello world!

Just a test

- https://ror.org/029chgv08
- [](ror:029chgv08)
- [link text](ror:029chgv08)
```

<img width="1006" height="754" alt="Image" src="https://github.com/user-attachments/assets/88ba858c-9c62-4efe-a9ab-2ebc27264d78" />

agoose77:
Nice catch @JimMadge! Looks like the ROR API has updated. We'll get a fix out in the next release (hopefully today!)

JimMadge:
@agoose77 That was quick :tada: 🤩

PR #2718, opened 2026-02-25
:globe_with_meridians:  Use v2 ROR API

agoose77:
Fixes https://github.com/jupyter-book/mystmd/issues/2717

### DIFF
--- a/.changeset/flat-ends-clap.md
+++ b/.changeset/flat-ends-clap.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Use V2 ROR API


packages/myst-cli/src/transforms/ror.ts
```diff
@@ -105,22 +109,30 @@ export async function transformLinkedRORs(
...
+
+      // If the link text is the ROR, update with a organization name
+      node.children = [{ type: 'text', value: displayEntry.value }];
       return true;
     }),
-  ]);
-  if (number > 0) {
-    session.log.info(toc(`🪄  Linked ${plural('%s ROR(s)', number)} in %s for ${path}`));
+  );
+  const nLinked = statuses.reduce((total, current) => total + +current, 0);
```
agoose77, line 133: I changed this to use a more functional style.

choldgraf:
I restarted the test but it's still failing - any idea why that might be?

I am wondering if there's value in adding a test for things like the external APIs that we depend on. That way we can discover breakage like this without users needing to hit it first.

I also know we don't want to run external API hits on every test run probably. What if we had a github workflow that ran once a day and simply hit all of the APIs we depend on to ensure they continue to resolve in the same way? I don't think it's strictly blocking this PR but it _would_ be nice if we could add a test somehow for this in the future

agoose77:
I'm not sure why, but I didn't see any API issues in this PR.

We do need better tests to catch things like this.

choldgraf:
@agoose77 one idea is we could build the docs as part of one test, and then parse the resolved HTML or .json or something on a specific page that has all of the external API deps demo-ed on it

PR #2725, opened 2026-02-27
Fix bug by adding multi-line to role parsing

### LINKED ISSUE #2724: markdown-it-myst: Inline roles fail to parse when backtick body spans multiple lines
## Summary

The inline role parser in `markdown-it-myst` fails to recognize roles whose backtick-delimited body spans more than one line. For example:

```markdown
See {ref}`Long Reference
Title` for details.
```

This is silently treated as plain text instead of being parsed as a role. The raw `` {ref}`...` `` markup leaks through to output formats (HTML, ipynb, etc.).

## Root Cause

The `ROLE_PATTERN` regex in [`packages/markdown-it-myst/src/roles.ts` (line 35)](https://github.com/jupyter-book/mystmd/blob/main/packages/markdown-it-myst/src/roles.ts#L35) uses `.+?` to capture the content between backticks:

```js
_x = new RegExp('^\\{\\s*([^}]+?)\\s*\\}(`+)(?!`)(.+?)(?<!`)\\2(?!`)');
```

By default, `.` does **not** match `\n`, so the regex fails when the role body contains a newline. The Safari fallback on [line 39](https://github.com/jupyter-book/mystmd/blob/main/packages/markdown-it-myst/src/roles.ts#L39) has the same limitation.

### 1. markdown-it delivers multi-line content to inline rules

When markdown-it tokenizes a paragraph that spans multiple source lines, the inline token's `content` preserves the `\n`:

```js
const md = new MarkdownIt();
const tokens = md.parse('Hello {ref}`foo\nbar` world', {});
const inlineToken = tokens.find(t => t.type === 'inline');
console.log(inlineToken.content);
// → "Hello {ref}`foo\nbar` world"   (newline preserved)
```

The parser infrastructure supports multi-line inline content — only the role regex blocks it.

### 2. Standard CommonMark code spans already work across lines

markdown-it's built-in `backticks` rule handles newlines correctly:

```js
const tokens = md.parse('Hello `foo\nbar` world', {});
// → code_inline token with content "foo bar"
```

Per the [CommonMark spec §6.1](https://spec.commonmark.org/0.31.2/#code-spans), line endings in code spans are treated as spaces. Since MyST roles use the same backtick delimiters, users naturally expect the same cross-line behavior.

### 3. reStructuredText allows it

The [RST spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules) explicitly supports inline markup (including interpreted text / roles) spanning multiple lines within a paragraph. Reference names are documented as "whitespace-neutral." Since MyST roles are the direct equivalent of RST's `:role:`text``, this is a compatibility gap.

### 4. Real-world impact

This was discovered while investigating [QuantEcon/mystmd#8](https://github.com/QuantEcon/mystmd/issues/8). In the [QuantEcon lectures](https://github.com/QuantEcon/lecture-python-intro.notebooks) source files, some `{ref}` roles have their body split across two lines (due to line-wrapping conventions). During `myst build --ipynb`, these roles are not parsed and the raw `` {ref}`...` `` syntax leaks into the notebook markdown cells as plain text.

## Suggested Fix

Add the `s` (dotAll) flag to both regex variants so that `.` matches `\n`:

```diff
  try {
-   _x = new RegExp('^\\{\\s*([^}]+?)\\s*\\}(`+)(?!`)(.+?)(?<!`)\\2(?!`)');
+   _x = new RegExp('^\\{\\s*([^}]+?)\\s*\\}(`+)(?!`)(.+?)(?<!`)\\2(?!`)', 's');
  } catch (err) {
-   _x = /^\{\s*([^}]+?)\s*\}(`+)(?!`)(.+?)\2(?!`)/;
+   _x = /^\{\s*([^}]+?)\s*\}(`+)(?!`)(.+?)\2(?!`)/s;
  }
```

**Why this is safe:** markdown-it processes each paragraph's inline content independently. A blank line (paragraph break) will never appear inside an inline token's `content`, so there is no risk of a role accidentally spanning across paragraphs.

**Additional consideration:** Following CommonMark code span behavior, newlines within the matched role body should probably be normalized to spaces during post-processing (similar to how code spans work), though this is a design decision for the maintainers.

## Reproduction

```markdown
This {ref}`works on one line`.

But {ref}`this does
not work`.
```

The first role parses correctly. The second is treated as plain text.

choldgraf:
This is a simple addition to the regex so that we can parse role content that spans multiple lines. This brings it inline with how we process multi-line literals like this:

```
`open
closed`
```

It does this by letting the regex wildcard match `\n` characters. So I think this makes our behavior more consistent

While I was at it, I added a little bit of docs about how `markdown-it-myst` relates to the internal `mystmd` document engine.

---

- closes #2724

### DIFF
--- a/.changeset/big-knives-fly.md
+++ b/.changeset/big-knives-fly.md
@@ -1,0 +1,5 @@
+---
+"markdown-it-myst": patch
+---
+
+Add multi-line to role parsing


agoose77:
I'm mindful that this kind of innocuous change can have unseen consequences on existing projects ... but let's merge it and wait for any reports.

choldgraf:
Yeah I was worried about that too - my rationale for this is that it brings role parsing in-line with in-line literal parsing, and for that reason it felt more like a bug to fix than a feature to change. ie to me if you expect this to work:

```
Some `foo
bar`
```
then users should also expect this to work unless we explicitly tell people it won't:

```
Some {thing}`foo
bar`
```

PR #2730, opened 2026-03-02
Document corresponding author behavior

choldgraf:
This documents how e-mails can be shown for multiple authors by setting `corresponding: true` for many. It clears up some confusion a user [reported in Discord here](https://discord.com/channels/1083088970059096114/1126886232060870678/1478054121209729034).

### DIFF
--- a/docs/authorship.md
+++ b/docs/authorship.md
@@ -192,6 +192,7 @@ See [the frontmatter affiliations table](#table-frontmatter-affiliations) for a 
 ## Display author emails
 
 An author's `email` is only shown if `corresponding: true` is also set.
+If `corresponding` is unset, the first author in the list who has an email is chosen as the corresponding author.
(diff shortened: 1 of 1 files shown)


docs/authorship.md
```diff
@@ -188,7 +188,23 @@ affiliations:
 
 See [the frontmatter affiliations table](#table-frontmatter-affiliations) for a complete list of the affiliations metadata you can use.
 
-## Add social media links 
+(author:email)=
+## Display author emails
+
+An author's `email` is only shown if `corresponding: true` is also set.
```
rowanc1, line 194: Something here about the default if unset:


```suggestion
An author's `email` is only shown if `corresponding: true` is also set.
If `corresponding` is unset, the first author in the list who has an email is chosen as the corresponding author.
```

choldgraf:
thank you for the clarification! merged that in!

PR #2736, opened 2026-03-03
🧪 Add names and a changeset to CICD action

choldgraf:
This adds a changesets check to our linter workflow, and also adds names to workflow steps so they're more human readable.


rowanc1 (approved):

rowanc1:
Picking up timeout failure in #2759.

PR #2740, opened 2026-03-05
🧹 Fix json type annotation in anywidget directive

fwkoch:
The `ReturnType<typeof JSON5.parse>` annotation was resolving to `unknown` and causing ts build errors downstream. This type annotation now matches the expected type of `AnyWidget.model`. Potentially we should add some guards against `JSON5.parse` resolving to [array, string, null, etc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#return_value), but for now I just wanted to ship this type change quickly without any code changes.


fwkoch:
This is a single type annotation fix; no code change and no changeset required. I'm going to go ahead and merge.

PR #2745, opened 2026-03-06
👨🏻‍💻 Allow code-cell directive to use code options

### LINKED ISSUE #2744: Allow line numbers to show up in code cells for notebooks (rather than just code blocks)
We allow code blocks to include [highlighting and line numbers](https://mystmd.org/guide/code#numbering-and-highlighting) via config like this:

````md
```{code}
:linenos:
:emphasize-lines: 2,3
...
````

However it's not currently possible to trigger this behavior in notebook code cells either from `.ipynb` or from `{code-cell}` directives. This feels like a natural case where showing, highlighting, and referring to lines helps a lot, so we should find a way to enable this here as well.

### PR DESCRIPTION
**Pull request by choldgraf**
This connects the `CODE` directive options to the `CODE CELL` directive, so it can re-use things like linenos.

It does *not* add this functionality to `ipynb` cell parsing, only the `{code-cell}` directive. I think ipynb would need to hook this into the comment parsing and that felt more complex, though if somebody knows an easy way to do this I could add it to this PR

---

- closes #2744

### DIFF
--- a/packages/myst-directives/src/code.ts
+++ b/packages/myst-directives/src/code.ts
@@ -252,7 +252,7 @@ export const codeCellDirective: DirectiveSpec = {
       lang: data.arg as string,
       executable: true,
       value: (data.body ?? '') as string,
-      ...getCodeBlockOptions(data, vfile)
+      ...getCodeBlockOptions(data, vfile),
     };
     const outputs = {
       type: 'outputs',

### CODE REVIEW

File: packages/myst-directives/src/code.spec.ts [hunk 1]
```diff
@@ -114,3 +114,21 @@ describe('Code block options', () => {
...
+      node: {} as any,
+      arg: 'python',
+      body: 'print("hello")',
+      options: { linenos: true, 'emphasize-lines': '2,3' },
+    };
+    const result = codeCellDirective.run!(data, vfile, { parseMyst: () => [] } as any);
+    const block = result[0];
+    const code = block.children![0];
+    expect(code.showLineNumbers).toBe(true);
+    expect(code.emphasizeLines).toEqual([2, 3]);
```
stefanv, line 132: I'm not familiar with how directives get tested, so would be good to have someone's eyes on this to check if it is the right mechanism.

packages/myst-directives/src/code.ts
```diff
@@ -246,11 +247,13 @@ export const codeCellDirective: DirectiveSpec = {
     doc: 'The code to be executed and displayed.',
   },
   run(data, vfile): GenericNode[] {
+    const opts = getCodeBlockOptions(data, vfile);
     const code: Code = {
       type: 'code',
       lang: data.arg as string,
       executable: true,
       value: (data.body ?? '') as string,
+      ...opts,
```
stefanv:
```suggestion
      ...getCodeBlockOptions(data, vfile)
```

Not sure we benefit right now, in terms of readability, from assigning it first.

stefanv:
Netlify failure is unrelated.

packages/myst-directives/src/code.spec.ts
fwkoch: This seems fine - we have very few tests in the `myst-directives` package (directives are mostly tested at a higher level, e.g. in `myst-to-md`) and none of the existing tests in `myst-directives` actually invoke `run`. But there's no reason not to do that! Builds confidence that the ~2 line code change actually does some heavy lifting.

packages/myst-directives/src/code.ts
```diff
@@ -246,11 +247,13 @@ export const codeCellDirective: DirectiveSpec = {
     doc: 'The code to be executed and displayed.',
   },
   run(data, vfile): GenericNode[] {
+    const opts = getCodeBlockOptions(data, vfile);
```
choldgraf:
```suggestion
```

choldgraf:
thanks for the reviews! I've made @stefanv 's suggestion to not assign to `opts` first

packages/myst-directives/src/code.ts
```diff
@@ -251,6 +252,7 @@ export const codeCellDirective: DirectiveSpec = {
       lang: data.arg as string,
       executable: true,
       value: (data.body ?? '') as string,
+      ...getCodeBlockOptions(data, vfile)
```
choldgraf:
```suggestion
      ...getCodeBlockOptions(data, vfile),
```

PR #2748, opened 2026-03-06
🫓 npm to bun

stefanv:
This PR switches to the `bun` runtime *and* package manager, for faster installs.

Bun is intended to be a drop-in replacement for Node.js (and npm)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,10 @@ jobs:
   test:
     if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
-    name: Testing
+    strategy:
+      matrix:
+        node: ["20", "22", "24"]
+    name: Testing with bun and node ${{ matrix.node }}
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


packages/myst-execute/tests/run.spec.ts
```diff
@@ -77,6 +77,10 @@ class SessionManagerFactory {
 
 const sessionManagerFactory = new SessionManagerFactory();
 
+beforeAll(async () => {
+  await sessionManagerFactory.load();
+}, 60_000);
```
agoose77, line 82: Great idea!

.github/workflows/ci.yml
```diff
@@ -14,45 +14,40 @@ jobs:
...
       - name: Ensure changeset has the right structure and doesn't contain ignored packages
-        run: npm run version
+        run: bun run version
   test:
     if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node: ['20', '22.4.x', '24']
-    name: Testing on node ${{ matrix.node }}
```
rowanc1, line 36: I think we still want some tests running on node.

That is still the way we deploy, so I think it would be a good idea to have at least the CLI integration tests on this matrix and some tests in node.

rowanc1:
I moved some tests over to bun, much faster locally in those packages and noticeable. Shaved about 15s off the CI times.

We could switch from pip --> uv to go faster ... 

I do think we should pick up node testing in some respect before merging.

PR #2758, opened 2026-03-10
🪟 Windows paths on upgrade, resolve paths correctly

rowanc1:
This replaces the backslashes with unix-style paths. There is also a warning now that has a note for users who do this on their own.

Fixes jupyter-book/jupyter-book#2599

### DIFF
--- a/packages/myst-cli/src/init/jupyter-book/toc.ts
+++ b/packages/myst-cli/src/init/jupyter-book/toc.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 import { resolveExtension } from '../../utils/resolveExtension.js';
-import { join, relative } from 'node:path';
+import path, { join, relative } from 'node:path';
 import { cwd } from 'node:process';
 import type { ISession } from '../../session/types.js';
 import type { Entry as MySTEntry, ParentEntry as MySTParentEntry } from 'myst-toc';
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/init/jupyter-book/toc.ts
```diff
@@ -201,7 +201,8 @@ function assertNever(): never {
 function maybeResolveDocument(dir: string, name: string, session: ISession): string {
   const resolved = resolveExtension(join(dir, name));
   if (resolved) {
-    return relative(dir, resolved);
+    // Use forward slashes for cross-platform compatibility (Windows path.relative uses backslashes)
+    return relative(dir, resolved).replace(/\\/g, '/');
```
rowanc1: This is the fix for the bug.

packages/myst-cli/src/utils/resolveExtension.ts
```diff
@@ -91,7 +91,13 @@ export function resolveExtension(
       }
     });
     if (normalizedMatches.length === 0 && path.extname(file)) {
-      warnFn(`Table of contents entry does not exist: ${file}`, 'error');
+      warnFn(
+        `Table of contents entry does not exist: ${file}`,
+        'error',
+        file.includes('\\')
+          ? 'Always use UNIX-style path separators (forward slashes) in the table of contents.'
+          : undefined,
```
rowanc1, line 99: This is a documentation fix for users that shows on the console.

artoftheblue:
ooh thanks, I believe this is something I have required some time ago

PR #2766, opened 2026-03-16
Support `columns` on `grid-item` directive

stefanv:
- Pass through the `columns` option
- Render `grid-item` as `grid-item` in AST, not as div

xref: https://github.com/jupyter-book/myst-theme/pull/838

closes: #1141

TODO:

- [x] List but comment out Sphinx Design options that are not currently implemented
- [ ] Consider generating documentation from / pointing to this PR to show which places need to be touched when updating a directive

### DIFF
--- a/packages/myst-ext-grid/src/index.ts
+++ b/packages/myst-ext-grid/src/index.ts
@@ -50,6 +50,7 @@ export const gridDirective: DirectiveSpec = {
 
 export const gridItemDirective: DirectiveSpec = {
   name: 'grid-item',
+  // Other options from sphinx-design we leave below for reference in case we want to support in the future
   // options:
   // // https://sphinx-design.readthedocs.io/en/furo-theme/grids.html#grid-item-options
   // margin


docs/dropdowns-cards-and-tabs.md
```diff
@@ -102,11 +102,29 @@ A button is an element with text content that triggers an action to navigate to
 
 Grids allow you to structure arbitrary chunks of content in a grid-like system.
 
-To generate a grid, use the ` ```{grid} ` wrapper directive along with ` ```{card} ` directives inside.
+To generate a grid, use the `:::{grid} ...` wrapper directive along with `:::{grid-item}` or `:::{card}` directives inside.
```
choldgraf, line 105: sidenote: we should probably standardize this in our docs

choldgraf (approved):
This LGTM - I think it's safe to merge when you're happy with it. A few quick non-blocking comments:

- Could we add minimal tests or warnings for the new functionality and the most common use-case bugs we expect users to accidentally trigger? e.g., what if columns is above 12 if a user thinks that might be possible?
- The confusion that triggered this was because there's a discrepancy between the functionality of the sphinx "grids" implementation and this one. Is there any other _obvious_ thing we should add a brief warning for? Doesn't strictly need to be in this PR but it might be a nice opportunity to clear up confusion

docs/dropdowns-cards-and-tabs.md
agoose77: We should. Any time it's MyST Markdown in the body, we should prefer `:::`. But we can do that elsewhere.

packages/myst-ext-grid/src/index.ts
```diff
@@ -50,6 +50,14 @@ export const gridDirective: DirectiveSpec = {
 
 export const gridItemDirective: DirectiveSpec = {
   name: 'grid-item',
+  // options:
```
choldgraf, line 54:
```suggestion
  // Other options from sphinx-design we leave below for reference in case we want to support in the future
  // options:
```

stefanv:
Ah, yes, you're right that we should warn if things look sus in column totals. But thanks for getting this in already!

Issue #1196, opened 2024-05-09
Keep links stable for static assets


choldgraf:
Currently when static assets are linked to a MyST site build, their URL gets generated with a hash that is part of the filename. This is done to ensure that the cache is busted if the file is updated.

For example, using `` {download}`myfile.png` `` results in a filename like `mysite.org/build/myfile-c7e3595ff6c09f5cc22cc52b8177cd4c.png)`

The problem with this is that if you want to directly link to a static asset elsewhere, the link will be invalidated as soon as the MyST site is rebuilt, because the hash is included in the filename itself.

It would be useful if the URL for static assets was persistent as long as the user didn't change the filename, so that external links to them remained valid.

### Potential solution: cache busting 

One option is to use a **query parameter** to bust caches, rather than putting the hash in the filename.

Essentially you'd move the hash from this:

```
mysite.org/build/myfile-c7e3595ff6c09f5cc22cc52b8177cd4c.png
```

to this:

```
mysite.org/build/myfile.png?v=c7e3595ff6c09f5cc22cc52b8177cd4c
```

The hash becomes part of a "query" parameter. You can thus still link directly to the file (`mysite.org/build/myfile.png`) and links on the MyST site will include the hash via the query parameter so you'll cache-bust as expected.

### References

- [Here's a StackOverflow post about it](https://stackoverflow.com/questions/9692665/cache-busting-via-params).
- Issue where this raised a problem: https://github.com/2i2c-org/2i2c-org.github.io/issues/238

cboettig:
💯 very much this!!  

also, what are you using for hashing?  is this just the md5sum of the file itself?

rowanc1:
Yes, just the md5!

choldgraf:
I spent a little bit of time thinking about this over the weekend, and realized it's a little bit more complex than just "hash all the files one way or another". We probably want some kinds of files hashed one way, and other files hashed another way. Here's a quick proposal and framing from me (copied into the issue body as well)

---

There are a few different kinds of links that we might want to apply this to, and we may _not_ want to apply this to *all* links. File-based cache busting is the "gold standard" for web applications because there's no way that an outdated file will be served because its content is what makes up the hash.

So I think this is a reasonable framing to use:

- Use URL query hashing (`?v=[hash]`) when we primarily expect *users* will download in a non cache-y way (e.g. clicking a link, downloading with a python script, etc).
[...]

Issue #2714, opened 2026-02-19
Add new interface for resolving static files in directives


agoose77:
## Context
MyST directives like `image` currently implement dedicated transforms that select `image` and handle their `url` arguments accordingly. This pattern (creating a transform for each node type) is cumbersome, and the present logic is bespoke to each node.

I recently added local file support to anywidgets, and realised that it is mostly a duplicate of image handling.

## Proposal
I propose that we do several things:
1. Add a new `path` option-type to directives, so that we can indicate in the docs which arguments are path-like, _and_ perform basic validation. 
2. Add a new `ISession.resolveAssetPath` function that:
   - Fetches remote assets into the static folder
   - Copies local files into the static folder
   - etc.
3. Expose this function to directives (and roles?)

This would move resolution earlier in the transforms — to the directive itself. I think that's OK — it eliminates the need for a separate transform _and_ we can provide a no-op function in contexts where this doesn't make sense e.g. the MyST sandbox.

(3) suggests that parsing now has side-effects. I think that's probably OK?

rowanc1:
I had a similar need wanting to create an audio node, and have that copied over to static storage, validated etc. This likely also solves (or improves slightly) our CSV-table directive challenges we had previously.

Very supportive of this direction.

choldgraf:
This sounds great - I think the `resolveAssetPath` function could also help us resolve things like these two:

- https://github.com/jupyter-book/mystmd/issues/2700 (partially, we'd still need to add the `head` tagging)
- https://github.com/jupyter-book/mystmd/issues/1921 (partially, we'd need to add the UX in config)

stevejpurves:
Note just as an additional use case in addition to the issues linked are - that @agoose77's current workaround is serving what anywidget needs for the esm & css files/urls but there is (near) future notion of static assets that need to be shipped along with the widget that are specified from an array/list/comma-separated-list of files/urls (we dropped this from v1). 

This might be image assets for a completely custom widget which could be handled in the same way OR image assets from a (viz) library that an anywidget wraps, that has the additional requirement that paths should not be hashes i.e. that viz library expects certain well known paths.

choldgraf:
Just noting that I would love to use this opportunity to get this issue resolved, since we are all thinking about and interested in the anywidget stuff. Opening up these little plugin / developer bottlenecks will really help increase the flexibility of myst for the power user / dev community, I think.

choldgraf:
[...]

PR #2770, opened 2026-03-18
Add option to specify static files for site builds

### LINKED ISSUE #1255: how to organize static assets in the filesystem tree
loosely related to #1250:

while we're at it talking about static assets, I take this chance to outline another possible improvement as compared to jb1
right now I have some books that are made of a non-trivial fs tree; I mean not all the source notebooks are located in unique folder; think for example of a book made of tens of programing exercises, each of which is made of a main README and related artefacts (media, data, ...)
for convenience, this use case maps into an fs tree with, say, one folder per exercise (all of them being in turn possibly organized into subfolders if need be)

in such cases when using sphinx and jb1, I *think* I had to group all the static assets into a unique toplevel `_static` folder

I guess my point here is, it might help to 
- have `mystmd` let users put their static assets anywhere in the tree and not necessarily at the toplevel - as it would help keep related stuff together
- regardless, whether or not this feature is available (I can understand if it's not deemed high priority of course), the documentation could ideally have a section 'how do I manage and refer to my static assets' so that users know right away how to deal with this sort of questions

with jb1, it felt like many aspects, such as this one, were understandably not developed in the jb1 doc because they actually pertained to sphinx, I guess what I'm trying to say is, that is changing with mystmd rebuilding the whole stack from the ground up

### LINKED ISSUE #2863: Allow users to specify static folders or files and have entire structure copied over to HTML output

stefanv:
This introduces the `project: static_files:` key. If that key is set to:

- a/b/foo.txt: foo.txt is copied to the build dir
- a/b: b is copied (with content) to the build dir
- a: a is copied (with content) to the build dir

We still lack:

1. <strike>A way to link to such artifacts directly, without hashing (see https://github.com/jupyter-book/mystmd/issues/1196).</strike> <-- tested, works perfectly fine
2. A way to add static files generated during plugin run (possible to modify site options directly?)

Associated PR in myst-theme: https://github.com/jupyter-book/myst-theme/pull/887

---

- closes https://github.com/jupyter-book/mystmd/issues/1255
- closes https://github.com/jupyter-book/mystmd/issues/2863


packages/myst-cli/src/process/site.ts
```diff
@@ -628,6 +630,13 @@ export async function processProject(
         }
       }),
     );
+    // Write all static files
```
choldgraf, line 632: A few quick questions:

- Am I right that this will copy the file into the `_build` output with its original filename, but not hierarchy of folder (e.g. `a/foo.md` will be copied to `_build/html/foo.md`?
- Why not `_build/html/_static/foo.md` (or some other standard static location?)
- Should we update the `{download}` role to use this mechanism instead of hashing the file?
- This would let you link directly to a static file, but linking in markdown would result in a "file not found" error of some sort is that right?

packages/myst-cli/src/build/html/index.ts
```diff
@@ -181,6 +181,12 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
 
   // Copy all of the static assets
   fs.copySync(session.publicPath(), path.join(htmlDir, 'build'));
+
+  // Copy user static files to html build root
+  const siteConfig = selectors.selectCurrentSiteConfig(session.store.getState());
+  for (const fn of siteConfig?.static_files ?? []) {
+    fs.copySync(fn, path.join(htmlDir, path.basename(fn)));
```
agoose77: Suggestion - we should check to avoid clobbering existing static files.

packages/myst-cli/src/process/site.ts
agoose77: No to `download` for now, my vote!

choldgraf:
@stefanv this feels quite close! can you add some documentation so that we can consider this good to go? I don't think there's anything blocking in the reviews etc

stefanv:
1. Will add docs.
2. How do we link to these static files?

choldgraf:
> How do we link to these static files?

I think just ensuring that the file is present in the output build tree is probably enough to be useful for now. The combination of:

- Not hashed
- Included in _build/html somewhere

Is probably enough IMO.

If we wanted to enable linking, we could recommend that users use `{download}` for this (I'd be +1 on allowing them to use `hash=False` on those as well, or even use that as the default). Or for now we could just expect that they'll need to figure out the links on their own, and leave that for a future improvement.

cycomachead:
I think this would be useful enough for my cases. FWIW, for many assets, I think I don't want any hashes and don't even always need links. 

I'm deploying my book to GitHub Pages, but with a custom domain name. The `CNAME` file must be present at the root of the site exactly as is. (It's quite easy to add `echo "domain" > _build/html/` to the CI workflow, but it's also nice to explicitly declare it in the yaml. I personally also like shipping (unlinked) README and LICENSE files in the build incase someone is looking at the gh-pages branch or grabs a build zip. Again, not terribly hard to do myself, but nicer to declare.

agoose77:
@stefanv in the community meeting, my understanding was that we were going to park this approach and look to directly tackle the hashing problem in mystmd directly (& using `download` roles). Did you change your mind?

stefanv:
@agoose77 What I thought we were going to do was add functionality to use query string hashing instead of filename hashing, to make it easier to link to these files.

choldgraf:
I pushed a commit that added a bit of docs for this! One confusion I ran into was how to differentiate the reason you'd use:

- Static folder listing, like we have here
- The `download:` configuration

I tried to add an explanation, it doesn't feel super convincing to me (why would an average user care about content hashing?) but at least it's something we can iterate on!

re @stefanv 's comment, I'd also be a fan of assuming query-string hashing for any user-facing download functionality or config, and retaining the filename hashing only for web assets that *must* be invalidated if a new version is served, because I think query string hashing is a little less consistent. IMO none of that needs to block this PR

So:

- CSS and JS files needed to build MyST: filename hashing, no change from now
- Everything else: query hashing so filenames stay the same
- Over time: identify other things that need filename hashing and make it possible to do so iteratively

choldgraf:
@stefanv what would you like to do with this PR? I can't tell if we're blocked on a decision or a design discussion, or if this is just waiting for a final review and merge. Could you advise?

espg:
I would love to see this merged. I ported my blog over from sphinx to myst last week, and am currently pinned to this PR as hard deployment dependency... since I can't for the life of me figure out any other way to reference a banner on every page of the site without having a `static_files` option.

RRiva:
@espg, me too 🙂 for the time being, I ended up making a new repo that only publishes what I need. Then, I link to it using the `iframe` directive.

packages/myst-cli/src/process/site.ts
```diff
@@ -627,6 +629,13 @@ export async function processProject(
...
+    const staticFiles = opts?.staticFiles ?? [];
+    await Promise.all(
```
agoose77: This will fail when any copy operation fails. I think that's a good thing, but surfacing it explicitly.

rowanc1:
I haven't tested or dived in too far to the PR. Just a quick design question, if we have:

```
images/foo.png
images/folder/foo.png (different image, same file name)
```

How does that get copied into the build folder?

docs/website-downloads.md
```diff
@@ -31,8 +29,38 @@ For example:
...
+
+To do so, add a list of paths under `site:` in your `myst.yml`:
+
+```{code-block} yaml
+:filename: myst.yml
+
+site:
+  static_files:
+    - path/to/CNAME # A file name, for example
+    - path/to/assets # A folder name, for example
```
rowanc1: Has there been thought put onto if these could be on the `project` instead? Or if this has any relation to the project `resources` (those are static files that are bundled and cached and retain their file paths...)

stefanv:
> I haven't tested or dived in too far to the PR. Just a quick design question, if we have:
> 
> ```
> images/foo.png
> images/folder/foo.png (different image, same file name)
> ```
> 
> How does that get copied into the build folder?

Currently you can specify folders or filenames; if you specify the folder, the foo.png will be nested, as expected.

docs/website-downloads.md
stefanv: I see no mention of resources at https://mystmd.org/guide/settings — can those be set by users? I think this could (and should probably) live under project, although for some formats (PDF etc.) it won't make much sense. How would you call it?

stefanv:
Looks like I missed the discussion on downloads. But here's my 2c: that's a rather confusing mechanism. Allowing users to mirror a folder into a static location, and then referring directly to the files within it is a lot easier to grok. Also, currently the downloaded filename is the hashed filename—pretty sure no-one wants that.

Given that the downloads functionality already exists, how would we use it to implement this feature?

1. Allow a `folder` instead of just `file`
2. For the user to know the path to provide to the `{download}` directive, we'd have to disable hashing (or switch to query hashing—we should do that anyway).

So, would the team prefer for me to try that route?

docs/website-downloads.md
rowanc1: There are some _very small_ docs here:
https://mystmd.org/guide/frontmatter#all-available-frontmatter-fields

The main goal of this is to setting is to bring files into the `meca` bundle so that there can be a reproducible build sent to binder or other services (e.g. scripts and data sets that would otherwise be skipped in compiling the `build` folder.

I could make arguments either way that this is separate or the same sort of feature. I lean _slightly_ more towards them being the same thing though.

If this is the implementation path chosen, I think adding to `project` rather than `site` is better. I think `site` should be left for downstream customizations (through the `template.yml`) — when we put things on `site` that means all templates have to adapt and start to use those and exclude `static_files` from the template, etc.

choldgraf:
I think you should follow your intuition @stefanv - I think the pre&existing download functionality would be more useful if it didn't hash the filenames for most users, but I'm happy if we don't touch it for the sake of getting this one in and unblocking this functionality. We can always iterate on the download functionality from there

packages/myst-cli/src/build/html/index.ts
stefanv: We could, but that's not how copy typically works :) 🦶 🔫

packages/myst-cli/src/process/site.ts
stefanv: Not sure I want to tell people where to place their files. If they want to dump it at the root 🤷‍♂️

stefanv:
This should be ready for review. In a follow-up, I could give `{download}` a hash parameter, so that you can create a nice download button for a static asset. And, perhaps that should default to false? We can also consider in that follow-up PR whether to use renaming or query hashing.

docs/website-downloads.md
```diff
@@ -4,21 +4,22 @@ short_title: Downloads & Static Files
...
-MyST will generate a _hashed filename_ for any files bundled with your site.
-For example: `myfile.[HASH].png`.
-This means download URLs will not be persistent if the content changes.
+- **`{download}` role** — an inline role used directly in content to create a download link at the point of use.
+  MyST content-hashes the filename (e.g. `myfile.[HASH].png`) so downstream caches invalidate when the file changes.
+  See [the `{download}` role section](#download-role).
+- **`downloads:` frontmatter** — page or project configuration that populates the page's download panel in the site UI (not inline in content).
+  Can reference local files, export IDs, or remote URLs.
+  See [the `downloads:` frontmatter section](#download-link).
+- **`project.static_files`** — copies files or folders into the build output at a stable, predictable URL, without hashing and without rendering any link.
```
choldgraf: I wonder if this should come first - I suspect it's the default behavior people would want (no hashing to have a predictable link)

docs/website-downloads.md
```diff
@@ -31,8 +32,48 @@ For example:
...
+```{code-block} yaml
+:filename: myst.yml
+
+project:
+  static_files:
+    - path/to/CNAME # A file name, for example
+    - path/to/assets # A folder name, for example
+```
+
+Each entry is copied into the **root** of the build output (`_build/html/`).
```
choldgraf, line 52: How hard would it be to expose the function this uses for _plugins_? This would be super useful...

If it's non-trivial, then we can open an issue to follow up on this

packages/myst-cli/src/build/html/index.ts
```diff
@@ -193,6 +193,16 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
 
   // Copy all of the static assets
   fs.copySync(session.publicPath(), path.join(htmlDir, 'build'));
+
+  // Copy user static files to html build root
+  const siteConfig = selectors.selectCurrentSiteConfig(session.store.getState());
+  for (const proj of siteConfig?.projects ?? []) {
+    if (!proj.path) continue;
+    const projectConfig = selectors.selectLocalProjectConfig(session.store.getState(), proj.path);
+    for (const fn of projectConfig?.static_files ?? []) {
```
choldgraf: Do we confirm somewhere that these are all relative to the project root? Maybe we should catch "filenotfound" erorrs and provide a helpful error message along these lines - I could see people foot-gunning by not getting the relative file paths right

packages/myst-cli/src/process/site.ts
```diff
@@ -627,6 +628,17 @@ export async function processProject(
...
     );
+    // Write all static files
+    const projectConfig = selectors.selectLocalProjectConfig(
+      session.store.getState(),
+      siteProject.path,
+    );
+    const staticFiles = projectConfig?.static_files ?? [];
+    await Promise.all(
+      staticFiles.map(async (fn) => {
+        fs.copySync(fn, join(session.publicPath(), basename(fn)));
```
choldgraf: Can you explain in comments how this is different from the copy logic you've got in the file just above in `packages/myst-cli/src/build/html/index.ts`, so maintainers know how they relate to or complement one another?

packages/myst-frontmatter/src/project/validators.ts
```diff
@@ -300,6 +300,15 @@ export function validateProjectFrontmatterKeys(
       },
     );
   }
+  if (defined(value.static_files)) {
```
choldgraf, line 303: could you add a simple test for:

- An individual file
- A folder

?

If it's really complicated to do this with our current testing infrastructure, that's OK to skip it so we can unblock this..

choldgraf (approved):
In general this looks good to me! I added a few comments and suggestions in there - I'll hit "approve" w/ the idea that I trust you to merge when you feel you've addressed the comments to an acceptable level. I'm also happy to keep iterating w/ you on feedback if you prefer!

Thanks for implementing this - I think it'll be super useful for communities that use their websites as a way to share resources via files etc.

agoose77 (approved):
Subject to Chris' suggestions, LGTM!

packages/myst-cli/src/process/site.ts
stefanv: @choldgraf `a/foo.md` will be copied without hierarchy, but if you specify `a` it will copy `a` as is.

With `_static` I don't know if you mean `_static` as the input that gets copied, or the target folder. IMO less useful to specify the target folder. Would you prefer the `_static` folder be the implicit input?

I think `{download}` is separate from this, since those files don't have to go to any specific location.

Hopefully markdown links will work; I can double check.

choldgraf: I'm going to resolve this comment as it is pretty outdated! The other commits I pushed and my comments in the latest review are the best place to focus time!

docs/website-downloads.md
stefanv: I have another PR for that written, but good to track it lest I forget.

packages/myst-cli/src/build/html/index.ts
stefanv: <strike>It is relative to wherever it is specified. E.g., if you use it in your global project options, it is relative to the project root, but if you specify it in a page's options, it is relative to that page's location. This made intuitive sense to me, so I did it that way :)</strike>

<strike>Also, through some dark witchery, it even works correctly with ToC. For websites, in particular, I quite dislike ToCs but :shrug:</strike>

Digging into this a bit more; I tripped myself up by testing that `./foo.pdf` was downloadable, but of course that works completely independently of static files :sweat_smile:

packages/myst-cli/src/process/site.ts
stefanv: Same logic, but one is for a site build and one is for a static build. I don't think we can rely on the webscraping to harvest all the static files—what if the user doesn't reference the file?

packages/myst-cli/src/build/html/index.ts
stefanv: Correction on my previous messages here: you cannot use this from within a page, it is only on a global project basis.

stefanv: I will add a warning on entries not found.

packages/myst-cli/src/process/site.ts
stefanv: But I unified the two functions :+1:

packages/myst-cli/src/process/site.ts
stefanv: Added a warning for the case where the input file isn't present.

stefanv:
OK @choldgraf addressed your review comments!

stefanv:
Our test suite is pretty flaky; will put it on the list :)

docs/website-downloads.md
choldgraf: maybe we can track that here: https://github.com/jupyter-book/mystmd/issues/2714

packages/myst-cli/src/utils/copyStaticFiles.ts
```diff
@@ -0,0 +1,20 @@
...
+import fs from 'fs-extra';
+import type { ISession } from '../session/types.js';
+
+export function copyStaticFiles(
+  session: ISession,
+  staticFiles: string[],
+  destDir: string,
+  projectPath: string,
+) {
+  for (const fn of staticFiles) {
```
choldgraf, line 11: I think that here, we should resolve this relative to `projectPath` first, because I'm _pretty sure_ it's possible to nest multiple projects under `projects:`, and if so, then it means that the `cwd` will be different from the root of the `myst.yml` in each project.

I'm also realizing this "multi project" feature is not documented anywhere, so I opened this one:

- https://github.com/jupyter-book/mystmd/issues/2886

rowanc1: That multi-project feature has been deprecated and mostly removed, but not fully from all code paths:

https://github.com/jupyter-book/mystmd/issues/1103

Project composition happens at a higher level now than in myst (e.g. github pages, routing, etc.) in favour of simplifying the myst code (some of which still needs to work through) including flattening the myst keys. etc.

stefanv:
There was a race condition in the test suite; the last commit addresses that (we could pull it out into a separate PR).

choldgraf:
@stefanv can you fix the lint errors and then we can merge this one?!?! Let's :shipit: !

stefanv:
Done.

choldgraf:
We discussed this today and decided it was ready to go! I'm gonna merge this one in, thanks so much @stefanv for all the iteration here!

Follow up is over at myst-theme:

- https://github.com/jupyter-book/myst-theme/pull/887

PR #2774, opened 2026-03-18
Fix \string macro leaking its name into image filenames in tex-to-myst

stefanv:
While converting a LaTeX book, exported from LyX, I've noticed that the following filename patterns fail in includegraphix:

- `something\string~.pdf`: should become `something~.pdf` but is interpreted as `somethingstring~.pdf`
- `"something~".pdf`: should become `"something~.pdf"` but instead the quotes are kept and confuses the filesystem.

*Technically* some operating systems can have quotes in the filenames, but this would mess LaTeX up anyway.

### DIFF
--- a/.changeset/great-cows-film.md
+++ b/.changeset/great-cows-film.md
@@ -1,0 +1,6 @@
+---
+"tex-to-myst": patch
+---
+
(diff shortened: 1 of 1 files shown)


rowanc1:
Nice! Thanks for the tests as well!!

PR #2783, opened 2026-03-24
🆕 Emit Node20 for mystmd

agoose77:
Now that our minimum version of Node.js is Node.js 20, and we have a policy of three trailing LTS versions (which I can't find the provenance for, but I remember discussing it with core team members), we should bump our esbuild target.


rowanc1:
Nice, thanks!

Issue #2784, opened 2026-03-24
Plugins not available when parsing `site.parts` from extended configs


choldgraf:
When a config file uses `extends` to inherit from another config, if you define a `site.parts` key in the target of `extends:`, and that key depends on the `plugins:` of the parent `myst.yml` file, then the plugin won't yet be loaded when the part is parsed.

## How to reproduce

Here's roughly what I did that ran into this. In this case, the "parts" content seems to be parsed *first*, and then the plugins are loaded later.

**`site.yml`**:

```yaml
version: 1
site:
  parts:
    footer: |
      ```{footer}
      title: My Project
      ```
```

**`myst.yml`** (project config):

```yaml
version: 1
extends:
  - site.yml
project:
  plugins:
    - https://github.com/jupyter-book/myst-plugins/releases/download/footer-latest/index.mjs
```

## What's going on?

I think somewhere around this section:

https://github.com/jupyter-book/mystmd/blob/a3a84a304a73a27cb730b53e41dfdd27db87e412/packages/myst-cli/src/process/file.ts#L403

We must be parsing the "`extends:`" config _as we're loading it_. It feels like instead we should build up the full `myst.yml` after all "extending" has been done, and _then_ parse it into config.

agoose77:
@fwkoch and I discussed this in the dev meeting today. As @choldgraf surmises,
> We must be parsing the "extends:" config as we're loading it. It feels like instead we should build up the full myst.yml after all "extending" has been done, and then parse it into config.

Our view is that `extends` is really not designed to be used as "extend someone else's MyST project" rather than "extend _our own_ MyST configuration that we have lifted out". The nuance is that we don't want to evaluate each myst.yml in its own context, but rather in the context of the final project. After all, `extends` doesn't really support file paths for e.g. parts because there's no way to look up the file, unless it's a URL. 

@fwkoch and I suggested simply that we defer parsing of parts until we have a final myst.yml.

ref: https://github.com/jupyter-book/mystmd/issues/2691

choldgraf:
This is closed by #2789 !

PR #2789, opened 2026-03-26
🔌 Defer config-level part resolution until plugins are all loaded

fwkoch:
Attempts to address: https://github.com/jupyter-book/mystmd/issues/2784

### DIFF
--- a/packages/mystmd/tests/extend-part-plugin/abc.mjs
+++ b/packages/mystmd/tests/extend-part-plugin/abc.mjs
@@ -10,7 +10,7 @@ var abcDirective = {
   run(data) {
     const count = Number(data.arg) || 1;
     const alphabet = "abcdefghijklmnopqrstuvwxyz";
-    return Array.from({ length: count }, () => u("paragraph", [{ type: "text", value: alphabet }]));
+    return Array.from({ length: count }, () => ({type: "paragraph", children: [{ type: "text", value: alphabet }]}));
   }
 };
 


fwkoch:
Tests are failing but CI is passing. 😕

image

fwkoch:
Just needed to chain test commands with `&&` instead of `;` 🎉

fwkoch:
Thanks @stefanv - not a problem at all, added docs to the functions I added / touched!

packages/mystmd/tests/extend-part-plugin/abc.mjs
```diff
@@ -0,0 +1,34 @@
...
+  } else {
+    Object.assign(node, props);
+  }
+  if (Array.isArray(value)) {
+    node.children = value;
+  } else if (value !== void 0 && value !== null) {
+    node.value = String(value);
+  }
+  return node;
+}
```
agoose77:
```suggestion
```

packages/mystmd/tests/extend-part-plugin/abc.mjs
```diff
@@ -0,0 +1,34 @@
...
+var abcDirective = {
+  name: "abc",
+  doc: "A directive that will generate deterministic alphabet text.",
+  alias: ["alphabet"],
+  arg: { type: Number, doc: "The number of paragraphs to generate." },
+  options: {},
+  run(data) {
+    const count = Number(data.arg) || 1;
+    const alphabet = "abcdefghijklmnopqrstuvwxyz";
+    return Array.from({ length: count }, () => u("paragraph", [{ type: "text", value: alphabet }]));
```
agoose77:
```suggestion
    return Array.from({ length: count }, () => ({type: "paragraph", children: [{ type: "text", value: alphabet }]}));
```

PR #2794, opened 2026-04-02
Clean up gitlab docs a bit

FreekPols:


FreekPols:
@choldgraf (not sure whether I could have used the previous merge request), but the fix is here :)

choldgraf (approved):
Thanks for cleaning this up!

PR #2798, opened 2026-04-03
📖 Install NodeJS guide: Removed unnecessary verb on a sentence

Polirecyliente:
Good day team, [in this guide about installing NodeJS](https://mystmd.org/guide/install-node#node-via-conda-mamba-all-platforms), there is a small grammatical error:

    Use the following command can be used to lock down the version you are installing

**Grammatical reasoning**

In the English language, each simple sentence contains 1 main verb. The quoted text contains 2 main verbs: `Use` and `can`. It seems that the author wanted to use `can`, and later forgot to delete the other verb.

**Resulting text**

    The following command can be used to lock down the version you are installing

This reads like a proper guide, indicating to the reader that the following command locks down the NodeJS version.

Thank you so much for your attention 🙏


rowanc1:
Thanks!

Issue #2042, opened 2025-05-20
files included as sections rather than chapters in LaTeX book template.


FreekPols:
When building a LaTeX PDF (using the book document class) from MyST markdown content, the individual markdown files are included as sections rather than chapters.

![Image](https://github.com/user-attachments/assets/4a5955ef-1eb0-4791-a465-445bd2e2f441)

While this structure may make sense for scientific articles—where "Introduction", "Methods", and "Results" are typically considered sections—for a book-format output, it would be more appropriate for each markdown file to be treated as a chapter.

**Expected Behavior**

When using the book format, each top-level markdown file (# Title_) should be rendered as a \chapter{} in LaTeX, rather than a \section{} (## ).

schmoelder:
Since I'm still having this issue, I started digging a bit. Admittedly, I'm not super experienced with the internals but I could imagine that the following might be causing the issue:

https://github.com/jupyter-book/mystmd/blob/f56139a7558e803dbdc5e18b13e7eabc976d617c/packages/myst-to-tex/src/index.ts#L132

Congrats to the team for publishing v2.0, btw. It's a great milestone and I'm looking forward to all the new stuff.

Issue #2002, opened 2025-05-04
Numbering of headers and titles


vangeest:
## Description
Numbering of headings is not working as documented.

I'll provide 2 examples which can be reproduced in https://mystmd.org/sandbox

### Example 1
```
---
numbering:
  headings:
    enabled: true 
---
# top level heading
## sub level heading
### sub-sub level heading
```

Renders as
```
top level heading
1 sub level heading
1.1 sub-sub level heading
```
It should render as
```
1. top level heading
1.1 sub level heading
1.1.1 sub-sub level heading
```

### Example 2
```
---
title: "my title"
numbering:
  title:
    enabled: true
  heading_1:
    enabled: true 
  heading_2:
    enabled: true 
  heading_3:
    enabled: true 
---
# top level heading
## sub level heading
### sub-sub level heading
```

Renders as
```
my title
2 top level heading
2.1 sub level heading
2.1.1 sub-sub level heading
```

It should render either as
```
1. my title
1.1 top level heading
1.1.1 sub level heading
1.1.1.1 sub-sub level heading
```
or as:
```
1. my title
2. top level heading
2.1 sub level heading
2.1.1 sub-sub level heading
 ```

## Proposed solution
Adjust code to work according to documentation

## Additional notes
tested with Myst v1.3.26
documentation link: https://mystmd.org/guide/cross-references#title-numbering

FreekPols:
To add to this: 
- it would be nice to include the numbering at the myst.yml file (as used in JB1.0) rather than at file level.

May be related to this issue:
- when converting to pdf (book), chapters are interpreted as sections. The toc always starts with 0.1 rather than 1.0

michaelaye:
maybe when enforced with `numbering: title: true`, there should be a number in front of the title, but otherwise this seems to follow the Chicago Manual of Style for publications where any frontmatter material like title pages etc. do not receive numbers and the numbering only starts with the beginning of sections or chapters. In other words, there is definitely a need for not numbering the title.

vangeest:
@michaelaye I could personally live with titles not being able to be numbered and documentation being brought in line with that.

It's the inconsistency in the way that numbering at the title and headings seems to be mixed up that bothers me:
- The title shows unnumbered and the first heading_1 starts at number 2 (an attempt to fix this with `start: 1` in the fronmatter did not work).
- When `numbering: title: true` is ommitted, the titles remain unnumbered, but heading_1 is not numbered either.

[...]

PR #2806, opened 2026-04-09
Add additional documentation for LaTeX PDFs

Songtech-0912:
This adds additional information regarding the table of contents structure of LaTeX PDF exports, fixing unexpected behavior where LaTeX pdfs show section numbers as "0.X.Y" in the table of contents. For more information, see https://github.com/myst-templates/plain_latex_book/pull/4, where this PR came from.

### DIFF
--- a/docs/creating-pdf-documents.md
+++ b/docs/creating-pdf-documents.md
@@ -99,6 +99,76 @@ Ensure that you download a full distribution with appropriate libraries installe
 
 (rendering-pdfs-with-typst)=
 
+### LaTeX export gotchas
+
+By default, MyST-rendered PDFs will use the `\section` command as the top-level heading in LaTeX. This is standard for scientific papers and short reports. However, if you are writing a book or long document (e.g. with Jupyter Book), this may cause errors as LaTeX expects the `\chapter` or `\part` command to be top-level. MyST can accomodate for this difference by manually setting the `level` key to each file you wish to export, as per the table below:
+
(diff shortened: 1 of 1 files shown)


schmoelder:
Very nice to see that this is being addressed. Thanks a lot!

For reference, these two issues seem to be related:
- #2002
- #2042

cycomachead:
Thank you! This is very helpful for a book project I am working on!

Just musing... It would be great if we could find a way to them keep the regular toc and the PDF export ones in sync, but I've moved all this config to separate yaml files:

```yml
extends:
  - toc.yml
  - toc_pdf.yml
  - toc_pdf_alternates.yml
```

This will at least make diff'ing the list of files a little easier over time. :)

agoose77:
> Thank you! This is very helpful for a book project I am working on!
> 
> Just musing... It would be great if we could find a way to them keep the regular toc and the PDF export ones in sync, but I've moved all this config to separate yaml files:
> 
> ```yaml
> extends:
>   - toc.yml
>   - toc_pdf.yml
>   - toc_pdf_alternates.yml
> ```
> 
> This will at least make diff'ing the list of files a little easier over time. :)

I agree with this point — we would like to give this some design thought!

choldgraf (approved):
Many thank for this! It's really helpful when people share their config setups for common gotchas. I think the point @agoose77 makes is about a follow-up improvement, so I'll merge this one as an incremental improvement for now, thanks!

PR #2807, opened 2026-04-09
Document credit role more clearly

choldgraf:
This just adds a short documentation section in `authorship.md` about credit roles so that it is more discoverable


rowanc1:
Nice!

Issue #2809, opened 2026-04-11
anywidget nodes cause fatal silent errors during MECA exports


stevejpurves:
## Description

When we include anywidgets in a myst project, exports to at least meca fail silently.

I have seen this in CI build and the only error is:

```
Error: ENOENT: no such file or directory, scandir '/Users/gvarnavides/Documents/myst-sites/presentation-20260408-sem-phase-retrieval/_build/temp/mystsCp7LY/files'
    at Object.readdirSync (node:fs:1590:26)
```

When running a normal build locally I see the following warning

```
⛔️ 03.transmission-vs-reflection.md container of kind figure contains no valid content besides caption
   Valid content types include image, referenced notebook cell, table, code, iframe, subfigure
```

## Proposed solution

It seems that some block level nodes are whitelisted for inclusion in figures, where `anywidgets` have not been added to that list. we should add it to that list.

## Additional notes

* Discovered by @gvarnavides and jointly debugged
* Fix identified and in progress, PR incoming

gvarnavi:
Thanks @stevejpurves! Was just about to open this issue -- glad to see you beat me to it, and already have a fix!

Perhaps useful to note the workaround for anyone deploying to curvenote, the `any:widget` directive works fine.

PR #2810, opened 2026-04-13
🔧 fix exports when anywidgets are in containers

stevejpurves:
- 🔧 no longer reading dir that may not exist
- 🏎️ moved to non-blocking fs calls, as any sync call blocks the whole of node
- 🖼️ add anywidget as a subfigure type, while here because they could be

### DIFF
--- a/packages/myst-cli/src/transforms/anywidgets.ts
+++ b/packages/myst-cli/src/transforms/anywidgets.ts
@@ -1,4 +1,4 @@
-import fs from 'node:fs';
+import { access, readdir } from 'node:fs/promises';
 import type { GenericParent } from 'myst-common';
 import { RuleId } from 'myst-common';
 import { computeHash, hashAndCopyStaticFile, isUrl } from 'myst-cli-utils';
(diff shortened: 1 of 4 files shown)


packages/myst-cli/src/transforms/anywidgets.ts
```diff
@@ -50,9 +50,15 @@ export async function transformWidgetStaticAssetsToDisk(
         const stem = computeHash(attrPath);
 
         // Check whether file with stem exists (but unknown extension)
-        const exists = fs.existsSync(writeFolder);
-        const existingName = fs.readdirSync(writeFolder).find((f) => path.parse(f).name === stem);
```
stevejpurves, line 54: This specifically causes #2809

packages/myst-cli/src/transforms/anywidgets.ts
```diff
@@ -50,9 +50,15 @@ export async function transformWidgetStaticAssetsToDisk(
...
-        if (exists && existingName !== undefined) {
+        let existingName: string | undefined;
+        try {
+          const entries = await readdir(writeFolder);
```
stevejpurves, line 55: moving to non-blocking as good node practice

agoose77: Are these two bugfixes in one PR? If so, this transform logic is a duplicate of that for image handling — we should update the image transforms too.

For the original impl I just copied the code because we'll rewrite all of this down the road for standardised asset handling.

PR #2819, opened 2026-04-21
Don't check links with strict mode

### LINKED ISSUE #2122: not mandate `check-links` if `strict` is set
## Proposal

As [discussed on discord](https://discord.com/channels/1083088970059096114/1126886232060870678/1359885379783622666) it would be useful if there was:

> a way to turn off the external link check but keep the internal link check with warnings raised to errors 

> ...if a contributor contributes something with a messed up internal link, we want it to fail CI, but if an external link has rotted in the meantime we don't want it to fail CI just because of that

@agoose77:
> We could add a --no-check-links flag. 
> But, I wonder whether we should just not mandate check-links if strict is set. 

## Additional notes

@agoose77 [here](https://discord.com/channels/1083088970059096114/1126886232060870678/1359894267480244307): 
> If we like that UX, though, the deprecation path would be:
>
>`myst build --strict`
> shows a warning about `--check-links`
>
> `myst build --strict --check-links`
> is silent
>
> `MYST_CHECK_LINKS=no`
> `myst build --strict`
> turns off the link checker until we leave the deprecation period

_(Potentially a different issue, but including for completeness)_
@agoose77 [here](https://discord.com/channels/1083088970059096114/1126886232060870678/1359937353187332348)
> Something else I'd been toying with is the idea of first-seen links vs old-links -- something we've seen before should not fail CI in the same way as a new, bad link

### LINKED ISSUE #2818: External link checking is problematic; let's separate that out from `--strict`
## The problem 

This overlaps strongly with https://github.com/jupyter-book/mystmd/issues/2122 and https://github.com/jupyter-book/mystmd/issues/2406. I'm opening a new issue here basically just to emphasize that this is a top priority right now for Project Pythia and we'd love to help converge on a solution.

There is currently no way to disable or configure external link checking when building with `--strict`. Since we moved to use `--strict` for all cookbook builds [a couple of months ago](https://github.com/ProjectPythia/cookbook-actions/pull/160), we have experienced a lot of headaches related to external link checking:
- Many perfectly valid links cause failures (I think because more and more web servers are rejecting these queries somehow). https://github.com/jupyter-book/mystmd/issues/2322 is a useful workaround but becomes very labor-intensive when we're managing hundreds of cookbooks repos with thousands of links
- Sometimes (and in some cases, unpredictably) the build times can become extraordinarily long as the link checker seems to hang indefinitely. This has led to huge increases in our consumption of Actions minutes.
- Failure of nightly cookbook builds because of temporary outages of external sites is not really our desired behavior.

These problems have become pervasive enough that Project Pythia has had to [disable strict building as default](https://github.com/ProjectPythia/cookbook-actions/pull/180) once again , hopefully temporarily. This means degrading our quality control since we will be allowing broken notebook code to be deployed on our cookbook sites.

## Potential solution

To me the most sensible solution was suggested [here in 2122](https://github.com/jupyter-book/mystmd/issues/2122#issuecomment-4025722144) by @choldgraf:
- `myst build --strict` - error enforcement only, including internal links and cross-references
- `myst build --check-links` - external link checking only
- `myst build --strict --check-links` - do both

Other suggestions for how to configure `-strict` have been made, and they would get the job done. But it makes a lot of semantic sense to me to treat the external link checking entirely separately as in the above. In fact, this would get us back to functionality that was available in Sphinx-based JupyterBook v1.

I'm not sure anyone on our team has the Javascript chops to open a PR for this. But we would love to help the community converge on the right solution and get it implemented.

choldgraf:
This separates the `--strict` and the `--check-links` logic in our build. It means that:

1. Links will only be checked if `--check-links` is explicitly given
2. Link warnings will only result in a fail if `--strict` is given

It also cleans up the documentation about this in general a little bit, so that information is a bit clearer and easier to find.

This _doesn't_ implement a deprecation pathway. Is that a huge problem? (personally, I think it'd be OK to not do this with a deprecation pathway, but if folks feel strongly I'm fine w/ it)

**Note**: The failing test doesn't seem to be related to this:

```
 FAIL  src/transforms/doi.spec.ts > DOI Resolvers for 'BibTeX' > markdown link with strange characters resolves
```

---

- resolves https://github.com/jupyter-book/mystmd/issues/2122
- resolves https://github.com/jupyter-book/mystmd/issues/2818

### DIFF
--- a/packages/myst-cli/src/process/site.ts
+++ b/packages/myst-cli/src/process/site.ts
@@ -545,7 +545,6 @@ export async function processProject(
     execute,
     maxSizeWebp,
     checkLinks,
-    strict,
   } = opts || {};
   if (!siteProject.path) {
     const slugSuffix = siteProject.slug ? `: ${siteProject.slug}` : '';
@@ -604,7 +603,7 @@ export async function processProject(
     pagesToTransform.map((page) =>
       postProcessMdast(session, {
         file: page.file,
-        checkLinks: checkLinks || strict,
+        checkLinks,
         pageReferenceStates,
         extraLinkTransformers,
         site: true,
--- a/packages/mystmd/tests/endToEnd.spec.ts
+++ b/packages/mystmd/tests/endToEnd.spec.ts
@@ -13,6 +13,7 @@ type TestCase = {
   env?: Record<string, any>;
   timeout?: number;
   command: string;
+  expectFailure?: boolean;
   outputs: {
     path: string;
     content?: string;
@@ -43,7 +44,7 @@ describe.concurrent('End-to-end cli export tests', { timeout: TIMEOUT }, () => {
   const cases = loadCases('exports.yml');
   test.each(
     cases.filter((c) => !only || c.title === only).map((c): [string, TestCase] => [c.title, c]),
-  )('%s', async (_, { cwd, env, command, outputs, timeout }) => {
+  )('%s', async (_, { cwd, env, command, expectFailure, outputs, timeout }) => {
     // Clean expected outputs if they already exist
     await Promise.all(
       outputs.map(async (output) => {
@@ -52,8 +53,13 @@ describe.concurrent('End-to-end cli export tests', { timeout: TIMEOUT }, () => {
         }
       }),
     );
-    // Run CLI command
-    await exec(command, { cwd: resolve(cwd), env: { ...process.env, ...env }, timeout });
+    // Run CLI command and assert non-zero exit when expectFailure is set
+    const run = exec(command, { cwd: resolve(cwd), env: { ...process.env, ...env }, timeout });
+    if (expectFailure) {
+      await expect(run).rejects.toThrow();
+    } else {
+      await run;
+    }
     // Expect correct output
     outputs.forEach((output) => {
       expect(fs.existsSync(resolve(output.path))).toBeTruthy();
--- a/packages/mystmd/tests/exports.yml
+++ b/packages/mystmd/tests/exports.yml
@@ -183,6 +183,11 @@ cases:
     cwd: strict-no-check-links
     command: myst build --strict
     outputs: []
+  - title: Check-links with strict fails on broken links
+    cwd: strict-no-check-links
+    command: myst build --check-links --strict
+    expectFailure: true
+    outputs: []
   - title: Basic site build
     cwd: basic-site
     command: myst build
--- a/packages/mystmd/tests/exports.yml
+++ b/packages/mystmd/tests/exports.yml
@@ -181,11 +181,11 @@ cases:
         content: dois/outputs/dois.bib
   - title: Strict build does not check external links
     cwd: strict-no-check-links
-    command: myst build --strict
+    command: myst build --site --strict
     outputs: []
   - title: Check-links with strict fails on broken links
     cwd: strict-no-check-links
-    command: myst build --check-links --strict
+    command: myst build --site --check-links --strict
     expectFailure: true
     outputs: []
   - title: Basic site build
--- a/packages/mystmd/tests/exports.yml
+++ b/packages/mystmd/tests/exports.yml
@@ -179,6 +179,10 @@ cases:
         content: dois/outputs/dois.tex
       - path: dois/_build/main.bib
         content: dois/outputs/dois.bib
+  - title: Strict build does not check external links
+    cwd: strict-no-check-links
+    command: myst build --strict
+    outputs: []
   - title: Basic site build
     cwd: basic-site
     command: myst build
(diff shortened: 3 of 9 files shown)


packages/myst-cli/src/process/site.ts
```diff
@@ -604,7 +603,7 @@ export async function processProject(
     pagesToTransform.map((page) =>
       postProcessMdast(session, {
         file: page.file,
-        checkLinks: checkLinks || strict,
+        checkLinks,
```
choldgraf, line 606: This is the thing that changes behavior, all the other stuff is documentation and testing

rowanc1: Confirmed on my review that this is the only substantial change!

rowanc1 (approved):
Looks good to me, haven't tested locally. But tests from CI looks good.

packages/mystmd/tests/endToEnd.spec.ts
```diff
@@ -13,6 +13,7 @@ type TestCase = {
   env?: Record<string, any>;
   timeout?: number;
   command: string;
+  expectFailure?: boolean;
```
choldgraf, line 16: This is just a little flag for our tests so that we can test an "expect this to fail" case for the link check...we might be able to use this elsewhere too but I just wanted to keep it targeted for now

packages/mystmd/tests/exports.yml
```diff
@@ -179,6 +179,15 @@ cases:
         content: dois/outputs/dois.tex
       - path: dois/_build/main.bib
         content: dois/outputs/dois.bib
+  - title: Strict build does not check external links
```
choldgraf, line 182: This build has a broken link...the first test passes because we aren't passing `--check-links` and the second one should fail.

agoose77:
LGTM! Thanks Chris! Thanks for reviewing Rowan!

da5nsy:
Hey folks, we're testing this [over on TTW](https://github.com/the-turing-way/the-turing-way/pull/4619) and getting some residual issues.
It looks like it is still checking (some?) DOIs (and failing on some, which seem to be available when manually tested but maybe 404d at build?).

```
...
🪄  Linked 3 DOIs from doi.org in 14 s for reproducible-research/overview/overview-barriers.md
📖 Built reproducible-research/overview/overview-barriers.md in 22 s.
🪄  Linked 2 DOIs from doi.org in 14 s for foreword/governance/ways-of-working.md
📖 Built foreword/governance/ways-of-working.md in 22 s.
🪄 Updated link text for 2 external MyST references in 977 ms
🪄 Updated link text for 1 external MyST reference in 961 ms
🪄 Updated link text for 2 external MyST references in 1.15 s
🪄 Updated link text for 3 external MyST references in 1.19 s
📚 Built 485 pages for project in 1.05 min.

reproducible-research/rdm/rdm-repository.md
  - ⛔ Citation data from doi.org was not available or malformed for doi:10.1007/s11192-022-04346-1

Site has 1 error and 0 warning, stopping build.
make: *** [Makefile:33: ci] Error 1
Error: Process completed with exit code 2.
```

bsipocz:
I regularly see sporadic DOI issues like this, usually a restart resolves the build. I suppose there maybe some rate limit set at doi.org?

da5nsy:
> I regularly see sporadic DOI issues like this, usually a restart resolves the build. I suppose there maybe some rate limit set at doi.org?

That may be true, but this PR is called "Don't check links with strict mode", and this result shows that some links are still being checked in strict mode. And also that those link-check failures are failing the build in strict mode.

dcamron:
To my understanding this is a specific output of the [automatic biblio generation and validation from DOI links](https://mystmd.org/guide/citations#doi-links) on build.

The [manual biblio generation recommendation](https://mystmd.org/guide/citations#writing-dois-to-bibtex) from the docs might be a good solution for those particular headaches.

choldgraf:
Something tells me that the DOI checking is a different mechanism than the more general link-checking pathway, since DOIs are treated as special things in MyST as opposed to regular links. Could you open an issue scoped to something like `Allow users to disable DOI checking for some or all DOI links`?

bsipocz:
Thanks for pointing to the bibtex route, but that doesn't work for us, at least not until we have way more cross referencing that are seriously affecting build time. We only have a handful of DOIs, so and this erroring out is sporadic enough that it's totally fine to live with it.

dcamron:
> different mechanism

To that end, if it's helpful: I think this can also be handled in [myst.yml `error_rules`](https://mystmd.org/guide/settings#setting-error-rules) under `doi-link-valid`, either through individual exceptions or altogether.

FWIW I do think it could be useful to clarify and/or control this within the umbrella of link checking or a similar CLI flag, and I relate to the confusion.

PR #2822, opened 2026-04-21
Add special case for sept for DOI metadata

### LINKED ISSUE #2821: Failing DOI test on main
We have a failing DOI test, I think because something has changed upstream about the data that's returned.

- In [this step](https://github.com/jupyter-book/mystmd/actions/runs/24656668717/job/72091792289#step:4:5669)
- We check this URL: https://doi.org/10.1002/(SICI)1096-987X(199709)18:12%3C1450::AID-JCC3%3E3.0.CO;2-I
- Somethings is going wrong with the data comparison when it returns

choldgraf:
It seems like doi.org is returning `sept` instead of `sep` and this was causing our bibtex parsing to skip it... this normalizes `sept` to `sep` if it encounters it.

Ideally, citation-js would know how to parse `sept` but this feels like a reasonable fix to add here in the meantime. I'm not sure why the data would suddenly change on us, so I noted that we might need to extend the function if we notice other kinds of changes to the DOI data structure

---
- closes https://github.com/jupyter-book/mystmd/issues/2821

### DIFF
--- a/.changeset/quiet-papayas-teach.md
+++ b/.changeset/quiet-papayas-teach.md
@@ -1,0 +1,5 @@
+---
+"citation-js-utils": patch
+---
+
+Add special case for sept
--- a/packages/citation-js-utils/src/index.ts
+++ b/packages/citation-js-utils/src/index.ts
@@ -180,13 +180,25 @@ function parseCitationStyle(style: string): string {
 }
 
 /**
+ * Normalize non-standard BibTeX month string macros that citation-js does not recognize.
+ * Add to this if we find other edge-cases / one-offs to fix.
+ */
+function normalizeBibTeXMonths(source: string): string {
+  // Crossref's BibTeX export from doi.org sometimes emits `month=sept` for
+  // September. citation-js drops it when this happens because it only
+  // expects "sep". So convert sept -> sep below.
+  source = source.replace(/(\bmonth\s*=\s*\{?)sept(\b)/gi, '$1sep$2');
+  return source;
+}
+
+/**
  * Parse a BibTeX string into an array of CSL items
  *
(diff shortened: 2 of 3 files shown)


rowanc1:
Another fix could be to drop the month from our tests and report the bug upstream? Our citation formatters do not show the month, so I don't think our users would notice the difference.

The upstream code is here for parsing:

https://github.com/citation-js/citation-js/blob/25037b55f70e0a44d26a40da715ff5f3c426b9dd/packages/plugin-bibtex/src/mapping/shared.js#L88

The data change is coming from the crossref API, so this is maybe more months as well that they also changed in the lastest release.

My gut feeling is that it doesn't feel great to do just `sept --> sep` and also doesn't feel great to take on responsibilities of bibtex parsing/preprocessing. I didn't look into the Regepxp too deeply, but there might be other unintended consequences to it?

Overall though, I think we can just drop this part of the test, as most of our users are not going to be effected.

choldgraf:
I thought about that too - if you don't think a user would hit this then IMO we should drop the `9` from the test. I was assuming that a user would hit this and that's why I took this path. What do you think?

rowanc1:
Month info doesn't show up anywhere in our UIs, so I think there is zero user impact. We do not expose the csl anywhere, so this is entirely internal:

- bibtex adds a `t` in this case (if they exported, rather than generate on the fly), CSL internally drops the month, our UI doesn't change.

I think we should change the test, not take on more responsibility!

It probably is worth a follow up with at least citationjs.

choldgraf:
that makes sense to me! closing this in favor of

- https://github.com/jupyter-book/mystmd/pull/2823

PR #2823, opened 2026-04-22
🐛 Reduce scope of citation tests for `sept` months

### LINKED ISSUE #2821: Failing DOI test on main

rowanc1:
An alternative PR to #2822 that reduces the scope of our citation parsing to ignore a month.

We don't use this in the UI, so we can just roll with the changes that crossref are doing that are somewhat incompatible with citation-js.

- closes #2821

### DIFF
--- a/packages/myst-cli/src/transforms/doi.spec.ts
+++ b/packages/myst-cli/src/transforms/doi.spec.ts
@@ -38,7 +38,7 @@ const BARTELS_1997_CSL_JSON = [
     'container-title': 'Journal of Computational Chemistry',
     issue: '12',
     issued: {
-      'date-parts': [[1997, 9]],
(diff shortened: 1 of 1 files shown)


choldgraf (approved):
LGTM - I opened up an upstream issue here

- https://github.com/citation-js/citation-js/issues/263

PR #2827, opened 2026-04-24
Expose content server with sockets on starting dev server

stevejpurves:
This exposes the content server details throught the `startServer` opening up more possibilities for the cli, and clients of the `myst-cli` library to interact directly with the content server and front ends via wockets.


agoose77:
This looks fine — exporting more information so that downstream libraries like Curvenote CLI can consume the server. This is a breaking change, but it's in an internal module that I don't think we make major compatibility statements for.

PR #2832, opened 2026-04-28
Add banner to the documentation

FreekPols:
Added description of banner at two places to fit with footer description.


choldgraf (approved):
This looks good! Thanks :-) 

want to add this to the jupyter-book docs as well? I think it could be as simple as a brief mention of banners and then linking to this page

PR #2834, opened 2026-04-29
Await signals to terminate processes

### LINKED ISSUE #2833: Lingering npm processes
## Description
I first noticed this behavior starting around Sunday or so. When I `myst build --html --ci`, it leaves behind an `npm` process. I've also noticed this on more than one system.

```shell
$ ps auwwx | grep -c [n]pm
0
$ myst build --html --ci > /dev/null 2>&1
$ ps auwwx | grep -c [n]pm
1
$ myst build --html --ci > /dev/null 2>&1
$ ps auwwx | grep -c [n]pm
2
$ myst build --html --ci > /dev/null 2>&1
$ ps auwwx | grep -c [n]pm
3
```

The command is `npm run start`.

This also happens if I omit the `--ci` flags and the shell redirection.

I have myst 1.8.3, npm 10.5.0.

I happened to have noticed it when a system that runs myst from cron became bogged down from all of the myst processes. I'm not sure when the behavior actually started.

choldgraf:
This adds an await step to our code that kills processes during the build. I think that it will resolve #2833. I think what happened was that we were exiting the parent mystmd process _before_ confirming that all the child-processes had been signaled for being killed, and as a result some of them were sticking around instead of being killed. This adds logic to make sure we don't move on until those signals have been sent.

it seemed to work locally, but @ryanlovett could you take a look and see if you can reproduce with this PR? I'm not sure how to test this one...

---

- resolves #2833

### DIFF
--- a/.changeset/green-hairs-wonder.md
+++ b/.changeset/green-hairs-wonder.md
@@ -1,0 +1,6 @@
+---
+"myst-cli-utils": patch
+"myst-cli": patch
+---
+
+Await signals to terminate processes
--- a/packages/myst-cli-utils/src/exec.ts
+++ b/packages/myst-cli-utils/src/exec.ts
@@ -48,17 +48,27 @@ export function makeExecutable(
 }
 
 /**
- * On Linux, child processes will not be terminated when attempting to kill their parent
- * This function uses `tree-kill` to kill the process and all its child processes
+ * On Unix-like systems, killing a process does not propagate to its children,
+ * they get re-parented and keep running. This uses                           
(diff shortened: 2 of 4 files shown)


ryanlovett:
Thanks Chris! I've never built MyST tools locally, but I'll try following the instructions at https://mystmd.org/guide/developer#developer-mystmd.

agoose77:
This should do it! If we recursively sent `kill`, I don't think we'd have to wait, but we're trying to be graceful with SIGTERM and therefore need to ensure everything is cleaned up.

PR #2848, opened 2026-05-05
📝 Experience Report Template Issue

rowanc1:
This came up in a recent community call that we can do a better job at requesting experience reports from the community on how they are using JB2/MyST and what can be improved along the way to smooth out the experience.

This puts together another issue template that can surface those and direct with some questions towards the UX/workflow that people are using.


choldgraf (approved):
This looks good to me - IMO we could merge it in and cross-link it from our docs etc and iterate from there. I'll merge tomorrow if nobody else objects or requests changes

rowanc1:
Nice. Looks good online in the template!

stefanv:
Looks great; let's cross-link!

PR #2851, opened 2026-05-14
fix(myst-cli): append index-redirect script to end of <head> instead of prepending

chad-earthscope:
Our deploys using mystmd display a white "blink" on page load, the initial page or during transition to sub pages.  The effect is more visible in some browsers than others, with Brave browser showing a pronounced "blink".  This is visible on https://myst-theme.netlify.app/.

Also, the site loads with script errors:

Screenshot 2026-05-14 at 15 18 06

This PR fixes this behavior.

Previously the index-redirect script was prepended to every built index.html's <head>, landing before SSR's first <meta charSet>. React's hydrator saw a fiber/DOM type mismatch, unmounted the SSR tree, and re-rendered client-only, visible as a multi-hundred-millisecond white flash on every page load.

Appending the script after </head>'s last SSR-rendered child keeps /foo/index.html → /foo/ redirect (it still runs before any Remix bundle scripts) while leaving every fiber position React expects intact. Trailing extras in <head> are tolerated by the hydrator.


chad-earthscope:
A recording of the blink, or flash, that I believe is addressed by this PR:
https://github.com/user-attachments/assets/dc139953-5c96-42db-91e3-e1e0d33adff9

This rendered with Safari, but is the same on Brave browser.  Less pronounced on Firefox.

choldgraf (approved):
Many thanks for this - I took a look and it makes sense to me! I did a little bit of investigating and I think what's going on is something like this (sort-of rewriting what you said above in my own words to make sure I understand it):

By injecting the script at the start of `<head>`, we were bumping down all of the other parts of head that React was  already tracking, and thus React saw a bunch of mismatched DOM errors. The React version we use is less strict about things at the *end* of head, and so moving this to the end means that React doesn't error, and instead treats it as a dangling part of `<head>` that it doesn't need to track.

I also realized that the HTML generation module was a little bit opaque in its relation to static builds and myst-theme, so I added a short docstring and linked it from our dev docs to make it easier to run this down in the future. 

I tested this and confirmed that there aren't remix errors on the built docs here!

Will merge and thanks for the bugfix @chad-earthscope !

chad-earthscope:
Much better said, thanks @choldgraf!

PR #2855, opened 2026-05-16
:left_right_arrow: Add `alternatives` directive and AST node

agoose77:
This PR adds an `alternatives` node, initially scoped to Images and Anywidgets. Whilst the concept of "preferred representation" is well defined with MIME types, it's trivial enough to generalise to other node types such as HTML and anywidgets. To begin with, we only add anywidgets, because HTML is transformed into AST by default.

example-alt

Once a user has authored an alternation, e.g.
```markdown
:::{alternatives}
![](foo.png)
![](bar.svg)
:::
```

The exporter must choose which to ultimately render. 

See https://hackmd.io/jDApA05kQFO76R3g0VOPDA for a detailed long-term plan.

## To Do
- [ ] Add renderers for `alternatives` in built-in MyST exporters.
- [ ] Add renderers for `alternatives` in MyST Theme.

## Motivation

- We want to parse outputs from code cells. 
- Each output can have multiple representations (MIME types)
- Each exporter has its own preference for MIME types
- We need a way to represent this choice in the AST

## Out of scope
This PR does not update the AST built from `` wildcard syntax. This _should_ be changed in future, once this mechanism is more well established.

This PR does not go beyond images and anywidgets. In the HackMD linked above, I set out a view that we should support more complicated alternations like:
```markdown
::::{alternatives}

:::{figure} globe.png
A static image of the world.
:::

:::{figure} spinning-globe.gif
A GIF of the world.
:::

::::
```

But I want to add that later as an extension of this mechanism. See https://agoose77.github.io/demo-myst-alternatives/ for a demonstration of this using MyST Theme components.

### DIFF
--- a/docs/alternative-content.md
+++ b/docs/alternative-content.md
@@ -1,0 +1,26 @@
+---
+title: Alternatives
+description: Specify alternative representations for videos and animated graphics.
+---
+
+The MyST Document Engine supports multiple kinds of export targets, such as PDF via Typst, and web deployments. Each of these has its own preferred set of image formats that it can display, and certain content may not be supported by one or more exporters (such as HTML). 
+
+Authors can use the {myst:directive}`alternatives` directive to define several representations of the same conceptual entity in order to allow different exports to choose their preferred representation:
+
+````{myst}
+:::{alternatives}
+![](https://gifdb.com/images/high/earth-spinning-and-rotating-zxmey0cjachu1buc.gif)
+![](http://pngimg.com/uploads/earth/earth_PNG39.png)
+:::
+````
+
+In the above example, exporters (such as the PDF exporter) that do not understand GIF images will choose the PNG representation instead. Meanwhile, exporters that support GIF images _may_ choose the GIF _if_ it is preferred over the PNG format (which it usually is). The decision about which content to render lies with the exporter, not the MyST engine.
+
+For now, the `alternatives` directive only supports the following content:
+1. Images
+2. Videos
+3. Anywidgets
+
+:::{note}
+In future, we expect to open this up to more complex types such as figures with different images and captions.
+:::
--- a/packages/myst-cli/src/transforms/images.ts
+++ b/packages/myst-cli/src/transforms/images.ts
@@ -169,20 +169,40 @@ export function transformImagesWithoutExt(
     const wildcardRegex = /\.\*$/;
     const imagePath = path.join(path.dirname(file), image.url).replace(wildcardRegex, '');
     if (!fs.existsSync(imagePath)) {
-      const sortedExtensions = [
+      const sortedExtensions = new Set([
         // Valid extensions
         ...(opts?.imageExtensions ?? []),
         // Convertible extensions
         ...Object.keys(conversionFnLookup),
         // All known extensions
         ...KNOWN_IMAGE_EXTENSIONS,
-      ];
-      const extension = sortedExtensions.find((ext) => fs.existsSync(imagePath + ext));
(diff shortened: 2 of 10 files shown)


packages/myst-cli/src/transforms/images.ts
```diff
@@ -169,20 +169,40 @@ export function transformImagesWithoutExt(
...
+            url,
+          })),
+        };
+
+        // Clear the current object
+        Object.keys(image).forEach((k) => {
+          delete image[k];
+        });
+        // Replace with the import
+        Object.assign(image, alternatives);
```
agoose77, line 205: I initially started with this approach. I don't know if I want to do this in the first PR though — I'll probably revert.

packages/myst-directives/src/alternatives.ts
```diff
@@ -0,0 +1,28 @@
...
+  body: {
+    type: 'myst',
+    doc: 'The body of the alternative. If there is no title and the body starts with bold text or a heading, that content will be used as the alternative title.',
+  },
+  run(data: DirectiveData): GenericNode[] {
+    const root = {
+      type: 'root',
+      children: (data.body as GenericNode[]) ?? [],
+    };
+    const children = selectAll('image,anywidget', root);
```
rowanc1: I think this should probably be done later in a transform. E.g. `<img ` is likely valid, but won't be picked up here because it is an html node. I think you are also reaching inside a `mystDirective` for the anywidget, and I think the parser stage should export those consistently.

docs/alternative-content.md
```diff
@@ -0,0 +1,26 @@
...
+Authors can use the {myst:directive}`alternatives` directive to define several representations of the same conceptual entity in order to allow different exports to choose their preferred representation:
+
+````{myst}
+:::{alternatives}
+![](https://gifdb.com/images/high/earth-spinning-and-rotating-zxmey0cjachu1buc.gif)
+![](http://pngimg.com/uploads/earth/earth_PNG39.png)
+:::
+````
+
+In the above example, exporters (such as the PDF exporter) that do not understand GIF images will choose the PNG representation instead. Meanwhile, exporters that support GIF images _may_ choose the GIF _if_ it is preferred over the PNG format (which it usually is). The decision about which content to render lies with the exporter, not the MyST engine.
```
rowanc1, line 17: Is there an ordering component to this? i.e. you put the gif first here, is that order aiming to be meaningful?

packages/myst-cli/src/transforms/images.ts
rowanc1, line 165: We may need to consider an image without an extension that is inside an alternative already...!

rowanc1 (commented):
A few early comments!

docs/alternative-content.md
agoose77: So far, no. The intention is to declare alternative representations with sufficient information that the renderer can choose accordingly.

I don't yet think it's totally clear whether there _is_ an ordering that the engine can define before knowing the renderer. If we don't define an order, then we don't have to get it right.

The obvious example of non-singular ordering is if two backends both support an image format, but have different preferences for it.

packages/myst-directives/src/alternatives.ts
agoose77: I suspect that's the right move too — the validation is slightly fiddlier at this stage.

packages/myst-cli/src/transforms/images.ts
agoose77: Yes, need to check whether this logic will be re-entrant somehow.

PR #2862, opened 2026-05-22
Add extra guidance around changesets and myst-cli bumping

choldgraf:
This does some extra clean-up to put guardrails around the confusion we had when we made a release without `myst-cli` in the changeset. It:

1. Adds a little job to the changeset release PR to check if `myst-cli` is in there, and will fail CI there if so, with a link to the docs about this in the error message. That way we have a failure signal _before action_ rather than after, and a person can read up on the docs and decide if they want to merge anyway.
2. Adds some more documentation describing why this is set up this way


.github/workflows/ci.yml
```diff
@@ -27,6 +27,13 @@ jobs:
         run: bun run lint:circular
       - name: Ensure changeset has the right structure and doesn't contain ignored packages
         run: bun run version
+      - name: Ensure release PR consumes a myst-cli changeset
```
rowanc1: Can we move this to a different workflow so we can see it from the UI and it runs much faster?

docs/developer.md
```diff
@@ -560,7 +560,9 @@ When you publish a release, you upload a new version of the tool for package man
 
 - **Find the changesets PR**. This contains a list of the version updates that will be included with this release. [Here's an example of a release PR](https://github.com/jupyter-book/mystmd/pull/1896).
 - Review the changeset PR.
-  - Ensure that `mystmd` is in the changesets. Or you are intentionally **not** releasing `mystmd` (this generally shouldn't happen), in which case the python and release notes are expected to fail.
+  - Ensure that `myst-cli` is in the changesets (which will release `mystmd`).
```
rowanc1: This can be mystmd or myst-cli, can we update the tests to check for one or the other?

.github/workflows/ci.yml
```diff
@@ -27,6 +27,13 @@ jobs:
...
+        if: github.event_name == 'pull_request' && github.head_ref == 'changeset-release/main'
+        run: |
+          if ! git diff HEAD^1 -- .changeset | grep -q myst-cli; then
```
rowanc1: Can this also be mystmd?

Maybe there is a way to use changesets CLI to do this?

This would trigger if the regexp is in the body of the changeset, so a bit fragile.

rowanc1 (commented):
Left a few comments. I think it should be in a different workflow at least. Can ship it then, but if you have time improving the lookup to just be in the head of the changeset, not the body.

Can you also check that it is only looking at the MD files in that folder? The json in there will always have a myst-cli.

.github/workflows/ci.yml
rowanc1: This I think at least needs to be at least limited to MD files.

choldgraf:
Yep sounds good - let me take another pass at it, thanks for the comments!

docs/developer.md
choldgraf: ah sorry - I thought that `myst-cli` *was* mystmd in changesets language

rowanc1: There are two packages, `mystmd` imports `myst-cli`. `myst-cli` contains the functions that can be used in other places, scripts, etc. the `mystmd` just creates a binary, no way to import from there, no dependencies.

choldgraf: I meant at the level of "what a person puts in a changeset" - I hadn't seen a person put this:

```
---
'mystmd': patch
---
```

only this

```
---
'myst-cli': patch
---
```

But this brings up a good point, I think we can just test the .json of the mystmd in the changesets PR rather than parsing the markdown yaml header!

rowanc1: You can do both. `myst-cli` is going to be more common.

choldgraf:
ok the latest commit moves the workflow to a new file, and also updates the logic so that it parses the `.json` of `mystmd` and checks that the version is different from what it was before. That's less hacky than parsing a bunch of markdown YAML headers haha

docs/developer.md
```diff
@@ -755,6 +757,20 @@ For now, we try to abide by the following rules for version bumps:
 - **minor**: Backward incompatible change to the JavaScript API, for example, changing the call signature or deleting an exported function. These can be a headache for developers consuming MyST libraries, but they do not break MyST content.
 - **patch**: For now, everything else is a patch: bug fixes, new features, refactors. This means some patch releases have a huge, positive impact on users and other patch releases are basically invisible.
 
+(changesets:myst-cli)=
+#### What if changesets don't bump `mystmd` or `myst-cli`?
```
rowanc1:
```suggestion
#### What if changesets doesn't bump `mystmd` or `myst-cli`?
```

.github/workflows/check-release-pr.yml
```diff
@@ -0,0 +1,26 @@
...
+      # So this checks that the .json has been updated, and if not raises an error to
+      # make sure the merger does this intentionally!
+      - name: Compare mystmd version on main vs this PR
+        run: |
+          old=$(git show HEAD^1:packages/mystmd/package.json | jq -r .version)
+          new=$(jq -r .version packages/mystmd/package.json)
+          if [ "$old" = "$new" ]; then
+            echo "::error::No changeset bumps 'mystmd' or 'myst-cli' (mystmd still $old). See https://mystmd.org/guide/developer#changesets-myst-cli"
+            exit 1
+          fi
```
rowanc1, line 26: Looks good, haven't tested.

rowanc1 (approved):
Haven't tested, but looks good from code. Thanks for pulling it out to a separate workflow, and testing the resulting .json is much better.

rowanc1:
Lets put it in and see if it works!

PR #2865, opened 2026-05-26
Add a brief guide about AST structure and metadata

choldgraf:
This adds a page to the `myst-specification` documentation to explain our AST structure a little bit more, and give developer-level information for those who are interested in building their own parsers, renderers, etc. I tries to clarify some of the hidden information in there, help others know where to look, etc.

We should still improve the ways that we talk about the myst spec vs. what's in mystmd. It's hard for readers to know things like "is the project xref data store part of the spec, or just a mystmd feature?" similarly for the specific parts we support (e.g. "abstract"), but I didn't want to make this PR too complex. If others are interested in that content, I'd be happy to add it.

Would love feedback from a user-perspective (@ciyer ? and `@sinclair` if I can figure out their username 😅) - it'd also be helpful to get a review for overall correctness from another core team member to make sure I am not getting something wrong.


ciyer (commented):
This is super helpful, and clarifies several questions I had had before reading. 🙌 

The only thing I would ask for is to provide some information about how the spec is versioned and how that is communicated in the AST.

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+Every node in the AST is a JSON object with a `type` field that names what kind of node it is.
+Beyond `type`, a node has either a `value` or `children`:
+
+- **Leaf nodes** have a `value` (typically a string).
+- **Parent nodes** have a `children` array (more nodes).
+
+Nodes may also carry additional properties - `url` on a link, `depth` on a heading, `identifier` on a cross-reference, etc.
+These properties are specific to each node type and are documented in the [node reference](./myst.schema.md).
+
+Here is the smallest meaningful AST: a single line of text.
```
parmentelat: you mean the smallest meaningful MyST, right ?

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+
+```{myst}
+Hello, world!
+```
+
+Click the **AST** tab above to see the AST. (The tab displays YAML for readability; MyST sites serve it as JSON.)
+You'll find a `root` node (always the outermost node) wrapping a `paragraph`, which wraps a `text` leaf.
```
parmentelat, line 47: maybe already mention here that https://mystmd.org/sandbox allows to see the AST of any myst fragment ?

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+### Which stage will you encounter?
+
+- The **spec's [test cases](https://unpkg.com/browse/myst-spec/dist/)** describe the **PRE** form. They are the canonical "source -> AST" pairs that any conforming parser must produce.
+- The [node reference](./myst.schema.md) documents node *types* that may appear in either stage. Some properties (like `resolved` on `crossReference`, or computed numbering fields) appear only after transforms.
+
+In short:
+
+- A parser produces PRE. Match the test cases.
+- A renderer receives POST. It's what `myst build` emits.
+- A transform takes PRE-like input and produces POST-like output.
```
parmentelat: naive/stupid question surely, but how can we invoke/chain several transforms then if input and output are of different types ?

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+
+When directive parsing fails, the POST AST signals it in one of two ways:
+
+- **Unknown directive name** (no plugin registered): the node stays as a `mystDirective` leaf, with the original source text in its `value`.
+- **Invalid arguments or body**: the node's type becomes `mystDirectiveError`, and its children are dropped. Renderers can use this to display error messages in the output.
+
+## Plugins often emit a custom node that is transformed
+
+A common plugin pattern uses a directive (or role) to emit a *placeholder* node, then a separate transform to expand that placeholder into real content.
+This is useful for anything you can't do inside `run()`: async work, anything that needs the whole document for context, or anything that depends on cross-references being resolved first.
```
parmentelat: I did not get that sentence; what does `run()` refer to exactly ?
in any case I'd suggest to move this below the example

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+
++++
+
+A second chunk.
+```
+
+In the AST (POST), each chunk ends up inside a `block` node. Recognized keys (`kind`, `class`, `label`, `identifier`) are moved onto the `block` node, anything else is put under `block.data`. See [](./features/blocks.md) for the formal block specification.
+
+## "Parts": named regions of a page
+
```
parmentelat, line 155: disclaimer: I have note been able to follow this section on parts (have never been exposed to that..)

parmentelat (commented):
from a candid user: not being able to judge accuracy or completeness

nevertheless I find this addition truly relevant, we're much better off with it than without it !

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+If you just want to author MyST documents, you probably don't need this page - head to the [MyST Guide](xref:guide).
+If you want to build a renderer, write a parser, or generate AST from a plugin, read on.
+:::
+
+## The spec and mystmd
+
+This primer covers both **the MyST Specification** (this site) and **mystmd** (the official reference implementation):
+
+- The **spec** defines what a conforming MyST parser must produce - node types, their properties, and how they correspond to source. It evolves through the [MyST Enhancement Proposal process](https://mep.mystmd.org).
+- **mystmd** implements the spec, then goes further: it runs transforms that resolve cross-references and apply numbering, emits additional node types that aren't yet in the spec, and applies its own conventions (like named parts of a document).
```
sinclairtarget: Would it make sense to characterize `mystmd` as an _application_ vs. a parser? (Do you think of it that way?)

I wonder if that would help motivate these additional transforms that `mystmd` does. That's how I've been thinking of it—`mystmd` does this additional stuff because it is, among other things, a static site generator on top of just a MyST parser.

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+
+If you're writing a parser, stick to the spec. mystmd's conventions (like specific named parts) aren't part of the spec, but they're still useful to know, since most MyST content today comes from mystmd. This page flags which is which as we go.
+
+## The shape of a node
+
+Every node in the AST is a JSON object with a `type` field that names what kind of node it is.
+Beyond `type`, a node has either a `value` or `children`:
+
+- **Leaf nodes** have a `value` (typically a string).
```
sinclairtarget: Is this true? I guess I don't know what's actually in `mystmd` but I can think of at least one leaf node without a `value` field according to the available node type index: A link definition has a `label`, `url`, and `title` but no `value` field

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+
+Directives change shape between PRE and POST (roles behave similarly):
+
+```{myst}
+:::{note}
+Heads up!
+:::
+```
+
+In **PRE**, you'll see a `mystDirective` wrapper (name: `note`) containing an `admonition` child. The wrapper records the directive that produced it. The child is what its `run` function returned.
```
sinclairtarget: The reference here to a `run` function seems unnecessarily specific to the existing implementation

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+The markup characters (`**`, `*`) are gone from the source text. They've been turned into nested nodes (e.g. `*word*` becomes a `type: emphasis` node wrapping a `text` node).
+
+(pre-and-post)=
+## Two stages: PRE and POST
+
+In the `myst-cli`, a MyST document is processed in two phases, and the AST has a different shape after each one.
+The short-hand for this is `PRE`- and `POST`-transforms.
+
+1. **PRE**: the output of *parsing*. The parser walks the source and expands directives and roles by calling their `run` functions. Cross-references still look like ordinary `link` nodes (e.g. `url: "#some-label"`). Directives appear as `mystDirective` wrapper nodes containing their expanded children. Numbering has not yet been applied.
+2. **POST**: the output after [transforms](xref:guide#develop:transforms) run over the PRE tree. Link nodes pointing at labels become `crossReference` nodes with resolved targets. Most directive wrappers are unwrapped, leaving only their semantic children. Numbering and enumeration are applied. This is the form most renderers consume.
```
sinclairtarget, line 67: I'm wondering whether, given this definition of "POST," it's really a concept that is relevant to document engines other than the reference one.

What "PRE" means is very clear. It's the shape of the AST required by the spec test cases. "POST" seems to mean "this grab-bag of transforms that happen to currently be implemented by `mystmd` before rendering."

I could imagine applications that want to do some but not all of these POST transforms. Or want to do additional transforms specific to the application before rendering. [This](https://github.com/sinclairtarget/aweigh), for example, is a silly little application that adds the anchor emoji at the beginning of all link text in a MyST document. Adding the emoji is a "transform" on the AST (at least, I think of it that way), and it's done before rendering, but does that mean the AST is a "POST" AST? Or does it not count because a transform was done that `mystmd` doesn't do?

Just throwing out a wild idea: Could it make more sense to instead define the "spec-compatible" AST (what we call "PRE" now) and then certain "canonical transforms" one can do on the tree, which might also have their own test cases in the spec? Whether or not other document engines do these transforms is up to them, but if they are done, they must be done in the same way `mystmd` does them.

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+{
+  "kind": "Article",
+  "frontmatter": { "title": "...", "authors": [...] },
+  "mdast": { "type": "root", "children": [ ... ] },
+  "references": { ... }
+}
+```
+
+The `mdast` field is the tree you've been inspecting in the AST tabs above.
+The `references` field holds cross-document data like resolved cross-references and citations, which a renderer needs when working across pages.
```
sinclairtarget, line 136: I'm a little unclear from this section what the relationship is between the AST, the spec, and this JSON structure for a built `mystmd` page.

Frontmatter is not a part of the MyST spec, right? And this JSON structure seems specific to the concerns of `mystmd` the application / static site builder. Presumably a MyST parser wouldn't have to worry about it?

sinclairtarget (commented):
This is wonderful! Clears up a lot of the questions I had and I imagine will be super helpful to anyone else trying to implement a parser

packages/myst-spec/docs/ast-primer.md
rowanc1: Yes - very much think of mystmd as a command-line application that does a lot more than parsing! We also ship `myst-parser` as one of the packages, and `myst-transforms` which implement a lot of the transformations that are inside of the application.

packages/myst-spec/docs/ast-primer.md
rowanc1: A link almost always has `children` (text nodes, which are the words shown on the page that you click on).

The only `leaf` nodes that I can think of that doesn't have a `value` is an iframe and an image. So that actually might be worth updating here.

rowanc1:
```suggestion
- **Leaf nodes** typically have a `value` (typically a string).
```

packages/myst-spec/docs/ast-primer.md
rowanc1: The below example is interactive, so will show the ast if you click on it... and that is explained below!

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+A paragraph with **bold** and *emphasis*.
+```
+
+In the AST tab, the paragraph's `children` is a flat sequence of `text`, `strong`, and `emphasis` nodes. Each `strong` and `emphasis` wraps its own `text` node.
+The markup characters (`**`, `*`) are gone from the source text. They've been turned into nested nodes (e.g. `*word*` becomes a `type: emphasis` node wrapping a `text` node).
+
+(pre-and-post)=
+## Two stages: PRE and POST
+
+In the `myst-cli`, a MyST document is processed in two phases, and the AST has a different shape after each one.
```
rowanc1:
```suggestion
In the `myst-cli`, a MyST document is processed through a series of transformations, and the AST has a different shape after each one.
```

packages/myst-spec/docs/ast-primer.md
```diff
@@ -0,0 +1,223 @@
...
+The short-hand for this is `PRE`- and `POST`-transforms.
```
rowanc1:
```suggestion
The short-hand for this in the online demo is `PRE`- and `POST`-transforms. You can get very granular by applying various [`myst-transforms`](https://npmjs.com/package/myst-transforms) in your own libraries or packages.
```

packages/myst-spec/docs/ast-primer.md
rowanc1:
```suggestion
- A transform pipeline applies a series of transformations that change the tree from `PRE` to `POST`
```

Maybe this is more clear, that it starts with PRE, applies a series of transformations, and then those show something closer to the POST (depending on which transforms you applied...).

packages/myst-spec/docs/ast-primer.md
rowanc1:
```suggestion
In **PRE**, you'll see a `mystDirective` wrapper (name: `note`) containing an `admonition` child. The wrapper records the directive that produced it. The child is what its `run` function returned from a `mystmd` plugin.
```

packages/myst-spec/docs/ast-primer.md
rowanc1: The `run` function is referring to mystmd plugins here:
https://mystmd.org/guide/javascript-plugins#defining-a-new-directive

We could update that to make it more clear...

packages/myst-spec/docs/ast-primer.md
rowanc1: @choldgraf I think we could safely delete the parts explainer from here, I think it is beyond the AST-primer and getting into mystmd specifics potentially...

rowanc1 (commented):
Left a few comments that can hopefully be used to refine.

packages/myst-spec/docs/ast-primer.md
choldgraf: OK cool - let me see if there's a natural space for this in the mystmd guide docs and we can cross-reference it as an example of something that's more myst-cli rather than myst-spec

choldgraf: I've moved the discussion of parts to the mystmd guide, so that it's just mentioned briefly in the spec and then cross-referenced there

packages/myst-spec/docs/ast-primer.md
choldgraf: I've tried to clarify the relationship between frontmatter/parts and the AST a bit more - that stuff is now documented in the mystmd guide rather than the spec!

packages/myst-spec/docs/ast-primer.md
choldgraf: I tried to clarify the difference between PRE and POST along with the additions that @rowanc1 made so hopefully this is a little bit clearer. I suspect we'll be able to improve this further, @sinclairtarget maybe you could open up a follow up PR to keep improving as you try things out?

packages/myst-spec/docs/ast-primer.md
choldgraf: added some more language to clarify this

packages/myst-spec/docs/ast-primer.md
choldgraf: tried to clarify this a bit

packages/myst-spec/docs/ast-primer.md
choldgraf: Added mention of the sandbox earlier

choldgraf (commented):
OK I think that this is pretty much ready to go - I've addressed each of the comments here, and resolved the ones that I thought were pretty clear. There's one remaining comment about how we discuss PRE/POST and transforms in general, but if possible could we bump that to a follow-up PR?

packages/myst-spec/docs/ast-primer.md
sinclairtarget: Happy to do that!

sinclairtarget:
Looks great to me! Thanks again for adding this

agoose77:
I'm merging this — we shouldn't block on the clear good-enough consensus here.

PR #2873, opened 2026-05-28
Document site options config w page frontmatter

### LINKED ISSUE #2869: Not clearly documented how to disable the Secondary Sidebar per page
### Proposal

There are some pages where I think the right hand side sidebar is more confusing or just looks odd. I'd be nice to turn it off on a per-page basis. 

I tried various versions of `hide_outline` in front matter and couldn't figure out if this works...

### Additional notes

_No response_

choldgraf:
This clarifies how site options are specified when using page frontmatter, and adds a bunch of cross-references so this information is easier to find.

---

- closes #2869 (I think, assuming this is the problem!)

### DIFF
--- a/docs/website-templates.md
+++ b/docs/website-templates.md
@@ -123,7 +123,9 @@ If both `getting-started.md` and `getting-started/index.md` exist, the sibling f
 ### Page Options
 
 Any option from `site.options` can be overridden per-page in the frontmatter under the `site` key.
-Note that the nesting is different: options are placed directly under `site:` in page frontmatter, not under `site.options:`.
+
+(page-site-options-note)=
+When setting an option from a page, omit the `options:` key - put it directly under `site:`, not `site.options:`. See [](#site-options-page).
 
 ```{code-block} yaml
 :filename: my-page.md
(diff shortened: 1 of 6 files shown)


stefanv:
I think the idea here is right but I'm not sure about the execution. Do we want to manually sprinkle these throughout? Maybe a glossary term, or a standard way to link to a section that explains it? Maybe we always use a standard phrase that links there. A sentence for each occurrence feels like a bit much?

docs/website-templates.md
```diff
@@ -118,6 +118,8 @@ getting-started/
 If both `getting-started.md` and `getting-started/index.md` exist, the sibling file takes the slug `getting-started` first, and the folder's index file gets a deduplicated slug like `getting-started-1`.
 :::
 
+(site-options-page)=
+
 ### Page Options
```
cycomachead, line 123: Ah this was the section I missed. I think this is helpful, mostly linking to this section

choldgraf:
Good point, I was wondering the same thing - what if we defined it in one place and used an "embed" role to reuse it throughout?

docs/website-templates.md
rowanc1: We should probably at least warn on page level `site.options` and alternatively unpack a level higher...

rowanc1 (approved):
This looks good. I think there is a follow up in code that we could do though as well!

choldgraf:
ok I pushed a little commit to re-use via embedding per @stefanv 's comments - will merge this one now as an iterative improvement, but definitely agree a warning would be helpful here!

Tracking that here:

- https://github.com/jupyter-book/mystmd/issues/2893

PR #2874, opened 2026-05-28
Add kernel startup timeout and retry + choose open port on execute

### LINKED ISSUE #2870: Flaky test for concurrent execution (and maybe a real bug)
There's a flaky test that occasionally fails unpredictably. It happens in the concurrent execution code.

[Here's a link to a flaking test workflow](https://github.com/jupyter-book/mystmd/pull/2851)

Error message in test:

```
   × End-to-end cli export tests > Concurrent execution 15024ms
     → Command failed: myst build --site --execute --execute-parallel 2

   ✓ End-to-end cli export tests > Plugins load before parts in extend yml files  1966ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  tests/endToEnd.spec.ts > End-to-end cli export tests > Concurrent execution
Error: Command failed: myst build --site --execute --execute-parallel 2

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed | 1 passed (2)
      Tests  1 failed | 53 passed (54)
   Start at  15:56:35
   Duration  62.41s (transform 5.42s, setup 0ms, collect 19.07s, tests 64.58s, environment 1ms, prepare 638ms)

error: script "test" exited with code 1
Error:  command finished with error: command (/home/runner/work/mystmd/mystmd/packages/mystmd) /home/runner/.bun/bin/bun run test exited (1)
 ERROR  run failed: command  exited (1)
error: script "test" exited with code 1
```

### References

- I thought we'd addressed this by https://github.com/jupyter-book/mystmd/pull/2853 but I guess not

choldgraf:
This PR aims to clear up two more sources of flakiness around execution.

1. When we have many concurrent kernels being spun up (either in multiple tests run concurrently, or one test with multiple kernels spun up at once), in some cases there seems to be port collisions where a port isn't correctly reported. This PR adds a little function to figure out a port that is open, and then explicitly use that. It should ensure that the execution logic never requests a port that is actually unavailable. I tried to document the rationale in the docstrings so it's clearer why we're doing it this way! 
2. Sometimes, a kernel seems to hang on _startup_ (ie it never starts). Usually startup is < 1 second. This adds a 10 second timeout, and 3 retries, to the "ask for a kernel" step. I think this should remove another class of build hangs. We might be able to shrink the timeout since it seems like the kernel startup was "less than a second, or never" but I didn't know if this would be different on other machines etc

---

- closes #2870

### DIFF
--- a/packages/myst-execute/src/kernel.ts
+++ b/packages/myst-execute/src/kernel.ts
@@ -2,20 +2,30 @@ import type { KernelSpec } from 'myst-frontmatter';
 import type { Kernel, KernelMessage, SessionManager } from '@jupyterlab/services';
 import type { IOutput } from '@jupyterlab/nbformat';
 import type { IExpressionResult } from 'myst-common';
+import type { Logger } from 'myst-cli-utils';
 import type { VFile } from 'vfile';
 import path from 'node:path';
 import assert from 'node:assert';
+import { setTimeout as delay } from 'node:timers/promises';
 
 type ISessionConnection = Awaited<ReturnType<SessionManager['startNew']>>;
 type ISessionConnectionWithKernel = ISessionConnection & {
   kernel: NonNullable<ISessionConnection['kernel']>;
 };
 
+// A kernel can start but never finish connecting, which would hang the build
+// forever. Wait for it to be ready, and restart it a few times if it isn't.
+// Healthy kernels connect in about a second, so this timeout is intentionally
+// short to recover quickly.
+const KERNEL_READY_TIMEOUT_MS = 10_000;
+const KERNEL_READY_ATTEMPTS = 3;
+
 export async function createKernelConnection(
   sessionManager: SessionManager,
   basePath: string,
   kernelspec: KernelSpec,
   vfile: VFile,
+  log?: Logger,
 ): Promise<ISessionConnectionWithKernel | undefined> {
   const sessionOpts = {
     type: 'notebook',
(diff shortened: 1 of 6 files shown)


packages/myst-execute/src/manager.ts
```diff
@@ -65,6 +66,26 @@ export async function findExistingJupyterServer(
...
+ * 2. Tracks the port number.
+ * 3. Closes the server (so that port is released)
+ * 4. Returns the port number so we know it's just been made available.
+ *
+ * We use that port in the Jupyter Server launch, so we are confident it is open.
+ */
+function pickUnusedPort(): Promise<number> {
+  return new Promise((resolve) => {
+    // By listening to port 0 it'll assign an open port, which we use later
+    const srv = createServer().listen(0, () => {
```
stefanv: Presumably should connect to 127.0.0.1 here.

But, why not provide port 0 to the Jupyter server and ask it what port it bound on?

choldgraf:
@stefanv - I've added in `127...` explicitly - I just didn't use it to keep LOC simple. I had tried using port=0 for the jupyter invocation, but Jupyter doesn't do what you'd think it would - jupyter server says it's _running_ at port:0, it doesn't grab a random port like you'd expect. That's why we gotta do it with this kinda hacky function

choldgraf:
I added another little bit of logic to add a kernel start timeout + a few retries - I think that's another class of flakiness in execution

packages/myst-execute/src/manager.ts
```diff
@@ -65,6 +66,26 @@ export async function findExistingJupyterServer(
...
+ * Ask the OS for an unused TCP port. Used so concurrent invocations of
+ * jupyter execution don't race on the same default port (8888). This function
+ * 1. Opens a server and requests an unused port.
+ * 2. Tracks the port number.
+ * 3. Closes the server (so that port is released)
+ * 4. Returns the port number so we know it's just been made available.
+ *
+ * We use that port in the Jupyter Server launch, so we are confident it is open.
+ */
+function pickUnusedPort(): Promise<number> {
```
rowanc1: Can we use this pattern, same as the start.ts:

```
import getPort, { portNumbers } from 'get-port';

await getPort({ port: portNumbers(3100, 3200) })
```

packages/myst-execute/src/kernel.ts
```diff
@@ -25,11 +35,25 @@ export async function createKernelConnection(
...
+    if (ready) {
+      return connection as any;
+    }
+    log?.warn(
+      `Kernel "${kernelspec.name}" did not become ready (attempt ${attempt}/${KERNEL_READY_ATTEMPTS}); restarting it`,
+    );
+    await connection.shutdown().catch(() => undefined);
   }
-  return connection as any;
+  return undefined;
```
rowanc1, line 56: Is this intentional?

rowanc1 (commented):
Can we use `get-port` package? Otherwise looks pretty solid, I haven't tested, but the code-review looks solid.

choldgraf:
We discussed this in the collab cafe @agoose77 made a good point that, even if the jupyter server is released, some users might be using an old version. So I am going to try and action @rowanc1 's suggestion and merge this in if the tests are happy, and we can see if this makes them happy

choldgraf:
Rowan said we're good in Discord, so I'm gonna merge this one in and we'll see if this makes things less flaky 🤞

stefanv:
Thanks, Chris!

PR #2878, opened 2026-05-29
Only copy thebe JS when necessary

stefanv:
Currently, we copy those ~120 JS files on each build (~3MB!).

This patch checks the `project: thebe: ...` (or `project: jupyter: ...` alias) configuration, and only copies thebe JS if it exists.

/cc @henryiii


rowanc1:
Nice — this looks solid! Merged!

PR #2880, opened 2026-05-29
📖 Document JupyterHub PDF export tools

### LINKED ISSUE #237: Document minimal environment for using latex and other tools in a JupyterHub
@fperez is doing some work at the moment investigating this! Hopefully we can circle back and get this into the docs for other people who are creating whole environments.

- https://github.com/berkeley-dsep-infra/datahub/pull/4277

Also add `imagemagik` for image conversions (optional, but helpful).

sgibler-droid:
## Summary
- include `imagemagick` in the conda-forge PDF export dependency install
- add a short note for shared JupyterHub / managed environments with minimal PDF export tools
- fix the ImageMagick spelling/link text in the figures dependency note

Fixes #237.

## Checks
- `npm exec --yes prettier@latest -- --check docs/installing.md docs/figures.md`
- `git diff --check`

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -105,14 +105,7 @@ $ mamba install -c conda-forge texlive-core latexmk imagemagick
 
 `imagemagick` is optional, but recommended if your documents include images that need conversion for PDF output.
 
+For Debian or Ubuntu based systems, the equivalent packages commonly include `texlive-xetex`, `texlive-fonts-recommended`, `texlive-plain-generic`, `latexmk`, and `imagemagick`.
+
 ::::
 :::::
-
-If you maintain a shared JupyterHub or other managed environment, install the PDF export tools into the user environment rather than expecting users to add them one-by-one. A minimal environment for MyST PDF exports should include:
-
-- `mystmd`
-- a working $\LaTeX$ distribution, such as `texlive-core` from `conda-forge`
-- `latexmk`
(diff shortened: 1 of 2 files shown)


docs/installing.md
```diff
@@ -97,11 +97,22 @@ mamba 1.5.8
...
-$ mamba install -c conda-forge texlive-core latexmk
+$ mamba install -c conda-forge texlive-core latexmk imagemagick
 ```
 
+`imagemagick` is optional, but recommended if your documents include images that need conversion for PDF output.
+
 ::::
 :::::
+
+If you maintain a shared JupyterHub or other managed environment, install the PDF export tools into the user environment rather than expecting users to add them one-by-one. A minimal environment for MyST PDF exports should include:
```
choldgraf: Rather than having a little section just for jupyterhub, I think it'd be cleaner if we added a different page for something like "Using MyST in JupyterHub" and include special information there for this use case there

docs/installing.md
```diff
@@ -97,11 +97,22 @@ mamba 1.5.8
...
+
+- `mystmd`
```
choldgraf: I don't think that we need to re-list all the same packages just below the above list we already provide

rowanc1:
Thanks! If we need to separate out for JupyterHub specifically, we can aim that at a new PR — the updated changes looked like solid additions to the docs, so merged them in!

@sgibler-droid thanks for the improvements!!

PR #2881, opened 2026-05-29
Ignore ./worktrees dir in linting/formatting

stefanv:
This is a selfish PR, but I (against all advice!) store my worktrees inside my project directories (otherwise it just gets too crazy in `..`). So, requesting if we can ignore `./worktrees`, but totally understood if that's considered too specific/noisy.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,1 @@
-*.md
 worktrees/
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ yalc.lock
 
 # vim swap files
 *.swp
+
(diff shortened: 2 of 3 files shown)


rowanc1:
Nice.

rowanc1 (approved):

choldgraf (approved):

choldgraf:
I have no idea how you do this without confusing the heck out of your code editor 😅 but +1 from me

stefanv:
My code editor isn't all that fancy :P

I have my local grep configured to also ignore `./worktrees`.

PR #2882, opened 2026-05-29
Add Developer API docs

stefanv:
I generated these docs using Claude Sonnet 4.6, iterating over each section to resolve anything that didn't maker sense to me. I find this to be a useful overview of the myst-specific concepts a developer would need to know to start tinkering with the engine.

The one section that I think is particularly weak is the `ISessionWithCache` — if anyone can enlighten me as to what it is exactly, why it is helpful, and how to use it, that'd be great.

### DIFF
--- a/docs/developer-api.md
+++ b/docs/developer-api.md
@@ -1,0 +1,633 @@
+---
+title: Developer API Reference
+subtitle: How the sausage is made
+short_title: Developer APIs
+description: Reference for MyST's internal APIs — Session, Plugin, AST, transforms, links, config, execution, and logging.
+---
+
+This page documents the TypeScript APIs used when extending or integrating with MyST. Familiarity with TypeScript and [unified](https://unifiedjs.com/) is assumed.
+
+## Build pipeline
+
+Understanding where each API sits in the build makes the rest of this page easier to read.
+
+```{mermaid}
+flowchart TD
+    src["Source files<br>(.md / .ipynb / myst.yml)"]
+    out["Render<br>(site JSON · PDF · JATS · DOCX)"]
+    p1["🔌 DirectiveSpec · RoleSpec<br>TransformSpec stage:'document'"]
+    p2["🔌 TransformSpec stage:'project'<br>LinkTransformer"]
+
+    subgraph per_file ["**Per-file**"]
+        parse["Parse → MDAST<br>(VFile + PreRendererData)"]
+        exec["Execute notebooks<br>(myst-execute · ICache)"]
+        dtrans["Unified transform pipeline<br>→ RendererData"]
+        parse --> exec --> dtrans
+    end
+
+    subgraph project ["**Project**"]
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -3,6 +3,8 @@ title: MyST Markdown Tools
 description: MyST (Markedly Structured Text) is designed to create publication-quality documents written entirely in Markdown.
 ---
 
+Look at [my report](./my_report.pdf).
+
```
fwkoch, line 7: This looks accidental?

```suggestion
```

docs/developer-api.md
```diff
@@ -0,0 +1,633 @@
...
+  plugins: ValidatedMystPlugin | undefined;
+  loadPlugins(plugins: PluginInfo[]): Promise<MystPlugin>;
+  getAllWarnings(ruleId: RuleId): (BuildWarning & { file: string })[];
+  jupyterSessionManager(): Promise<SessionManager | undefined>;
+  dispose(): void;
+  fetch(url: URL | RequestInfo, init?: RequestInit): Promise<Response>;
+};
+```
+
+### Cache layer (`ISessionWithCache`)
```
fwkoch, line 91: @stefanv - regarding your question about this:

`ISessionWithCache` is an extended `Session` object with more fields slapped on, which store a bunch of cached, runtime data. This lets us expose a cleaner `ISession` for consumption outside MyST, then, within MyST, call `castSession` to re-type it as the session with all the cache fields...

In practice, what "cleaner" means is less dependencies: In order to implement `ISessionWithCache`, you need to import many more `myst` packages than if you are just implementing `ISession`.

There's also an argument that `ISession` is closer to the interface we want to expose and `ISessionWithCache` is internal, but I don't think that argument is particularly motivating. Neither of these interfaces have been given a ton of thought regarding external consumption, and `castSession` is available to consumers of `myst-cli` so both interfaces are easily available.

fwkoch (commented):
I haven't read the entire api docs page, but in general, I'm in favor of more human-reviewed docs.

Scanning through this, the most glaring missing piece is anchors for cross-references. We need to be able to easily link into these docs.

choldgraf:
@stefanv how would you like to resolve this PR? Here's my suggestion for things to change:

- I think that this is a little bit wordy still - there are some parts of this that feel very "copy / pasted from the code" and I feel like we could take another pass from a model explicitly with the goal of sharpening / shrinking the changes.
- I also would like to explore if there are places where we could factor out this documentation into different pages so that it's not just one gigantic "developer API" page -> for example, a bunch of the plugin API docs could be in the "javascript-plugins" page, or a subset of that section.
- Addressing @fwkoch 's comment w/ the extra information he gave.

Beyond that, I'd be fine just merging this and iteratively improving the docs from there. What do you think?

stefanv:
I am happy to do some further editing here! I was taking the temperature of the room, so I expected to change it quite a bit. I'll put this on my to-do for next week.

PR #2883, opened 2026-05-29
fix: handle \left|\right| delimiters and \label in Typst math export

### LINKED ISSUE #2876: Typst export: sized `|` delimiter inside `\frac` fails with "Undefined left bracket"
This report was produced using Claude Opus 4.8 after I encountered hard errors in exporting to typst (failed to produce .typ file). It may be valuable due to the specificity of its diagnosis.

### Versions

- mystmd: v1.9.1
- typst: 0.14.2
- OS: macOS (Darwin)

## Description

When a math expression contains a *sized* vertical-bar delimiter — `\left|…\right|`, `\bigl|…\bigr|`, or `\lvert`/`\rvert` (including via macros such as `\abs{…}` or `\norm{…}`) — **inside a `\frac`/`\dfrac` argument**, the tex→Typst conversion aborts with:

```
Error: Undefined left bracket: |a
```

Because this is a hard error, the whole export is aborted and **no `.typ` source (and no PDF) is written**.

Other delimiters (`\left(…\right)`, `\left[…\right]`) in the same position convert fine, and vertical bars convert fine everywhere *except* as a `\frac` argument, so the problem is specific to bar-type delimiters inside a fraction.

### Minimal working example

`myst.yml`:
```yaml
version: 1
project:
  exports:
    - format: typst
      output: out.typ
  toc:
    - file: home.md
```

`home.md`:
```markdown
# Test

$\dfrac{\left|a\right|}{b}$
```

Run:
```
myst build --typst
```
### Actual behavior

```
Error: Undefined left bracket: |a
```
No `out.typ` is produced.

### Scope observed (boundary of the bug)

| Construct | Result |
|---|---|
| `$\left\|a\right\|$` (standalone) | OK |
| `$\left\|\frac{a}{b}\right\|$` (frac *inside* the bars) | OK |
| `$\dfrac{\left(a\right)}{b}$`, `$\dfrac{\left[a\right]}{b}$` | OK |
| `$\dfrac{\left\|a\right\|}{b}$` | **Error: Undefined left bracket** |
| `$\dfrac{\bigl\|a\bigr\|}{b}$` | **Error** |
| `$\dfrac{\lvert a\rvert}{b}$` (macro `\abs`) | **Error: Undefined left bracket: lvert** |
| `$\dfrac{\|a\|}{b}$` (bare bars) | OK |

## Proposed solution

A workaround is to use bare `|…|` (or `\|…\|`) instead of sized bars when they are a fraction argument, but this is not a true fix.

## Additional notes

Probably related to #2269 and #1889.

stefanv:
This PR addresses two primary issues when exporting LaTeX math to Typst:

1. Fixes "Undefined left bracket" error (gh-2876):
   The `tex-to-typst` converter misidentifies delimiters for `\left` and `\right`
   when they are immediately followed by other characters (e.g., `\left|a`).
   The preprocessor now wraps these delimiters in braces (e.g., `\left{|}`)
   to ensure the underlying parser treats the delimiter as a distinct unit.

2. Corrects `\label` conversion in math blocks:
   Raw conversion of `\label{...}` often produces invalid Typst syntax
   like `label(e q : 1)` inside math blocks. The preprocessor now:
   - Strips `\label{...}` from the LaTeX string before conversion.
   - Extracts the label identifier and applies it as a native Typst
     label (e.g., `$ ... $ <label>`) if the node lacks a label.
   - Ensures cross-references remain functional in the exported document.

Closes #2876

### DIFF
--- a/packages/myst-to-typst/src/math.ts
+++ b/packages/myst-to-typst/src/math.ts
@@ -66,11 +66,11 @@ const math: Handler = (node, state) => {
   const tex = node.typst
     ? mathValue
     : mathValue
-        .replace(/\\label\{([^}]+)\}/g, (_, l) => {
-          labelFromTex = l;
+        .replace(/\\label\{([^}]+)\}/g, (_: string, l: string) => {
+          if (!labelFromTex) labelFromTex = l;
           return '';
         })
-        .replace(/\\(left|right)\s*(?!\{)([^\s\\]|\\[a-zA-Z]+|\\.)/g, '\\$1{$2}');
+        .replace(/\\(left|right)\s*(\\\||\|)/g, '\\$1{$2}');
   const { value, macros } = node.typst
     ? { value: tex, macros: undefined } // No conversion needed for typst
     : texToTypst(tex); // Convert LaTeX to Typst
@@ -100,7 +100,7 @@ const inlineMath: Handler = (node, state) => {
     ? mathValue
     : mathValue
         .replace(/\\label\{([^}]+)\}/g, '')
-        .replace(/\\(left|right)\s*(?!\{)([^\s\\]|\\[a-zA-Z]+|\\.)/g, '\\$1{$2}');
+        .replace(/\\(left|right)\s*(\\\||\|)/g, '\\$1{$2}');
   const { value, macros } = node.typst
     ? { value: tex, macros: undefined } // No conversion needed for typst
     : texToTypst(tex); // Convert LaTeX to Typst
--- a/packages/myst-to-typst/src/math.ts
+++ b/packages/myst-to-typst/src/math.ts
@@ -62,14 +62,23 @@ export function resolveRecursiveCommands(plugins: MathPlugins): MathPlugins {
 const math: Handler = (node, state) => {
   // Use typst value if available, otherwise convert LaTeX
   const mathValue = node.typst || node.value;
+  let labelFromTex: string | undefined;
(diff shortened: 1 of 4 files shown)


packages/myst-to-typst/src/math.ts
```diff
@@ -62,14 +62,23 @@ export function resolveRecursiveCommands(plugins: MathPlugins): MathPlugins {
...
   // Use typst value if available, otherwise convert LaTeX
   const mathValue = node.typst || node.value;
+  let labelFromTex: string | undefined;
+  const tex = node.typst
+    ? mathValue
+    : mathValue
+        .replace(/\\label\{([^}]+)\}/g, (_: string, l: string) => {
+          if (!labelFromTex) labelFromTex = l;
+          return '';
+        })
```
rowanc1: I think we pull this out already, so surprised it is getting in here!

Did you come across this directly?

stefanv: Ran into this while setting up a test, but probably fed in a non-realistic input. I'll take it out.

rowanc1:
@stefanv fixed this upstream here:

https://github.com/continuous-foundation/tex-to-typst/pull/35


This is released as `v0.0.20`!

stefanv:
OK will pin to that.

stefanv:
See https://github.com/jupyter-book/mystmd/pull/2889

PR #2884, opened 2026-05-29
Accept myst.yaml as an alternative to myst.yml

stefanv:


packages/myst-cli/src/build/site/manifest.ts
```diff
@@ -149,15 +149,15 @@ export async function localToManifestProject(
...
+    projConfigFile ?? path.join(projectPath, 'myst.yml'),
     projFrontmatter,
     session.publicPath(),
     { altOutputFolder: '/', webp: true },
   );
   const thumbnail = await transformThumbnail(
     session,
     null,
-    path.join(projectPath, 'myst.yml'),
+    projConfigFile ?? path.join(projectPath, 'myst.yml'),
```
choldgraf, line 160: aren't these hard-coding myst.yml even though this PR is trying to make myst.yaml be valid too? Don't we need to add logic here to support myst.yaml cases?

choldgraf (commented):
LGTM in general though I have on quick comment!

Also thanks for adding a nice little test :-)

packages/myst-cli/src/build/site/manifest.ts
stefanv: `projConfigFile` comes from `selectLocalConfigFile`, which grabs it from the Redux store.

That gets populated in `loadConfig`, which uses `configFromPath` to determine the file location. That function, in turn, looks at `session.configFiles` (set upon Session initialization to the list we update here).

Seems like a long way for that information to travel, but I didn't want to try and simplify the data flow in this PR.

stefanv: (Oh, and the way `configFromPath` determines which one to use:

```ts
  const configs = session.configFiles
    .map((file) => {
      return resolve(path, file);
    })
    .filter((file) => {
      return fs.existsSync(file);
    });
```
)

I.e.: first to exist.

choldgraf: ohh I see so this just assumes its called `myst.yml` if `projectConfigFile` doesn't already exist, which it almost certainly will 99% of the time. If that's right then I don't think it's a big deal

yuvipanda:
Death to MS DOS conventions! 8.3 down down!

rowanc1:
Not sure why changesets bot didn't run here. Will add one.

PR #2887, opened 2026-06-01
Document multi-project configuration

### LINKED ISSUE #2886: Document multi-project configuration
In our [`deployment` page here](https://mystmd.org/guide/deployment#creating-static-html) we note that it's possible to have multiple projects defined in a `myst.yml` file. AFAICT this isn't documented anywhere else. It looks like this is indeed a "real" feature, so this is an issue tracking our need to document why it exists, what it's for, and how to use it.

choldgraf:
This adds a bit of documentation and a simple test for multi-project configuration.

---

- closes #2886

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -154,6 +154,37 @@ The behavior of each frontmatter field is hard-coded within MyST. These are the 
 
 +++
 
+(site-multiple-projects)=
+
+## Sites with multiple projects
+
+A MyST site is built from a single `site:` configuration combined with one or more projects.
+By default your site has a single project at its root, so most sites never need to think about this.
+
+To combine several projects into one site, list them under `site.projects` in your top-level `myst.yml`.
+Each entry has a `slug`, which becomes that project's URL prefix, and a `path` to the folder that contains it:
+
+```{code-block} yaml
+:filename: myst.yml
+version: 1
+site:
+  title: My research group
+  projects:
+    - slug: lab-notes
+      path: lab-notes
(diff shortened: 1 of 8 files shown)


rowanc1:
This functionality has been deprecated due to complexity. The approach of managing this in a loop outside was chosen instead (e.g. the multi-blog post approach in pythia/jupyter-book) and to focus the myst tooling on getting good at a single site.

choldgraf:
OK - then I will open a PR to make it clear that this is deprecated.

stefanv:
I think there are still remnants of this across the code base / in some docs, so we need to clean up a bit.

choldgraf:
yes - if it is technically possible, we should at least add a deprecation notice in the docs, code, and ideally a warning. I'm going to start with teh first two and see how hard it is to do the third

rowanc1:
100% there are some things to clean up. The config loading is pretty complex because of this at the moment (using redux, really hard to know where the config content is coming from.

From a user stand point, I think a good signal would be to collapse the myst.yml to be flat (i.e. the keys on the myst.yml are things on the project as it is now 1:1). That work is in https://github.com/jupyter-book/mystmd/issues/1308.

choldgraf:
right now i'm just going to focus on documenting stuff - agree that #1308 would be useful!

stefanv:
1308 looks like a very sensible enhancement; I could work on that?

choldgraf:
Go for it! I'm just going to document etc not change any implementation

PR #2889, opened 2026-06-01
Pin new tex-to-typst

### LINKED ISSUE #2876: Typst export: sized `|` delimiter inside `\frac` fails with "Undefined left bracket"

stefanv:
Fix \left| and \right| delimeters (rely on upstream fix in tex-to-typst)

Closes #2876

### DIFF
--- a/packages/myst-to-typst/package.json
+++ b/packages/myst-to-typst/package.json
@@ -40,7 +40,7 @@
     "myst-common": "^1.9.4",
     "myst-frontmatter": "^1.9.4",
     "myst-spec-ext": "^1.9.4",
-    "tex-to-typst": "^0.0.18",
+    "tex-to-typst": "^0.0.20",
     "unist-util-select": "^4.0.3",
     "vfile-reporter": "^7.0.4"
   }
--- a/.changeset/chubby-nails-look.md
+++ b/.changeset/chubby-nails-look.md
@@ -1,0 +1,7 @@
+---
+'myst-to-typst': patch
+'myst-cli': patch
+'mystmd': patch
+---
+
+Fix \left| and \right| delimeters (rely on upstream fix in tex-to-typst)


stefanv:
No idea where this test failure comes from.

packages/myst-to-typst/package.json
```diff
@@ -40,7 +40,7 @@
     "myst-common": "^1.9.4",
     "myst-frontmatter": "^1.9.4",
     "myst-spec-ext": "^1.9.4",
-    "tex-to-typst": "^0.0.18",
+    "tex-to-typst": "^0.0.20",
```
choldgraf, line 43: Could we mark the ones that are owned by the CSF here in a comment, so that in the future we know to try updating things upstream there instead of in the jupyter-book repos? You can either do that here or in a follow-up PR if we think it's a good idea

rowanc1:
Test failure was unrelated - timeout on jupyter execute.

packages/myst-to-typst/package.json
stefanv: I think it's a good idea! I tried to make the fix in the wrong place because I didn't realize that dependency was in our broader ecosystem. Not sure where this would catch a person's eye, though? I don't typically read dependency files for that information.

PR #2890, opened 2026-06-02
🛠️ Increase test timeout

rowanc1:
Jupyter takes a while, doubling the test timeout as I am seeing this fail in a bunch of places recently.

### DIFF
--- a/packages/mystmd/tests/endToEnd.spec.ts
+++ b/packages/mystmd/tests/endToEnd.spec.ts
@@ -39,7 +39,7 @@ function cleanHashes(text: string) {
 }
 
 const only = '';
-const TIMEOUT = 35000; // Long-ish to allow for the execution tests
+const TIMEOUT = 65000; // Long-ish to allow for the execution tests
 describe.concurrent('End-to-end cli export tests', { timeout: TIMEOUT }, () => {
   const cases = loadCases('exports.yml');
   test.each(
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
I think this is safe to merge - I'm not convinced it'll solve our problems, as there seems to be some deeper kernel startup bug going on where it hangs indefinitely rather than just taking a long time but if this timeout bump fixes a % of the flakiness that would be great.

We can follow up w/ this PR which has a deeper fix for the issues (though it's really hard to know if this is actually the core problem). That one adds a timeout for _kernel startup_ so that if kernel startup is hanging, we fail and try again quickly.

- https://github.com/jupyter-book/mystmd/pull/2874

rowanc1:
Ok - lets merge this one fast. At the least it is harmless, and at most it will shine a light on #2874 (or similar) that something is going wrong with Jupyter startup... and we need to fix that!

packages/mystmd/tests/endToEnd.spec.ts
```diff
@@ -39,7 +39,7 @@ function cleanHashes(text: string) {
 }
 
 const only = '';
-const TIMEOUT = 35000; // Long-ish to allow for the execution tests
+const TIMEOUT = 65000; // Long-ish to allow for the execution tests
 describe.concurrent('End-to-end cli export tests', { timeout: TIMEOUT }, () => {
```
rowanc1, line 43: @choldgraf this `concurrent` might be the cause jupyter clobbering ports all at once, we run quite a few tests at the same time, and Jupyter will compete for the same port. Then there is a failure somewhere and we never connect, then timeout.

So this actually might not be a good way forward...

rowanc1:
Going to close this in favour of #2874

PR #2891, opened 2026-06-02
🐛 Citation `@mention` followed by html parsing bug

rowanc1:
See jupyterlab/jupyter-chat#440

### DIFF
--- a/packages/markdown-it-myst/src/citations.ts
+++ b/packages/markdown-it-myst/src/citations.ts
@@ -31,9 +31,9 @@ type InlineState = {
 
 export const citationsPlugin: PluginWithOptions = (md) => {
   const regexes = {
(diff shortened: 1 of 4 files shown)


packages/markdown-it-myst/src/citations.ts
```diff
@@ -31,9 +31,9 @@ type InlineState = {
 
 export const citationsPlugin: PluginWithOptions = (md) => {
   const regexes = {
-    citation: /^([^^-]|[^^].+?)?(-)?@([\w][\w:.#$%&\-+?<>~/]*)(.+)?$/,
+    citation: /^([^^-]|[^^].+?)?(-)?@([\w][\w:.#$%&\-+?~/]*)(.+)?$/,
```
rowanc1, line 34: I don't think there is ever a time that we should need `<>` in the label of a citation.

stefanv (approved):

stefanv:
Seems reasonable!

PR #2896, opened 2026-06-03
Document the a11y workflow

choldgraf:
This adds a little bit of docs about our new a11y workflow from @pancakereport!

### DIFF
--- a/docs/accessibility-and-performance.md
+++ b/docs/accessibility-and-performance.md
@@ -50,9 +50,10 @@ In April 2026 we completed a focused round of work to [align with WCAG 2.1 AA](h
 On every push to `main`, `myst-theme` runs an automated accessibility audit in CI.
 It builds the current themes and runs [`berkeley-cdss/myst-a11y`](https://github.com/berkeley-cdss/myst-a11y) (which wraps [axe-core](https://github.com/dequelabs/axe-core)) against the [reference pages](https://myst-theme.netlify.app), checking WCAG 2.0 and 2.1 (A & AA).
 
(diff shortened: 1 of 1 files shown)


docs/accessibility-and-performance.md
```diff
@@ -44,6 +44,16 @@ The MyST themes ship with these accessibility behaviors:
 
 In April 2026 we completed a focused round of work to [align with WCAG 2.1 AA](https://github.com/jupyter-book/mystmd/issues/2802). We know there is always ongoing work to be done to improve accessibility and welcome both guidance and contributions, see [](#a11y:contribute).
 
+(a11y:monitor)=
+### How we monitor accessibility
+
+On every push to `main`, `myst-theme` runs an automated accessibility audit in CI.
+It builds the current themes and runs [`berkeley-cdss/myst-a11y`](https://github.com/berkeley-cdss/myst-a11y) (which wraps [axe-core](https://github.com/dequelabs/axe-core)) against the [reference pages](https://myst-theme.netlify.app), checking WCAG 2.0 and 2.1 (A & AA).
+
+Results are surfaced in a single live tracking issue labeled [`a11y-audit`](https://github.com/jupyter-book/myst-theme/labels/a11y-audit).
```
stefanv: Could you clarify: is there one issue that keeps being updated? If so, can we link to the issue directly? Or is it posted to a new one after each audit, in which case linking to the label makes sense.

stefanv (approved):
Looks good to me; one minor clarification please.

docs/accessibility-and-performance.md
choldgraf: I just tried to clarify a bit - that work?

PR #2900, opened 2026-06-04
Warn on potentially unintended lockfile changes

stefanv:


.github/workflows/lockfile-check.yml
```diff
@@ -0,0 +1,84 @@
+name: Lockfile check
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened]
+
+jobs:
+  lockfile-check:
```
rowanc1, line 9: Maybe if we can have a label on here or something to say "changes-packages" so we can pass if the label is there?

PR #2903, opened 2026-06-05
Revert thebe

### LINKED ISSUE #2901: Cell outputs no longer render unless thebe is loaded
I believe we introduced a bug in #2878 , because thebe is more necessary than we appreciated.

We use `thebe-core` for rendering all jupyter output mimetypes, but that PR only bundles it if `thebe/jupyter: true`. As a result, if that condition isn't met, thebe won't be loaded, and the myst theme won't know how to render cell outputs.

I think we should change the "should we bundle thebe" logic by crawling the AST and seeing if there are any output nodes there. That said, I suspect we'd then lose the ability to render hover-preview or "embed" cell outputs, so maybe we should just load thebe all the time like before, and add a comment in our code to explain *why* it's always loaded

- Demonstrating this in https://github.com/jupyter-book/myst-theme/pull/899

choldgraf:
This reverts thebe's conditional logic and adds a comment for why we always load thebe so this is clearer for our maintainers

- closes #2901

### DIFF
--- a/.changeset/thin-chefs-glow.md
+++ b/.changeset/thin-chefs-glow.md
@@ -1,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Revert thebe
--- a/.changeset/thin-chefs-glow.md
+++ b/.changeset/thin-chefs-glow.md
@@ -2,4 +2,4 @@
 "myst-cli": patch
 ---
 
-Revert thebe
+Revert stopping of thebe-copy for static build assets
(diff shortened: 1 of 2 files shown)


.changeset/thin-chefs-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"myst-cli": patch
+---
+
+Revert thebe
```
rowanc1:
```suggestion
Revert stopping of thebe-copy for static build assets
```

PR #2908, opened 2026-06-09
⚛️ Fuse `myst-spec-ext` into `myst-spec`

### LINKED ISSUE #2583: Deprecate `myst-spec-ext` package
Following #2563, we'll have no strong reason to keep `myst-spec-ext` around.

We can deprecate `myst-spec-ext` by:
1. Merging the types into `myst-spec`
2. Authoring an MEP to reflect the spec update
3. Marking `myst-spec` as deprecated

## Definition of done
- [ ] `myst-spec-ext` is published as deprecated

agoose77:
Closes #2583 by deprecating `myst-spec-ext`

Builds upon #2914 

## Changes
1. Move `myst-spec-ext` types into `myst-spec`
2. Deprecate `myst-spec-ext` exports

## What's Next?

Once we've deduplicated `myst-spec-ext` and `myst-spec`, we can move to an `mdast`-based type system.


packages/myst-spec-ext/src/types.ts
```diff
@@ -1,319 +1,3 @@
...
-  type: 'algorithmLine';
-  indent?: number;
-  enumerator?: string;
-};
-
-export type InlineMath = SpecInlineMath &
-  Target & {
-    /** Typst-specific math content. If not provided, LaTeX content will be converted to Typst. */
-    typst?: string;
-  };
```
rowanc1, line 75: There is some valuable stuff in here. Are you in progress of moving it to MyST spec?

agoose77: Yes, the goal (which should mostly be realised already) is to encode this as the present state of myst-spec. Later, we can introduce a new MEP to officially update the spec (bless its current state).

I can see `InlineMath` in `myst-spec` in this PR with `typst` as required. I did a `diff` against origin/main, and confirmed that the `myst-spec-ext` that I had access to in the original PR (that this PR was cherry picked from) matches origin/main, so I _should_ have ported everything over.

packages/myst-spec-ext/src/types.ts
```diff
@@ -362,3 +46,5 @@ export type MystSearchIndex = {
   version: '1';
   records: SearchRecord[];
 };
+
+export type * from 'myst-spec';
```
rowanc1: Can we import explicitly and deprecate on export from this library?

packages/myst-spec-ext/src/types.ts
rowanc1: Ok, was hidden on my phone this morning!

packages/myst-spec/src/index.ts
```diff
@@ -0,0 +1,775 @@
...
+  Image,
+  ListContent,
+  Literal,
+  Parent,
+  PhrasingContent,
+  Root,
+  StaticPhrasingContent,
+  Table,
+  TopLevelContent as MdastTopLevelContent,
+} from 'mdast';
```
rowanc1: Was this what was here before with the auto generation?

Or is this an addition on changing the myst-spec types to mdast?

If we can split this into a few PRs, that would be more clear what the change is and I think be faster to merge in.

packages/myst-spec/index.ts
```diff
@@ -1,407 +0,0 @@
...
-
-type JsonTestCase = {
-  title: string;
-  mdast: Record<string, any>;
-  myst: string;
-  html?: string;
-};
-
-const directory = join('docs', 'examples');
-const files: string[] = readdirSync(directory).filter((name) => name.endsWith('.yml'));
```
rowanc1: This PR should suggest a new route to documentation I suppose?

packages/myst-spec-ext/src/types.ts
agoose77: We certainly can!

packages/myst-spec/src/index.ts
agoose77: It's a mix — this PR introduces `@types/mdast`-based interfaces. We can split that into two, as per Discord.

packages/myst-spec/index.ts
agoose77: Not this PR, I think, but yes.

packages/myst-spec-ext/src/index.ts
```diff
@@ -1 +1,59 @@
...
+  TabSet,
+  Underline,
+} from 'myst-spec';
+
+/**
+ * @deprecated Use the export from `myst-spec`
+ *
+ * This is a duplicate declaration, due to esbuild/bun problems
+ */
+export enum SourceFileKind {
```
agoose77, line 147: Simply exporting this from `myst-spec` breaks esbuild for some reason.

agoose77:
@rowanc1 what's your goal with deprecating `myst-spec-ext` here? I ask because I don't think it's very easy to deprecate import aliases — I think you have to simply declare them twice, with the `myst-spec-ext` declarations having deprecation annotations.

I don't really want to maintain two sets of definitions. If we can't use deprecation, we could just delete `myst-spec-ext` altogether, but I imagine you want to keep publishing it for a while because downstream tools might be pulling it in with a version constraint that requires us to keep publishing?

If so, can we just not deprecate in code, and instead add this to our release notes. Curvenote already know it's going to be deprecated, which is the main user(?)

rowanc1:
I think you can re-export the type with a deprecated comment?

This gives a few cycles of seeing the code marked as deprecated before a hard switch.

I will try and see if I can do it easily.

rowanc1:
This is a pattern that works:

```
import type { Outputs as SpecOutputs } from 'myst-spec';

/** @deprecated Use `Outputs` from `myst-spec` instead */
export type Outputs = SpecOutputs;
```

agoose77:
> This is a pattern that works:
> 
> ```
> import type { Outputs as SpecOutputs } from 'myst-spec';
> 
> /** @deprecated Use `Outputs` from `myst-spec` instead */
> export type Outputs = SpecOutputs;
> ```

Aha, that was on my next to-try list! Fab!

rowanc1:
I also **really** like the small nature of this PR. It is clearly safe, we can merge it in early, then there is a follow on PR to move myst-spec-ext imports to myst-spec. And we know we can ship both at the same time and be confident of zero downstream impact of the refactor.

choldgraf (approved):
If I understand correctly, this is basically just:

- moving the myst-spec-ext definitions into myst-spec
- re-using the myst-spec definitions in myst-spec-ext so that we can deprecate without deleting it now

That sounds like a nice step forward to me! +1 if my assumptions are correct

I also agree it is very helpful to get these smaller PRs done independently, to make it easier to diff and understand the more complex PR (which I assume will be the mdast stuff)

rowanc1:
I just saw that this isn't pointing to main though. Can we do this as a first step against main and get that over the line @agoose77?

agoose77:
@rowanc1 I'd prefer to merge #2914 first, which is what this PR is based upon. 

That PR is mostly a no-op that removes the JSON Schema → TS step and replaces it with the TS directly.

rowanc1:
Ok - first one merged. Can you update this one to have those last commits?

agoose77:
Good to go? @rowanc1

packages/myst-spec-ext/src/index.ts
```diff
@@ -1 +1,59 @@
-export * from './types.js';
+export type {
```
rowanc1: @agoose77 this is no longer deprecating on export? Did that commit get removed when you force-pushed?

agoose77: Huh, that's weird! Thanks for catching that, it was late when I FPd.

PR #2948, opened 2026-06-10
Convert pyproject.toml dependencies from approx equal to lower bound

### LINKED ISSUE #2082: pip install failed due to "platformdirs~=4.2.2" in mystmd-py
## Description
The latest version of `platformdirs` is 4.3.8 but in `mystmd-py`'s pyproject.toml, the version of `platformdirs` is locked to   "platformdirs~=4.2.2", which requires version <4.3.0.
The result is, any python package using `platformdirs` >=4.3.0 won't be able to `pip install mystmd`

Is there a strong reason for this ~= version lock?

## Proposed solution
change it to "platformdirs>=4.2.2", not sure if it will break anything though.

coretl:
This allows them to coexist more easily with other python deps in a single venv

Fixes #2082

### DIFF
--- a/packages/mystmd-py/pyproject.toml
+++ b/packages/mystmd-py/pyproject.toml
@@ -43,8 +43,8 @@ env-var = "MYSTMD_PACKAGE_VARIANT"
 # Ensure that we only bring in nodeenv for PyPI
 [tool.hatch.build.hooks.selector.variants.pypi]
 dependencies = [
-  "platformdirs~=4.2.2",
-  "nodeenv~=1.9.1"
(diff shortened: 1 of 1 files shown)


coretl:
I'm guessing that I need to add a label to this PR, but I don't seem to have permission to do this?

agoose77:
@coretl typically we want people to install this tool in its own environment. But, as discussed in #2082, we can try loosening them for the benefits that it brings users.

You can add a label in future by naming the branch `feat/XXX` or `fix/XXX`

PR #2950, opened 2026-06-10
Document multi-page transform pattern

choldgraf:
This adds simple documentation for a very useful pattern that I've found pretty hard to discover on your own, so I figured it'd help to have a little note in the docs about it, and a link to a working example

### DIFF
--- a/docs/javascript-plugins.md
+++ b/docs/javascript-plugins.md
@@ -85,6 +85,12 @@ I am **special bold text**, whilst I am **normal bold text**
 
 I am **special bold text**, whilst I am **normal bold text**
 
+(plugins:multi-page)=
+:::{seealso} Aggregate information across multiple pages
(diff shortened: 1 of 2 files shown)


stefanv:
This pattern was brought to light (saying that, because I'm sure it was known to *some* developers already) in a discussion between @agoose77, @harold-campbell, and myself at the weekly working meeting. It became apparent when looking at the diagram in #2882, which I hope we also get merged in some shape or form.

fwkoch:
This PR just adds a small note that indicates available plugin functionality and links to an example.

There's certainly potential to expand this more, as our plugin interfaces mature and harden, e.g. discuss specifically `stage: document` vs `stage: project`, what each of those mean more specifically, etc.

But for now, I think this small note is an improvement!

fwkoch (approved):

Issue #2773, opened 2026-03-18
Allow users to reverse how fence length maps onto nested directives


choldgraf:
Currently you can only nest directives by making the *outer* directives longer. E.g. you start with:

````
:::{dropdown}
Text
:::
````

You realize you want to make the thing inside a note as well, now you have to grow the outer fence and then add the inner fence:

````
::::{dropdown}
:::{note}
Text
:::
::::
````

This becomes a huge pain when you have a lot of nested notes.

### Let users reverse this logic!

This problem would go away if you could allow the following:

````
:::{dropdown}
::::{note}
Text
::::
:::
````

Such that outer fences had to be *lower* in length than inner fences for them to be nested.

**Comment by stefanv** [comment]
I think this is an important usability improvement. When the user starts typing, they don't *know* how many levels deep a directive will go. Requiring the outermost layer to match, with number of colons, the inner nesting, means you have to jump back to the outer directive repeatedly. No fun.

Maybe we add an option like `reversed-fences: true`?

**Comment by choldgraf** [comment]
This is maybe a crazy idea but is there some way we could accomplish this without a feature flag? I dunno how the regex/parsing works for these but I wonder if we could define the logic such that you nest code fences as long as they aren't the same length as one that's already open, rather than strictly larger or smaller. Does that make sense? So both of these would work:

`````
::::{note} one
:::{note} two
:::
::::

:::{note} one
::::{note} two
::::
:::

or maybe even

::::{note} one
:::{note} two
:::::{note} three
:::::
:::
:::::
`````

is that crazy? I think it might break strict commonmark rules...

**Comment by stefanv** [comment]
Turns out this is not part of the Commonmark standard. Also, some implementations don't care about the number of colons. In some implementations they don't even have to match.

In pandoc, e.g., you can use any number, and then `:::` (or more) always closes the nearest block.

See https://pandoc.org/demo/example33/8.18-divs-and-spans.html

**Comment by agoose77** [comment]
Colon fences in MyST are not CommonMark. We might have pressure to keep their UX consistent with fenced code blocks (and fenced admonitions), but that's a different matter!

Additionally, we don't call them div-fences either — Pandoc, djot, etc. treat 
```
:::
A div
:::
```
[...]

PR #2953, opened 2026-06-11
📦 Vendor `markdown-it-myst-extras` into `markdown-it-myst`

agoose77:
We spoke a while ago about doing this in the context of colon-fence handling. Presently, the `markdown-it-myst-extras` package lives under the `executablebooks` repo. There are still other plugins that live there, but I think for now that's OK — we might want to build them in our monorepo going forward.

I decided to deprecate the default export — it feels unusual to only enable half of the plugins. Feel free to push back there.

c.f. https://github.com/jupyter-book/mystmd/issues/2773


packages/myst-parser/src/plugins.ts
```diff
@@ -5,8 +5,14 @@ export { default as frontMatterPlugin } from 'markdown-it-front-matter';
...
 export { default as deflistPlugin } from 'markdown-it-deflist';
-export { mystPlugin, citationsPlugin } from 'markdown-it-myst';
-export { mystBlockPlugin, colonFencePlugin } from 'markdown-it-myst-extras';
+export {
+  rolePlugin,
+  directivePlugin,
+  citationsPlugin,
+  blockPlugin,
+  colonFencePlugin,
+  mystPlugin,
```
agoose77, line 14: This `mystPlugin` is deprecated, but we'll leave the export for a bit longer.

rowanc1:
package.json drops the import?

packages/myst-parser/src/plugins.ts
rowanc1, line 14:
```suggestion
```

I don't think we need to re-export this — deprecated!

rowanc1: Cool. Sorry, didn't see it on my diff. This sounds good.

rowanc1 (approved):
This looks good to go from code review. @agoose77 if you are confident this is faithfully pulling those tools in, lets merge!

PR #2957, opened 2026-06-15
Use availableParallelism rather than CPUs

agoose77:
I had hoped that this would give physical cores, rather than overcommitting CPUs. On my system, however, it's still reporting threads.

Nevertheless, it's good to use a more appropriate API.

### DIFF
--- a/.changeset/vast-brooms-clap.md
+++ b/.changeset/vast-brooms-clap.md
@@ -1,0 +1,5 @@
+---
+'mystmd': patch
+---
+
+Fix parallelism count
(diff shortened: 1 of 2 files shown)


.changeset/vast-brooms-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'mystmd': patch
+---
+
+Fix parallelism count
```
agoose77:
```suggestion
Use NodeJS parallelism parallelism feature
```

Issue #2593, opened 2025-12-16
Link to mybinder.org trigger DOI


rgaiacs:
This is related to https://github.com/jupyter-book/mystmd/issues/2209

## Description

https://github.com/jupyterhub/team-compass/blob/cbed9d2ab7cd65e939e318083ae12d4375964a17/docs/meetings/community/2019-11-21.md includes

```
The Jupyter Notebook we demoed: https://mybinder.org/v2/zenodo/10.7910/DVN/EOYZKH/
```

The link to mybinder.org is causing the error

```
⛔️ meetings/community/2019-11-21.md:100 Could not find DOI "https://mybinder.org/v2/zenodo/10.7910/DVN/EOYZKH/" from doi.org as doi:10.7910/DVN
   Please check the DOI and, if correct, add citation info to local BibTeX file
```

## Proposed solution

Add mybinder.org to the list of websites that might look like a DOI but are not.

fwkoch:
This is resolved by https://github.com/jupyter-book/mystmd/pull/2960

Issue #2230, opened 2025-08-07
Complex DOIs not resolving as expected


bl-freeman:
## Description
Apologies if I missed an existing open issue that covers this. In the [citations and bibliography](https://mystmd.org/guide/citations) guide, a note is made for dealing with complex DOIs. For both suggestions—shortdoi and encoded url—the rendered link is invalid despite the workarounds. I see this mentioned for shortdoi in [#1074](https://github.com/jupyter-book/mystmd/issues/1074#issue-2231476318) but the issue was marked resolved by [#1075](https://github.com/jupyter-book/mystmd/pull/1075). Clicking on the hyperlink returns a "DOI not found," as does copying the full url and pasting into the address bar. Copying _only_ the DOI will at least return the proper reference via a web search.

## Proposed solution
Having the adjusted link (either shortdoi or encoded url) appear instead of the complex DOI would avoid having an unusable hyperlink and allows users to simply click the link as they would any other rendered non-complex DOI.
## Additional notes
This is the resolved citation using the shortdoi workaround as seen in the guide. 
<img width="874" height="484" alt="Image" src="https://github.com/user-attachments/assets/85095d9f-fd7f-4718-b7d0-f4da68ba8157" />

giswqs:
I just noticed the same issue. Many of the DOIs of my papers are not resolving as expected. 
https://gishub.org/research/

PR #2960, opened 2026-06-16
🔗 Relax overaggressive DOI coercion

fwkoch:
This PR addresses the issue where arbitrary links that end in a DOI are coerced to citations linking back to doi.org. Issues bringing this up:
- https://github.com/jupyter-book/mystmd/issues/1629
- https://github.com/jupyter-book/mystmd/issues/2209
- https://github.com/jupyter-book/mystmd/issues/2593

My solution was simply adding a regex check for `doi.org` before passing off DOI validation to `doi-utils`. If it's a URL without `doi.org`, do not coerce it to a citation. (We can't use the `doi-utils` "strict" option, because that is too aggressive - disabling short dois and dois with non-html-safe characters.)

This PR _changes the default behavior_ of these URLs. I think this is fair - unwanted, unpreventable coercion of a URL feels very bad. If someone wants to re-enable the legacy behavior, though, there is an option in `myst.yml`:

```
project:
  options:
    infer_dois_from_urls: true
```

Note - the scope of this PR does not include manual disabling of individual URLs/DOIs (https://github.com/jupyter-book/mystmd/issues/2406), disabling citations per page (https://github.com/jupyter-book/mystmd/issues/1629#issuecomment-2569363875), edge case behavior for non-DOI doi.org URLs (https://github.com/jupyter-book/mystmd/issues/828), or rendering issues of complex DOIs (https://github.com/jupyter-book/mystmd/issues/2230).


bsipocz:
> If it's a URL without doi.org, do not coerce it to a citation.

But could it still be a citation and not just a random URL mention? 

(E.g. the astronomical journal reference in one of the issues above is definitely used in a way as a \cite or reference so we should definitely pick it up and include in the references section at the bottom of the rendered page)

fwkoch:
Hi @bsipocz - Thanks for the comment!

Trying to make sure I understand correctly: are you suggesting something like `[Figure 10](https://iopscience.iop.org/article/10.3847/1538-3881/ace32f#ajace32ff10)` both (1) keeps the link to the figure and (2) adds a citation via DOI since part of the URL looks like a DOI?

My feeling is that any "arbitrary url -> inferred DOI" is too implicit / magic. In the example above, the DOI does not resolve, so we can't automatically do anything interesting with the DOI, it will just look broken. Even if it did resolve, it's impossible to know the author's intent: Does this URL just coincidentally have a DOI-looking part? Or do they want to cite it?

I think it's reasonable to require `[Figure 10](...) [@doi:10.3847/...]`.

---

Or - Am I misunderstanding and you are suggesting something like: `` {cite}`https://iopscience.iop.org/article/10.3847/1538-3881/ace32f` `` should add a citation for that page (independent of any DOI resolution), without requiring a bibtex entry? I think that one is totally reasonable - authors are explicitly asking for a citation. We shouldn't try to do anything magic, but we should build a reference entry from the URL as best as possible.

If this is your suggestion, I agree it could be interesting, but it feels like a different feature, e.g. "Pull citation info from URL page metadata" or something.

packages/myst-transforms/src/links/doi.ts
```diff
@@ -7,16 +7,44 @@ import { updateLinkTextIfEmpty } from './utils.js';
 
 const TRANSFORM_SOURCE = 'LinkTransform:DOITransformer';
 
+/** Matches doi.org, dx.doi.org, and www.doi.org — same hosts as doi-utils doiOrg resolver. */
+const DOI_ORG_HOSTNAME = /(?:dx\.)?(?:www\.)?doi\.org/i;
```
choldgraf: Probably not a huge deal, but I think this would also match things like `notadoi.org` - can we tighten up the regex a little to make sure we select what we want?

choldgraf (approved):
This looks pretty good to me - thank you for the tests! I had one comment about the regex specificity. But in general LGTM when you are comfortable merging.

Can we think of any other case where we'd _want the default_ URL parsing to infer a DOI? Any other doi-like service that's similar to doi.org? If there's not an obvious one then I think it's safe to just leave this as-is but thought I should ask haha

bsipocz:
> Or - Am I misunderstanding and you are suggesting something like: {cite}`https://iopscience.iop.org/article/10.3847/1538-3881/ace32f` should add a citation for that page (independent of any DOI resolution), without requiring a bibtex entry? I think that one is totally reasonable - authors are explicitly asking for a citation. We shouldn't try to do anything magic, but we should build a reference entry from the URL as best as possible.

Yes, I was thinking this. I definitely not suggesting any magic to resolve to a DOI, but also don't really want to maintain a separate bibtex file. 
If you say that we need to link to a valid DOI to make the citation appear in the references, I think that could be OK, but it will need to be documented, but I also believe that we could have citations to locations that doesn't have DOIs yet we want them to be listed under references (I'm mostly thinking of software citations, etc.)

choldgraf:
Are there certain URL prefixes that we know will result in a resolve able DOI? Eg if nature has a standard prefix etc. then we could add them all to the regex and have that as something we can add to over time

packages/myst-transforms/src/links/doi.ts
fwkoch: Good catch! I just ditched the regex entirely for now, since really we are only looking for those 3 hostnames.

(I pulled this out of `doi-utils` - https://github.com/continuous-foundation/doi-utils/blob/main/src/resolvers.ts#L10 - should fix it up there too... 😅)

fwkoch:
@bsipocz - Ok, amazing. Yeah, I agree maintaining separate bibtex files is pretty annoying, and I think we could do interesting things with the `cite` role: Re-introduce doi inferring, e.g. if you cite a publisher url. Cite a webpage directly, pulling metadata from the HTML. Even define specific cite fields in-line as arguments to the role, almost like putting a mini bibtex entry in-line...

@choldgraf - Thanks for the review! I don't _think_ there are other URLs where we would want the default `link` parsing to be straight to DOI citation, especially in the way we are currently doing it, where avoiding that behavior is impossible. There are some other indexing services, like PubMed, arXiv, Open Alex... but DOIs are so ubiquitous that almost all those articles will also have a DOI. If you want to use the shortcut link syntax to get a citation, you have to use the DOI url. If you want a citation from a PubMed url (or publisher url or something), you may be able to use the `{cite}` syntax, then we can support all sorts of cool things in there.

Bottom line for this PR though, even without touching the `cite` role, we are making something previously impossible possible: A user who wants to _link_ to a URL that looks like it has a DOI in it previously couldn't, and now they can. A user who wanted to _cite_ the same URL previously could, and now they still can, they just need to use the DOI directly.

Anyway - I'm going to merge and close the 3 issues linked at the top. Then I'll open another issue to capture this discussion!

PR #2970, opened 2026-06-19
Switch to `.ipynb` build cache

### LINKED ISSUE #2963: Use `.ipynb` for the cache representation of execution
### Proposal

Right now, we use a bespoke representation for execution caches. Given that it's arbitrary, I'd like to change this to generate an `.ipynb` file. This would be nice, because then we can integrate with tools like `jupytext` and other output-aware programs that can read outputs (even if they can't display the `eval` results).

We would only generate code cells, and the cells would be empty. This keeps the cache to a reasonable size.

### Additional notes

_No response_

agoose77:
At #myst-education-2026, when watching @mwouts talk about "hydrating" text notebooks with outputs from `.ipynb` files, it occurred to me that this might be an opportunity to rethink our execution cache.

MyST's cache is stupid and simple, because that's all we needed. However, the format we selected was bespoke for convenience, rather than because it's a good format. I think we could change this so that people like @bsipocz have an easier time investigating the build cache.

Closes #2963 

## Content
1. Defines a cache "document" that stores outputs and metadata. This does not store _inputs_, because I don't think we need to waste space on that. 
2. Implements support for integrating this new cache document into existing legacy caches.
3. Implements a new `ipynb` cache abstraction that can write `.ipynb` files instead of JSON / our legacy format.
4. Implements a tiered cache abstraction so that this new format can be used alongside existing caches.

The `.ipynb` cache is stupid. It writes `IOutput[]` arrays to code-cell outputs, and the same for expressions. This means that you _don't_ need JupyterLab MyST to view the expression outputs — the goal is interop with notebook viewing tools rather than as a core unit of sharing. 

## Next Steps
1. Modify `myst-spec` (and add migrations) to rename `jupyter_data` and `result` to `jupyterData` in both cases. This was an ugly oversight we should have made in the last migration.


packages/myst-cli/src/process/mdast.ts
```diff
@@ -191,7 +196,10 @@ export async function transformMdast(
       session.log.debug(`▶️  Executing: ${fileName}`);
       await kernelExecutionTransform(mdast, vfile, {
         basePath: session.sourcePath(),
-        cache: new LocalDiskCache<(IExpressionResult | IOutput[])[]>(cachePath),
+        cache: new TieredExecutionCache(
+          new NotebookExecutionCache(new LocalDiskCache(cachePath, '.ipynb')),
+          new LegacyExecutionCache(cachePath),
+        ),
```
agoose77, line 202: An `.ipynb`-first cache that also supports the legacy cache format.

packages/myst-cli/src/process/notebook.ts
```diff
@@ -92,9 +92,7 @@ function embedInlineExpressions(
 ) {
   const inlineNodes = selectAll('inlineExpression', block) as InlineExpression[];
   inlineNodes.forEach((inlineExpression) => {
-    const data = findExpression(userExpressions, inlineExpression.value);
-    if (!data) return;
-    inlineExpression.result = data.result as unknown as Record<string, unknown>;
+    inlineExpression.result = findExpression(userExpressions, inlineExpression.value)?.result;
```
agoose77, line 95: Refactoring.

packages/myst-cli/src/transforms/inlineExpressions.ts
```diff
@@ -79,9 +69,6 @@ function renderExpression(node: InlineExpression, file: VFile): StaticPhrasingCo
 export function transformRenderInlineExpressions(mdast: GenericParent, file: VFile) {
   const inlineNodes = selectAll('inlineExpression', mdast) as InlineExpression[];
   inlineNodes.forEach((inlineExpression) => {
-    if (!inlineExpression.result) {
```
agoose77, line 82: Refactoring.

packages/myst-common/src/types.ts
```diff
@@ -173,31 +171,6 @@ export enum AdmonitionKind {
...
-  /**
-   * Exception value
-   */
-  evalue: string;
-  /**
-   * Traceback
-   */
-  traceback: string[];
-}
-
```
agoose77, line 198: Move this to our spec (it's part of the `inlineExpression` node type)

packages/myst-execute/src/execute.ts
```diff
@@ -40,7 +40,10 @@ export async function computeExecutableNodes(
       const code = select('code', matchedNode) as Code;
       const { status, outputs } = await executeCodeCell(kernel, code.value);
       // Cache result
-      results.push(outputs);
+      results.push({
+        type: 'code',
+        responses: outputs,
+      });
```
agoose77, line 46: Use the new (thin) `ExecutionResult` abstraction rather than arbitrary array-of-output or expression-display types.

rowanc1:
Another thing that would be awesome to surface in the cache is execution date and compute time. Not sure if that fits here.

packages/myst-execute/src/kernel.ts
```diff
@@ -153,6 +155,8 @@ export async function evaluateInlineExpression(kernel: Kernel.IKernelConnection,
     user_expressions: {
       expr: expr,
     },
+    allow_stdin: false,
+    stop_on_error: false,
```
agoose77, line 159: These are changes that mean execution won't ask (and get stuck) for STDIN. We also don't want to stop for queued execution if there's an error (although, that's not supported here anyway).

packages/myst-execute/src/transform.ts
```diff
@@ -160,7 +164,14 @@ export async function kernelExecutionTransform(tree: GenericParent, vfile: VFile
     });
     // Populate cache if things were successful
     if (!errorOccurred) {
-      opts.cache.set(cacheKey, results);
+      opts.cache.set(cacheKey, {
+        context: {
+          kernelspec,
+          path: path.basename(opts.basePath, vfile.path),
+          timestamp: new Date().toISOString(),
+        },
```
agoose77, line 177: Our WIP cache context. This is untyped because we may continue to add things like GIT rev, etc.

packages/myst-spec/src/ext.ts
```diff
@@ -314,3 +339,15 @@ export type AnyWidget = {
...
+ * Type narrowed Block to contain code-cells
+ */
+
+export type CodeBlock = Block & {
+  kind: 'code';
+  data?: {
+    tags?: string[];
+  };
+  children: Code[];
+};
```
agoose77, line 353: Although `Block` is already in our spec, we do now rely on `CodeBlock` internally for representing

agoose77:
> Another thing that would be awesome to surface in the cache is execution date and compute time. Not sure if that fits here.

Yes, I agree. I've already put the execution date in. I was originally going to put computation metrics in later — @bsipocz asked for better computation info (even down to the cell level). However, in its current shape, document timing is probably trivial here.

agoose77:
@rowanc1 @stefanv the CI fails because `bun run test` doesn't work, whilst `bunx vitest` or `npx vitest` does.

Since we switched to Bun, it's been feeling a little "fragile" at times. Things seem to break or need fixing in modules unrelated to direct code changes. I wonder if this is an us problem, or some fragility in Bun.

rowanc1:
Latest failure is network related to doi.org. Seems unrelated to bun.

agoose77:
@rowanc1 that's odd ... it fails locally for me due to vitest. 

Actually, this reminds me, I think the main `bun` issue is that it's highly sensitive to the bun version. We might need to pick a particular one, or figure out the compat range.

agoose77:
Merging this, as someone else has taken a look!

choldgraf:
Did compute time get in? If not I'll create an issue to track that. In JB1 we had a directive for showing a compute time summary of all notebooks that quantecon found useful so I just wanna track that!

Also to be clear: even if not implemented, thanks for moving forward on this! I am happy to see quick merges and fewer dangling PRs rather than trying to get things perfect 🙂

agoose77:
Yes — there's a timer stored in the metadata: https://github.com/jupyter-book/mystmd/pull/2970/changes#diff-44df3d7ff82ef8d793958328f60376770d7b7cdd918b18116591602a68c5be0dR176

Issue #2962, opened 2026-06-16
Implement support for composing MyST applications from components


agoose77:
## Motivation
We want to expose more of MyST's internals to extension and plugin authors. This requires agreement on APIs, and a sense of what to expose. The present design of MyST does not readily admit natural APIs for access to things like parse state, exporters, etc. In addition, we would like to later be able to swap out implementations of features like exporters without recompiling the application. This requires strong isolation between functionality, and conformance to an interface that supports runtime providing of functionality.

## User Stories
1. "I am a core developer. I want to refactor MyST internals to make them more modular and extensible" — Clear mechanism for wiring up internal functionality — no more "prop drilling" of classes, just top-level DI.
2. "I am a core developer. I want to expose more of MyST to plugin authors. I don't know what APIs to expose, or how to build and consume them internally" — Natural way to isolate functionality — helps us to figure out an API by isolating functionality into smaller parts, gradually, over time.

## Proposal
1. Do a tiny amount of work to load "extension" via Lumino's plugin registry.
2. Port one existing internal feature to this, e.g. the `myst-execute` executor.
3. Iteratively (over time) expose more features like this.
4. At some point, support Module Federation to allow third-party extensions to interact with core app.

> [!Important]
> We'll need a better name for this _kind_ of plugin, e.g. "component", "extension", or "core plugin"

### Some terms

- [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection)
- [Lumino](https://github.com/jupyterlab/lumino) — The JupyterLab underlying system app framework
- [Module Federation](https://module-federation.io/guide/start/)

### How JupyterLab Does DI

JupyterLab uses the Lumino plugin registry to implement a DI framework. One extension (e.g. the file browser) asks for the implementation of a token (e.g. the contents manager `Contents.IDrive`), another _provides_ that implementation (e.g. [the JupyterLab Drive](https://github.com/jupyterlab/jupyterlab/blob/34ca6a0dddc3e73d9b4f8f8c8f006d0073aa9e8c/packages/services-extension/src/index.ts#L155-L170) or [the JupyterLite Drive](https://github.com/jupyterlite/jupyterlite/blob/a0a668efc8d41d5d2235fcd9d1e61deeddd9b82d/packages/services-extension/src/index.ts#L85-L110)). Clearly from this example, this pattern makes it easy to build different _distributions_ of JupyterLab that support different features.

<details><summary>Details</summary>
 
[...]

PR #2972, opened 2026-06-19
[Spike] Demonstrate using Lumino plugin registry

agoose77:
This PR uses the Lumino plugin registry to:

1. Define the Jupyter Session Manager factory as several possible plugins
2. Consume this plugin(s) in the main session

At this stage, this will feel like boilerplate because _it is_. Once you start to have more plugins, it quickly becomes clear what the wins are

Demonstration of #2962


packages/myst-cli/src/session/session.ts
```diff
@@ -127,6 +140,32 @@ export class Session implements ISession {
...
+    }
+    // Now register another plugin to ask for this implementation
+    this.$registry.registerPlugin({
+      id: 'myst-cli:consume-session-manager-factory',
+      autoStart: true,
+      requires: [ISessionManagerFactory],
+      activate: (app: ISession, factory: ISessionManagerFactory): void => {
+        (app as any).setSessionManagerFactory(factory);
+      },
+    } satisfies IPlugin<ISession, void>);
```
agoose77, line 159: This is not normally required, because you should architect the main app to not itself need the plugins. However, this is a low-effort retrofit for demo purposes, and it's fine.

packages/myst-execute/src/manager.ts
```diff
@@ -145,3 +148,104 @@ export async function launchJupyterServer(
...
+ */
+export const existingSessionManagerFactoryPlugin: IPlugin<ISession, ISessionManagerFactory> = {
+  id: 'myst-execute:existing-session-manager-factory',
+  autoStart: true,
+  provides: ISessionManagerFactory,
+  activate: (session: ISession): ExistingSessionManagerFactory => {
+    console.log('Activated existing session manager provider');
+    return new ExistingSessionManagerFactory(session);
+  },
+};
```
agoose77, line 242: Define a plugin that provides a concrete implementation for the `ISessionManagerFactory` token

packages/myst-execute/src/manager.ts
```diff
@@ -145,3 +148,104 @@ export async function launchJupyterServer(
...
+ */
+export const newSessionManagerFactoryPlugin: IPlugin<ISession, ISessionManagerFactory> = {
+  id: 'myst-execute:new-session-manager-factory',
+  autoStart: true,
+  provides: ISessionManagerFactory,
+  activate: (session: ISession): NewSessionManagerFactory => {
+    console.log('Activated new session manager provider');
+    return new NewSessionManagerFactory(session);
+  },
+};
```
agoose77, line 254: Define a competing implementation for the same token.

packages/myst-execute/src/types.ts
```diff
@@ -1,6 +1,25 @@
...
+export interface ISessionManagerFactory {
+  getSessionManager(): Promise<SessionManager | undefined>;
+}
+
+/**
+ * Token object for requesting / providing an implementation
+ */
+export const ISessionManagerFactory = new Token<ISessionManagerFactory>(
+  'myst-execute:ISessionManagerFactory',
+);
```
agoose77, line 21: Define a token that, when resolved, will implement some interface.

packages/myst-cli/src/session/session.ts
agoose77, line 159: The `registerPlugins` part is usually performed after loading them via module federation, which lets package managers add new plugins to an existing installation of MyST.

Given that we're retrofitting plugins/extensions, though, it's worth wiring up the plugins by hand to begin with (so that we don't also need to add module federation support).

agoose77:
Inspired by https://github.com/jupyterlab/lumino/pull/827

PR #2975, opened 2026-06-23
Give notebook cells stable deep-link anchors

### LINKED ISSUE #518: Linking to a cell by cell ID
JupyterLab and nbconvert allow users to link to a specific cell by it's ID by using hash fragment like `#cell-id=my-cell-id` where `my-cell-id` is an example cell ID. JupyterBook nor MyST-NB does not support it yet (https://github.com/executablebooks/MyST-NB/issues/310) although a pull request is open (https://github.com/executablebooks/MyST-NB/pull/519).

Given that there is a new notebook parser in this repository, I wonder if it could facilitate implementation of such a feature.

## Proposal

Adopt the same hash fragment scheme for HTML output as other projects.

## Additional notes

It appears that support for a kind of cell IDs was implemented in https://github.com/executablebooks/mystmd/pull/361 but using a different scheme: `nb-cell-0-code` (and `nb-cell-0-output` for output). It appears that cell index was used there instead of cell ID which would not be a stable identifier when user moves cells around. For that I would propose using `#cell-index=0` format where 0 represents first cell (whether 0- or 1-based). Also, please note that in Jupyter notebooks any single cell may have multiple outputs, so cell ID cannot identify output (but barely a list of outputs).

Darshan808:
### Description
Right now, deep-linking to notebook cells is unreliable. We were discarding the nbformat cell id and falling back to a random nanoid, so `#<id>` links changed on every build.

This PR promotes notebook cell ids to stable anchors:

- `notebook.ts` preserves `cell.id` when reading `.ipynb` files.
- `blocks.ts` maps it to `identifier` (normalized) and `html_id` (verbatim).
- `myst-to-html` emits the anchor on block containers, so static HTML exports
  get stable links too.

### DIFF
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -33,9 +33,7 @@ function blockParent(cell: ICell, children: GenericNode[]) {
   const kind = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
   const data = JSON.parse(JSON.stringify(cell.metadata));
   // Stash the nbformat cell `id` so it can be promoted to a durable anchor
-  // downstream (see blockMetadataTransform). This is read verbatim from the
-  // .ipynb JSON, so its presence signals a trusted, file-persisted id: markdown
-  // notebooks never reach this reader, and ids mystmd fabricates live elsewhere.
+  // downstream
   const cellId = (cell as { id?: unknown }).id;
   if (typeof cellId === 'string' && cellId.length > 0) {
     data._jupyterCellId = cellId;
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -31,7 +31,16 @@ import type { PageFrontmatter } from 'myst-frontmatter';
 
 function blockParent(cell: ICell, children: GenericNode[]) {
   const kind = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
-  return { type: 'block', kind, data: JSON.parse(JSON.stringify(cell.metadata)), children };
+  const data = JSON.parse(JSON.stringify(cell.metadata));
+  // Stash the nbformat cell `id` so it can be promoted to a durable anchor
+  // downstream (see blockMetadataTransform). This is read verbatim from the
+  // .ipynb JSON, so its presence signals a trusted, file-persisted id: markdown
+  // notebooks never reach this reader, and ids mystmd fabricates live elsewhere.
+  const cellId = (cell as { id?: unknown }).id;
+  if (typeof cellId === 'string' && cellId.length > 0) {
+    data._jupyterCellId = cellId;
+  }
+  return { type: 'block', kind, data, children };
 }
 
 /**
--- a/packages/myst-cli/src/process/notebook.ts
+++ b/packages/myst-cli/src/process/notebook.ts
@@ -1,4 +1,4 @@
-import { NotebookCell, RuleId, fileWarn } from 'myst-common';
+import { NotebookCell, RuleId, fileWarn, normalizeLabel } from 'myst-common';
 import type { GenericNode, GenericParent } from 'myst-common';
 import { selectAll } from 'unist-util-select';
 import { nanoid } from 'nanoid';
@@ -16,7 +16,7 @@ import { logMessagesFromVFile } from '../utils/logging.js';
 import type { ISession } from '../session/types.js';
 import { BASE64_HEADER_SPLIT } from '../transforms/images.js';
 import { parseMyst } from './myst.js';
-import type { Code, InlineExpression } from 'myst-spec';
+import type { Code, InlineExpression, Block } from 'myst-spec';
 import { findExpression, metadataSection } from '../transforms/inlineExpressions.js';
 import type { IUserExpressionMetadata } from '../transforms/inlineExpressions.js';
 import { frontmatterValidationOpts } from '../frontmatter.js';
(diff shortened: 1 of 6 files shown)


agoose77:
Thanks for this PR @Darshan808! I'll take a look this week unless someone gets in first!

krassowski (commented):
Can it use `#cell-id=my-cell-id` schema to align with the format used by nbconvert and JupyterLab?

Darshan808:
Yes, ultimately we will support `#cell-id=<id>`. This PR enables mystmd to attach the bare nbformat cell id as the cell's stable anchor (it has to stay bare, because the same id is reused in JATS/XML where `=` is illegal). The `cell-id=` part is handled in `myst-theme` ([this pr](https://github.com/jupyter-book/myst-theme/pull/915)) with a small client-side hash handler that strips the prefix and scrolls.

agoose77:
@Darshan808 can you clarify your relationship with Jupyter Book? Your wording here
> Yes, ultimately we will support #cell-id=<id>. 

slightly implies that you have decision making authority for the project.

I think this is just a nuance of language, and I'm very appreciative of your contribution here — but I want to understand and clarify any confusion as the distinction is important.

Darshan808:
That was just an imprecise choice of words on my part. I don't have any decision-making authority for Jupyter Book.
By "we" I meant the broader Jupyter Book/MyST community and the direction this work is heading, not myself or the project's maintainers.
Sorry for the confusion.

krassowski:
Disclosure: I work with @Darshan808

> By "we" I meant the broader Jupyter Book/MyST community and the direction this work is heading, not myself or the project's maintainers.

I don't think one can speak for the broader community either, but certainly for some parts of the userbase as to what they seem desirable (i.e. having a stable way to link to different parts of a notebook across Jupyter interfaces). I think what would be fair to say in the original comment would be:

> Yes, ultimately we will try contribute support for `#cell-id=<id>`.

I for one was reaching out to a few members of MyST and JupyterBook community for years, sadly this was previously blocked by complexity of implementing this in Sphinx and possibly lack of impetus:
- https://github.com/executablebooks/MyST-NB/issues/310
- https://github.com/jupyter-book/mystmd/issues/518
- https://github.com/executablebooks/MyST-NB/pull/519 (it looks like someone switched it to draft probably because it picked up a conflict, but it was ready for review which never came; I heard in private conversations at the time that this was due to folks focusing on the JS rewrite of MyST)
- https://github.com/jupyter-book/jupyter-book/issues/1812

In the process multiple community members engaged with the discussion, supporting the idea or welcoming the PRs as seen in the issues linked above (Chris Sewell, Chris Holdgraf and Rowan Cockett), which might give the impression that there is some level of agreement in the community. Of course, whatever was expressed by individuals years ago does not imply a project-level agreement at the time, nor that they would still support this given the time that has passed. @agoose77 apologies that this was misinterpreted and misrepresented above.

agoose77:
@Darshan808 @krassowski thanks both for clarifying things! I really do not want this to get overblown — I just wanted to clear up what was being said. I wasn't aware that @krassowski works with @Darshan808, this adds some helpful context.

On this PR and the goals — Jupyter Book is a Jupyter subproject and we're very keen to ensure that we provide a good integration with the conventions and tooling like Jupyter Lab. So, I wholeheartedly agree with the idea from @krassowski that Jupyter Book is keen to move this forward. :heart:

krassowski:
Is there anything I could help with to move this forward?

agoose77:
@krassowski good timing — I was just circling back. Let me pick this up tomorrow (sets reminder).

packages/myst-cli/src/process/notebook.ts
```diff
@@ -31,7 +31,14 @@ import type { PageFrontmatter } from 'myst-frontmatter';
 
 function blockParent(cell: ICell, children: GenericNode[]) {
   const kind = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
-  return { type: 'block', kind, data: JSON.parse(JSON.stringify(cell.metadata)), children };
+  const data = JSON.parse(JSON.stringify(cell.metadata));
+  // Stash the nbformat cell `id` so it can be promoted to a durable anchor
+  // downstream
+  const cellId = (cell as { id?: unknown }).id;
+  if (typeof cellId === 'string' && cellId.length > 0) {
+    data._jupyterCellId = cellId;
```
agoose77: I think we might want to rename this field later on, but it's fine for now!

agoose77:
Thank you so much for this PR!

I've updated this PR to simplify the label propagation given my knowledge of our intention behind labels and how this state is managed. It was a good opportunity to clean something up!

packages/myst-cli/src/process/notebook.ts
```diff
@@ -29,9 +29,31 @@ import {
...
   const kind = cell.cell_type === CELL_TYPES.code ? NotebookCell.code : NotebookCell.content;
-  return { type: 'block', kind, data: JSON.parse(JSON.stringify(cell.metadata)), children };
+  const block: LabeledBlock = {
+    type: 'block',
+    kind,
+    data: JSON.parse(JSON.stringify(cell.metadata)),
+    children: children as any,
+  };
+  const cellId = (cell as { id?: unknown }).id;
+  if (typeof cellId === 'string' && cellId.length > 0) {
```
agoose77, line 47: Now we only handle cell IDs during parsing of an ipynb file. The intention here is to avoid stashing something — we can directly set the ID and html_id.

packages/myst-transforms/src/blocks.ts
```diff
@@ -79,7 +81,11 @@ export function blockMetadataTransform(mdast: GenericParent, file: VFile) {
         delete block.data.label;
       }
     }
-    if (block.identifier) {
+
+    // Did the user set a label? Choose to only propagate IDs if so.
+    const hasExplicitLabel = typeof block.label === 'string';
```
agoose77, line 86: In choose to check the final `block.label`, because that feels slightly more explicit than the intermediate `block.data`

agoose77 (approved):

krassowski:
Thank you!

Issue #2973, opened 2026-06-22
Explicitly mentioned TOC entries are not excluded from glob pattern


mtryan83:
## Description

According to the Table of Contents documentation on [glob pattern matching](https://mystmd.org/guide/table-of-contents#glob-pattern-matching):
> Files that are listed explicitly in the TOC will be ignored by the pattern

To my understanding, that means if I have a directory structure like so:
```
- /
  - Index.md
  - Test1.md
  - Test2.md
```

The `toc.yml`:
```yaml
version: 1
project:
  toc:
    - file: Index.md
    - file: Test1.md
    - pattern: ./*.md
```
should be equivalent to
```yaml
version: 1
project:
  toc:
    - file: Index.md
    - file: Test1.md
    - file: Test2.md
```

Instead, the generated output looks closer to
```yaml
version: 1
project:
  toc:
    - file: Index.md
    - file: Test1.md
    - file: Index.md
    - file: Test1.md
    - file: Test2.md
```

I _think_ the problem is somewhere in the `patternsToFileEntries` or `listExplicitFiles` functions, but given my unfamiliarity with typescript, I'm not sure where.

## Proposed ~~solution~~ outcome
I don't have the solution, but I believe the correct outcome should be the first output:

```yaml
version: 1
project:
  toc:
    - file: Index.md
    - file: Test1.md
    - file: Test2.md
```

## Additional notes
N/A

PR #2978, opened 2026-06-25
Properly discard glob entries that are explicitly enumerated in ToC

parmentelat:
fixes issue #2973

the 'ignore' context that carries already met file entries contains resolved entries, so we need to resolve the item in consideration if we want to find it in that collection

### DIFF
--- a/.changeset/eighty-dots-train.md
+++ b/.changeset/eighty-dots-train.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': minor
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/eighty-dots-train.md
```diff
@@ -0,0 +1,5 @@
+---
+'myst-cli': minor
```
agoose77:
```suggestion
'myst-cli': patch
```

PR #2979, opened 2026-06-26
WIP: Project state in plugins

### LINKED ISSUE #1616: Allow plugins access to the MyST documents state
I was trying to write a custom plugin that returned a list of blog pages in my site, similar to this issue:

- #840 

However I couldn't figure out (and think it may not be possible) to define a `.mjs` plugin that can access the information needed to do this.

It would be helpful to enable / document how MyST plugins can access the list of all parsed documents in a MyST build, along with their metadata and/or content.

rowanc1:
Working in person with @stefanv on blog listing through improving the global project state that is available to plugins.

image

This puts more information in the global plugin (`document` phase is most reliable), and can be exposed through plugins in a bit of a simpler way. This can help support blog listings, and many other full-project views of parsed content that do not rely on parsing the `md` files independently.

---

- closes https://github.com/jupyter-book/mystmd/issues/1616


examples/plugin-tagged/myst.yml
```diff
@@ -0,0 +1,13 @@
...
+version: 1
+project:
+  title: Tagged Pages Plugin Example
+  plugins:
+    - plugin.mjs
+  toc:
+    - file: index.md
+    - file: apples.md
+    - file: oranges.md
+    - file: bananas.md
```
rowanc1, line 11: Not sure if this is worth putting in here, but sometimes it is nice to jump into a specific example?

packages/myst-common/src/types.ts
```diff
@@ -115,7 +115,62 @@ export type RoleSpec = {
...
+   * The current project, including the list of all pages and their tags, as
+   * well as the site configuration nested under `project.site`.
+   *
+   * This is only fully populated for `project` stage transforms; during
+   * `document` stage transforms other pages may not have been processed yet.
+   */
+  project?: SessionProject;
+};
+
+export type PluginUtils = { select: Select; selectAll: SelectAll; unstableSession: SessionAPI };
```
rowanc1, line 173: This is the main change, that we pass more information to the plugins.

packages/myst-cli/src/process/mdast.ts
```diff
@@ -108,6 +112,76 @@ export type TransformFn = (
...
+ *
+ * This gathers the list of all pages in the project (with their tags, slugs, and
+ * resolved urls) along with the current site configuration so that plugins, e.g. a
+ * directive that lists pages with a given tag, can operate on project-wide data.
+ *
+ * Page tags and urls are populated during `transformMdast`; for `project` stage
+ * transforms (running in `postProcessMdast`) all pages have been processed so this
+ * data is complete.
+ */
+function getSessionProject(session: ISession, projectPath?: string): SessionProject | undefined {
```
rowanc1, line 126: Here we actually construct this. This function is pretty messy, and the goal is to resolve a clean API to consumers through the plugins.

rowanc1 (commented):
Comments!

packages/myst-common/src/types.ts
choldgraf: so if I understand, PluginUtils now gets an extra `SessionAPI` object, and within that object, plugin authors can access the `SessionProject`, and within that they can access the `SessionSite` and list of `SessionPage` objects?

packages/myst-cli/src/process/mdast.ts
```diff
@@ -220,9 +294,17 @@ export async function transformMdast(
     .use(inlineMathSimplificationPlugin, { replaceSymbol: false })
     .use(mathPlugin, { macros: frontmatter.math });
   // Load custom transform plugins
+  const documentSessionProject = getSessionProject(session, projectPath);
```
choldgraf, line 297:
```suggestion
/** Document transform stage */
// Load custom transform plugins
  const documentSessionProject = getSessionProject(session, projectPath);
```

choldgraf: Just to make it clear this is document stage (I think it is?)

packages/myst-cli/src/process/mdast.ts
```diff
@@ -376,9 +458,17 @@ export async function postProcessMdast(
   await transformMystXRefs(session, vfile, mdast, frontmatter);
   await embedTransform(session, mdast, file, dependencies, state);
   const pipe = unified();
+  const projectSessionProject = getSessionProject(session, projectPath);
   session.plugins?.transforms.forEach((t) => {
```
choldgraf, line 462:
```suggestion
  /** Project transform stage (document stage is above) */
  const projectSessionProject = getSessionProject(session, projectPath);
  session.plugins?.transforms.forEach((t) => {
```

choldgraf: also not critical - just trying to make the macro-structure of this a bit easier to parse

packages/myst-cli/src/process/mdast.ts
```diff
@@ -376,9 +458,17 @@ export async function postProcessMdast(
...
     if (t.stage !== 'project') return;
-    pipe.use(t.plugin, undefined, pluginUtils);
+    pipe.use(t.plugin, undefined, {
+      select,
+      selectAll,
+      unstableSession: {
```
choldgraf, line 467: Quick check: if old plugins don't include `unstableSession` will they break until they update? Could we ensure it's optional so this doesn't happen?

examples/plugin-tagged/plugin.mjs
```diff
@@ -0,0 +1,105 @@
...
+  return { type: 'card', url: page.url, children };
+}
+
+/**
+ * Project-stage transform that replaces each placeholder node with a grid of
+ * cards for every page in the project carrying the requested tag.
+ */
+function taggedTransform(opts, utils) {
+  return async (mdast) => {
+    const { selectAll, unstableSession } = utils;
```
choldgraf, line 69: `unstableSession` feels a little idiosyncratic to me - could we try to think of a more intuitive naming for plugin authors? Based on what's inside, my brain immediately went to something like `sessionContext` or `buildContext` or something like that...I feel like the `unstable` _could_ make sense, but is also a bit verbose and may not be necessary as long as we document this functionality clearly enough (ie, note that it is unstable, and that certain moments in the build chain are more unstable than others)

packages/myst-common/src/types.ts
```diff
@@ -115,7 +115,62 @@ export type RoleSpec = {
...
 
-export type PluginUtils = { select: Select; selectAll: SelectAll };
+/**
+ * Page information exposed to plugins via the unstable session API.
+ *
+ * This is the full page frontmatter (title, description, tags, date, authors,
+ * affiliations, contributors, doi, etc.) extended with the page's location
+ * within the site.
+ */
+export type SessionPage = PageFrontmatter & {
```
choldgraf, line 125: Does this open a pathway towards us getting the `AST` from the page as well? I know that here it's just frontmatter, but I think it'd be quite useful to get information in the AST from other pages. For example, that "collect abbreviations" PR that @haroldcampbell opened up (#2985 ).

packages/myst-common/src/types.ts
```diff
@@ -115,7 +115,62 @@ export type RoleSpec = {
...
+  title?: string;
+  description?: string;
+  options?: Record<string, any>;
+  nav?: any[];
+  actions?: any[];
+  domains?: string[];
+  template?: string;
+};
+
+/** Minimal project information exposed to plugins via the unstable session API. */
```
choldgraf, line 148: why not pass the whole project config? I've seen several plugins want to have access to other config in `myst.yml` (though, in many cases they are defining their _own_ config in `myst.yml` which might get dropped when it passes through the validator...).

I think this isn't strictly needed for this PR but want to flag it in case it is a design decision to leave open a pathway for doing this in a follow up

packages/myst-cli/src/process/mdast.ts
```diff
@@ -220,9 +294,17 @@ export async function transformMdast(
...
   session.plugins?.transforms.forEach((t) => {
     if (t.stage !== 'document') return;
-    pipe.use(t.plugin, undefined, pluginUtils);
+    pipe.use(t.plugin, undefined, {
+      select,
+      selectAll,
+      unstableSession: {
+        page: { slug: pageSlug, frontmatter },
+        project: documentSessionProject,
+      },
```
choldgraf, line 306: Or if it's not easy to provide the AST to transforms, could we add `parseMyst` function to the transforms and then they could parse stuff themselves? (I think this would be useful either way)

choldgraf (commented):
This seems like a really nice start, and this would be a really useful addition for plugin authors. I linked https://github.com/jupyter-book/mystmd/issues/1616 , and added a few comments throughout.

In general, I think this looks pretty good - we should add a little bit of plugin dev docs so authors can learn more about this, and there were a few pieces that I thought would be particularly useful to _add_, though not critical as long as we don't close a door on them. If we don't have bandwidth for that right now, my vote would be to just get this in so we can experiment with it.

examples/plugin-tagged/plugin.mjs
agoose77: I do like the explicitness of "unstable". We could also use a namespace, e.g. 

```
utils.hazmat.session.
```

packages/myst-common/src/types.ts
```diff
@@ -115,7 +115,62 @@ export type RoleSpec = {
...
+export type SessionProject = {
+  slug?: string;
+  index?: string;
+  title?: string;
+  pages: SessionPage[];
```
rowanc1, line 153: I think this should be one `pages[]` at the top. The slug/index/title should be the first one in here.

examples/plugin-tagged/plugin.mjs
choldgraf: I won't die on this hill if y'all like `unstableFoo` :-)

PR #2981, opened 2026-07-01
Fix IReplyAbortContent type

stefanv:
@jupyterlab/services@7.6 widened `IReplyAbortContent.status` to
'abort' | 'aborted'; the local variable type needs to match.

### DIFF
--- a/packages/myst-execute/src/kernel.ts
+++ b/packages/myst-execute/src/kernel.ts
@@ -133,7 +133,7 @@ export async function executeCodeCell(kernel: Kernel.IKernelConnection, code: st
       }
     }
   };
-  let status: 'abort' | 'error' | 'ok' | undefined;
+  let status: 'abort' | 'aborted' | 'error' | 'ok' | undefined;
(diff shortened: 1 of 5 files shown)


stefanv:
See https://github.com/jupyter-book/mystmd/pull/2982 for CI fix

packages/myst-execute/src/kernel.ts
```diff
@@ -133,7 +133,7 @@ export async function executeCodeCell(kernel: Kernel.IKernelConnection, code: st
       }
     }
   };
-  let status: 'abort' | 'error' | 'ok' | undefined;
+  let status: 'abort' | 'aborted' | 'error' | 'ok' | undefined;
```
agoose77: Can we just pull the type from the upstream?

stefanv: That's a good idea!

stefanv:
@agoose77 Now implements your suggestion.

PR #2982, opened 2026-07-01
fix: tolerate failure on unconfigured favicon

stefanv:
Without a configured favicon, the theme falls back to fetching a default icon from mystmd.org. When that site is down (as recently happened on CI), it breaks the build. This PR adds an "optional" flag to a route, so that if that route cannot be fetched the static build simply continues.

We only enable that flag for the fallback favicon.ico, since it really isn't important to have it.

Also add error listeners on the binary response/write streams so a premature stream close rejects the write promise instead of hanging indefinitely.


stefanv:
All PRs are currently failing CI, including mystmd, so going to merge this to get things going. Happy to revert and implement an alternative.

PR #2983, opened 2026-07-01
Add proof:kind spelling for proof directives and ref role

stefanv:
Adds `proof:kind` (e.g. `{proof:theorem}`) and `{proof:ref}` as the recommended spellings, generated alongside the legacy `prf:kind`/`prf:ref` aliases so existing content keeps building. Docs now only show the `proof:` forms.

### DIFF
--- a/.changeset/yellow-bikes-carry.md
+++ b/.changeset/yellow-bikes-carry.md
@@ -1,0 +1,6 @@
+---
+'myst-ext-proof': patch
+'myst-roles': patch
+---
+
+Add proof:_ as alias for prf:_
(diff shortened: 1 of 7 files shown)


stefanv:
After merging, update myst-theme example docs (like color picker) to use `proof` instead of `prf`.

rowanc1 (approved):
Nice - code review looks good to go.

This makes a bunch more sense, and is backwards compatible with sphinx-proof as well, which is where this came from.

choldgraf (approved):
I think proof is pretty short, and more precise, so I'm +1 on this! Given that @rowanc1 already approved as well, I'll merge! :-)

PR #2986, opened 2026-07-03
Decode HTML entities in CSL text fields

stefanv:
CrossRef's CSL-JSON metadata sometimes contains HTML entities in title and journal name fields.

Thanks to @jarrodmillman for the report.


choldgraf (approved):
The test looks good to me - if I understand, this won't change anything for the vast majority of cases but now if there html stuff in the metadata, it'll get converted into a sensible character.

One double check: there's no native html decode we could use? The "he" package is the most lightweight way here? Just double checking before we add a new dep but if you think it's the right decision imo, go ahead and merge!

stefanv:
That's a good question, Chris; I'm not sure — it didn't look like any of other pre-existing deps did it, and it's a bit surprising that there's no "batteries included" solution. Maybe @agoose77 or @fwkoch knows of something?

PR #2987, opened 2026-07-03
Switch bun to isolated linker mode

stefanv:
Sets linker = "isolated" in bunfig.toml so each package's node_modules only contains what it actually declares, instead of a hoisted flat tree. This surfaced ~30 phantom dependencies across the workspace (types/packages used but not declared) which are now fixed.

Temporarily works around a bug in meca/jats-utils/jats-xml (misdeclared as devDependencies upstream, PR pending at
https://github.com/continuous-foundation/jats/pull/109)

The workarounds:

- bumped jats-xml to the already-fixed 1.1.7
- added esbuild --alias flags so mystmd's bundled CLI resolves myst-cli-utils/myst-common correctly regardless
- disabled (renamed .skip) three test files that load meca/jats-utils directly and have no bundler to alias around; jats-to-myst's test script now uses `--pass-with-no-tests` to make it green


stefanv:
@agoose77 I noticed that you use pnpm for everything, and I recalled (incorrectly) that pnpm had less strict handling of dependencies. Turns out *bun* has less strict handling, contrary to what I originally intended to implement.

This fixes that for bun; and, should we wish to transition to pnpm in the future, it would open the path to that.

There is a related PR (https://github.com/continuous-foundation/jats/pull/109) that needs to be merged in CSF upstream repos and released before this should go in, so I can revert the hacks I made because of import problems in those packages.

rowanc1:
@fwkoch released v1.1.8 on Friday, this should be unblocked a bit more @stefanv!

rowanc1 (approved):

Issue #2955, opened 2026-06-15
xref is decapitalizing references


jorgensd:
With the minimal setup `intro.md`
```markdown
# My reference

[Actual docs](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray)
[xref](xref:numpy#numpy.typing.NDArray)
```
and the config file
```yml
version: 1
project:
  id: 179c4232-4898-41ad-9741-c6af05dc0479
  title: xref test
  toc:
    - file: intro.md

  references:
    numpy: https://numpy.org/doc/stable/
site:
  template: book-theme
```
i get the following references

- Actual docs: https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray
- xref: https://numpy.org/doc/stable/reference/typing.html#numpy.typing.ndarray
where one observes that `NDArray` has been converted to `ndarray`, which therefore doesn't reference the correct section of the documentation.

coretl:
I think this is duplicate of #1758

Issue #1758, opened 2025-01-10
Intersphinx inventory fragment IDs get over-normalized.


ebolyen:
## Description
It appears that `xref:sphinx` style link nodes have their `url` attribute normalized, whereas the original `urlSource` preserves case.
Here is an example:

```
<xref:qiime2#qiime2.plugin.Bool>
```
This correctly matches the intersphinx inventory and myst resolves it, but produces the following AST:
```json
{
    "type": "link",
    "url": "https://develop.qiime2.org/en/latest/plugins/references/api/types.html#qiime2.plugin.bool",
    "position": {
        "start": {
            "line": 5,
            "column": 1
        },
        "end": {
            "line": 5,
            "column": 1
        }
    },
    "children": [
        {
            "type": "text",
            "value": "qiime2",
            "key": "WmC8YrGWys"
        }
    ],
    "urlSource": "xref:qiime2#qiime2.plugin.Bool",
    "internal": false,
    "protocol": "xref:sphinx",
    "key": "PAt4Q0ZuxY"
}
```

This normalized `url` does not function as the fragment has changed from `Bool` to `bool`.
In the browsers I've checked (Firefox and Chrome), URL fragments are case-sensitive.

## Proposed solution
Don't normalize the fragment id under the `xref:sphinx` protocol.

## Additional notes
```
§ jupyter-book --version
v2.0.0-a0
```
myst.yml references section
```
  references:
    qiime2: https://develop.qiime2.org/en/latest/
```

ebolyen:
Oh and the `children` property makes it look like this normalization must have happened relatively early, since I would expect something other than the `project` label to make it into link's text.

ebolyen:
Looks like it may be this line in the `intersphinx` package:
https://github.com/continuous-foundation/intersphinx/blob/main/src/intersphinx.ts#L157

It's been that way for a few years, @rowanc1, do you recall the reason for this?

```
§ intersphinx list https://develop.qiime2.org/en/latest/ --domain py:data --includes Bool
py:data  (qiime2.plugin.Bool)
  https://develop.qiime2.org/en/latest/plugins/references/api/types.html#qiime2.plugin.bool

```

vs
```
§ python -m sphinx.ext.intersphinx Downloads/objects.inv | grep -C1 Bool
py:data
    qiime2.plugin.Bool                                                               : plugins/references/api/types.html#qiime2.plugin.Bool
[...]

Issue #1259, opened 2024-05-29
Add support for API documentation generation


agoose77:
## Proposal

We would like to be able to use MyST for API documentation. It's already very close to possible; adding support for declaring functions, classes, etc, just requires some new directives and some labelling logic. To support more advanced documentation tools, like `sphinx-autodoc`, we could leverage Sphinx itself.

## Related issues:
- https://github.com/executablebooks/mystmd/issues/293

choldgraf:
Here's a related issue that is specifically about docstrings https://github.com/jupyter-book/mystmd/issues/1858

rowanc1:
From @fwkoch on Discord:

> This is a demo of some work @stefanv and I did at the Scientific Python Developer Summit. The goal was to get autogenerated Python docs rendering in MyST, a critical step to enable websites with technical API docs to upgrade to MyST. Let me know if you have any thoughts or questions!
> https://www.loom.com/share/589a947389fb49829c9b712b796d8a7d

echarles:
> This is a demo of some work @stefanv and I did at the Scientific Python Developer Summit.

Looks awesome! Is there a link to a repo with the source of that work?

rowanc1:
The Python code and the MyST plugin are here: https://github.com/stefanv/npdoc2json

romanlutz:
@rowanc1 any plans for Google-style docstrings?

From a cursory glance, it seems like it may be as simple as just replacing the Python script with logic to parse the Google-style docstrings. I might give this a try one of these days and report back 🙂

atharvaaalok:
Has there been any progress on generating API documentation using docstrings?
If so, is there any resource/webpage that I could use to create my own MyST based API docs?
Cheers!

stefanv:
@romanlutz Yeah, exactly! The beauty of this demonstrator is that it is pretty easy to modify to work with any format, as long as you can parse it. I will put a few more cycles into this soon, perhaps try it on scikit-image or similar.

atharvaaalok:
@stefanv Is there anything planned for incorporating the functionality of building documentation through an API call in the near future as a proper feature in MyST?

julian-belina:
Are there any plans to make this possible? V2 looks much nicer, but without the ability to automatically generate API documentation, it's not suitable for my use cases.

choldgraf:
[...]

PR #2989, opened 2026-07-06
🐛 Preserve matched target identifier for case-sensitive cross-references

### LINKED ISSUE #2988: Links to case-sensitive targets resolve to the wrong (lowercased) target
## Description

`resolveReferenceLinksTransform` already looks up link targets by their **verbatim** identifier before falling back to the normalized (lowercased) form ([`packages/myst-transforms/src/enumerate.ts`](https://github.com/jupyter-book/mystmd/blob/main/packages/myst-transforms/src/enumerate.ts)):

```ts
const target = opts.state.getTarget(identifier) ?? opts.state.getTarget(reference?.identifier);
```

…but then unconditionally overwrites the result with the normalized identifier:

```ts
xref.identifier = reference.identifier; // always the lowercased form
```

So when two case-distinct targets exist, a link to the exact-case one *finds* the right target and then silently re-points at its lowercase sibling. The same pattern exists in `resolveUnlinkedCitations` a few lines below.

## Reproduction

Case-distinct targets can't be authored with `(label)=` (those are normalized by design, which is appropriate for prose). But any mdast node carrying an `identifier` registers as a target, which is exactly what an API-docs plugin does. Minimal plugin directive registering `sample.Match` (a class) and `sample.match` (a function) with exact-case `identifier`/`html_id`, then:

```markdown
Class reference: [](#sample.Match)

Function reference: [](#sample.match)
```

Built page AST (mystmd 1.10.1 and current `main`):

```text
crossReference identifiers: ["sample.match", "sample.match"]   # both point at the function
```

The complete runnable reproducer (plugin + project + assertion script) is in https://github.com/coretl/mystmd-repro-repo, whose CI matrix demonstrates the bug on released 1.10.1 **and** `main`, and correct behaviour with the one-line fix: https://github.com/coretl/mystmd-repro-repo/actions

## Proposed fix

Keep the identifier of the target that actually matched, falling back to the normalized identifier as before:

```ts
xref.identifier = target?.node.identifier ?? reference.identifier;
```

Ordinary prose labels are unaffected: their stored identifier *is* the normalized identifier, so the two expressions are identical for every target that exists today; only targets registered with intentionally case-sensitive identifiers change behaviour. Fix + regression test on this branch, all `myst-transforms` tests passing: https://github.com/coretl/mystmd/tree/fix/xref-identifier-case. Happy to open it as a PR.

## Context

This is the local-reference sibling of the intersphinx `$`-anchor lowercasing bug (#1758, #2955; fix pending in continuous-foundation/intersphinx#6). That bug and this one are the last blockers for case-sensitive Python API cross-references (`re.match` vs `re.Match`); see the discussion in #1259. We (Diamond Light Source / bluesky) are migrating [ophyd-async](https://github.com/bluesky/ophyd-async)'s API docs from Sphinx to MyST and hit both.

```text
§ myst --version
v1.10.1 (and main @ bde9e9c0)
```

## AI Disclaimer

The code and this description were generated with AI and reviewed/tweaked by the author.

coretl:
Fixes #2988.

## The bug

`resolveReferenceLinksTransform` looks up link targets by their **verbatim** identifier before falling back to the normalized (lowercased) form:

```ts
const target = opts.state.getTarget(identifier) ?? opts.state.getTarget(reference?.identifier);
```

...but then unconditionally overwrites the resulting cross-reference with the normalized identifier:

```ts
xref.identifier = reference.identifier; // always the lowercased form
```

So when two case-distinct targets exist, a link to the exact-case one *finds* the right target and then silently re-points at its lowercase sibling. With a plugin directive registering `sample.Match` (a class) and `sample.match` (a function) as targets, both `[](#sample.Match)` and `[](#sample.match)` end up as crossReferences to `sample.match`. `resolveUnlinkedCitations` has the same pattern a few lines below.

## The fix

Keep the identifier of the target that actually matched, falling back to the normalized identifier as before:

```ts
xref.identifier = target?.node.identifier ?? reference.identifier;
```

Ordinary prose labels are unaffected: their stored identifier *is* the normalized identifier (`normalizeLabel` at creation time), so the two expressions are identical for every target that exists today. Only targets registered with intentionally case-sensitive identifiers (which today can only come from plugins setting `identifier` directly on nodes) change behaviour.

## Changes

- `packages/myst-transforms/src/enumerate.ts`: use the matched target's identifier in `resolveReferenceLinksTransform` and `resolveUnlinkedCitations`.
- `packages/myst-transforms/src/enumerate.spec.ts`: regression test with two case-distinct targets; fails without the fix (both resolve to `sample.match`), passes with it. All myst-transforms tests pass.
- changeset (patch).

## Independent CI reproduction

https://github.com/coretl/mystmd-repro-repo runs a full end-to-end reproducer (plugin + project + AST assertion) as a matrix: released 1.10.1 and upstream `main` asserting the bug, this branch asserting the fix: https://github.com/coretl/mystmd-repro-repo/actions

## Context

This is the local-reference sibling of the intersphinx `$`-anchor lowercasing bug (#1758, #2955), just fixed in continuous-foundation/intersphinx#6 (released as intersphinx 1.1.0). Together they make case-sensitive references correct end-to-end, which matters for Python API docs where `re.match` (function) and `re.Match` (class) are different objects; see the discussion in #1259. We (Diamond Light Source / bluesky) hit both while migrating [ophyd-async](https://github.com/bluesky/ophyd-async)'s API docs from Sphinx to MyST.

## AI Disclaimer

The code and this description were generated with AI and reviewed/tweaked by the author.


choldgraf (approved):
This seems good to me - thanks for adding a simple test for this as well. Given that it brings this behavior in line with another part of the ref transformer, it feels safe to merge to me. IMO this is also closer to what one would expect given that we haven't documented labels as case insensitive

rowanc1:
@fwkoch can you take a look?

packages/myst-transforms/src/enumerate.ts
```diff
@@ -803,7 +803,11 @@ export const resolveReferenceLinksTransform = (tree: GenericParent, opts: StateR
     // Change the link into a cross-reference!
     const xref = link as unknown as CrossReference;
     xref.type = 'crossReference';
-    xref.identifier = reference.identifier;
+    // Prefer the identifier of the target that actually matched: getTarget
```
stefanv: Our logic is pretty yucky here. Wondering if we should update the `getTarget` API a bit to do case-insensitive matching + return of the found target label. That way the logic can become:

```ts
target, reference.identifier = opts.state.getTarget(identifier);
```

or similar.

What do you think @fwkoch?

choldgraf: I do agree with this as well 🙂 is the author is up for a little clean up I'm happy to iterate via this PR.

I think that we want is case _sensitive_ matching though, right?

coretl: I've had a go at folding the normalization into `getTarget`. Changing the signature of `getTarget` seemed too big a change for me to make without understanding the codebase better. Is that any cleaner?

fwkoch (commented):
Thanks for this @coretl! This is a subtle bug I agree we should fix. From my perspective, the point of normalization is two-fold: (1) making non-standard characters in `identifier` safe for reuse in other contexts and (2) allowing user-friendly coercion for minor mismatches rather than hard failures.

This PR addresses (2) in a nice way - it allows careful authors to set up robust case-sensitive xref/target pairs, but it still falls back to the "user-friendly" normailze-if-there-is-not-a-case-sensitive-match behavior.

However, it opens up (1) to more risk I think - if an author sets up matching xref/target pairs with non-standard characters, it won't be normalized at all and _may_ cause issues in other contexts where these characters break the xrefs.

I'm wondering if your fix should _only_ address case mismatch... That would cover your `sample.Match` example. (It's also worth pointing out - in other contexts - web, typst, tex - cross-references are all case-sensitive. So supporting case but not special characters could make sense.) If we are going this route, it would probably change the implementation a bit, where `normalizeLabel` (and probably `createHtmlId`) no longer includes `toLowerCase()` and we move the `lower` fallback to `getTarget` (so we never check against raw identifier; instead we check against normalized-but-not-lowered value then normalized-and-lowered value).

Does that track and make sense...? (Happy to discuss further if you think I'm missing the mark 😅)

coretl:
Thanks for the reply @fwkoch. I'll try to answer it, but I work mainly in low level python, so a PR in TypeScript about front end concepts is stretching my knowledge somewhat! I leaned heavily on AI to try and explain the details to me, but I may have missed some important context.

Were you suggesting that prose labels should also become case sensitive as @choldgraf suggested in https://github.com/jupyter-book/mystmd/issues/2988#issuecomment-4894040315?

If so, then wouldn't that break the html anchors of every existing mystmd site that happened to use uppercase characters in the label text? Is there a use case for having case preserving anchors generated from prose references?

The only reason I ask is that it doesn't actually seem required for this change, as long as the producer assigns a valid identifier, it is used without lowercasing (this PR stops the reference resolver forcing it to lowercase), which fixes the issue for any plugin which wants to make case sensitive references.

<details>
<summary><b>More details (AI generated)</b></summary>

### Why this PR doesn't need prose labels to change

**A preset `identifier` is already honored verbatim.** Every `normalizeLabel` call in the pipeline derives an identifier from a `label` (or the `:label:` option), and that is the only path that lowercases. A node whose `identifier` is set directly is stored in `state.targets` verbatim by `addTarget`, never re-lowercased. So a plugin that assigns a valid exact-case identifier (e.g. `sample.Match` py:class vs `sample.match` py:function) gets two distinct targets today.

**The only real bug was resolution.** `resolveReferenceLinksTransform` / `resolveUnlinkedCitations` found the correct case-sensitive target via `getTarget`, but then re-stamped the resulting `crossReference` with the normalized (lowercased) identifier, so `#sample.Match` silently re-pointed at `sample.match`. This PR keeps the matched target's own identifier. If we agree that "a preset `identifier` is honored verbatim" is the intended contract, I'm happy to formalize it with a test (two case-distinct targets registering verbatim and keeping their `html_id`, i.e. the exact fields `writeObjectsInv` reads) so a future refactor can't regress it.

**The export side is already case-preserving.** `writeObjectsInv` keys each `std:label` row by the target's identifier and locates it by `html_id`; intersphinx `setEntry` stores explicit names and anchors verbatim (it only ever lowercased the `$` shorthand, which the writer doesn't emit, and that `$` case is fixed in the latest intersphinx release for the consume side). So the two objects come out as distinct exact-case rows.

### Why making prose labels case sensitive would break existing anchors

Prose identifiers are lowercased at creation, by `normalizeLabel` and `createHtmlId`. A heading `## My Section` currently produces anchor `#my-section`. Dropping the `toLowerCase()` (the choldgraf/fwkoch idea) would make it `#My-Section`, so:

- **Every existing site with uppercase in a heading or label gets different anchor slugs** (`#my-section` -> `#My-Section`), breaking external deep-links and any hand-written `#anchor` links.
- `createHtmlId`'s character class would also need widening to `[A-Za-z0-9-]`, or uppercase letters get replaced with `-`.
- `getTarget` would need a lowercased fallback index (not just `targets[normalized.toLowerCase()]`) so old-style `#my-section` links to `## My Section` still resolve once stored keys are mixed-case.
- Duplicate-identifier detection changes: `## Foo` and `## foo` would stop colliding.

### Is there a use case for case-preserving prose anchors?

There's a reasonable one: web, typst, and tex cross-references are all case sensitive, and Python's own docs serve `re.Match` and `re.match` as distinct anchors. So it's defensible as a deliberate, separate change. It just isn't required to let a plugin produce case-sensitive references, and it's a breaking change to existing sites, so it seems better kept out of this fix.

</details>

choldgraf:
I just pushed a few small improvements to this branch:

- added a two-pass system so that we first search all pages for exact matches, then do another pass for normalized matches if we don't find an exact match. That way we can prefer exact matches across the whole site, even if a normalized match exists somewhere (that's probably relevant for things like `#Classname`
- made the function for exact matching more explicit
- added some docstrings and tried to comment a little bit more so it's clearer how the logic works

I feel like this is safe to merge - it seems like we _don't_ want to change the HTML labels since that'd be a bigger user-facing shift, and since HTML is already getting its normalized labels via something other than the identifier, that behavior won't change.

PR #2997, opened 2026-07-09
Fix documentation plugins

### LINKED ISSUE #2993: Multiple "Unknown Directive" errors in latest deployment of mystmd docs
I noticed that the [Reference / Directives](https://mystmd.org/guide/directives) and [Reference / Roles](https://mystmd.org/guide/roles) pages are full of "Unknown Directive" errors:

Image

I'm not sure whether it's a configuration issue or what, but this seems related to [a few hundred errors reported in the logs](https://github.com/jupyter-book/mystmd/actions/runs/28960869026/job/85931774633#step:7:22) for the latest run of the "Deploy Docs" GitHub Action (although the workflow overall seems to have reported success).

My apologies if this is not the right place to report this situation.

stefanv:
We used to rely on hoisting of dependencies from the root of this
repo, but that broke when switching bun to standalone mode.

This PR declares explicit dependencies for the docs (by name, not version---we simply use the versions in the root repo), and adds a plugin build step to the docs build.

We could, with some shenanigans, get away without the build step, but I think (a) it is cleaner and (b) better to eat our own dogfood the way a user would have to.

Closes #2993

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -315,6 +315,8 @@ The build process uses unix commands that might not work properly on Windows.
 When building on Windows, use either WSL or a unix-like shell (such as Git Bash or MSYS2).
 ```
 
+(local-docs-workflows)=
+
 ### Local docs workflows (preview/build)
 
 To run a MyST server with your local changes and a preview of the documentation, run:
@@ -322,6 +324,10 @@ To run a MyST server with your local changes and a preview of the documentation,
 ```shell
 $ bun run docs
 ```
+
+This builds the `mystmd` packages, and starts a live-reloading preview server.
+
+The `docs/` plugins import several `myst-*` packages as `workspace:*` dependencies, so that they build against the source versions in this repository (not against npm). If you only change the plugins in `docs/*.mjs`, you will need to rebuild the packages and restart the server.
 
 ## Local development: myst-theme
 
--- a/bun.lock
+++ b/bun.lock
@@ -28,6 +28,26 @@
         "vitest": "^3.2.4",
       },
     },
+    "docs": {
+      "name": "mystmd-docs",
+      "version": "0.0.0",
+      "dependencies": {
+        "myst-common": "workspace:*",
+        "myst-directives": "workspace:*",
+        "myst-ext-button": "workspace:*",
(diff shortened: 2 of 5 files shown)


fperez:
I can't really offer a proper review, but I figured I'd test the workflow from the PR, and with only `npm` and `bun` installed on my system, everything worked for the initial install/link steps, but when I tried the plugin build step I got:

```bash
(main)blanca[docs]> bun run build:plugins
error: Script not found "build:plugins"
```

If this was user error on my end, ignore! But I was copy/pasting the steps from the PR and that didn't work :) 

This was on macOS Sequoia 15.7.7, with bun and npm installed via homebrew and not much else (no big conda env active).

stefanv:
That command is supposed to be run from the docs folder, but the "cd" may have gotten lost in my instructions!

fperez:
Nope, I was in the docs folder as you can see ( `[docs]` in my bash prompt is my CWD).

fperez:
I did the cd step (I also tried backtracking one dir up just in case... This is my bash history for that session:

```bash
git clone https://github.com/jupyter-book/mystmd
cd mystmd/
bun install
bun run build
bun run link
brew install npm
bun run link
cd docs/
bun run build:plugins
```

stefanv:
Looks like we may not need the build step, so will revert those docs changes.

rowanc1:
I simplified to not use esbuild.

docs/developer.md
```diff
@@ -323,6 +325,10 @@ To run a MyST server with your local changes and a preview of the documentation,
 $ bun run docs
 ```
 
+This builds the `mystmd` packages, and starts a live-reloading preview server.
+
+The `docs/` plugins import several `myst-*` packages as `workspace:*` dependencies, so that they build against the source versions in this repository (not against npm). If you only change the plugins in `docs/*.mjs`, you will need to rebuild the packages and restart the server.
```
stefanv, line 330: Maybe a remnant of previous stuff here. Only need to rebuild if packages are changed.

stefanv:
Great that we can get away without building; thank you for noticing that, Rowan.

rowanc1:
Docs are fixed!!

PR #2999, opened 2026-07-10
Add avif and webm to the list of supported extensions

### LINKED ISSUE #2996: .avif and .webm extension support
### Proposal

[AV1 video](https://en.wikipedia.org/wiki/AV1) [AVIF (AV1 encoded picture)](https://en.wikipedia.org/wiki/AVIF) offer better compression for images and videos compared to other state of the art formats. If I add an AV1 encoded video in a `webm` container or an AV1 encoded image with `avif` extension as follows:

```
![](/src/include/video.webm)
```

I get the error message similar to.

```
⛔️ src/index.md:9 Unsupported image extension ".webm" may not correctly render.
```

Would be great if mystmd would also support these extensions, which are supported by wide-spread browsers.

### Additional notes

Workarounds: Overriding local files with `<video controls>` not possible, because it is converted to an `img` tag. But using an URL and `iframe` works:

```
:::{iframe} https://upload.wikimedia.org/wikipedia/commons/f/f4/Demo_Video_Tutorial.webm
:::
```

goekce:
Fixes #2996


packages/myst-cli/src/utils/resolveExtension.ts
```diff
@@ -13,13 +13,15 @@ export enum ImageExtensions {
...
+  avif = '.avif',
   mp4 = '.mp4', // A moving image!
   mov = '.mov',
   avi = '.avi',
+  webm = '.webm',
 }
 
 export const KNOWN_IMAGE_EXTENSIONS = Object.values(ImageExtensions);
-export const KNOWN_VIDEO_EXTENSIONS = ['.mp4', '.mov', '.avi'];
+export const KNOWN_VIDEO_EXTENSIONS = ['.mp4', '.mov', '.avi', '.webm'];
```
rowanc1, line 24: Does avif need to be here as well?

goekce:
Related PR in the default theme: https://github.com/jupyter-book/myst-theme/pull/926

packages/myst-cli/src/utils/resolveExtension.ts
goekce: avif extension is used by AV1 encoded pictures. AV1 encoded videos can be integrated both to webm and mp4, which are *container formats*.

goekce: So avif should not need be there.

PR #3002, opened 2026-07-15
Add public manifest to site builds

### LINKED ISSUE #3003: Implement manifest of public files in MyST site build
I recently built https://github.com/agoose77/jupyter-book-site-renderer, an implementation of MyST's AST → HTML rendering in Python. I noticed when building this that for static builds, having a CDN is not enough → it is not possible to exhaustively discover what is in `public/` unless you walk the entire AST, and in future plugins may populate static in dynamic ways.

I propose that we simply implement some kind of manifest that lists the contents of `public/`. It should be a new file to avoid bloating `config.json`, and it should be trivial to read. This would make it possible for site renderers to simply consume this manifest and mirror each file.

I also mentioned this here: https://github.com/jupyter-book/myst-theme/pull/595#issuecomment-2923050770

agoose77:
Themes that wish to build static sites from site builds do not have enough information to do so without having file-system access to the CDN directory. Namely, they do not have any way to determine the set of files under `public/`.

For example, see https://github.com/jupyter-book/mystmd/issues/3003

This PR adds a simple `public.json` JSON list of paths under public.

Closes https://github.com/jupyter-book/mystmd/issues/3003

### DIFF
--- a/.changeset/grumpy-lions-push.md
+++ b/.changeset/grumpy-lions-push.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
+Add public.json file manifest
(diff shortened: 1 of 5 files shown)


choldgraf (commented):
This looks useful! Could you document that this exists and what it should be used for? Or do you think this is a "anybody would expect this file to exist and we're just behind in not putting it there" kind of thing?

PR #3015, opened 2026-07-28
Update GH pages example workflow

JimMadge:
The workflow you can create with `myst init --gh-pages` had some quite old versions for actions and node.
I have bumped those here.

I was thinking there might be a more sustainable way to do this. The numbers aren't used elsewhere so moving them to a struct doesn't add much value. I'm not sure if some process which checks the versions in the string against what is on the marketplace is really worth the complexity.

### DIFF
--- a/.changeset/neat-shirts-attend.md
+++ b/.changeset/neat-shirts-attend.md
@@ -1,0 +1,5 @@
+---
+'myst-cli': patch
+---
+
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
Many thanks for the improvement! I agree there's probably a more robust way to solve this - for now I'll merge this because it's a nice fix!

PR #3019, opened 2026-08-03
Change execution session type to console

agoose77:
`jupyter-server-documents` complains about being unable to open the file associated with a MyST kernel (when using a remote kernel). This is because MyST expects to drive the execution, whereas JSD wants to execute notebooks locally.

This PR changes the type to console, which indicates avoids this error condition.


agoose77:
The test failures look unrelated.

PR #3023, opened 2026-08-07
Fix: normalize kernel path with forward slashes (Windows)

### LINKED ISSUE #3010: Windows: Build executes notebooks from the root rather than notebook location
## Description

tl;dr building and executing docs locally executes relative to `myst.yml` (as expected), but on CI runs from the parent directory of the notebook.  [This continues to happen now on napari/workshops](https://github.com/napari/workshops/actions/runs/29611364305/job/87986566890?pr=66).

cc @stefanv by request as we discussed this in person at SciPy

From debugging in February (sorry I didn't make an issue sooner, it's me, not you) I made a test repo ([timmonko/napari-jupyterbook](https://github.com/TimMonko/napari-jupyterbook)) to figure out what was happening on CI compare to local commands of something like `jupyter-book build` or `pixi run <some-task-that-builds>` where either `start` or `build` both ave the same error. This [files.md](https://github.com/TimMonko/napari-jupyterbook/blob/main/notebooks/files.md) shows the problem and is laid out in prose here: https://github.com/TimMonko/napari-jupyterbook#file-paths-in-notebooks
Ultimately:
```
# Locally (pixi run start --execute or build)
Current working directory:  C:\Users\...\napari-jupyterbook
Does notebooks/data exist?  True
Does data exist?  False

# On CI (pixi run build)
Current working directory:  /home/runner/work/napari-jupyterbook/napari-jupyterbook/notebooks
Does notebooks/data exist?  False
Does data exist?  True
```

## Proposed solution

The goal is that CI and local will be identical. It would make the most sense to me that execution would run relative to myst.yml, as occurs locally

## Additional notes

My workaround has been 
```python
from pathlib import Path

# Works whether CWD is the repo root or the notebooks/ subdirectory
data_dir = next(p for p in [Path('data'), Path('notebooks/data')] if p.exists())
```
certainly not ideal 🤷

TimMonko:
# References and Relevant Issues

Closes #3010

# Description

When executing notebooks on Windows, the kernel starts from the root rather than the notebook's directory. On Unix systems, the kernel starts from the notebook's directory. This is because `path.relative` returns native separators (backslashes on Windows), and Jupyter Server API paths are `/`-delimited. With a Windows path, Jupyter Server silently falls back to the server root.

[Jupyter server docs explaining the path resolution](https://jupyter-server.readthedocs.io/en/latest/developers/contents.html#api-paths)
And I think this is the actual [juypter-server forward-slash code](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/contents/filemanager.py#L1150-L1156) being executed by myst-execute.

I made https://github.com/TimMonko/myst-cwd-repro to demonstrate the issue -- and building with bun (my first time) and running this repo works with the fix.

I'm not sure if I can demonstrate this with a regression test because CI is only Linux it appears, so I guess you could mock the Windows path? Anyways, it's why I left a thorough comment in the code.

I used Deepseek to help me navigate mystmd. To be honest, I've never looked at typescript in my life so please be honest with me.

### DIFF
--- a/packages/myst-execute/src/kernel.ts
+++ b/packages/myst-execute/src/kernel.ts
@@ -27,8 +27,13 @@ export async function createKernelConnection(
   vfile: VFile,
   log?: Logger,
 ): Promise<ISessionConnectionWithKernel | undefined> {
+  // Jupyter Server API paths are `/`-delimited, but `path.relative` returns
+  // native separators (backslashes on Windows).
+  // With a windows path, jupyter_server silently fall back to the server root.
+  // Normalize to a forward-slash path so kernels start in same location on all platforms.
+  const sessionPath = path.relative(basePath, vfile.path).split(path.sep).join('/');
   const sessionOpts = {
-    path: path.relative(basePath, vfile.path),
+    path: sessionPath,
     name: path.basename(vfile.path),
     type: 'console',
     kernel: {


choldgraf (approved):
This feels like a safe change to me, thanks for the comments and keeping this tightly focused

agoose77:
Excellent — good spot that the paths are unix-like, not platform-specific!

PR #3027, opened 2026-08-17
Add `html` rendering pathway

agoose77:
Our existing HTML rendering pathway is facilitated by `mystmd`. There's no direct need to do things this way. This PR adds an opt-in rendering pathway for themes to implement HTML rendering, which permits a theme to e.g. use a SSG approach.

### DIFF
--- a/packages/myst-cli/src/build/html/index.ts
+++ b/packages/myst-cli/src/build/html/index.ts
@@ -169,6 +169,8 @@ function getBaseUrl(session: ISession): string | undefined {
  */
 export async function buildHtml(session: ISession, opts: StartOptions) {
   const template = await getSiteTemplate(session, opts);
+  // Ask template for command to render itself into HTML
+  const renderCommand = (template.getValidatedTemplateYml().build as any)?.render;
   // The BASE_URL env variable allows for mounting the site in a folder, e.g., github pages
   const baseurl = getBaseUrl(session);
   // Note, this process is really only for Remix templates
@@ -176,18 +178,27 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
   const htmlDir = path.join(session.buildPath(), 'html');
   fs.rmSync(htmlDir, { recursive: true, force: true });
   fs.mkdirSync(htmlDir, { recursive: true });
-  const appServer = await startServer(session, { ...opts, buildStatic: true, baseurl });
+  const appServer = await startServer(session, {
+    ...opts,
+    buildStatic: true,
+    baseurl,
+    // We do not need a running site if the template knows how to render itself
+    headless: renderCommand !== undefined,
+  });
   if (!appServer) return;
+
   const host = `http://localhost:${appServer.port}`;
 
-  const renderCommand = (template.getValidatedTemplateYml().build as any)?.render;
+  // Use the template to render itself
   if (renderCommand !== undefined) {
     // Run pre-rendering
     await makeExecutable(renderCommand, createNpmLogger(session), {
       cwd: template.templatePath,
       env: { ...process.env, BUILD_DIRECTORY: htmlDir, CONTENT_CDN: host },
     })();
-  } else {
+  }
+  // Fallback on fetch-based rendering (deprecated)
+  else {
     const routes = await currentSiteRoutes(session, host, baseurl);
 
     // Fetch all HTML pages and assets by the template
--- a/packages/myst-cli/src/build/html/index.ts
+++ b/packages/myst-cli/src/build/html/index.ts
@@ -10,7 +10,7 @@
 
 import fs from 'fs-extra';
 import path from 'node:path';
-import { writeFileToFolder } from 'myst-cli-utils';
+import { writeFileToFolder, createNpmLogger, makeExecutable } from 'myst-cli-utils';
 import type { MystXRefs } from 'myst-transforms';
 import type { ISession } from '../../session/types.js';
 import type { StartOptions } from '../site/start.js';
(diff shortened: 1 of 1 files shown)


packages/myst-cli/src/build/html/index.ts
```diff
@@ -177,54 +179,81 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
...
 
-  // Copy all of the static assets
-  fs.copySync(session.publicPath(), path.join(htmlDir, 'build'));
+    // Copy the files for the template used.
+    //
+    // This always includes the thebe JS chunks, even when no project enables
+    // `thebe`/`jupyter`. The myst-theme uses thebe-core to render Jupyter cell
+    // outputs, so these chunks are required for outputs to render at all.
+    const templateBuildDir = path.join(template.templatePath, 'public');
+    fs.copySync(templateBuildDir, htmlDir);
```
rowanc1, line 261: This should probably be the responsibility of theme render function if provided of ensuring all assets are available. (Might need to pass template directory in as env variable?)

packages/myst-cli/src/build/html/index.ts
```diff
@@ -177,54 +179,81 @@ export async function buildHtml(session: ISession, opts: StartOptions) {
...
+    fs.copySync(path.join(session.sitePath(), 'config.json'), path.join(htmlDir, 'config.json'));
+
+    // We need to go through and change all links to the right folder
+    rewriteAssetsFolder(htmlDir, baseurl);
+  }
+  await appServer.stop();
```
rowanc1, line 267: If the template is in charge of rendering html, then we shouldn't have to start the server (some could probably do it without a server, but it should be a template choice).

rowanc1 (commented):
A few comments when looking on my phone on pushing more responsibility to the render function if it is provided.

I would say we go all or nothing TBH. If there is a render html, it should do everything (including favicon, sitemap, js copy). "Given this content make it an html site in this directory".

agoose77:
@rowanc1 yes - this is a WIP that I wanted to push whilst I'm OoO! 

There are nuances around non route paths, but I've made some progress on the theme side to just proxy those routes such that we should be able to eliminate special cases on the mystmd side

packages/myst-cli/src/build/html/index.ts
agoose77: Yep, we only want the content server. To-do!

packages/myst-cli/src/build/html/index.ts
agoose77: This code-path is only for the existing rendering (if a template doesn't support HTML rendering). My view is that we deprecate this, but remove it later.

packages/myst-cli/src/build/html/index.ts
```diff
@@ -70,10 +70,12 @@ export async function currentSiteRoutes(
           };
         }),
         // Download other assets
-        ...['robots.txt', 'myst-theme.css', 'sitemap.xml', 'sitemap_style.xsl'].map((asset) => ({
-          url: `${host}/${asset}`,
-          path: asset,
-        })),
+        ...['robots.txt', 'myst-theme.css', 'sitemap.xml', 'sitemap_style.xsl', 'objects.inv'].map(
```
agoose77: Fixes missing download in old render pathway.

packages/myst-cli/src/build/html/index.ts
```diff
@@ -167,64 +167,104 @@ function getBaseUrl(session: ISession): string | undefined {
...
+    );
+
+    // Copy all of the static assets
+    fs.copySync(session.publicPath(), path.join(htmlDir, 'build'));
+    fs.copySync(
+      path.join(session.sitePath(), 'myst.search.json'),
+      path.join(htmlDir, 'myst.search.json'),
+    );
+
+    // NOTE: HTML static output needs to patch the contents, this is done on the fly by the server
```
stefanv: Can you say for what purpose?

packages/myst-cli/src/build/html/index.ts
```diff
@@ -167,64 +167,104 @@ function getBaseUrl(session: ISession): string | undefined {
...
+  // Use the template to render itself
+  if (renderCommand !== undefined) {
+    // Run pre-rendering
+    const cdn_url = `http://localhost:${appServer.contentServer.port}`;
+    await makeExecutable(renderCommand, createNpmLogger(session), {
+      cwd: template.templatePath,
+      env: { ...process.env, BUILD_DIRECTORY: htmlDir, CONTENT_CDN: cdn_url },
+    })();
+  }
+  // Fallback on fetch-based rendering (deprecated)
```
stefanv, line 197: Shall we add a deprecation notice here for theme authors? Or at least add a TODO sometwhere (do we have a TODO.txt?) so we remember to deprecate this later?

packages/myst-cli/src/build/html/index.ts
stefanv: Did this get addressed?

stefanv (approved):
Looks good overall, Angus; I've left some small comments and one check. Go ahead and merge when you feel this is ready, and we'll try it out.

packages/myst-cli/src/build/html/index.ts
agoose77: It might be interesting to do something like https://github.com/alstr/todo-to-issue-action

I'm not worried in this case, though — we'll visit this code again when we do things like generalise the exporters.

packages/myst-cli/src/build/html/index.ts
agoose77:
```suggestion
    // MyST content-server `myst.xref.json` objects contain `/content`/
    // as a marker for the "root" URL of the deployed site.
    // Meanwhile, MyST tooling expects links to be relative to the root.
    // We therefore must strip off `/content/` such that this holds true.
```

packages/myst-cli/src/build/html/index.ts
agoose77: I think so, it was originally starting the theme as well as content server (by accident)

agoose77:
I've actioned @stefanv's comments, and am self-merging.

choldgraf:
:shipit:

PR #3040, opened 2026-08-27
Improve LaTeX citations

### LINKED ISSUE #858: In text citation issue when building PDF
My in text citations are working fine on the local web server, but when building a PDF, some of the citation options seem to be ignored. For example, none of the prefix or suffixes appear, or options like "-" for only showing the year are ignored. 

I've tried add suffix/prefix with both of the following format:
```
{cite:p}`{see}@https://doi.org/10.1126/science.1235798{fig 1}`
```
```
[e.g. @@https://doi.org/10.1126/science.1235798, pg. 22]
```

## Additional notes

I'm using mystmd v1.1.37, and I've tried both the `plain_latex` and `lapreprint` templates.

arthurrump:
This PR aims to improve the handling of citations in LaTeX, both for standard natbib and for biblatex. I based the testcases on the list provided in #858, so it now correctly handles narrative and parenthetical citations, with partial citations where supported and prefixes and suffixes for all configurations.

I included a few cases that are representable in the AST, but for which I could not find a valid syntax. This includes multiple narrative citations (`\citet{jon90, jam91}`), partial author citations and partial year citations in narrative form. These cases are marked with `# myst: ??` in the testcases.

I decided to move all logic to the `cite` handler and let `citeGroup` be a simple pass-through to all children, to avoid a duplication of logic in both. The downside is that the `cite` node changes behaviour not only based on its parent (which was already the case), but also based on its siblings. Overall I think I prefer this over extracting some of the logic and then using that in both the `citeGroup` and in `cite`. An alternative could be to enforce that `cite`s are always in a `citeGroup`, but given that most of the relevant properties are on `cite`, I don't think that option makes much sense.

Fixes #858

### DIFF
--- a/.changeset/six-words-watch.md
+++ b/.changeset/six-words-watch.md
@@ -1,0 +1,5 @@
+---
+'myst-to-tex': patch
+---
+
+Improve LaTeX citation handling
--- a/packages/myst-to-tex/src/index.ts
+++ b/packages/myst-to-tex/src/index.ts
@@ -403,29 +403,71 @@ const handlers: Record<string, Handler> = {
     state.write(text.replace(/%s/g, `\\ref{${id}}`));
   },
   citeGroup(node, state) {
-    if (state.options.citestyle === 'numerical-only') {
-      state.write('\\cite{');
-    } else if (state.options.bibliography === 'biblatex') {
-      const command = node.kind === 'narrative' ? 'textcite' : 'parencite';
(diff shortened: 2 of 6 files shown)


agoose77:
Thanks for this PR! We're currently preparing for a release, and I doubt this one will make it. But we aim to release frequently, so it shouldn't be held up too long.

PR #3041, opened 2026-09-01
Fail site builds when no site configuration is found

### LINKED ISSUE #3026: `myst build --html` returns non-zero exit code if no configuration file (`myst.yml`) was found
Behavior:
```
pushd $(mktemp -d)
virtualenv venv
. ./venv/bin/activate
pip install mystmd
myst build --html

echo $?
```
Output:
```
building myst-cli session with API URL: https://api.mystmd.org
(node:897905) ExperimentalWarning: localStorage is not available because --localstorage-file was not provided.
(Use `node --trace-warnings ...` to show where the warning was created)
🍡 Execution parallelism set to: 11
🌎 No site configuration found.

0
```
This leads to confusion in CI jobs. The build job succeeds even no output was generated.

## Proposed solution

Return non-zero exit code.

## Additional notes

```
myst --version   
v1.10.1
```

BetterAndBetterII:
`myst build --html` (and `--site` / `--all`) in a directory with no `myst.yml` printed "No site configuration found" and still exited 0, so CI treated a no-op as success.

When a site build is requested and no site configuration is loaded, `build()` now throws after the existing log. `clirun` already maps a thrown error to `process.exit(1)`. Other `myst build` invocations that are not asking for a site still succeed.

Tests cover `--html`, `--site`, and `--all` in an empty temp directory.

Fixes #3026

### DIFF
--- a/packages/myst-cli/src/build/build.ts
+++ b/packages/myst-cli/src/build/build.ts
@@ -222,6 +222,13 @@ export async function build(session: ISession, files: string[], opts: BuildOpts)
   const { site, all, watch, writeDOIBib } = opts;
   const performSiteBuild = all || (files.length === 0 && exportSite(session, opts)) || writeDOIBib;
   const exportOptionsList = await collectAllBuildExportOptions(session, files, opts);
+  const state = session.store.getState();
+  const siteConfig = selectors.selectCurrentSiteConfig(state);
+  const projectConfig = selectors.selectCurrentProjectConfig(state);
+  if (files.length === 0 && !siteConfig && !projectConfig) {
+    session.log.info('No configuration found.');
+    throw new Error(`No configuration found. First run '${binaryName()} init'`);
+  }
   // TODO: generalize and pull this out!
   const buildLog: Record<string, any> = {
     input: {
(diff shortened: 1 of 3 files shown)


packages/myst-cli/src/build/build.ts
```diff
@@ -272,6 +272,9 @@ export async function build(session: ISession, files: string[], opts: BuildOpts)
     if (!siteConfig) {
       session.log.info('🌎 No site configuration found.');
       session.log.debug(`To build a site, first run '${binaryName()} init --site'`);
+      throw new Error(
+        `No site configuration found. To build a site, first run '${binaryName()} init --site'`,
```
stefanv, line 283: Why only for sites? Shouldn't we return a non-zero exit code for any type of build with a missing config?

BetterAndBetterII: Expanded this: any `myst build` with no files listed and no project or site configuration now throws, so `clirun` exits 1. That includes `--pdf` / `--docx` / `--tex` and a bare `myst build`, not only `--html` / `--site` / `--all`.

A project-only config still fails site builds the same way as before (`No site configuration found`). Listing explicit files is unchanged.

PR #3042, opened 2026-09-01
feat: configure public site URLs

### LINKED ISSUE #2695: sitemap starts with localhost
Hi, I was about to submit the sitemap to Google and Bing, when I noticed that all links are in the form
```
<loc>http://localhost:3000/...</loc>
```
Shouldn't they start with the `BASE_URL`?

I'm using jupyter-book 2.1.2, which is based on mystmd 1.8.0.

cnotin:
Fixes #2695

Summary:
- Add and validate site.url.
- Infer the routing and asset base path from site.url or SITE_URL, and reject conflicting BASE_URL values.
- Preserve the site URL in the manifest, document the deployment override, and warn when static builds lack a public URL.

Testing:
- npx vitest run packages/myst-config/src/site/site.spec.ts
- npx bun run build:esm in packages/myst-config
- npx prettier --check (changed files)


stefanv:
My initial thought: it is worth adding a canonical_url field *per page*.

I'm not equally sure whether it makes as sense to add site-wide; would you ever try to set it to a different domain than where the site is being hosted? Because the theme already knows how to introspect the hosted domain name.

Canonical URLs that point outside the current domain should not be added to sitemap.

cnotin:
I agree that per-page canonical URLs are a separate SEO feature and should not determine sitemap entries. This proposal is not intended to set page-level canonical metadata. It supplies the public deployment base URL required by static builds, where the renderer only receives `localhost:3000` and cannot infer the final host. Same goal for my other PR: https://github.com/jupyter-book/myst-theme/pull/950

For dynamic serving, we retain the request origin. For static builds, `SITE_URL` / configuration identifies the domain where that generated sitemap and robots file will actually be served. The sitemap then remains on that explicitly declared public site base.

Would `site.public_url` or `site.url` be clearer than `site.canonical_url`, to avoid conflating it with per-page canonical links?

stefanv:
Yes, in that case I think canonical URL is overloaded; `site.url` would be more explicit.

stefanv:
If we do set `site.url`, do we still need BASE_URL? Do we infer BASE_URL? What if they conflict?

cnotin:
Good points! `BASE_URL` isn't needed anymore but I suppose that many users are already using it and you don't want a regression so it'll have to remain supported. I suggest to normalize both to detect and error when there's a conflict.

cnotin:
Implemented this in the latest commit: `site.url` is now the complete public deployment address, and its pathname supplies the effective base URL for routing and assets. `BASE_URL` remains supported for existing path-only deployments and for sites without `site.url`, but an explicitly set value now fails the build if it differs from the path in `site.url` (or its `SITE_URL` override). This avoids silently generating asset links and sitemap URLs for different deployment paths.

stefanv:
Thank you for the update; may I ask that you also review https://mystmd.org/guide/contributing#ai-policy please.

cnotin:
Sorry that I missed that AI policy. Indeed this is an AI-assisted work using Codex with GPT 5.6 Terra/Sol.
I've reviewed and tested the code myself but I'll get a 2nd deeper look.

cnotin:
As per the AI policy, I've reviewed manually the code and it looks good to me. I've also steered it until it reached a state that satisfied me.

choldgraf:
We discussed this a bit in the working meeting! Could you explore a few things to improve this PR?

- It feels like we could simplify the UX by allowing `BASEURL` to accept the fully URL (`https://example.com/foo`) instead of just the subfolder (`/foo`).
- Could you first confirm that this is how other SSG-style projects handle site urls? Here were a few examples we discussed for reference:
  - Hugo: https://gohugo.io/methods/site/baseurl/
  - Sphinx: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl [sphinx 
- If that is indeed sensible, could you re-work this PR so that we expand the capabilities of `BASEURL` as an environment variable?
- For now, don't introduce a new configuration pathway, and let's keep it to environment variables
- In a follow up PR, we'd definitely like to introduce more ways to set configuration other than just environment variables, so that's where we could add support for something like `site: baseurl:` in the config (or maybe even via the command line?)

Edit: just to clarify after Stefans comment: thank you for doing this! It indeed looks helpful and moving in the right direction, we appreciate it 🙂

stefanv:
@cnotin I should add that the reason we discussed this in the working meeting this morning is because you brought this PR to a viable state, have been careful in exploring the edge cases, and it looked like it was ready to consider for inclusion; so thank you for that!

And the reason we are asking about exploring whether to add this functionality to the base URL, is because we strive to keep a balance between keeping the implementation as simple as possible while supporting the desired use-cases.

But, to be clear, in principle we are agreed to support the sitemap use case. We just have to figure out exactly how.

cnotin:
Thank you for your feedbacks and pushing further. I appreciate that you're interested by this. For the background, I noticed the issue on a [website using MyST](https://spook.boo/) for which I suggested [an ugly, but effective, hack](https://github.com/frenck/spook/pull/1568). And then I thought it'd be better to suggest a proper fix upstream :)

And for this project I fixed, the env var will be enough. So the config item can wait I agree.


For the benchmark of the other projets:

* For Hugo, Base URL [the full URL](https://gohugo.io/methods/site/baseurl/) (host + subfolder)
* For Sphinx, it's also a [full URL](https://sphinx-llms-txt.readthedocs.io/en/latest/advanced-configuration.html#using-html-base-url)
* And for ReadTheDocs, it's also the [full URL](https://docs.readthedocs.com/platform/stable/reference/environment-variables.html#envvar-READTHEDOCS_CANONICAL_URL)
* However, for Jekyll (which I'm more familiar with), base URL is [just the subfolder](https://mademistakes.com/mastering-jekyll/site-url-baseurl/) so that makes a counter-example!

So we can say that base URL = full URL since it's more common.

I'm just afraid of regression for folks already using `BASE_URL` in MyST, with a subfolder value [as currently documented](https://mystmd.org/guide/deployment#deploy-base-url). But I can try to see to deal correctly with both cases.

cnotin:
Re-switched to a common helper module in myst-common to factorize the code with myst-theme

PR #3045, opened 2026-09-01
Add replacement for path.parse that handles multi-extensions like .tar.gz

### LINKED ISSUE #3043: Download Filename Hashing breaks multi-part file extensions
Filename content-hashing as done by the [`{download}`-role](https://mystmd.org/guide/website-downloads) breaks multi-part file extensions such as in `.tar.gz` gzipped tarballs.

Observed with MystMD v1.10.1.

## Description

I have a situation where I'd like to offer a file called `archive.tar.gz` for download, but MyST renames it to `archive-[CONTENT_HASH].gz`, omitting the `.tar` portion of the filename. As a consequence, archiver GUI tools (at least [Ark](https://apps.kde.org/ark/), in my case) don't recognize the file as a gzipped tarball anymore, and users become unable to open the archive.

## Proposed solution

As far as I understand, the problem would be solved by the query-based solution for cache busting discussed in #1196. Also, an option to disable cache-busting for a specific download (#709) would also work, but seems like a rather dirty workaround.

stefanv:
Closes https://github.com/jupyter-book/mystmd/issues/3043

AI disclosure: I iterated on the parseMultiExt design with both Gemini and Claude Opus 5, and had Claude Opus 5 write the tests.

### DIFF
--- a/packages/myst-cli-utils/src/filesystem.ts
+++ b/packages/myst-cli-utils/src/filesystem.ts
@@ -23,7 +23,7 @@ export function writeFileToFolder(
 
 /* Replacement for path.parse that handles multi-extensions, like .tar.gz */
 /* Could make this more complex and filter edge cases like report.2026.final.pdf */
-function parseMultiExt(filepath: string) {
+export function parseMultiExt(filepath: string) {
   const parsed = path.parse(filepath);
   // Everything from the first dot on is the extension; a leading dot is
   // part of the name, so `.bashrc` has no extension.
--- a/packages/myst-cli-utils/src/filesystem.ts
+++ b/packages/myst-cli-utils/src/filesystem.ts
@@ -21,6 +21,18 @@ export function writeFileToFolder(
   fs.writeFileSync(filename, data, opts);
 }
 
+/* Replacement for path.parse that handles multi-extensions, like .tar.gz */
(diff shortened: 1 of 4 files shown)


packages/myst-cli-utils/src/filesystem.ts
```diff
@@ -21,6 +21,18 @@ export function writeFileToFolder(
   fs.writeFileSync(filename, data, opts);
 }
 
+/* Replacement for path.parse that handles multi-extensions, like .tar.gz */
+/* Could make this more complex and filter edge cases like report.2026.final.pdf */
+export function parseMultiExt(filepath: string) {
```
agoose77, line 26: I was worried about the fragility of this, but we're not _actually_ worried about whether `foo.bar.baz` is `foo .bar.baz` or `foo.bar .baz`. Instead, this function is ensuring that we set a max split of 1, so that we definitely have a valid filepath.

agoose77 (approved):

packages/myst-cli-utils/src/filesystem.ts
stefanv: I think the original issue did have a concern around how the extension gets split, because it affects the extension of the downloaded artifact.

PR #3047, opened 2026-09-02
fix(myst-to-typst): use label() for cross-references whose identifier has spaces

### LINKED ISSUE #2984: Typst PDF export breaks for multi-word glossary `{term}` references
## Description
Glossary `{term}` references work correctly in the HTML build, but fail during Typst PDF export when the target glossary term contains spaces.

Example MyST source:

```md
In a notebook, {term}`execution order <Execution order>` matters.
````

The Typst export generates:

```typst
#link(<term-execution order>)[execution order]
```

This is invalid Typst syntax because angle-bracket labels cannot contain spaces. Typst then fails with errors such as:

```text
error: unclosed label
error: expected expression
```

The same issue occurs for other multi-word glossary terms, for example:

```typst
#link(<term-computer programming>)[programming]
#link(<term-assignment operator>)[assignment operator]
#link(<term-code cell>)[code cell]
#link(<term-snake case>)[snake\_case]
```

Single-word glossary terms such as `{term}`Program`` seem to export without this specific syntax problem.

## Proposed solution

Slugify glossary term labels before writing them into Typst angle-bracket labels.

For example, the Typst export should generate:

```typst
#link(<term-execution-order>)[execution order]
```

instead of:

```typst
#link(<term-execution order>)[execution order]
```

This would make the generated labels valid Typst syntax and consistent with common anchor-style label generation.

## Additional notes

This appears to be specific to Typst PDF export. The same glossary references work correctly in the HTML build.

Current workarounds are to avoid `{term}` roles for multi-word glossary terms, replace them with ordinary Markdown links, or post-process the generated temporary `.typ` files. None of these are ideal because they either lose semantic glossary references or require patching generated files in `_build/temp`.

Jorge-Polanco-Roque:
Fixes #2984

## Problem

Multi-word glossary `{term}` references break Typst PDF export. The cross-reference emits `#link(<term-Execution order>)`, but Typst angle-bracket labels cannot contain spaces, so the export fails with `unclosed label` / `expected expression`. The reference target is already written as `#label("...")`, so the two forms did not match.

## Fix

When the identifier is not label-safe (`/^[a-zA-Z0-9_\-:.]+$/`), emit `label("...")` (and `#ref(label(...))` for the no-children shorthand branch) — the same pattern `cite` already uses in this file. Identifier-safe references are unchanged, and this matches the target's `#label("...")` so the reference resolves.

## Testing

Added a `cross-references.yml` case for a spaced identifier; it fails without the fix (produces `<term-Execution order>`). The full `myst-to-typst` suite (71 tests), `tsc`, prettier and eslint all pass. Included a changeset.

_Disclosure (per the project's AI policy): this change was prepared with AI assistance; I reviewed and tested every line, and wrote this description myself._

### DIFF
--- a/.changeset/typst-label-spaces.md
+++ b/.changeset/typst-label-spaces.md
@@ -1,0 +1,5 @@
+---
+'myst-to-typst': patch
+---
+
+Fix Typst PDF export of cross-references whose identifier contains spaces (e.g. multi-word glossary terms). These now emit `label("...")` instead of an invalid angle-bracket label, matching how the reference target is written.
--- a/packages/myst-to-typst/src/index.ts
+++ b/packages/myst-to-typst/src/index.ts
@@ -361,10 +361,18 @@ const handlers: Record<string, Handler> = {
       return;
     }
     const id = node.identifier;
+    // Typst angle-bracket labels and the `@label` shorthand only accept
+    // identifier-safe names, but cross-reference targets (e.g. glossary terms)
+    // can contain spaces and are emitted as `#label("...")`. Match that form
+    // here so the reference resolves instead of producing invalid Typst.
+    const needsLabel = !/^[a-zA-Z0-9_\-:.]+$/.test(id ?? '');
+    const idRef = needsLabel ? `label("${id}")` : `<${id}>`;
     if (node.children && node.children.length > 0) {
-      state.write(`#link(<${id}>)[`);
+      state.write(`#link(${idRef})[`);
       state.renderChildren(node);
       state.write(']');
+    } else if (needsLabel) {
+      state.write(`#ref(${idRef})`);
     } else {
       // Note that we don't need to protect against the previous character as text
       const next = nextCharacterIsText(parent, node);
--- a/packages/myst-to-typst/src/index.ts
+++ b/packages/myst-to-typst/src/index.ts
@@ -366,7 +366,10 @@ const handlers: Record<string, Handler> = {
     // can contain spaces and are emitted as `#label("...")`. Match that form
     // here so the reference resolves instead of producing invalid Typst.
     const needsLabel = !/^[a-zA-Z0-9_\-:.]+$/.test(id ?? '');
-    const idRef = needsLabel ? `label("${id}")` : `<${id}>`;
+    // Escape backslashes and quotes so an identifier with special characters
+    // cannot break out of the Typst string literal.
+    const labelArg = (id ?? '').replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+    const idRef = needsLabel ? `label("${labelArg}")` : `<${id}>`;
     if (node.children && node.children.length > 0) {
       state.write(`#link(${idRef})[`);
       state.renderChildren(node);
--- a/packages/myst-to-typst/tests/cross-references.yml
+++ b/packages/myst-to-typst/tests/cross-references.yml
@@ -95,3 +95,16 @@ cases:
               value: ' matters.'
     typst: |-
       In a notebook, #link(label("term-Execution order"))[execution order] matters.
+  - title: reference identifier containing a quote is escaped in the Typst label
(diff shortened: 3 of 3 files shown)


Jorge-Polanco-Roque:
Gentle nudge on this one, it's been a couple weeks. It fixes the Typst export breaking on glossary cross-references whose identifier has spaces — those were coming out as invalid Typst (`<term x y>`), and the fix uses `label()` so the target actually matches. It's small and self-contained, with a test that fails without it, and I checked it end-to-end against the bun/turbo build. Whenever someone has a minute to take a look I'd really appreciate it. Thanks!

stefanv:
Thanks for the fix @Jorge-Polanco-Roque; I'm going to make a release imminently, and will get this PR reviewed for it.

stefanv:
I took the liberty to push some fixes to this PR; please take a look and see if you're OK with them.

stefanv:
I have some more changes coming in; found another bug.

agoose77:
@stefanv on mobile - a commit body states that heading and paragraph targets are always valid names, is that the case?

Jorge-Polanco-Roque:
Author here — jumping in since this touches my original change. The question's really for @stefanv on his commit, but for context on the reference side:

My original fix didn't key off the node type at all. It just checks whether the *identifier itself* is label-safe and picks the emission form from that:

```js
const needsLabel = !/^[a-zA-Z0-9_\-:.]+$/.test(id ?? '');
// label("...") when not safe, <id> when safe
```

So headings and paragraphs went through the same guard as glossary terms — nothing assumed their targets are always valid names. Auto-generated heading slugs are kebab-case and safe in practice, but a user-defined label/target (e.g. `(My Target)=`) can contain spaces or other characters outside that set, so I kept the check universal instead of special-casing node types. It's basically free and sidesteps exactly this "is it always valid?" question.

If it'd help settle it, I'm happy to add a cross-reference test with a spaced heading/paragraph target so the behavior is pinned down either way. Thanks for picking this up for the release!

Jorge-Polanco-Roque:
Following up on @agoose77's question — I checked, and target names aren't always safe: the same problem exists on the **target** side, not just the reference.

`paragraph` and `heading` emit their target as `` <${identifier}> `` directly, without the label-safe guard I added for references. So a paragraph/heading whose identifier has spaces comes out as `<my target>`, which is invalid Typst — the same failure mode as the original glossary bug, just on the definition side instead of the reference.

I've got a small local test that pins this down (a paragraph target with a spaced label) plus a one-helper fix that routes both `paragraph` and `heading` targets through the same `label("...")` guard. Test fails without the fix, passes with it; lint/tsc green.

@stefanv this might be the "other bug" you mentioned — happy to push the test + fix, or hold off if you're already on it. Just let me know which is less duplicated effort.

stefanv:
Yeah, basically I tried the original fix and noticed that there were cases that weren't covered (for which I added tests). And I wasn't confident about the escaping approach.

The first attempt was a bit wild, so I tried to reign it in a bit. You are welcome to drop or squash the commits I made, to your own preference.

I would suggest keeping the tests and ensuring that those types of examples all pass with a real typst compile.

I also considered how the ultimate rendering looked: there are shorter paths that get a functional render, but it's nice if it is human readable (i.e. only touch labels when necessary).

stefanv:
By the way, it took me a long time of staring to see what `label("...")` meant. It is useful to clarify to future readers the distinction between triangle bracket and label-function syntax. I would use concrete examples.

stefanv:
> @stefanv this might be the "other bug" you mentioned — happy to push the test + fix, or hold off if you're already on it. Just let me know which is less duplicated effort.

Is this bug present on the current state of the branch? If so, go ahead and add the test.

agoose77:
This LGTM!

stefanv:
Thanks @Jorge-Polanco-Roque!

PR #3049, opened 2026-09-10
Add plausible analytics

jasongrout:
I think it may be this simple, based on https://github.com/jupyter/enhancement-proposals/pull/147/changes#r3606847629?

I created a new jupyterbook.org plausible intake already, to which we can push all jupyter book metrics.

See also https://github.com/jupyter-book/jupyterbook.org/pull/33 and https://github.com/jupyter-book/mystmd.org/pull/39

Stats are available at https://plausible.io/jupyterbook.org


stefanv:
Plausible doesn't have a free plan, right? We have https://views.scientific-python.org as an alternative, if that is the case.

choldgraf:
Nope - jupyter pays for its own plan. If plausible isn't hard to self host honestly that may be a reasonable idea at some point.

jasongrout:
> Plausible doesn't have a free plan, right? We have https://views.scientific-python.org as an alternative, if that is the case.

@stefanv, you use Plausible now? Is that a change from last year? How much traffic does your plausible instance manage?

I think I remember you telling us last year when we were investigating how to do metrics, and we asked about using the scientific-python metrics, that we should just pay Plausible and be done with it :).

FYI, we are paying on the order of $2200/year right now for Plausible.

jasongrout:
It seems that the plausible js is not showing up on page at https://mystmd.org/guide. I do see `"analytics_plausible":"jupyterbook.org"` showing up json data in the page, but it seems that whatever turns the config into a script tag on the page isn't working?

I'll let you all take it from here - not sure if something needs to be upgraded, or a theme needs to be modified, etc., or maybe I'm just misunderstanding how it gets deployed.

